PR #25, opened 2018-11-23
Solidity compiler custom arguments

Shirikatsu:
Current solidity compilation invokes solc. Solc can take more arguments that aren't able to be passed in for compilation and these have been added.

Motivation:
Compilation of contracts are restricted to the scope of the directory and all subdirectories of the target contract file by solc. Passing `--allow-paths` option and/or remapping arguments for relative paths allows solc to access other directories. Thus we need a way to pass custom arguments to the solc compiler. This is primarily used by the Ion CLI for contract compilation for deployment/execution. This does not seem to affect the client implementation.

### DIFF
--- a/cmd/abigen/main.go
+++ b/cmd/abigen/main.go
@@ -85,7 +85,7 @@ func main() {
 		var contracts map[string]*compiler.Contract
 		var err error
 		if *solFlag != "" {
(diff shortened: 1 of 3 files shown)


aneequesafdar:
@Shirikatsu do you still want this pull request to be merged into master?

Shirikatsu:
Yes please, if it doesn't break anything and still looks good.

PR #41, opened 2019-02-12
Autonity default eips

Shirikatsu:
Added default autonity protocol minimum configuration for genesis block. It ensures that all EIPs byzantium and below are included in the genesis block. These override any genesis.json specification of the affected protocol changes and is unopinionated about constantinople and beyond.

The motivation was to ensure Ion compatibility out of the box as the Ion stack fails when deployed to a chain that does not have the byzantium fork active due to missing EIP.


aneequesafdar:
some of the code is outdated and we have defaults for autonity contract now so I think it should be closed

PR #52, opened 2019-03-27
Restore tests and corresponding code

JekaMas:
* all tests that not skipped or ignored are fixed
* `make test` works fine
* to run all Geth tests you need to pull `testdata` submodule: `git submodule update --init --recursive` and then `make test`

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -17,7 +17,6 @@
 package backend
 
 import (
-	"bytes"
 	"crypto/ecdsa"
 	"math/big"
 	"sync"
@@ -36,7 +35,6 @@ import (
 	"github.com/clearmatics/autonity/event"
 	"github.com/clearmatics/autonity/log"
 	"github.com/clearmatics/autonity/params"
-	"github.com/clearmatics/autonity/rlp"
 	"github.com/hashicorp/golang-lru"
 )
 
@@ -361,47 +359,4 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
 		return false
 	}
 	return sb.hasBadBlock(hash)
-}
-
-func GetGenesisAndKeys(n int) (*core.Genesis, []*ecdsa.PrivateKey) {
-	// Setup validators
-	var nodeKeys = make([]*ecdsa.PrivateKey, n)
-	var addrs = make([]common.Address, n)
-	for i := 0; i < n; i++ {
-		nodeKeys[i], _ = crypto.GenerateKey()
-		addrs[i] = crypto.PubkeyToAddress(nodeKeys[i].PublicKey)
-	}
-
-	// generate genesis block
-	genesis := core.DefaultGenesisBlock()
-	genesis.Config = params.TestChainConfig
-	// force enable Istanbul engine
-	genesis.Config.Istanbul = &params.IstanbulConfig{}
-	genesis.Config.Ethash = nil
-	genesis.Difficulty = defaultDifficulty
-	genesis.Nonce = emptyNonce.Uint64()
-	genesis.Mixhash = types.IstanbulDigest
-
-	AppendValidators(genesis, addrs)
-	return genesis, nodeKeys
-}
-
-func AppendValidators(genesis *core.Genesis, addrs []common.Address) {
-
-	if len(genesis.ExtraData) < types.IstanbulExtraVanity {
-		genesis.ExtraData = append(genesis.ExtraData, bytes.Repeat([]byte{0x00}, types.IstanbulExtraVanity)...)
-	}
-	genesis.ExtraData = genesis.ExtraData[:types.IstanbulExtraVanity]
-
-	ist := &types.IstanbulExtra{
-		Validators:    addrs,
-		Seal:          []byte{},
-		CommittedSeal: [][]byte{},
-	}
-
-	istPayload, err := rlp.EncodeToBytes(&ist)
-	if err != nil {
-		panic("failed to encode istanbul extra")
-	}
-	genesis.ExtraData = append(genesis.ExtraData, istPayload...)
 }
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -17,6 +17,7 @@
 package backend
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"math/big"
 	"sync"
@@ -29,13 +30,14 @@ import (
 	"github.com/clearmatics/autonity/consensus/istanbul/validator"
 	"github.com/clearmatics/autonity/core"
 	"github.com/clearmatics/autonity/core/types"
+	"github.com/clearmatics/autonity/core/vm"
 	"github.com/clearmatics/autonity/crypto"
 	"github.com/clearmatics/autonity/ethdb"
 	"github.com/clearmatics/autonity/event"
 	"github.com/clearmatics/autonity/log"
-	"github.com/clearmatics/autonity/core/vm"
 	"github.com/clearmatics/autonity/params"
-	lru "github.com/hashicorp/golang-lru"
+	"github.com/clearmatics/autonity/rlp"
+	"github.com/hashicorp/golang-lru"
 )
 
 const (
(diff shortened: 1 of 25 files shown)


consensus/istanbul/backend/backend_test.go
```diff
@@ -19,16 +19,15 @@ package backend
 import (
```
maxrobot: small comment but go standard dictates packages should be added as such:
```
import (
	"bytes"
	"crypto/ecdsa"
	"sort"
	"strings"
	"testing"
	"time"

	"github.com/clearmatics/autonity/consensus/istanbul"
	"github.com/clearmatics/autonity/consensus/istanbul/validator"
	"github.com/clearmatics/autonity/core/types"
	"github.com/clearmatics/autonity/crypto"
)
```

JekaMas: Sure, will check other files in PR for `fmt` and `goimports`

core/blockchain.go
```diff
@@ -949,7 +949,13 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
 	}
 
 	// Call network permissioning logic before committing the state
-	bc.updateEnodesWhitelist(state, block)
+	if bc.chainConfig.Clique != nil {
```
yazzaoui: Our approach for network permissioning is only compatible with BFT-style consensus algorithms where finality is guaranteed (no soft forks). We had no requirements on supporting clique and open networks, that's why I didn't put any condition when I wrote the code.. You should anyway replace that by ` bc.chainConfig.Istanbul != nil`

JekaMas: Will try use the condition. I know that the goal is to support only *BFT, however we have a log panicing tests that is not good. Also it's better to have such conditions in case we want add a new consensus.

core/blockchain.go
```diff
@@ -949,7 +949,13 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
...
+		err = bc.updateEnodesWhitelist(state, block)
+		if err != nil && err != GlienickeContractError {
```
yazzaoui: I think we should abort if there is a problem with the permissioning contract, that's why I had `log.Crit` called in updateEnodesWhitelist

core/evm.go
```diff
@@ -87,6 +87,9 @@ func GetHashFn(ref *types.Header, chain ChainContext) func(n uint64) common.Hash
 // CanTransfer checks whether there are enough funds in the address' account to make a transfer.
 // This does not take the necessary gas in to account to make the transfer valid.
 func CanTransfer(db vm.StateDB, addr common.Address, amount *big.Int) bool {
+	if amount.Sign() == 0 {
```
yazzaoui: May I ask why  ?

core/genesis.go
```diff
@@ -281,20 +281,25 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
...
 			return nil, errGenesisBadWhitelist
 		}
 		enodeWhiteList = append(enodeWhiteList, newEnode)
 	}
-	rawdb.WriteEnodeWhitelist(db, enodeWhiteList)
-	config := g.Config
-	if config == nil {
-		config = params.AllEthashProtocolChanges
+
+	if len(enodeWhiteList) != 0 {
```
yazzaoui: I think we might just want to abort if `len(enodeWhiteList) == 0`, unless we want the possibility to switch a private network to a public one in-flight and vice-versa, but again no requirements for that.

core/blockchain.go
JekaMas: This condition is for Clique consensus only. With the log.Crit it'll require whitelist all the time or panic. I think we shouldn't extend whitelist requirement to other consensus.

core/evm.go
JekaMas: Get whitelist call uses 0 amount to get the data.

eth/handler.go
```diff
@@ -258,7 +258,9 @@ func (pm *ProtocolManager) Stop() {
 
 	pm.txsSub.Unsubscribe()        // quits txBroadcastLoop
 	pm.minedBlockSub.Unsubscribe() // quits blockBroadcastLoop
-	pm.glienickeSub.Unsubscribe() // quits glienickeEventLoop
+	if !pm.openNetwork {
```
yazzaoui: my bad

eth/handler.go
```diff
@@ -132,7 +132,7 @@ func NewProtocolManager(config *params.ChainConfig, mode downloader.SyncMode, ne
 		quitSync:    make(chan struct{}),
 		engine:      engine,
 		openNetwork: openNetwork,
-		glienickeCh: make(chan core.GlienickeEvent),
+		glienickeCh: make(chan core.GlienickeEvent, 64),
```
yazzaoui: Interesting, any specific reason for that ?

consensus/istanbul/backend/backend.go
```diff
@@ -360,3 +362,46 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
 	}
 	return sb.hasBadBlock(hash)
 }
+
+func GetGenesisAndKeys(n int) (*core.Genesis, []*ecdsa.PrivateKey) {
```
yazzaoui: Seems like a test related function, better to put it in backend_test.go instead

core/blockchain.go
yazzaoui: but then how `updateEnodesWhitelist` is going to be called for IBFT ?

JekaMas: Sure, will push commit in a minute. Thanks.

core/genesis.go
JekaMas: We can pass here and check the consensus type on a higher level (backend) and panic there.

We're going to be a public repo with the good and solid codebase. I think one of the requirements is green tests and stable CI. It's just a step further.

consensus/istanbul/backend/backend.go
JekaMas: It's used from outside of `backend` package so we can't move it to `_test` file.

core/evm.go
yazzaoui: This is quite a big change in the evm, we are now allowing transactions of 0 eth . Not sure to understand what was the original problem

eth/handler.go
JekaMas: Only to not to have a blocking channel. Karalabe uses a lot 16,32,64 capacity for channels, so I decided to use the same as `make(chan accounts.WalletEvent, 16)`, `fetch := make(chan *request, 16)`, `matches := make(chan uint64, 64)`. We can use any capacity here.

core/evm.go
JekaMas: Ok, I've removed it.

consensus/istanbul/backend/backend.go
JekaMas: I've rewritten the related code and moved back this functions to the test file.

core/blockchain.go
JekaMas: I've changed the condition to `bc.chainConfig.Istanbul`

yazzaoui:
LGTM

JekaMas:
I think I'll wait for a couple reviews more.
@maxrobot

eth/helper_test.go
```diff
@@ -49,7 +49,7 @@ var (
 // newTestProtocolManager creates a new protocol manager for testing purposes,
 // with the given number of blocks already known, and potential notification
 // channels for different events.
-func newTestProtocolManager(mode downloader.SyncMode, blocks int, generator func(int, *core.BlockGen), newtx chan<- []*types.Transaction) (*ProtocolManager, *ethdb.MemDatabase, error) {
+func newTestProtocolManager(mode downloader.SyncMode, blocks int, generator func(int, *core.BlockGen), newtx chan<- []*types.Transaction, peers []string) (*ProtocolManager, *ethdb.MemDatabase, error) {
```
aneequesafdar: Would `initialWhitelistPeers` be a better name than `peers`? It was a bit misleading? possible change in #221

JekaMas: Let's rename it. I'd prefer to do refactoring in a new issue to not hold any other issues.

aneequesafdar: ok sure, #226

Issue #53, opened 2019-03-28
[CI] Tests are not stable due to data races in IBFT


JekaMas:
We should check a few IBFT data races to have more stable and reliable tests.
There are quite a few data races more in Autonity tests, however, most of them are from Geth codebase and should be fixed after rebasing to 1.8.23 - has only 2 data races - or can be ignored.

1. 
```
==================
WARNING: DATA RACE
Write at 0x000001479930 by goroutine 90:
  github.com/clearmatics/autonity/consensus/istanbul/backend.New()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/build/_workspace/src/github.com/clearmatics/autonity/consensus/istanbul/backend/backend.go:60 +0x2d9
  github.com/clearmatics/autonity/consensus/istanbul/backend.newBlockChain()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/build/_workspace/src/github.com/clearmatics/autonity/consensus/istanbul/backend/engine_test.go:49 +0x222
  github.com/clearmatics/autonity/consensus/istanbul/backend.TestVerifyHeaders()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/build/_workspace/src/github.com/clearmatics/autonity/consensus/istanbul/backend/engine_test.go:364 +0x4f
  testing.tRunner()
      /snap/go/3417/src/testing/testing.go:865 +0x163

Previous read at 0x000001479930 by goroutine 41:
  github.com/clearmatics/autonity/consensus/istanbul/backend.(*backend).Validators()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/build/_workspace/src/github.com/clearmatics/autonity/consensus/istanbul/backend/backend.go:129 +0x15b
  github.com/clearmatics/autonity/consensus/istanbul/core.(*core).startNewRound()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/build/_workspace/src/github.com/clearmatics/autonity/consensus/istanbul/core/core.go:231 +0x1abe
  github.com/clearmatics/autonity/consensus/istanbul/core.(*core).handleTimeoutMsg()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/build/_workspace/src/github.com/clearmatics/autonity/consensus/istanbul/core/handler.go:198 +0x1f9
  github.com/clearmatics/autonity/consensus/istanbul/core.(*core).handleEvents()
[...]

PR #55, opened 2019-03-29
Restore travis CI

### LINKED ISSUE #54: [CI] Restore travis CI

JekaMas:
Minimal CI with tests on push event.
Due to data races, a few tests and code in Istanbul consensus (engine, core) were unstable. So they are fixed in the PR.

closes #54 #53


consensus/istanbul/core/handler.go
```diff
@@ -74,11 +74,9 @@ func (c *core) handleEvents() {
 	// Clear state
 	defer func() {
 		c.current = nil
-		c.handlerWg.Done()
+		<-c.handlerStopCh
```
yazzaoui: What was the problem with the waitgroup ? Seems to me the same behaviour in practice

JekaMas: It's a data race to use wg in such way.

consensus/istanbul/core/core.go
```diff
@@ -75,11 +75,12 @@ type core struct {
...
 	backlogsMu *sync.Mutex
 
-	current   *roundState
-	handlerWg *sync.WaitGroup
+	current       *roundState
+	handlerStopCh chan struct{}
 
 	roundChangeSet   *roundChangeSet
 	roundChangeTimer *time.Timer
+	roundChangeMutex sync.RWMutex
```
yazzaoui: maybe better to rename `roundChangeMutex` to `roundChangeTimerMu` for same naming convention

JekaMas: Will do.

consensus/istanbul/core/handler.go
yazzaoui: I see, thanks !

consensus/istanbul/core/core.go
JekaMas: Fixed

yazzaoui:
LGTM

consensus/istanbul/core/core.go
```diff
@@ -311,6 +312,9 @@ func (c *core) stopFuturePreprepareTimer() {
 
 func (c *core) stopTimer() {
 	c.stopFuturePreprepareTimer()
+
+	c.roundChangeTimerMu.RLock()
+	defer c.roundChangeTimerMu.RUnlock()
```
aneequesafdar: Why is it a RLock/RUnlock and not Lock/Unlock?

JekaMas: Here is next few last next of code
```
if c.roundChangeTimer != nil {
		c.roundChangeTimer.Stop()
}
```

So we only need to read the value and check if it's nil or not.
If we use .Lock() here, we get panic, because the Stop method calls .Lock() too

aneequesafdar: The reason I thought that a Lock/Unlock should be used is because `c.roundChangeTimer.Stop()` changes the `c.roundChangeTimer`. I looked at the Stop() (its a function of the go timer package) but I couldn't find `c.roundChangeTimerMu.Lock()` being called, so why would it panic?

JekaMas: Ok, .Rlock() blocks .Lock() call and pass other .Rlock()`s. So `consensus/istanbul/core/core.go:335` will be blocked by this Rlock() and we don't need to use Lock() once again here

PR #61, opened 2019-04-04
Provide an option to specify white list item ip+port not only in Genesis file

JekaMas:
## Reason:
In some of use cases `genesis.json` can be generated before we will know about ip addresses and ports of peers. This information is surplus and it requires to know in advance the network information.

## Typical user cases
1. Test or local networks with any kind of containerization (kubernates, docker...)
2. Cloud solution without static IPs
3. As Alex said: application/pod could be died in any time and be automatically respawned in another node with another ip/ports. It is an essential mechanism for most failover systems.
4. 2+ remote clusters with external IPs managed by a provider

## Description
At the moment we only check IDs from the white list, however, in Genesis we require a list of enodes with IPs and ports.

The idea behind the PR:
1. enodes in Genesis can be without IP:port
2. if we get from white list enodes without IP:port, we should check only for node.IDs
3. if we get full enode with IP:port, then we should check full enode address.
4. user can specify a list of static peers in the file `static-node.json`, so we can change this file as frequently as we want without the need to change Genesis, update Glienicke contract or redeploy entire network

## Further steps
The PR is the first step in managing ips and addresses:
1. add an option to drop IPs and ports from `genesis.json`
2. add `enode` name resolving for `static-node.json` - dns working out of box.
3. add default port `30303`
4. addresses like `enode://....@validator01.name` should be valid and resolved by dns name `validator01.name` and default port

I've rewritten the example application to show the way to work with Genesis+`static-nodes.json`


JekaMas:
After the discussion, we decided to
1. implement domain name resolution
2. not to use `static-nodes.json` until we got real user cases
3. set up default port as `30303`
4. document that using an external list of static nodes through the flag or `static-nodes.json` is removed from autonity

aiman:
Do we need to add any new tests for hostname resolution?

JekaMas:
> Do we need to add any new tests for hostname resolution?

@aiman Got your point and pushed 2 commits with tests.

core/rawdb/accessors_chain.go
```diff
@@ -60,22 +56,26 @@ func WriteEnodeWhitelist(db DatabaseWriter, whitelist []*enode.Node) {
...
-		whitelist = append(whitelist, enode.MustParseV4(node))
+		enode, err := enode.ParseV4WithResolve(node)
+		if err != nil {
+			panic(err)
+		}
+		whitelist = append(whitelist, enode)
 	}
 
-	return whitelist
+	return whitelist, whitelistStorage
```
yazzaoui: Could you explain to me why you want to return the string list, is that because you are considering the case of enodes that failed to be resolved?

core/genesis.go
```diff
@@ -288,16 +291,17 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
...
+		newEnode, err := enode.ParseV4WithResolve(enodeString)
 		if err != nil {
+			log.Error("enode parse error", "err", err, "enode", enodeString)
 			return nil, errGenesisBadWhitelist
 		}
+		log.Info("Genesis Authorized Enode", "enode", enodeString, "parsed", newEnode)
 		enodeWhiteList = append(enodeWhiteList, newEnode)
 	}
 
 	if len(enodeWhiteList) != 0 {
```
yazzaoui: not really in the scope of the PR but I think we should panic if `len(enodeWhiteList) == 0`, what do you think?

p2p/enode/urlv4.go
```diff
@@ -118,13 +129,34 @@ func parseComplete(rawurl string) (*Node, error) {
...
 	}
+
 	if ip = net.ParseIP(host); ip == nil {
-		return nil, errors.New("invalid IP address")
+		if !resolve {
+			return nil, errors.New("invalid IP address", )
+		}
+		// if host is not IPV4/6, resolve host is a domain
+
+		hostName, err := net.LookupHost(host)
```
yazzaoui: why not using  `net.LookupIP` directly?

core/genesis.go
JekaMas: It's not clear why we would want to panic instead of normal error handling. If a user wants to panic using genesis.Commit(), they will use `MustCommit()`. If we add panic into `Commit` method, it'll broke API: Commit - returns error; MustCommit - throws panic.

core/rawdb/accessors_chain.go
JekaMas: The siting representation is used for `rawdb.WriteEnodeWhitelist` method, so the parameter was added to not to run `for _, enode := range whitelist { enode.ID().String()}` before `rawdb.WriteEnodeWhitelist` and to have a little bit cleaner code in `BlockChain.updateEnodesWhitelist()`

p2p/enode/urlv4.go
JekaMas: A good idea. The PR has been updated with new code.

core/genesis.go
yazzaoui: Yes, right, but what do you think makes technically more sense if the whitelist is empty: 
1 - Keep the code like it is, meaning the whitelist can potentially be empty and leading, meaning no one can connect the network. 
2 -  Return an error here and do something in `eth/backend.go` to panic 
3 - Switch to open network mode

JekaMas: It can't be empty - we return the error. And than it's on caller to define the behaviour according to received error.
If someone wants to forbide the error and handle panic, they can call MustCommit.

core/rawdb/accessors_chain.go
yazzaoui: I see, but now the code is becoming confusing in my opinion having to deal with two returned whitelists for a very little gain in efficiency. I would prefer really to stick with one or the other..

core/genesis.go
yazzaoui: Ok for a tempory fix I suggest then:
-  Return error if `len(enodeWhiteList) == 0`
- in `eth/backend.go New()`  check for empty whitelist error and forbid error if not open-network

core/rawdb/accessors_chain.go
JekaMas: Ok, maybe we could use a new type then?

p2p/server.go
```diff
@@ -357,6 +357,20 @@ func (src *Server) UpdateWhitelist(enodes []*enode.Node) {
...
 				found = true
+			}
+
+			// if IP:Port are specified in a white list, we check them
+			if err := whitelistedEnode.ValidateComplete(); err == nil {
+				if !connectedPeer.Node().IP().Equal(whitelistedEnode.IP()) {
+					found = false
+				}
+
+				if connectedPeer.Node().UDP() != whitelistedEnode.UDP() {
```
yazzaoui: Not sure if we should perform check on UDP/TCP listening ports, they will change depending on inbound or outbound connection.

core/genesis.go
yazzaoui: @aiman @cmjc

JekaMas: Agree. Anyway with `types.Nodes` we can add this check just in single place.

p2p/server.go
JekaMas: OK

core/network_permissioning.go
```diff
@@ -165,19 +163,10 @@ func (bc *BlockChain) callGlienickeContract(state *state.StateDB, header *types.
...
-	for i := range returnedEnodes {
-		newEnode, err := enode.ParseV4(returnedEnodes[i])
-		if err == nil {
-			enodes = append(enodes, newEnode)
-		} else {
-			log.Error("Invalid whitelisted enode", "returned enode", returnedEnodes[i], "error", err.Error())
-		}
-	}
-	return enodes, nil
+	return types.NewNodes(returnedEnodes, false), nil
```
yazzaoui: Do you know if there is a local caching mechanism in the net library for lookup requests? Now that I'm thinking this could be very inefficient to resolve all the enodes after every new block (each second in xBFT)

core/genesis.go
cmjc: @ya, @JekaMas , @aiman - 

* **should we support open network?** I think yes; this should be a toggle - e.g. a boolean flag. By default it is permissioned, if open you set the flag. If you want to interact with an open network, then as discussed with @ya yesterday pm, an interop scenario - an Ion interaction.

|Functional Area|Requirement|Description|
|-----|-----|-----|
|Identity / Access|Node permissioning - run network in whitelist or open mode|It SHOULD be possible to configure the network to run in an open mode where any node can join the network or in a permissioned mode where only whitelisted nodes can connect to the network. 'Open' is defined as no node-level permissioning - any node can join the network. The default setting should be whitelisted|


* **error handling of zero length whitelist?** If the mode is permissioned it's an error and the config is inconsistent, broken. It it's in an open mode, then the whitelist should be empty. I think the network shouldn't initiate and a bad genesis configuration error be thrown - dis/agree?

Can you clarify what you mean by _panic_ please?

Looking forward, product requirements:
* **error reporting requirements and alerting**. This has never been discussed except as 'monitoring and alerting'. I think of approaching this in terms of 'error reporting architecture' - i.e. a design or pattern for how and when we raise errors, error codes, error messages, and for raising alerts.

Do you have such a pattern? 
Do you agree this is useful or do you think this is too prescriptive? 

@ya - can we talk about this some time, please, get your thoughts on this.

Note: Florian and I have had a first meeting with Tudor to understand the scope of the network tools and data that DevOps are extracting;

JekaMas: @cmjc @yazzaoui I've added support of open network in very last commit. All checks are done in `types.Nodes` constructor.
However, I haven't added the check for non-empty white list for permissioned network. The code is really simple for `types.Nodes`:
```
if !openNetwork && len(strList) == 0 {
	panic("Invalid whitelist has been given. It should not be empty")
}
```
but fixing all the tests will take time. I think we could add the check in a new issue to keep the PR clean and simple.

core/network_permissioning.go
JekaMas: It doesn't cache requests, however, I've added the cache into `types.Nodes` type. Please, take a look.

JekaMas:
@yazzaoui @cmjc @aiman  Could you check last changes? it'd be great to get new comments or merge the PR.

yazzaoui:
sure @JekaMas I'm very busy with consensus stuff, will review as soon as I can, thank you for your understanding

JekaMas:
@yazzaoui The `openNetwork` parameter has been reverted from chainConfig. Could you take a look, when you have some time.

yazzaoui:
LGTM

p2p/enode/urlv4.go
```diff
@@ -118,13 +129,29 @@ func parseComplete(rawurl string) (*Node, error) {
...
+			return nil, errors.New("invalid IP address", )
+		}
+		// if host is not IPV4/6, resolve host is a domain
+
+		hostIPs, err := net.LookupIP(host)
+		if err != nil {
+			return nil, errors.New("invalid domain or IP address")
+		}
+		if len(hostIPs) > 0 {
+			ip = hostIPs[0]
```
aneequesafdar: What would happen if hostname resolved to the same ip address for multiple enodes (since we always take the first ip address)? is that ok?

PR #68, opened 2019-04-12
Provide validators as a list in genesis file

JekaMas:
Nowadays the only way to set up validators is encoding those IDs into `extraData` Genesis field.
DevOps could exclude a few operations from their pipeline if it'd possible to provide validators in human-readable format in Genesis.
The user documentation would be much simpler too. A user would need only to edit Genesis and run `autonity init path_to_genesis`.


core/genesis.go
```diff
@@ -260,6 +261,9 @@ func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
 	}
 	if g.Difficulty == nil {
 		head.Difficulty = params.GenesisDifficulty
+	}
+	if len(g.Validators) != 0 {
+
```
yazzaoui: Something is missing here

JekaMas: It was auto generated. Will check why go generate returned that code.

core/genesis.go
```diff
@@ -314,6 +341,20 @@ func (g *Genesis) MustCommit(db ethdb.Database) *types.Block {
...
 }
 
+func (g *Genesis) UpdateValidators(addressList []string) error {
+	var validators []common.Address
+	for _, address := range addressList {
+		validators = append(validators, common.HexToAddress(address))
+	}
+
+	var err error
+	if g.ExtraData, err = types.PrepareExtra(&g.ExtraData, validators); err != nil {
```
yazzaoui: Why is this function taking a pointer on a slice as an argument and not the slice directly?

JekaMas: The method uses append function that can change the underlying struct of slice. So if we want slice to be changed after the method call, we have to use it by reference.
A small example https://play.golang.org/p/Ex5cZhzPLhg

yazzaoui: Oh yes, sorry forgot that, thank you for the example !

core/genesis.go
JekaMas: I've checked that place. It'd a piece of dead code, the actual function was extracted into `setIstanbul`

core/network_permissioning.go
```diff
@@ -163,7 +163,7 @@ func (bc *BlockChain) callGlienickeContract(state *state.StateDB, header *types.
 	}
 
 	var returnedEnodes []string
-	if err := ABI.Unpack(returnedEnodes, "getWhitelist", ret); err != nil { // can't work with aliased types
+	if err := ABI.Unpack(&returnedEnodes, "getWhitelist", ret); err != nil { // can't work with aliased types
```
yazzaoui: Interesting, why this change?

yazzaoui:
LGTM

core/genesis.go
```diff
@@ -283,15 +297,44 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
...
+	if len(g.Validators) != 0 {
+		err := g.UpdateValidators(g.Validators)
+		if err != nil {
+			return fmt.Errorf("can't commit genesis block with incorrect validators: %s", err)
+		}
+	}
+	log.Info("starting Istanbul consensus", "extraData", common.Bytes2Hex(g.ExtraData))
+
+	// we have to use '1' to have TD == BlockNumber for xBFT consensus
+	g.Difficulty = big.NewInt(1)
```
aneequesafdar: Isn't the block number for genesis defined as 0? So why aren't we setting the difficulty to 1? Is it because difficulty 1 is inherited by block 1?

JekaMas: It is, however here we set `Difficulty` to 1, not the total difficulty. If we have `Difficulty==1` so for blockNumber==0 => TD=0, blockNumber==1 => TD=1, blockNumber==2 => TD=2 etc

IBFT relies on TD as equal to Block number, afaik.

PR #75, opened 2019-04-17
Remove old enode check

### LINKED ISSUE #74: Remove old enode check
A piece of code was restored after merge master

JekaMas:
An old piece of code was restored after last master merge - https://github.com/clearmatics/autonity/pull/61/files#diff-3e8301305b1dc6d3f96fbe3429cda929L284

@eastata mentioned to me that we need to not to panic if some of enodes won't resolve on `autonity init` step. So I changed `types.Nodes` to not to panic when we parse nodes to write whiteList to the DB, but to panics on errors if we're reading from the DB. @yazzaoui Could you review the recent changes?

closes #74

### DIFF
--- a/core/types/nodes.go
+++ b/core/types/nodes.go
@@ -25,10 +25,9 @@ func NewNodes(strList []string, openNetwork bool) *Nodes {
 			if !openNetwork {
 				panic(err)
 			}
-		} else {
-			n.List = append(n.List, newEnode)
-			n.StrList = append(n.StrList, enodeStr)
 		}
+		n.List = append(n.List, newEnode)
+		n.StrList = append(n.StrList, enodeStr)
 	}
 
 	return n
--- a/core/types/nodes.go
+++ b/core/types/nodes.go
@@ -1,10 +1,12 @@
 package types
 
 import (
+	"sync"
+	"sync/atomic"
+	"time"
+
 	"github.com/clearmatics/autonity/log"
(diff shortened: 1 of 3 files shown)


yazzaoui:
LGTM

core/types/nodes.go
```diff
@@ -1,50 +1,110 @@
...
-		newEnode, err := cache.Get(enodeStr, enode.ParseV4WithResolve)
-		if err != nil {
-			log.Error("Invalid whitelisted enode", "returned enode", enodeStr, "error", err.Error())
+		wg.Add(1)
 
-			if !openNetwork {
-				panic(err)
+		go func(enodeStr string) {
+			log.Error("performing", "node", enodeStr)
+			newEnode, err := cache.Get(enodeStr, getEnode)
```
yazzaoui: Why parallelizing here? Is cache access that slow?
Thinking in this case we might want as well to use a worker pool for dns resolve

yazzaoui:
LGTM

aneequesafdar:
The changes made in the pull request does not panic when whitelist is been read from the Database (as suggested) since the [panic](https://github.com/clearmatics/autonity/blob/9fffe9e239c97a6fe2f00586708eecefac8f58ef/core/types/nodes.go#L63) is in `types/nodes.go`'s `NewNodes()` and it has no context of from where it is been called from, rather `NewNodes` panics based on `openNetwork` parameter.

PR #88, opened 2019-04-26
Sort whitelist before contract deploy

### LINKED ISSUE #87: Sort whitelist before contract deploy
`backend.getValidators()` is called many times until Genesis is committed to the blockchain. And `genesis.Commit()` is called 2-3 times also.
So `DeployGlienickeContract` should be deployed each time with exactly the same code and data. In the other case, it can lead to inconsistent StateDb.

P.S. I've tried to go deeper to understand how this can happen that `genesis.Commit()` is called many times. However, after an hour, I realised that it can take the whole day or more to find the cause and anyway we're going to rewrite the consensus.

By far we'll see multiple `Deployed Glienicke Contract` in the logs.

JekaMas:
closes #87

### DIFF
--- a/core/network_permissioning.go
+++ b/core/network_permissioning.go
@@ -19,6 +19,7 @@ package core
 import (
 	"errors"
 	"math/big"
+	"sort"
 	"strings"
 
 	"github.com/clearmatics/autonity/accounts/abi"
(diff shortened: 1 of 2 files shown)


yazzaoui:
LGTM

PR #90, opened 2019-04-30
Tendermint package with proper steps names

### LINKED ISSUE #79: Create a new consensus package called Tendermint using IBFT implementation and rename component respectively
Rename IBFT references to Tendermint and changes the names of the stages:
- `Preprepare -> Proposal`.
- `Prepare -> Prevote`.
- `Commit -> Precommit`.
- `FinalCommited -> Commit`.

JekaMas:
closes #79


JekaMas:
@yazzaoui @aneequesafdar just a friendly reminder about the CR

aneequesafdar:
I think it might be on standby until the branching strategy is sorted.

JekaMas:
It's easy to change the target branch anytime. It won't change the PR's code at all.

consensus/tendermint/core/roundstate.go
```diff
@@ -27,13 +27,13 @@ import (
...
-// we need to keep a reference of preprepare in order to propose locked proposal when there is a lock and itself is the proposer
-func newRoundState(view *tendermint.View, validatorSet tendermint.ValidatorSet, lockedHash common.Hash, preprepare *tendermint.Preprepare, pendingRequest *tendermint.Request, hasBadProposal func(hash common.Hash) bool) *roundState {
+// lockedHash and proposal are for round change when lock exists,
+// we need to keep a reference of proposal in order to propose locked proposal when there is a lock and itself is the proposer
+func newRoundState(view *tendermint.View, validatorSet tendermint.ValidatorSet, lockedHash common.Hash, proposal *tendermint.Proposal, pendingRequest *tendermint.Request, hasBadProposal func(hash common.Hash) bool) *roundState {
 	return &roundState{
 		round:          view.Round,
 		sequence:       view.Sequence,
-		Preprepare:     preprepare,
+		proposal:       proposal,
```
aneequesafdar: What is the reason to make the proposal private?

consensus/tendermint/types.go
```diff
@@ -40,12 +40,12 @@ type Proposal interface {
 }
 
 type Request struct {
-	Proposal Proposal
+	Proposal ProposalBlock
```
aneequesafdar: Why not keep it as Proposal? with just Proposal Block we lose the View (also needs to be renamed/changed) associated with the request.

consensus/tendermint/core/roundstate.go
JekaMas: To nit to have a name conflict with Proposal method

consensus/tendermint/types.go
JekaMas: Because we now have a Proposal struct - that one with View.

consensus/tendermint/core/core.go
```diff
@@ -241,18 +241,18 @@ func (c *core) startNewRound(round *big.Int) {
...
 	if roundChange && c.isProposer() && c.current != nil {
 		// If it is locked, propose the old proposal
 		// If we have pending request, propose pending request
 		if c.current.IsHashLocked() {
 			r := &tendermint.Request{
-				Proposal: c.current.Proposal(), //c.current.Proposal would be the locked proposal by previous proposer, see updateRoundState
+				Proposal: c.current.Proposal().Proposal, //c.current.Proposal would be the locked proposal by previous proposer, see updateRoundState
 			}
-			c.sendPreprepare(r)
+			c.sendProposal(r)
```
aneequesafdar: With regards to this [comment](https://github.com/clearmatics/autonity/pull/90#discussion_r281587048) we can see that sending a proposal takes in a Request which now only has ProposalBlock. Thus when a proposal is received we lose the information about when the block was proposed and we need View to say which height and sequence the block was first proposed.

consensus/tendermint/core/roundstate.go
```diff
@@ -85,23 +85,23 @@ func (s *roundState) Subject() *tendermint.Subject {
...
+func (s *roundState) SetProposal(proposal *tendermint.Proposal) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	s.Preprepare = preprepare
+	s.proposal = proposal
 }
 
-func (s *roundState) Proposal() tendermint.Proposal {
+func (s *roundState) Proposal() *tendermint.Proposal {
```
aneequesafdar: Why did you make the Proposal a pointer?

consensus/tendermint/core/types.go
```diff
@@ -33,16 +33,16 @@ type State uint64
 
 const (
 	StateAcceptRequest State = iota
-	StatePreprepared
+	StateProposald
```
aneequesafdar: I think we need a better name for this.

consensus/tendermint/types.go
aneequesafdar: You are right, I became very confused by the names, too many proposals with different meanings.

consensus/tendermint/core/core.go
aneequesafdar: Please ignore, as above.

consensus/tendermint/core/types.go
JekaMas: You're right. We need a better naming. Will try to find it.

miner/worker.go
```diff
@@ -267,10 +267,10 @@ func (w *worker) pendingBlock() *types.Block {
 // start sets the running status as 1 and triggers new work submitting.
 func (w *worker) start() {
 	atomic.StoreInt32(&w.running, 1)
-	if istanbul, ok := w.engine.(consensus.Istanbul); ok {
+	if istanbul, ok := w.engine.(consensus.PoS); ok {
```
aneequesafdar: This should also be renamed to `pos`

consensus/tendermint/backend/engine_test.go
```diff
@@ -141,7 +141,7 @@ func makeBlock(chain *core.BlockChain, engine *backend, parent *types.Block) (*t
 
 func makeBlockWithoutSeal(chain *core.BlockChain, engine *backend, parent *types.Block) (*types.Block, error) {
 	header := makeHeader(parent, engine.config)
-	engine.Prevote(chain, header)
+	engine.Prepare(chain, header)
```
aneequesafdar: Seems to have been reverted

consensus/tendermint/backend/engine_test.go
```diff
@@ -160,12 +160,12 @@ func makeBlockWithoutSeal(chain *core.BlockChain, engine *backend, parent *types
 func TestPrevote(t *testing.T) {
 	chain, engine := newBlockChain(1)
 	header := makeHeader(chain.Genesis(), engine.config)
-	err := engine.Prevote(chain, header)
+	err := engine.Prepare(chain, header)
```
aneequesafdar: Seems to have been reverted

consensus/tendermint/backend/engine_test.go
```diff
@@ -160,12 +160,12 @@ func makeBlockWithoutSeal(chain *core.BlockChain, engine *backend, parent *types
...
 	if err != nil {
 		t.Errorf("error mismatch: have %v, want nil", err)
 	}
 	header.ParentHash = common.BytesToHash([]byte("1234567890"))
-	err = engine.Prevote(chain, header)
+	err = engine.Prepare(chain, header)
```
aneequesafdar: Seems to have been reverted

consensus/tendermint/backend/engine.go
```diff
@@ -539,10 +539,10 @@ func (sb *backend) retrieveValidators(header *types.Header, parents []*types.Hea
 
 	if len(parents) > 0 {
 		parent := parents[len(parents)-1]
-		var istanbulExtra *types.IstanbulExtra
-		istanbulExtra, err = types.ExtractIstanbulExtra(parent)
+		var tendermintExtra *types.IstanbulExtra
+		tendermintExtra, err = types.ExtractIstanbulExtra(parent)
```
aneequesafdar: Might be a good idea to change `ExtractIstanbulExtra` to `ExtractPOSExtra`

miner/worker.go
JekaMas: Do you mean 'POS'?

consensus/tendermint/backend/engine_test.go
JekaMas: see above

consensus/tendermint/backend/engine.go
JekaMas: A good idea

PR #96, opened 2019-05-12
changed UpdateValidators() to SetPoSExtraData()

aneequesafdar:
Name conflict with `updateValidators()` [here](https://github.com/clearmatics/autonity/blob/ba458153663289b0803fb967fa449923489ea985/cmd/autonity/chaincmd.go#L490)

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -303,7 +303,7 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
 //setPoS sets default PoS(IBFT or Tendermint) config values
 func (g *Genesis) setPoS() error {
 	if len(g.Validators) != 0 {
-		err := g.UpdateValidators(g.Validators)
+		err := g.SetPoSExtraData(g.Validators)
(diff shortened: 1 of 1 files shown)


JekaMas:
It's not very obvious what 'SetPoSExtraData()' stands for. Could we find a better name for it?
Maybe 'SetValidatorsExtraData" or something

aneequesafdar:
Based on the spec it is implied that extradata contains the validator set and since we change the reference from Istanbul to PoS, I felt it was more appropriate to change it to the above, how about `setPoSValidatorsExtraData()`?

JekaMas:
'setPoSValidatorsExtraData()' looks more explicit. Could you update the PR?

JekaMas:
By the way, why this is a name conflict? The autonoty source compiles well, one name is used as method, another one as a function in different context.

aneequesafdar:
Yes, you are right, there is no name conflict in the scope. I felt we could have used a better name since `updatedateValidator` was being used for cli command.

PR #98, opened 2019-05-13
Respect config values from genesis

### LINKED ISSUE #95: Changing config consensus to tendermint doesn't affect anything
It looks like that the default config is always used and user-defined config is ignored.

So `HelloWorld` application doesn't work with Tendermint consensus

JekaMas:
closes #95


aneequesafdar:
Can we please not push the helloworld changes for testing, please remove them. Otherwise, LGTM

eth/config.go
```diff
@@ -61,8 +61,6 @@ var DefaultConfig = Config{
 		Blocks:     20,
 		Percentile: 60,
 	},
-	Istanbul:    *istanbul.DefaultConfig,
```
aneequesafdar: `Istanbul istanbul.Config` `Tendermint tendermint.Config` should also be removed from the `eth.Config` definition in `eth/config.go`

eth/config.go
```diff
@@ -61,8 +61,6 @@ var DefaultConfig = Config{
...
-	OpenNetwork: false,
```
aneequesafdar: Why should this also be removed?

JekaMas: I've just thought, why not remove useless values from the config.

cmd/autonity/chaincmd.go
```diff
@@ -219,6 +224,59 @@ func initGenesis(ctx *cli.Context) error {
...
+		}
+		if len(genesis.Config.Tendermint.Deployer) == 0 || genesis.Config.Tendermint.Deployer == (common.Address{}) {
+			genesis.Config.Tendermint.Deployer = tendermint.DefaultConfig.Deployer
+		}
+		if genesis.Config.Tendermint.Epoch == 0 {
+			genesis.Config.Tendermint.Epoch = tendermint.DefaultConfig.Epoch
+		}
+	}
+}
+
```
aneequesafdar: Just an idea, instead of checking again their empty values would it be better to check that the fields are not default values? It could be the case that user-defined `Epoch = 0 and BlockPeriod = 0` but then they are set to default values, is this something we want?

eth/config.go
aneequesafdar: ok it should be fine since the default value for boolean is false

cmd/autonity/chaincmd.go
JekaMas: `Epoch = 0` and `BlockPeriod = 0` are not very meaningful if the one wants to have an epoch each block, he or she will use `Epoch = 1`.
The solution could be to use reference types `*int` instead of `int`.

PR #101, opened 2019-05-14
Test Network and add missing changes for #79

aneequesafdar:
Also gives another approach to #95


JekaMas:
So we need more flags for ABI, deployer, bytecode params? Also it rise question, why we want user to specify this by flags and not to use Genesis file with config inside it?

aneequesafdar:
Ok, I felt that we should at least start off from exactly where current IBFT is, so I have added the flags for consistency. 

I spent time trying to understand how all of the config work in autonity, here is a small breakdown:
```go
AutonityConfig
   |--> Config (eth.Config) 
             |--> Genesis
                      |--> ChainConfig
                                |--> EthashConfig
                                |--> CliqueConfig
                                |--> IstanbulConfig
                                |--> TendermintConfig
}
```
If you also notice that there is `Tendermint.Config` and `Istanbul.Config`, which is different from `IstanbulConfig` and `TendermintConfig` defined in `params/config.go`. The purpose of this is to allow for runtime config variability in some consensus parameters, such as the period between blocks, how large are timeouts for the rounds, etc.

If we don't have `Tendermint.Config` and `Istanbul.Config` and their respective flags we lose the ability to configure these value for an existing network. So for things like ABI, deployer and bytecode, we will not have the flags (which we know shouldn't be modified at runtime) and thus the `Tendermint.Config` and `ChainConfig.TendermintConfig` will now be different (which is not the case currently).

JekaMas:
I think we should start a separate conversation about the question do we need to let users change consensus params in runtime or it should be specified in the genesis and being charged by governance (maybe by smart contract).

And I'm not sure why do we need flags for some params and don't need for the rest.
The problem with flags is that a user should specify them for each start. Also the problem with genesis config is still present in the PR.

aneequesafdar:
Ok, let's leave the discussion about flags and what parameters can be changed at runtime for another day.

You are right that this doesn't solve the genesis config issue, I had misunderstood the problem you were solving.

PR #105, opened 2019-05-15
add docker local_test_network

aneequesafdar:


JekaMas:
I think we need https://github.com/clearmatics/autonity/pull/107 in case of slow travisCI
@aneequesafdar

aneequesafdar:
I also need to merge from tendermint branch

PR #124, opened 2019-06-04
[WIP] Fixes for the tendermint implementation

JekaMas:
fixes https://github.com/clearmatics/autonity/pull/116

### DIFF
--- a/consensus/tendermint/types.go
+++ b/consensus/tendermint/types.go
@@ -70,7 +70,11 @@ func (sub *Vote) EncodeRLP(w io.Writer) error {
 
 // DecodeRLP implements rlp.Decoder, and load the consensus fields from a RLP stream.
 func (sub *Vote) DecodeRLP(s *rlp.Stream) error {
-	var subject = new(Vote)
+	var subject struct {
+		Round             *big.Int
+		Height            *big.Int
+		ProposedBlockHash common.Hash
+	}
 
 	if err := s.Decode(&subject); err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


helloworld/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     volumes:
       - ./nodekey1:/nodekey:ro
       - ./keystore:/keystore:ro
-      - ./genesis-ibft.json:/genesis.json:ro
+      - ./genesis-tendermint.json:/genesis.json:ro
```
aneequesafdar: Can we please not change the helloworld dir, unless there is a specific reason to.

consensus/tendermint/types.go
```diff
@@ -70,7 +70,11 @@ func (sub *Vote) EncodeRLP(w io.Writer) error {
 
 // DecodeRLP implements rlp.Decoder, and load the consensus fields from a RLP stream.
 func (sub *Vote) DecodeRLP(s *rlp.Stream) error {
-	var subject = new(Vote)
+	var subject struct {
+		Round             *big.Int
+		Height            *big.Int
+		ProposedBlockHash common.Hash
+	}
```
aneequesafdar: How does the above change fix an infinite loop? This error will also be present in other DecodeRLP methods

aneequesafdar: I can update the rest

JekaMas: Nope. It introduces a new type without DecodeRLP.
Without it the DecodeRLP will be called until stack will be overflowed.

helloworld/docker-compose.yml
JekaMas: Ok, it was just for debug.

PR #134, opened 2019-06-17
Tendermint debug

aneequesafdar:


consensus/tendermint/core/core.go
```diff
@@ -222,6 +222,10 @@ func (c *core) commit() {
 
 // startRound starts a new round. if round equals to 0, it means to starts a new height
 func (c *core) startRound(round *big.Int) {
+	if r := recover(); r != nil {
+		c.logger.Error("panic in core.startRound", "panic", r)
+	}
+
```
aneequesafdar: Is this the correct way to call recover, it says [here](https://golang.org/ref/spec#Handling_panics) that 
```
The return value of recover is nil if any of the following conditions holds:

- panic's argument was nil;
- the goroutine is not panicking;
- recover was not called directly by a deferred function.
```
The last condition says that `recover()` is always nil if called outside of a deferred function. Here, it is not being called directly by a deferred function, so is it still working?

aneequesafdar: I think it should have been:
```go
defer func() {
		if r := recover(); r != nil {
			c.logger.Error("panic in core.handleEvents", "panic", r)
		}
}()
```

JekaMas: Yeap you're right. I've forgotten to add defer.

JekaMas:
@aneequesafdar It works?

JekaMas:
Yeap, it works!

consensus/tendermint/core/core.go
```diff
@@ -131,8 +133,9 @@ type core struct {
 
 	currentHeightRoundsStates map[int64]*roundState
 
-	latestPendingUnminedBlock *types.Block
-	firstUnminedBlockCh       chan struct{}
+	latestPendingUnminedBlock   *types.Block
```
JekaMas: Could you create an issue to simplify this part of the code?

consensus/tendermint/core/prevote.go
```diff
@@ -11,14 +11,18 @@ func (c *core) sendPrevote(isNil bool) {
...
+		Round:  big.NewInt(c.currentRoundState.Round().Int64()),
 		Height: big.NewInt(c.currentRoundState.Height().Int64()),
 	}
 
 	if isNil {
 		prevote.ProposedBlockHash = common.Hash{}
 	} else {
-		prevote.ProposedBlockHash = c.currentRoundState.Proposal().ProposalBlock.Hash()
+		if h := c.currentRoundState.GetCurrentProposalHash(); h == (common.Hash{}) {
+			c.logger.Error("sendPrecommit Proposal is empty! It should not be empty!")
```
JekaMas: sendPrecommit -> sendVote ?

consensus/tendermint/core/core.go
aneequesafdar: OK, I will create an issue

consensus/tendermint/core/prevote.go
aneequesafdar: So the issue here was that `c.currentRoundState.Proposal().ProposalBlock.Hash()` was giving nil point panic, and this should never be empty because if it is empty then `isNil` would be true and we should send prevote/precommit nil

PR #149, opened 2019-06-26
Fix unmined block data races

JekaMas:
closes https://github.com/clearmatics/autonity/issues/146
related to https://github.com/clearmatics/autonity/issues/141


aneequesafdar:
LGTM

Issue #141, opened 2019-06-25
Tendermint consensus data races


JekaMas:
Can be reproduced by `go test -race ./consensus/tendermint/... -v`
- [x] Change currentRoundState https://github.com/clearmatics/autonity/issues/142
- [x] Update currentRoundState https://github.com/clearmatics/autonity/issues/143
- [x] Update validators set https://github.com/clearmatics/autonity/issues/144
- [x] Update default validator https://github.com/clearmatics/autonity/issues/145
- [x] Unmined block https://github.com/clearmatics/autonity/issues/146

JekaMas:
done

Issue #142, opened 2019-06-25
CurrentRoundState data race


JekaMas:
```
==================
WARNING: DATA RACE
Read at 0x00c014845ee8 by goroutine 102:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).checkUnminedBlockMsg()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/unmined_block.go:43 +0x6a
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleUnminedBlock()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/unmined_block.go:11 +0x60
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleEvents()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:81 +0xc10

Previous write at 0x00c014845ee8 by goroutine 103:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).startRound()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/core.go:280 +0x95d

Goroutine 102 (running) created at:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:17 +0x5e
  github.com/clearmatics/autonity/consensus/tendermint/backend.(*Backend).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/backend/engine.go:483 +0x1ab
  github.com/clearmatics/autonity/consensus/tendermint/backend.newBlockChain()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/backend/backend_test.go:275 +0x4a6
  github.com/clearmatics/autonity/consensus/tendermint/backend.TestGetProposer()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/backend/backend_test.go:188 +0x4f
  testing.tRunner()
      /snap/go/current/src/testing/testing.go:865 +0x163

Goroutine 103 (running) created at:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:20 +0x9c
[...]

PR #150, opened 2019-06-27
Fix roundchange data races

### LINKED ISSUE #136: Refactor `latestPendingUnminedBlock` in tendermint/core.go
`latestPendingUnminedBlock` keeps track of the potential future proposal, we want `latestPendingUnminedBlock` to be the current height and most updated at the time of proposal so that we have latest transactions in the block. There is an event `NewUnmineBlockEvent` which signals to the consensus layer that a new "unmined" block is available and this event's handler updates the `core.latestPendingUnminedBlock`.

If the `c.latestPendingUnminedBlock == nil` at startup or `latestPendingUnminedBlock` height is not the same then start round will wait for a `NewUnminedBlockEvent`.

The current implementation uses a channel and boolean values to coordinate, however, it is not ideal. So we need to refactor how `c.latestPendingUnminedBlock` is updated.

aneequesafdar:
closes rest of the data races: #142 #143 
Also closes #136

### DIFF
--- a/consensus/tendermint/core/core.go
+++ b/consensus/tendermint/core/core.go
@@ -83,7 +83,6 @@ func New(backend tendermint.Backend, config *tendermint.Config) Engine {
 		backend:              backend,
 		backlogs:             make(map[tendermint.Validator]*prque.Prque),
 		pendingUnminedBlocks: prque.New(),
-		unminedBlockCh:       make(chan struct{}),
 		proposeTimeout:       new(timeout),
 		prevoteTimeout:       new(timeout),
 		precommitTimeout:     new(timeout),
@@ -127,11 +126,11 @@ type core struct {
 	lockedValue *types.Block
 	validValue  *types.Block
 
+	// TODO: change type to map[int64]roundState
 	currentHeightRoundsStates map[int64]*roundState
 
 	latestPendingUnminedBlock   *types.Block
 	latestPendingUnminedBlockMu sync.RWMutex
-	unminedBlockCh              chan struct{}
 
 	proposeTimeout   *timeout
 	prevoteTimeout   *timeout
@@ -266,6 +265,7 @@ func (c *core) startRound(round *big.Int) {
 	// Update current round state and the reference to c.currentHeightRoundsState
 	// We only add old round prevote messages to c.currentHeightRoundState, while future messages are sent to the backlog
 	// Which are processed when the step is set to StepAcceptProposal
+	// TODO: Update round state instead of creating new currentRoundState
 	c.currentRoundState = newRoundState(round, height, c.backend.HasBadProposal)
 	c.currentHeightRoundsStates[round.Int64()] = c.currentRoundState
 	c.sentProposal = false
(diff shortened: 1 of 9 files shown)


consensus/tendermint/core/core.go
```diff
@@ -284,12 +277,22 @@ func (c *core) startRound(round *big.Int) {
...
 
+	// Wait for c.handleNewUnminedBlockEvent() go routine to update c.latestPendingUnminedBlock
 	// Only wait for new unmined block if latestPendingUnminedBlock is nil or fo previous height
-	c.latestPendingUnminedBlockMu.RLock()
-	isMined := c.latestPendingUnminedBlock == nil || c.latestPendingUnminedBlock.Number() != c.currentRoundState.Height()
-	c.latestPendingUnminedBlockMu.RUnlock()
-	if isMined {
-		<-c.unminedBlockCh
+	if c.checkLatestPendingUnminedBlock() {
+		ticker := time.NewTicker(100 * time.Millisecond)
```
JekaMas: May its better to subscribe to 'newBlockEvent'(it's not a precise name)? Timers are ok for timeouts, gc, networking, retries.

consensus/tendermint/core/handler.go
```diff
@@ -11,90 +11,118 @@ import (
...
+		if r := recover(); r != nil {
+			debug.PrintStack()
+
+			c.logger.Crit("panic in core.handleNewUnminedBlockEvent", "panic", r)
+		}
+	}()
+
+	for e := range c.newUnminedBlockEventSub.Chan() {
+		// TODO: check is probably unnecessary
+		if newUnminedBlockEvent, ok := e.Data.(tendermint.NewUnminedBlockEvent); ok {
```
JekaMas: Yeap, we can remove the check

consensus/tendermint/core/core.go
aneequesafdar: Ok I will think about it a bit more

consensus/tendermint/core/handler.go
aneequesafdar: Actually, I just realised that the casting is required. I will remove the todo.

JekaMas: The casting is required, however the check is not.

consensus/tendermint/core/core.go
aneequesafdar: I don't want to create another message event for `NewUnminedBlockEvent`, channels should be sufficient, so its either this or the previous solution.

 The general idea is to have 2 goroutines, where one is always updating `latestUnminePendingBlock`, while the other is handling consensus event sequentially and anytime `latestUnminePendingBlock` is nil or of different height, we want to block and wait for the other goroutine. Which is why we need both locks and channels, lock to make sure `latestUnminePendingBlock` correctly set when the condition (`c.latestPendingUnminedBlock == nil || c.latestPendingUnminedBlock.Number().Int64() != c.currentRoundState.Height().Int64()`) is false, and channel to synchronise when the condition is true. In order to get rid of the channel, I set a ticker to check every 100 * Millisecond, which I know is not ideal because it could be set in between 100 ms.

The other problem is that passing the new unmined block through the channel is not useful because then we will be setting it twice, so passing it through the channel will mean that we will block one go routine and it will remain blocked until `startRound()` is called in the other goroutine. We can think about using buffered channels but then we will have to manage the priority which is already being done by `storeUnminedBlockMsg` and `processPendingUnminedBlock`. So, I cannot think of another way without changing the consensus architecture for handling events (which IMO is unnecessary for #136).

consensus/tendermint/core/handler.go
aneequesafdar: I see, I will update

consensus/tendermint/core/core.go
JekaMas: Ok, lets live with this solution for some time. I still think that this should be done without timers, flags, mutexes but only with one channel. However, we can wait for a while and think about it later.

JekaMas:
Are tests with 'race' flag green?

aneequesafdar:
Yes, but its not mining properly but that is another issue to fix

JekaMas:
My congratulations! now you have the first experience working with data races in golang. It's an important step.

aneequesafdar:
also closes #141

PR #153, opened 2019-06-28
Fix block mining

aneequesafdar:


JekaMas:
@aneequesafdar 
What should be reviewed I see only additional logs?

aneequesafdar:
Yeah, it's incomplete, I just created a pull request for future reference, that's why I didn't request a review yet.

Issue #156, opened 2019-07-01
Round doesn't change when proposer is down


aneequesafdar:
When we run `local_test_network` with 4 nodes only where the proposer is not started than consensus doesn't move to new round.

Here is a log from one of the node:

```
DEBUG[07-01|14:23:52.215|consensus/tendermint/core/core.go:280]       Starting new Round                       Height=1 Round=0
INFO [07-01|14:23:52.215|consensus/tendermint/backend/engine.go:352]  Soma Contract Deployer                   Address=0x0000000000000000000000000000000000000000
INFO [07-01|14:23:52.219|consensus/tendermint/backend/contract.go:72] Deployed Soma Governance Contract        Address=0xBd770416a3345F91E4B34576cb804a576fa48EB1
DEBUG[07-01|14:23:52.220|p2p/server.go:786]                           Adding p2p peer                          name=Autonity/v1.0.0-alph... addr=172.25.0.14:30303 peers=1
DEBUG[07-01|14:23:52.221|eth/handler.go:310]                          Ethereum peer connected                  id=5b68cfe243a23a70 conn=staticdial name=Autonity/v1.0.0-alpha-63b26399/linux-amd64/go1.11.11
DEBUG[07-01|14:23:52.221|core/types/nodes.go:39]                      performing                               node=enode://1f207dfb3bcbbd338fbc991ec13e40d204b58fe7275cea48cfeb53c2c24e1071e1b4ef2959325fe48a5893de8ff37c73a24a412f367e505e5dec832813da546a@172.25.0.12:30303
DEBUG[07-01|14:23:52.222|core/types/nodes.go:39]                      performing                               node=enode://3ce6c053cb563bfd94f4e0e248510a07ccee1bc836c9784da1816dba4b10564e7be1ba42e0bd8d73c8f6274f8e9878dc13814adb381c823264265c06048b4b59@172.25.0.15:30303
DEBUG[07-01|14:23:52.222|core/types/nodes.go:39]                      performing                               node=enode://d73b857969c86415c0c000371bcebd9ed3cca6c376032b3f65e58e9e2b79276fbc6f59eb1e22fcd6356ab95f42a666f70afd4985933bd8f3e05beb1a2bf8fdde@172.25.0.11:30303
[...]

PR #158, opened 2019-07-01
Fix round-change

aneequesafdar:


aneequesafdar:
renamed branch to `fix-round-change-#156` and new pull request is #159

PR #159, opened 2019-07-01
Fix round change

aneequesafdar:


JekaMas:
@aneequesafdar Are you sure that we don't need to have a timeout for the waiting a proposer? AFAIK we stuck if any proposer is down and it's the round when we're waiting it's proposed block.

aneequesafdar:
I am not sure what you mean, do you mean to have for a proposer who sends the proposal and is waiting for prevotes?

aneequesafdar:
please refer to `fix-round-change-#156-debug` for debug messages

local_test_network/docker-compose.yml
```diff
@@ -76,25 +76,6 @@ services:
     networks:
       chainnet:
         ipv4_address: 172.25.0.14
-  node5:
```
JekaMas: Why not keep a cluster of 5 nodes?

aneequesafdar: Not a good reason but the simplest case is for 4 nodes. Would you like to revert?

aneequesafdar: Also, on slower machines messages are passed around more quickly, I can do that locally and not push.

aneequesafdar: reverted

PR #173, opened 2019-07-04
Add Tmux support for local cluster

aneequesafdar:
This is taken from https://github.com/clearmatics/local-autonity-deployment-utility


aneequesafdar:
Cannot get it to work inside Goland so closing the pull request, please refer to  https://github.com/clearmatics/local-autonity-deployment-utility for future releases

PR #176, opened 2019-07-04
Stop tendermint consensus properly

### LINKED ISSUE #175: Stop tendermint consensus properly
We want to stop the consensus properly

JekaMas:
closes #175 

However, it only can help not sending messages after Stop command. It cannot and should do nothing with sync, because the db state changes only after backend.commit() call and all other messages and message handling do nothing with the local state (until we don't have WAL)

### DIFF
--- a/consensus/istanbul/backend.go
+++ b/consensus/istanbul/backend.go
@@ -17,6 +17,7 @@
 package istanbul
 
 import (
+	"context"
 	"math/big"
 	"time"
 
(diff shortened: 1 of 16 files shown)


aneequesafdar:
Any particular reason to include the context in Istanbul? the backend.go interface is not shared between Istanbul and Tendermint. Each has its own.

JekaMas:
@aneequesafdar My bad. I've reverted Istanbul changes.

JekaMas:
The nodes now stop properly also if resume a node start sync and validation with new errors though. Hovewer, the issue is seems solved.

Issue #172, opened 2019-07-03
A validator doesn't sync the state after restart


JekaMas:
The logs after stop, wait for a minute or so and start again:

```
Autonity INIT 172.25.0.14
INFO [07-03|17:42:19.120] Maximum peer count                       ETH=25 LES=0 total=25
INFO [07-03|17:42:19.121] Allocated cache and file handles         database=/autonity-data/autonity/chaindata cache=16 handles=16
Fatal: Failed to write genesis block: database already contains an incompatible genesis block (have 13f5475117722c48, new eec57094325cd8f6)
INFO [07-03|17:42:19.290] Persisted trie from memory database      nodes=11 size=1.67kB time=54.731µs gcnodes=0 gcsize=0.00B gctime=0s livenodes=1 livesize=0.00B
Fatal: Failed to write genesis block: database already contains an incompatible genesis block (have 13f5475117722c48, new eec57094325cd8f6)
Autonity START
DEBUG[07-03|17:42:19.391|autonity/main.go:237] Sanitizing Go's GC trigger               percent=100
INFO [07-03|17:42:19.393|cmd/utils/flags.go:968] Maximum peer count                       ETH=25 LES=0 total=25
DEBUG[07-03|17:42:19.393|accounts/keystore/file_cache.go:87] FS scan times                            list=80.596µs set=24.951µs diff=7.393µs
INFO [07-03|17:42:19.398|node/node.go:166]                   Starting peer-to-peer node               instance=Autonity/v1.0.0-alpha-9f8dad44/linux-amd64/go1.11.6
INFO [07-03|17:42:19.398|ethdb/database.go:73]               Allocated cache and file handles         database=/autonity-data/autonity/chaindata cache=512 handles=524288
INFO [07-03|17:42:19.508|eth/backend.go:142]                 Initialised chain configuration          config="{ChainID: 1991 Homestead: 0 DAO: <nil> DAOSupport: false EIP150: 0 EIP155: 0 EIP158: 0 Byzantium: <nil> Constantinople: <nil> Engine: tendermint}"
INFO [07-03|17:42:19.508|consensus/tendermint/backend/backend.go:59] Default Validator smart contract set 
INFO [07-03|17:42:19.508|eth/backend.go:178]                         Initialising Ethereum protocol           versions="[63 62]" network=1
WARN [07-03|17:42:19.560|core/blockchain.go:241]                     Head state missing, repairing chain      number=142 hash=e9142a…16c193
[...]

PR #180, opened 2019-07-09
Close autonity properly

### LINKED ISSUE #175: Stop tendermint consensus properly

aneequesafdar:
shares base with #176 and related to #172  and close #175

### DIFF
--- a/consensus/tendermint/core/precommit.go
+++ b/consensus/tendermint/core/precommit.go
@@ -1,13 +1,14 @@
 package core
 
 import (
+	"context"
 	"math/big"
 
 	"github.com/clearmatics/autonity/common"
 	"github.com/clearmatics/autonity/consensus/tendermint"
 )
 
-func (c *core) sendPrecommit(isNil bool) {
+func (c *core) sendPrecommit(ctx context.Context, isNil bool) {
 	logger := c.logger.New("step", c.currentRoundState.Step())
 
 	var precommit = &tendermint.Vote{
@@ -34,14 +35,14 @@ func (c *core) sendPrecommit(isNil bool) {
 	c.logPrecommitMessageEvent("MessageEvent(Precommit): Sent", precommit, c.address.String(), "broadcast")
 
 	c.sentPrecommit = true
-	c.broadcast(&message{
+	c.broadcast(ctx, &message{
 		Code: msgPrecommit,
 		Msg:  encodedVote,
 	})
 }
 
 // TODO: ensure to check the size of the committed seals as mentioned by Roberto in Correctness and Analysis of IBFT paper
-func (c *core) handlePrecommit(msg *message) error {
+func (c *core) handlePrecommit(ctx context.Context, msg *message) error {
 	var precommit *tendermint.Vote
 	err := msg.Decode(&precommit)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


consensus/tendermint/core/precommit.go
```diff
@@ -34,27 +35,27 @@ func (c *core) sendPrecommit(isNil bool) {
...
 	})
 }
 
 // TODO: ensure to check the size of the committed seals as mentioned by Roberto in Correctness and Analysis of IBFT paper
-func (c *core) handlePrecommit(msg *message) error {
-	var precommit *tendermint.Vote
-	err := msg.Decode(&precommit)
+func (c *core) handlePrecommit(ctx context.Context, msg *message) error {
+	var preCommit *tendermint.Vote
+	err := msg.Decode(&preCommit)
```
JekaMas: **preCommit type?

consensus/tendermint/core/prevote.go
```diff
@@ -34,79 +35,82 @@ func (c *core) sendPrevote(isNil bool) {
...
 		Msg:  encodedVote,
 	})
 }
 
-func (c *core) handlePrevote(msg *message) error {
-	var prevote *tendermint.Vote
-	err := msg.Decode(&prevote)
+func (c *core) handlePrevote(ctx context.Context, msg *message) error {
+	var preVote *tendermint.Vote
+	err := msg.Decode(&preVote)
```
JekaMas: **tendermint.Vote type?

consensus/tendermint/core/timeout.go
```diff
@@ -38,88 +47,87 @@ func (t *timeout) scheduleTimeout(stepTimeout time.Duration, round int64, height
...
+				return errNilPrecommitSent
+			case precommit:
+				return errMovedToNewRound
+			}
+		}
+	}
+	return nil
 }
 
+/////////////// On Timeout Functions ///////////////
```
JekaMas: Maybe we can extract this methods into a new file?

consensus/tendermint/core/timeout.go
```diff
@@ -38,88 +47,87 @@ func (t *timeout) scheduleTimeout(stepTimeout time.Duration, round int64, height
...
-		"msgHeight", timeout.heightWhenCalled,
-		"currentRound", c.currentRoundState.Round(),
-		"msgRound", timeout.roundWhenCalled,
-		"currentStep", c.currentRoundState.Step(),
-		"msgStep", timeout.step,
-	)
+func (t *timeout) stopTimer() error {
+	t.Lock()
+	defer t.Unlock()
+	if t.started {
```
JekaMas: Is it thread safe? Have you run tests with race flag?

consensus/tendermint/core/precommit.go
aneequesafdar: there was a name conflict with `precommit` in step.go so I just renamed it to `preCommit`, everything is same as before, it's just a rename

consensus/tendermint/core/prevote.go
aneequesafdar: Same as above

consensus/tendermint/core/timeout.go
aneequesafdar: Yep

consensus/tendermint/core/precommit.go
JekaMas: The renaming is ok. I worry is it ok to have double reference. It seems that decode should be called without '&'

consensus/tendermint/core/timeout.go
JekaMas: Nice!

consensus/tendermint/core/precommit.go
aneequesafdar: It's probably because msg.Decode is defined on message type, maybe we can try:
```
var preCommit tendermint.Vote
err := msg.Decode(preCommit)
```
Now that you mention, in this context, we don't need to use pointer values
```

Issue #120, opened 2019-05-29
Release first Tendermint implementation


JekaMas:
- [x] Fix the data races https://github.com/clearmatics/autonity/issues/141
- [x] Review first draft of Tendermint https://github.com/clearmatics/autonity/pull/116
- [x] Run local cluster with 1-10 machines with default config. Check that all nodes reach consensus and get recent state.
- [x] The same cluster, but start nodes with 1/2 RoundTimeout delay. Check that all nodes reach consensus and get recent state.
- [x] Run tests with the data race flag

JekaMas:
1. Dies with panic:
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0xffc77e, 0xe)
	/usr/local/go/src/runtime/panic.go:608 +0x72
runtime.newstack()
	/usr/local/go/src/runtime/stack.go:1008 +0x729
runtime.morestack()
	/usr/local/go/src/runtime/asm_amd64.s:429 +0x8f

goroutine 192 [running]:
type..hash.github.com/clearmatics/autonity/rlp.typekey(0xc0320003f0, 0xb1c3ade2, 0x0)
	<autogenerated>:1 +0x73 fp=0xc032000370 sp=0xc032000368 pc=0x59bdd3
runtime.mapaccess1(0xe900a0, 0xc0000aa2d0, 0xc0320003f0, 0x0)
	/usr/local/go/src/runtime/map.go:404 +0x71 fp=0xc0320003b8 sp=0xc032000370 pc=0x424a11
github.com/clearmatics/autonity/rlp.cachedTypeInfo(0x1266520, 0xecc0c0, 0x0, 0x0, 0x0, 0x0)
	/autonity/build/_workspace/src/github.com/clearmatics/autonity/rlp/typecache.go:61 +0xaf fp=0xc032000418 sp=0xc0320003b8 pc=0x5985df
github.com/clearmatics/autonity/rlp.(*Stream).Decode(0xc005856880, 0xe09220, 0xc053da6920, 0xecc0c0, 0xecc000)
	/autonity/build/_workspace/src/github.com/clearmatics/autonity/rlp/decode.go:805 +0x166 fp=0xc032000508 sp=0xc032000418 pc=0x5929d6
github.com/clearmatics/autonity/consensus/tendermint.(*Vote).DecodeRLP(0xc053dab410, 0xc005856880, 0xc053dab410, 0x7f5cc80262d8)
[...]

PR #183, opened 2019-07-10
After node restart genesis is different #182

JekaMas:
closes #182

Issue #108, opened 2019-05-15
Better Config Management


aneequesafdar:
Configurations in Autonity is complicated, in order to make sure intended config reach the correct level we need to better understand how they work, here is a brief breakdown:

```go
AutonityConfig
   |--> Config (eth.Config) 
             |--> Genesis
                      |--> ChainConfig
                                |--> EthashConfig
                                |--> CliqueConfig
                                |--> IstanbulConfig
                                |--> TendermintConfig
}
```
If you also notice that there is `Tendermint.Config` and `Istanbul.Config`, which is different from `IstanbulConfig` and `TendermintConfig` defined in `params/config.go`. The purpose of this is to allow for runtime config variability in some consensus parameters, such as the period between blocks, how large are timeouts for the rounds, etc.

If we don't have `Tendermint.Config` and `Istanbul.Config` and their respective flags we lose the ability to configure these value for an existing network. So for things like ABI, deployer and bytecode, we will not have the flags (which we know shouldn't be modified at runtime) and thus the `Tendermint.Config` and `ChainConfig.TendermintConfig` will now be different (which is not the case currently).

For the current implementation, [this code](https://github.com/clearmatics/autonity/blob/0ac6589dd5f90a00f1e69afb98c6b068fcc6b2bc/cmd/autonity/chaincmd.go#L227) needs to be refactored, and for the future flags and changes in configuration, the above considerations need to be made.

PR #185, opened 2019-07-10
Tendermint

aneequesafdar:
Renamed structs/files:
```
- preprepare -> propose
- prepare -> prevote
- commit -> precommit
- request -> unmined_block
- sequence -> height
```

Removed structs/files:
```
- all test files in core since the core logic changed
- errors.go
- types.go
- roundchnage.go
- events.go
- final_committed.go
```

New struct/files:
```
- timeout.go (in tendermint/core)
- Validators set (in tendermint/validator):
  - round_robin.go
  - round_robin_test.go
  - sticky.go
  - sticky_test.go
- validator_mock.go
```


aneequesafdar:
closes #120

shapeshed:
A couple of questions around this PR. 

1. Given some core files are changed does this affect the Istanbul implementation? We need to support that until we reach stability.
2. Have we applied the versioning scheme?

JekaMas:
1. In IBFT only renaming according to linters was done. The consensus and API are the same and untouched.
2. It's a good question. Not yet and I think we have some issues before applying versioning scheme. If we want to start from 1.5 version than we should give backwards capability guarantees. Hovewer, it's hard with the very first Tendermint implementation. We can that rise major version each time we break capability. I think it's nothing wrong having versions like 48.3.1. Or we can hold pre-release versioning like 0.n.m and rise only minor version in case of breaking capability.

Shirikatsu:
Exciting stuff guys. Quick question regarding the stuff in the `contracts` directory. Before we have the work around the Autonity contract, is the current Glienicke/Soma stuff still deploying and if its only IBFT compatible are we stopping deployment of these if we're launching the network under Tendermint?

Shirikatsu:
Also what's the state of the #121, #183? Are they planned to be merged into `tendermint` branch before it merges to master?

aneequesafdar:
@Shirikatsu 
Soma/Glienicke works exactly the same as it does on IBFT, nothing has changed in that regards. Its related code was copy-pasted.

#183 I still need to check whether that is still an issue. #121 depends on the voting power which will become more clear with autonity contract.We want to do small changes and push to master instead of such a huge change. These pull request are not critical to the algorithm and hence can be reviewed and merged after  this one.

consensus/istanbul/backend/backend.go
```diff
@@ -81,7 +88,7 @@ func New(config *istanbul.Config, privateKey *ecdsa.PrivateKey, db ethdb.Databas
 
 // ----------------------------------------------------------------------------
 
-type backend struct {
+type Backend struct {
```
yazzaoui: Why making the consensus backend public? Was it really necessary?

consensus/consensus.go
```diff
@@ -134,8 +134,8 @@ type PoW interface {
 	Hashrate() float64
 }
 
-// Istanbul is a consensus engine to avoid byzantine failure
-type Istanbul interface {
+// PoS is a consensus engine to avoid byzantine failure
+type PoS interface {
```
yazzaoui: Well, "BFT" would have been more appropriate since Tendermint and IBFT are not PoS by default, without weighted by stake consensus messages.

consensus/istanbul/backend/backend.go
```diff
@@ -132,31 +139,33 @@ func (sb *backend) Validators(number uint64) istanbul.ValidatorSet {
...
 	// send to others
-	sb.Gossip(valSet, payload)
+	_ = sb.Gossip(valSet, payload)
 	// send to self
 	msg := istanbul.MessageEvent{
 		Payload: payload,
 	}
-	go sb.istanbulEventMux.Post(msg)
+	go func() {
+		_ = sb.istanbulEventMux.Post(msg)
```
yazzaoui: Curious to understand the reason for that, is it because of a linter warning?

consensus/istanbul/backend/backend.go
aneequesafdar: This is linting https://github.com/clearmatics/autonity/pull/111/commits/9ee704cad36e0851d812d42e1c73c4edfae91e0c

consensus/consensus.go
aneequesafdar: Sure, please create an issue for renaming

consensus/istanbul/backend/backend.go
aneequesafdar: linters say you have to handle errors since you can't handle error when calling a function by creating a go-routine, you create a dummy function and satisfy the linter.

consensus/istanbul/backend/backend.go
aneequesafdar: Please checkout to the commit previous to that and run linters to see the error

consensus/istanbul/backend/backend.go
yazzaoui: This is quite an ugly hack IMHO, and there are many places in the code where the `go typemux.Post(msg)` pattern is being used. Better to not satisfy the linter in this case or better, use the TypeMuxSilent that you created?

aneequesafdar: This is an inherited problem from authors of IBFT and we didn't want to make semantical changes to IBFT, as expected by everyone espically George as asked above. This just syntactical sugar to satisfy the linter and ensure semantics remain the same.

 If we had used `TypeMuxSilent`  it changes the semantics you would have said: "was it really necessary"? like above, so either way, you would have had an issue. 

Are you are suggesting to sacrifice our code quality by not using linters to keep the syntax same from IBFT which we know is broken and is theoretically deprecated?

These are just syntax preferences and does not affect the semantics, therefore, it doesn't warrant too much of our time. If you would like to change syntax please make a pull request.

aneequesafdar: Also, `TypeMux` was deprecated (https://github.com/clearmatics/autonity/commit/9b62facdd4bdabfed5ef98d131686c4d2606083a) by Felix Lange (Geth core developer) 2.5 years ago. So we shouldn't even be using `TypeMux` (for which I have a TODO to change it). This is a non-issue.

consensus/tendermint/backend/backend.go
```diff
@@ -0,0 +1,638 @@
...
+func (sb *Backend) sendToResendCh(ctx context.Context, m messageToPeers) {
+	select {
+	case <-ctx.Done():
+		return
+	case sb.resend <- m:
+		//sent to channel
+	}
+}
+
+func (sb *Backend) trySend(ctx context.Context, msgToPeers messageToPeers) {
```
yazzaoui: This function is complicated and would benefit from explanations with comments.

cmd/autonity/chaincmd.go
```diff
@@ -219,6 +224,59 @@ func initGenesis(ctx *cli.Context) error {
 	return nil
 }
 
+func setupDefaults(genesis *core.Genesis) {
```
yazzaoui: Was this function created to solve https://github.com/clearmatics/autonity/issues/182 ?

consensus/tendermint/backend/backend.go
```diff
@@ -0,0 +1,638 @@
...
+)
+
+const (
+	// fetcherID is the ID indicates the block is from PoS engine
+	fetcherID = "tendermint"
+)
+
+// New creates an Ethereum Backend for PoS core engine.
+func New(config *tendermint.Config, privateKey *ecdsa.PrivateKey, db ethdb.Database, chainConfig *params.ChainConfig, vmConfig *vm.Config) *Backend {
+	if chainConfig.Tendermint.Epoch != 0 {
```
yazzaoui: This is now I think useless with the new `setupDefaults(genesis *core.Genesis)` function?

consensus/tendermint/backend/backend.go
```diff
@@ -0,0 +1,638 @@
...
+		config.Epoch = chainConfig.Tendermint.Epoch
+	}
+
+	if chainConfig.Tendermint.Bytecode != "" && chainConfig.Tendermint.ABI != "" {
```
yazzaoui: This condition is not reflected in `setupDefaults(genesis *core.Genesis)`

cmd/autonity/chaincmd.go
aneequesafdar: no #98

aneequesafdar: The issue was that I the genesis file defined values different from the defaults, the network would always favour defaults over genesis, so there wasn't really a point in defining values in genesis if it was getting overwritten by default which is clearly wrong. I opened this issue with a brief description of how configs work and to sort it out later #108

consensus/istanbul/backend/backend.go
JekaMas: @yazzaoui are things like this crucial for the review? I don't see how discussing such things could help us with Tendermint implementation.

consensus/tendermint/backend/backend.go
JekaMas: It's just a usual golang context pattern here. The function comment could looks like 'send the message to the resend channel if timeout didn't happened', hovewer it's literally the name and the signature of the function.

consensus/istanbul/backend/backend.go
JekaMas: @yazzaoui For me, we should define the goal of the first Tendermint implementation and this code review. We could discuss every variable and function name, but should we?

consensus/tendermint/core/timeout.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (c *core) handleTimeoutPrecommit(ctx context.Context, msg timeoutEvent) {
+	if msg.heightWhenCalled == c.currentRoundState.Height().Int64() && msg.roundWhenCalled == c.currentRoundState.Round().Int64() {
+		c.logTimeoutEvent("TimeoutEvent(Precommit): Received", "Precommit", msg)
+		c.startRound(ctx, new(big.Int).Add(c.currentRoundState.Round(), common.Big1))
+	}
+}
+
+/////////////// Calculate Timeout Duration Functions ///////////////
```
aesmonty: Could we consider adopting the `Timeout` initial values and increments from `Tendermint`? It's not optimal since our network assumptions might be different than theirs, but definitely better than setting completely arbitrary values.

Here are the Tendermint parameters for timeout parameters:

`// DefaultConsensusConfig returns a default configuration for the consensus service`
`func DefaultConsensusConfig() *ConsensusConfig {`
	`return &ConsensusConfig{`
		`WalPath:                     filepath.Join(defaultDataDir, "cs.wal", "wal"),`
		`TimeoutPropose:              3000 * time.Millisecond,`
		`TimeoutProposeDelta:         500 * time.Millisecond,`
		`TimeoutPrevote:              1000 * time.Millisecond,`
		`TimeoutPrevoteDelta:         500 * time.Millisecond,`
		`TimeoutPrecommit:            1000 * time.Millisecond,`
		`TimeoutPrecommitDelta:       500 * time.Millisecond,`
		`TimeoutCommit:               1000 * time.Millisecond,`
		`SkipTimeoutCommit:           false,`
		`CreateEmptyBlocks:           true,`
		`CreateEmptyBlocksInterval:   0 * time.Second,`
		`PeerGossipSleepDuration:     100 * time.Millisecond,`
		`PeerQueryMaj23SleepDuration: 2000 * time.Millisecond,`
	`}`
`}`

consensus/tendermint/core/backlog.go
```diff
@@ -0,0 +1,163 @@
...
+			msg := m.(*message)
+			var round, height *big.Int
+			switch msg.Code {
+			case msgProposal:
+				var m tendermint.Proposal
+				err := msg.Decode(&m)
+				if err == nil {
+					round, height = m.Round, m.Height
+				}
+				// for msgRoundChange, msgPrevote and msgPrecommit cases
```
aesmonty: Please fix comment. There is no `msgRoundChange` in Tendermint algorithm.

aneequesafdar: sure

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,327 @@
...
+	// errMovedToNewRound is returned when timer could be stopped in time
+	errMovedToNewRound = errors.New("timer expired and new round started")
+)
+
+type Engine interface {
+	Start() error
+	Stop() error
+}
+
+// New creates an Istanbul consensus core
```
aesmonty: Please change `Istanbul` for `Tendermint`.

consensus/tendermint/backend/backend.go
aneequesafdar: I think what @yazzaoui means is to have some comments regarding the rebroadcasting mechanism because from first glance it's not obvious what is happening.

consensus/tendermint/core/propose.go
```diff
@@ -0,0 +1,154 @@
...
+		return errNotFromProposer
+	}
+
+	// Verify the proposal we received
+	if duration, err := c.backend.Verify(*proposal.ProposalBlock); err != nil {
+		c.logger.Warn("Failed to verify proposal", "err", err, "duration", duration)
+		// if it's a future block, we will handle it again after the duration
+		// TIME FIELD OF HEADER CHECKED HERE - NOT HEIGHT
+		// TODO: implement wiggle time / median time
+		if err == consensus.ErrFutureBlock {
```
aesmonty: I am worried that there might be a DoS vulnerability exploitable by a Byzantine proposer.
In `handleProposal(...)` we just check if the proposal comes from the valid proposer for that round. If that proposal is future (in time field, not in view), we will store it in our backlog. However, we are not checking if we have already received a proposal from the proposer. Therefore, a byzantine proposer might be able to populate the backlog of other nodes with many future proposals for the same round.

Any ideas on this (I might be missing something)?

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		committedSeals := make([][]byte, c.currentRoundState.Precommits.VotesSize(proposal.ProposalBlock.Hash()))
+		for i, v := range c.currentRoundState.Precommits.Values(proposal.ProposalBlock.Hash()) {
+			committedSeals[i] = make([]byte, types.PoSExtraSeal)
+			copy(committedSeals[i][:], v.CommittedSeal[:])
+		}
+
+		if err := c.backend.Commit(*proposal.ProposalBlock, committedSeals); err != nil {
+			return
```
aesmonty: This error is not handled properly. Could we raise or log an error here, as in the other failure conditions?

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+	}
+}
+
+// startRound starts a new round. if round equals to 0, it means to starts a new height
+func (c *core) startRound(ctx context.Context, round *big.Int) {
```
aesmonty: Could we consider breaking down this function in different parts? It's massive and fulfilling several purposes.

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,327 @@
...
+	c.sentProposal = false
+	c.sentPrevote = false
+	c.sentPrecommit = false
+	c.setValidRoundAndValue = false
+	// c.setStep(propose) will process the pending unmined blocks sent by the backed.Seal() and set c.lastestPendingRequest
+	c.setStep(propose)
+
+	c.logger.Debug("Starting new Round", "Height", height, "Round", round)
+
+	var p *types.Block
```
aesmonty: I think a comment here would be quite helpful, explaning what's going on below, as it is quite important.

consensus/tendermint/core/propose.go
```diff
@@ -0,0 +1,154 @@
...
+		if vr == -1 {
+			// Line 22 in Algorithm 1 of The latest gossip on BFT consensus
+			if c.lockedRound.Int64() == vr || h == c.lockedValue.Hash() {
+				c.sendPrevote(ctx, false)
+			} else {
+				c.sendPrevote(ctx, true)
+			}
+			c.setStep(prevote)
+			// Line 28 in Algorithm 1 of The latest gossip on BFT consensus
+		} else if rs, ok := c.currentHeightRoundsStates[vr]; vr > -1 && vr < curR && ok && c.quorum(rs.Prevotes.VotesSize(h)) {
```
aesmonty: Consider using a helper function to make the code understandable, as this `else if` condition is quite difficult to parse.

consensus/tendermint/core/timeout.go
aneequesafdar: Ok, I will use their initial values

aneequesafdar: It might also be good to see their equation for incrementing the timeouts

consensus/tendermint/core/propose.go
JekaMas: Maybe this should be discussed as a new issue. The described code is the same as in IBFT and not related to Tendermint implementation.

aneequesafdar: You raise a good point, which is why need to do the `TODO`. However, we can add a check that a proposer is not sending a proposal more than once.

aneequesafdar: but I agree with @JekaMas that this should be a separate issue

aesmonty: Ok, I'll create it guys.

consensus/istanbul/backend/backend.go
aneequesafdar: Can we resolve this convo?

consensus/consensus.go
aneequesafdar: created #193 for this

consensus/istanbul/backend/backend.go
aneequesafdar: can we resolve this convo?

consensus/tendermint/backend/backend.go
aneequesafdar: #192

consensus/tendermint/backend/backend.go
```diff
@@ -0,0 +1,638 @@
...
+	workers.Add(numberOfWorkers)
+
+	for i := 0; i < numberOfWorkers; i++ {
+		go func(ctx context.Context) {
+			sb.workerSendLoop(ctx)
+			workers.Done()
+		}(ctx)
+	}
+
+	workers.Wait()
```
aneequesafdar: @yazzaoui pointed out that this waitGroup may be unnecessary since ReSend function is also called in a go routine.

consensus/istanbul/backend/backend.go
yazzaoui: No please until I check what was the issue with the linter. Limiting visibility and encapsulation is key for maintainability, there needs to be a good reason for exposing publicly this structure.

consensus/istanbul/backend/backend.go
yazzaoui: >So we shouldn't even be using TypeMux (for which I have a TODO to change it). This is a non-issue.

So what are you planning to do?

consensus/tendermint/core/timeout.go
aesmonty: @aneequesafdar It's also in the code snippet above. They just add the `delta` after each timeout increase.

cmd/autonity/chaincmd.go
yazzaoui: The old logic as I understood it is that we have indeed a default configuration being set for `Config.tendermint` but the `New` consensus engine function was supposed to set them to their correct value coming from `params.ChainConfig.TendermintConfig` .

consensus/istanbul/backend/backend.go
aneequesafdar: Use `// Deprecated: use Feed` as per [this comment](https://github.com/clearmatics/autonity/blob/9b62facdd4bdabfed5ef98d131686c4d2606083a/event/event.go#L40)

consensus/tendermint/core/propose.go
aneequesafdar: posting #194 for reference

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,327 @@
...
+		c.futureProposalTimer.Stop()
+	}
+}
+
+func (c *core) checkValidatorSignature(data []byte, sig []byte) (common.Address, error) {
+	return tendermint.CheckValidatorSignature(c.valSet.Copy(), data, sig)
+}
+
+func (c *core) quorum(i int) bool {
+	return float64(i) >= math.Ceil(float64(2)/float64(3)*float64(c.valSet.Size()))
```
yazzaoui: The formula is looking good, however, I'm keen to avoid performing equality comparisons when using floating-point values.
I would propose using that instead:
```
i >= int(math.Ceil(float64(2*c.valSet.Size())/float64(3)))
```

aneequesafdar: Please give a bit more context on floats equality comparisons

yazzaoui: https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
Your formula might be bugfree I'm really not sure

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,327 @@
...
+		c.lockedValue = nil
+		c.validRound = big.NewInt(-1)
+		c.validValue = nil
+
+		// Set validator set for Height
+		valSet := c.backend.Validators(height.Uint64())
+		c.valSet.set(valSet)
+
+		// Assuming that round == 0 only when the node moves to a new height
+		c.currentHeightRoundsStates = make(map[int64]roundState)
```
yazzaoui: `c.futureRoundsChange` needs to be reset here.

aneequesafdar: done

consensus/tendermint/core/core.go
```diff
@@ -0,0 +1,329 @@
...
+			delete(c.futureRoundsChange, r)
+		}
+	}
+
+	// Update current round state and add a copy to c.currentHeightRoundsState
+	// We only add old round prevote messages to c.currentHeightRoundState, while future messages are sent to the backlog
+	// Which are processed when the step is set to propose
+	if round.Int64() > 0 {
+		// This is a shallow copy, should be fine for now
+		c.currentHeightRoundsStates[round.Int64()] = *c.currentRoundState
```
aesmonty: Since you are passing `*c.currentRoundState` value, the `step` field in `currentRoundState` will not be updated in the map, right? Is this a problem?

consensus/istanbul/backend/backend.go
yazzaoui: Sounds good

consensus/tendermint/core/core.go
aesmonty: Resolved by https://github.com/clearmatics/autonity/commit/a3cfed63f45ebe5a4288100d92800f3599550704

aneequesafdar: The name has been updated to `c.currentHeightOldRoundStates` since we are copying old round states as we want to keep track of previous round `prevotes` and there was an index off by 1 error which has been fixed in the latest commit. Also, Step is of type uint64 so it is always passed by value not reference.

consensus/tendermint/core/core.go
yazzaoui: Ok I've tested them both and it appears as if there are no differences

consensus/tendermint/core/timeout.go
aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/ac5b4b495a14138cce3ff366a816e107266faa0a

consensus/tendermint/core/backlog.go
aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/ac5b4b495a14138cce3ff366a816e107266faa0a

consensus/tendermint/core/core.go
aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/ac5b4b495a14138cce3ff366a816e107266faa0a

consensus/tendermint/core/core.go
aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/75aa7f6cbe457f222336dc1c2738742b76f6d667

consensus/tendermint/core/core.go
aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/75aa7f6cbe457f222336dc1c2738742b76f6d667

consensus/tendermint/core/core.go
aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/75aa7f6cbe457f222336dc1c2738742b76f6d667

consensus/tendermint/backend.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// CheckSignature verifies the signature by checking if it's signed by
+	// the given validator
+	CheckSignature(data []byte, addr common.Address, sig []byte) error
+
+	// LastCommittedProposal retrieves latest committed proposal and the address of proposer
+	LastCommittedProposal() (*types.Block, common.Address)
+
+	// HasPropsal checks if the combination of the given hash and height matches any existing blocks
+	HasPropsal(hash common.Hash, number *big.Int) bool
```
aesmonty: @aneequesafdar this typo brings me a lot of memories...

consensus/tendermint/backend/backend.go
```diff
@@ -0,0 +1,638 @@
...
+// Address implements tendermint.Backend.Address
+func (sb *Backend) Address() common.Address {
+	return sb.address
+}
+
+func (sb *Backend) Validators(number uint64) tendermint.ValidatorSet {
+	validators, err := sb.retrieveSavedValidators(number, sb.blockchain)
+	proposerPolicy := sb.config.GetProposerPolicy()
+	if err != nil {
+		return validator.NewSet(nil, proposerPolicy)
```
aesmonty: Can the protocol continue with an empty validator set? What is the expected behaviour here?
Shouldn't we raise an error here already?

consensus/tendermint/backend.go
aneequesafdar: lol.
It's not being used, so I will remove it

aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/3ba70585ad0aa4f1fdd58c755ef33a13774ed641

consensus/tendermint/core/prevote.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err = c.checkMessage(preVote.Round, preVote.Height); err != nil {
+		// We want to store old round messages for future rounds since it is required for validRound
+		if err == errOldRoundMessage {
+			// The roundstate must exist as every roundstate is added to c.currentHeightRoundsState at startRound
+			// And we only process old rounds while future rounds messages are pushed on to the backlog
+			prevoteMS := c.currentHeightOldRoundsStates[preVote.Round.Int64()].Prevotes
+			if preVote.ProposedBlockHash == (common.Hash{}) {
+				prevoteMS.AddNilVote(*msg)
+			} else {
```
aesmonty: @aneequesafdar we need to review this logic. What happens when we receive a `ProposedBlockhash` which is different than our `c.currentProposalHash`? If we folow the Tdm protocol, this should count as a NIL vote, not trigger an error while handling the message.

aneequesafdar: This snippet of code only deals with old round messages, the error is to tell the calling function that it is an old round message, now the caller can do what they want with the error. There is no reference to `c.currentRoundProposalHash`, so I am not sure what you mean

aneequesafdar: The `ProposedBlockHash` is the hash we received in the `preVote`

consensus/tendermint/core/propose.go
```diff
@@ -0,0 +1,154 @@
...
+			c.stopFutureProposalTimer()
+			c.futureProposalTimer = time.AfterFunc(duration, func() {
+				_, sender := c.valSet.GetByAddress(msg.Address)
+				c.sendEvent(backlogEvent{
+					src: sender,
+					msg: msg,
+				})
+			})
+		}
+		return err
```
yazzaoui: We should broadcast nil if verification has failed.

aneequesafdar: That will happen naturally due to time out

aneequesafdar: Propose timeout is not cancelled until the proposal is accepted

consensus/tendermint/core/backlog.go
```diff
@@ -0,0 +1,163 @@
...
+				}
+				// for msgPrevote and msgPrecommit cases
+			default:
+				var sub tendermint.Vote
+				err := msg.Decode(&sub)
+				if err == nil {
+					round, height = sub.Round, sub.Height
+				}
+			}
+			if round == nil || height == nil {
```
aesmonty: @aneequesafdar This is check is useless, as after encoding, `round` and `height` are always going to be `big.Int` types. Also, the reference to "Nil view" does not apply anymore.

consensus/tendermint/core/propose.go
yazzaoui: That's unnecessary waiting, and it's in the spec L26 ?

aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/39ffba9ce7c52458b8952fd8fadc2f19360cd5ad

consensus/tendermint/core/propose.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// Check if the message comes from currentRoundState proposer
+	if !c.valSet.IsProposer(msg.Address) {
+		c.logger.Warn("Ignore proposal messages from non-proposer")
+		return errNotFromProposer
+	}
+
+	// Verify the proposal we received
+	if duration, err := c.backend.Verify(*proposal.ProposalBlock); err != nil {
```
yazzaoui: One of the big difference here with IBFT is that we must be sure that the block can be inserted to the chain after calling Verify().
In the current implementation, this verification is very limited, we're not checking the state, receipt and bloom roots and even if we have the parent in our chain. Luckily a special object exists for that called _BlockValidator_, but it would still need significant work to be integrated.

aesmonty: Does the _BlockValidator_ **guarantee** successful block insertion? My assumption is it does not since it is not processing the full state and may miss run out of gas errors. Am I correct? @yazzaoui

aneequesafdar: We have discussed this issue before and there could be a DDOS vector, which can be mitigated by if you have received a proposal you ignore the all future proposal of same round and height. The acceptance is defined by us as mentioned in the paper the `valid()` function is determined by implementers, so for the time being and for this pull request acceptable criteria would that used in Clique, we can create an issue for future discussion.

consensus/tendermint/backend/backend.go
aneequesafdar: fixed via https://github.com/clearmatics/autonity/pull/185/commits/3d4d55029cfb962cb943e0164ad1558363da81a1

consensus/tendermint/core/backlog.go
JekaMas: I think it's *big.Int types, so nil value is possible.

aneequesafdar: I will check by encoding and decoding a simple struct

consensus/tendermint/core/propose.go
aneequesafdar: We have decided that our definition of `valid()` will ensure that a block is insertable into the chain, therefore, we need to check that the proposal received by the nodes can be inserted into the chain.

consensus/tendermint/core/handler.go
```diff
@@ -0,0 +1,271 @@
...
+				if e := msg.Decode(&v); e != nil {
+					return errFailedDecodeVote
+				}
+				msgRound = v.Round.Int64()
+			}
+
+			c.futureRoundsChange[msgRound] = c.futureRoundsChange[msgRound] + 1
+			totalFutureRoundMessages := c.futureRoundsChange[msgRound]
+
+			if totalFutureRoundMessages >= int64(c.valSet.F()) {
```
yazzaoui: Should be greater strict.

aneequesafdar: done https://github.com/clearmatics/autonity/pull/185/commits/c7efa80aa4ca2b7650442cad8e8dd6e7cf80d3a0

consensus/tendermint/core/precommit.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (c *core) handlePrecommit(ctx context.Context, msg *message) error {
+	var preCommit tendermint.Vote
+	err := msg.Decode(&preCommit)
+	if err != nil {
+		return errFailedDecodePrecommit
+	}
+
+	if err := c.checkMessage(preCommit.Round, preCommit.Height); err != nil {
+		// We don't care about old round preCommit messages, otherwise we would not be in a new round rather a new height
```
yazzaoui: I don't get the point, according to L49 this should be round independant?

aneequesafdar: I am not sure what you mean?

This part of the code is saying that we don't care about old precommit messages from the previous round.

If I am wrong in assuming that can you please tell me why and what should be done instead?

If the comment is misleading can you please suggest an example to change it?

consensus/tendermint/core/propose.go
aneequesafdar: fixed by #220

consensus/tendermint/core/handler.go
```diff
@@ -0,0 +1,271 @@
...
+
+			} else {
+				var v tendermint.Vote
+				if e := msg.Decode(&v); e != nil {
+					return errFailedDecodeVote
+				}
+				msgRound = v.Round.Int64()
+			}
+
+			c.futureRoundsChange[msgRound] = c.futureRoundsChange[msgRound] + 1
```
yazzaoui: Double signing vulnerability:
- multiple future round messages from the same peer will increment `c.futureRoundsChange[msgRound]` 

Solution: 
- use a mapping of messageSet instead

consensus/tendermint/core/precommit.go
yazzaoui: precommit messages returning `errOldRoundMessage` must be considered. We can receive a delayed message making a quorum for an old round and in this case, we must decide on the previous proposal. This is making explicit L49 for the conditions triggering a decision by not referring to `roundp` but an arbitrary round `r` .

consensus/tendermint/core/message_set.go
```diff
@@ -0,0 +1,89 @@
...
+
+package core
+
+import (
+	"github.com/clearmatics/autonity/common"
+)
+
+func newMessageSet() messageSet {
+	return messageSet{
+		votes:    map[common.Hash]map[common.Address]message{},
```
yazzaoui: Not critical but useless, we can't have votes for different blocks in the same round, it would be byzantine behavior.

consensus/tendermint/core/precommit.go
```diff
@@ -0,0 +1,164 @@
...
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			c.commit()
+		}
+
+		// Line 47 in Algorithm 1 of The latest gossip on BFT consensus
+	} else if !c.precommitTimeout.timerStarted() && c.quorum(c.currentRoundState.Precommits.TotalSize()) {
```
yazzaoui: i think it is actually very problematic with respect of `TotalSize()`  taking into account every precommits irrespective of the block. Agree the specification is lacking here but that's an obvious vulnerability where an attacker can send 2F+1  differents precommits each of them for a different block. This can be solved by just denying votes that are not for the round proposed block.

consensus/tendermint/core/precommit.go
aneequesafdar: You are correct that it is important to handle old precommits.

consensus/tendermint/core/handler.go
aneequesafdar: That will make backlog redundant, so we need to better think about message set duplication.

consensus/tendermint/core/message_set.go
aneequesafdar: This was in anticipation for accountability and hasn't been fully through, and it doesn't have much impact on performance or otherwise. This is not "useless", would be useful in future fo accountability but since we don't have that I didn't spend too much time thinking about it. 

I hope you can be more polite in future.

consensus/tendermint/core/precommit.go
aneequesafdar: You are correct but for the happy path it doesn't have an impact and the intention for this body of work was to get the happy path working at minimum. Then in the next iteration, we focus on different levels of testing to catch all of these edges cases.

aneequesafdar:
This pull request was only intended to handle the "happy path", and before creating the pull request we said multiple times that it is the first iteration. Without having some basic test it doesn't make sense for me to handle all the edge cases because anything you fix now can be broken in future and we may not find out about it. Any cases we feel are not being handled by the current implementation we need to express them as unit tests and then we fix that issue.

It is difficult to simulate the behaviour we want in a running network, but it is possible to write unit test for handling messages, it is quite simple to send a different type of proposal, prevote, precommit messages and then see how our handler behaves and thus we can express a lot of cases.

It seems to me that the iterative approach is not appreciated  by @yazzaoui, and everything needs to be perfect, so for me, it doesn't make sense anymore to carry on with this pull request and once we have something more robust with test and further stability in line with the waterfall model, @yazzaoui can review this monolith of a code again in 2/3 months time. 

Therefore, I recommend to close this pull request and carry on with the testing.

consensus/tendermint/core/precommit.go
aneequesafdar: done via https://github.com/clearmatics/autonity/pull/185/commits/fe6c3bf9063007ac9c0119f0069ed431b3060387

consensus/tendermint/backend/backend_resend.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+type peerError struct {
+	error
+	addr common.Address
+}
+
+func delayBeforeResendPassed(msgToPeers messageToPeers) bool {
+	return !(time.Since(msgToPeers.lastTry).Truncate(time.Millisecond).Nanoseconds()/int64(time.Millisecond) < retryInterval &&
+		time.Since(msgToPeers.startTime).Truncate(time.Millisecond).Nanoseconds()/int64(time.Millisecond) > retryInterval)
```
yazzaoui: if `time.Since(msgToPeers.startTime)` is less than `retryInterval` then this function will return true, is it the intended behavior?

aneequesafdar: As far as I understand we only want to retry at a multiple of `retryInterval`. So the answer to the first bullet point is yes. Not sure how you mean to use `msgToPeers.lastTry` because it is being used on line 231. @JekaMas can you confirm?

JekaMas: > As far as I understand we only want to retry at a multiple of `retryInterval`. So the answer to the first bullet point is yes. Not sure how you mean by use `msgToPeers.lastTry` because it is being used on line 231. @JekaMas can you confirm?

Confirm.

yazzaoui: Thanks, I'm having some troubles understanding this logic, so, please tell me if I'm wrong :

consensus/tendermint/backend/backend_resend.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (sb *Backend) trySend(ctx context.Context, msgToPeers messageToPeers) {
+	if int(time.Since(msgToPeers.startTime).Seconds()) > TTL {
+		sb.logger.Trace("worker loop. messages TTL expired", "messages", msgToPeers)
+		return
+	}
+
+	if !delayBeforeResendPassed(msgToPeers) {
+		// send messages to the channel to further tries
```
yazzaoui: I think we might want to sleep here for some few milliseconds

consensus/tendermint/backend/backend_resend.go
```diff
@@ -0,0 +1,233 @@
...
+		var err error
+		var try int
+
+	SenderLoop:
+		for {
+			select {
+			case <-ticker.C:
+				try++
+
+				if err = p.Send(tendermintMsg, payload); err != nil {
```
yazzaoui: I'm having a hard time to figure out if `Send` returns no error does it means that `p` received the message?
https://golang.org/src/net/net.go?s=6460:6503#L175

consensus/tendermint/backend/backend_resend.go
JekaMas: We already have sleep at other place. Could you describe in more details why do we need one sleep more?

consensus/tendermint/backend/backend_resend.go
JekaMas: I think the right place is https://github.com/golang/go/blob/master/src/internal/poll/fd_unix.go#L254
If Write method returns no error it means that all bytes were written into the network.

yazzaoui: Ok thanks, this part of the code is very important as it is how we're implementing reliable broadcast and I will need to communicate our strategy to the researchers.

consensus/tendermint/backend/backend_resend.go
yazzaoui: I think while `delayBeforeResendPassed` is false, the message it would go back and forth between `sendToResendCh` and `workerSendLoop` exhausting one cpu core.

consensus/tendermint/backend/backend_resend.go
```diff
@@ -0,0 +1,233 @@
...
+			wg.Done()
+			sb.workerSendLoop(ctx)
+		}(ctx)
+	}
+
+	// we want to be sure that all workers started
+	wg.Wait()
+}
+
+func (sb *Backend) workerSendLoop(ctx context.Context) {
```
yazzaoui: Why using a worker pool pattern ?

consensus/tendermint/backend/backend_resend.go
yazzaoui: Ok thanks !

consensus/tendermint/backend/backend_resend.go
aneequesafdar: Btw we didn't change anything for `p.Send()` its the same from IBFT and the standard way how messages are sent to peers.

consensus/tendermint/backend/backend_resend.go
JekaMas: To not to stuck on one message. We're working here with network and it's impossible to predict how much time will take single `Send`. So if you have more workers you have less probability to stuck.

consensus/tendermint/backend/backend_resend.go
JekaMas: I'm not sure, could you check it writing a test?

consensus/tendermint/backend/backend_resend.go
aneequesafdar: And remember this is not an ideal solution but a temporary one to not get the network stuck. This consumes a lot of io resources, so in future, this will probably be changed.

consensus/tendermint/backend/backend_resend.go
aneequesafdar: I think I understand what you are trying to say but you are assuming that the resend channel is empty and once you send the messages back to the channel it will immediately be read, however, without empirical analysis it's not as simple to assume that. You are not considering how long the message will spend in the resend channel. If you have a large backlog of messages in the resend channel it will be problematic to sleep here because you may have waited longer than your retry interval (time spent in the resend channel + sleep > n*retryInterval). So I think given its a temp measure it should be fine. If you can express it through a test then it would be ideal.

Also, given our manual testing, usually messages are always sent to the peer on the first try if they are connected. In practice, this code only comes in to play when a node goes down or at the start of the network when all peers have not been connected. Please run the tendermint network with verbosity 5 to look at the logs for resend messages, the difference in logs when a node is not connected will be very apparent.

consensus/tendermint/backend/backend_resend.go
JekaMas: Agree. Until we faced with real bottleneck in the consensus in the network we shouldn't "optimize" things. We should start with something that works and step by step make it better for given requirements.

yazzaoui: Why not simply using one goroutine for one message sending attempt? Is it for optimization or something else?

JekaMas: What will happen if this goroutine will sending message for 3 seconds? We do have a bunch of peers and this peers are handling in separate goroutines fir the same reason.
The goal of this code is to have maximum probably of getting all messages without network issues.

aneequesafdar: It is quite hard to understand what you mean without some sample code. As you know a problem can be solved in multiple ways, the same goes for designing concurrent strategies please read/watch this for more info https://blog.golang.org/concurrency-is-not-parallelism

consensus/tendermint/backend/backend_resend.go
yazzaoui: I know, but since the underlying connection is TCP, a successful "send" might be taken as an ACK and I've seen some reliable broadcast protocol leveraging that, protocols like this [one](https://arxiv.org/pdf/1503.08934.pdf) or some presented [here](http://fileadmin.cs.lth.se/cs/Personal/Amr_Ergawy/dist-algos-slides/fourth-presentation.pdf)

consensus/tendermint/backend/backend_resend.go
yazzaoui: I think waiting 1 or 2 ms here could avoid some issues as the cpu is the bottleneck but like you said it's fine for now

consensus/tendermint/backend/backend_resend.go
yazzaoui: If we had an ACK that would be I think an implementation of [Automatic repeat request](https://en.wikipedia.org/wiki/Automatic_repeat_request) at the application layer.

aneequesafdar: IMO what we have here is not a strategy taken from a reliable broadcast mechanism, rather simply solving a problem for nodes which get stuck at star-up. Once the reliable broadcast mechanism and gossiping is decided on we can make the necessary changes.

aneequesafdar: This seems to be a criticism of Geth rather than Tendermint, it's important to understand these differences but shouldn't this be written elsewhere. We can really do much right now.

consensus/tendermint/backend/backend_resend.go
JekaMas: CPU is a bottleneck if you run local cluster. We don't know what will be a bottleneck on real system. So it's hard to predict something.

consensus/tendermint/backend/backend_resend.go
yazzaoui: I'm not saying that what you did is wrong, I'm just trying to understand the reasons behind the design choice of using a thread pool pattern: not rocket science but still adding overhead, and consensus deserves close scrutiny.

Very roughly something like this (very imperfect), but one goroutine per send attempt, would appear to be a much simpler implementation to understand and to achieve what you want to do:
```golang
func (sb *Backend) trySend(ctx context.Context, msg message) {
	connectedPeers, _ := sb.getPeers(msg)
	ctx, _ = context.WithTimeout(ctx, TTL*time.Second)
	for _, p := range connectedPeers {
		go func(p consensus.Peer){
			ticker := time.NewTicker(retryInterval * time.Millisecond)
			defer ticker.Stop()
			for {
				select {
				case <-ticker.C:
					if err := p.Send(tendermintMsg, msg); err == nil {
						return
					}
				case <-ctx.Done():
					return
				}
			}
		}(p)
	}
}
```

consensus/tendermint/backend/backend_resend.go
yazzaoui: Ok I understand but please could you just confirm that this mechanism could be avoided if we were sending the current height state to newly connected validators?

consensus/tendermint/backend/backend_resend.go
yazzaoui: yeah we've done some benchmarking last year on IBFT using several servers and it was clear that the bottleneck was CPU, should be the same for tdm, but it's fine we don't have requirements for performance yet.

consensus/tendermint/backend/backend_resend.go
aneequesafdar: Since the scenario you are mentioning is not implemented I cannot say anything because the depends on how its implemented and what issue arise during the implementation. Its too early to comment and there is no such requirement for it yet.

consensus/tendermint/backend/backend_resend.go
aneequesafdar: I think this will be a small bottleneck in the grand scheme of things, I am sure after profiling we can optimise other parts of the code base first before we start worrying about sending to a channel often. If you disagree, please create a pull request.

JekaMas: "yeah we've done some benchmarking last year on IBFT using several servers and it was clear that the bottleneck was CPU, should be the same for tdm, but it's fine we don't have requirements for performance yet."

consensus/tendermint/backend/backend_resend.go
aneequesafdar: It's a standard practice to use patterns, it's becoming difficult to keep up with what you want, when we keep things simple you want optimisations and when there is optimisation you want us to keep that simple.

We will always have a difference of opinion on how things are implemented, hope you have heard "if it ain't broke don't fix it". If you still want to change things please create a pull request.

consensus/tendermint/backend/backend_resend.go
aneequesafdar: if it becomes a problem we will fix it, simple

consensus/tendermint/backend/backend_resend.go
JekaMas: I'm not sure about the goals of this code review. Could you explain it to me?

I've chosen this pattern because it works well, we can control the number of goroutines, don't have goroutines leak and have a constant number of gorounines to solve this issue.

consensus/tendermint/backend/backend_resend.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if !delayBeforeResendPassed(msgToPeers) {
+		// send messages to the channel to further tries
+		sb.sendToResendCh(ctx, msgToPeers)
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, TTL*time.Second)
+	defer cancel()
```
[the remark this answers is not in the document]
aneequesafdar: @JekaMas Some questions:
- Is this defer function called after line 114 (I am guessing its not)?
- For peers who are not connected, we don't send resend them the messages because after line 139 the cancel function is called (line 118), this is problematic, or am I misunderstanding?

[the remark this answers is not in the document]
JekaMas: 1. You're right and it isn't called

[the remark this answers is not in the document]
aneequesafdar: @JekaMas What do you think of the second bullet point?

PR #188, opened 2019-07-12
changed pendingUnminedBlocks type to map[uint64]*types.Block and fix data race in isProposer

aneequesafdar:
related to #136 

There was a data race issue, where, if the height was not updated in time, the NewUnminedBlock will put the new unmined block in the backlog and just after releasing the lock the consensus event goroutine will update the state and then there would be a deadlock, because there will not be another event to send to update the lastest pending block.

The approach here is to use a map(`map[height]*unmineBlock`) instead of a priority queue, this would allow us to keep updating the map without any interference with the consensuEvents go routine and when the consensusEvent go routine requires the unmined block it will call `getUnminedBlock()` and if a block is available then it would be returned otherwise it will wait on `c.pendingUnminedBlockCh` and this will be sent over when a `NewUnminedBlockEvent` happens.

### DIFF
--- a/consensus/tendermint/core/core.go
+++ b/consensus/tendermint/core/core.go
@@ -282,8 +282,7 @@ func (c *core) startRound(ctx context.Context, round *big.Int) {
 		if c.validValue != nil {
 			p = c.validValue
 		} else {
-			unmineBlock := c.getUnminedBlock()
-			p = unmineBlock
+			p = c.getUnminedBlock()
 			if p == nil {
 				select {
 				case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


consensus/tendermint/core/core.go
```diff
@@ -278,18 +277,20 @@ func (c *core) startRound(ctx context.Context, round *big.Int) {
...
 	var p *types.Block
 	if c.isProposer() {
 		if c.validValue != nil {
 			p = c.validValue
 		} else {
-			c.latestPendingUnminedBlockMu.RLock()
-			p = c.latestPendingUnminedBlock
-			c.latestPendingUnminedBlockMu.RUnlock()
+			unmineBlock := c.getUnminedBlock()
+			p = unmineBlock
```
JekaMas: Why we need 'p' variable?

aneequesafdar: I think you mean unminedBlock is unnecessary, it should be
```go
p := c.getUnminedBlock()
if p == nil {
...
```

PR #200, opened 2019-07-18
added to genesis autonity contract section

b00ris:
Closes https://github.com/clearmatics/autonity/issues/186

Autonity contract section integration should be made after https://github.com/clearmatics/autonity/issues/169 in https://github.com/clearmatics/autonity/issues/191

### DIFF
--- a/core/autonity_contract_genesis.go
+++ b/core/autonity_contract_genesis.go
@@ -1,0 +1,102 @@
+package core
+
+import (
+	"errors"
+	"fmt"
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/p2p/enode"
+	"reflect"
+)
+
+const (
+	UserParticipant        UserType = "participant"
+	UserMember             UserType = "member"
+	UserGovernanceOperator UserType = "GovernanceOperator"
+)
+
+type UserType string
+
+func (ut UserType) IsValid() bool {
+	if ut == UserMember || ut == UserParticipant || ut == UserGovernanceOperator {
+		return true
+	}
+	return false
+}
+
+// Autonity contract config. It'is used for deployment.
+type AutonityContract struct {
+	// Address of the validator who deploys contract stored in bytecode
+	Deployer common.Address `json:"deployer" toml:",omitempty"`
+	// Bytecode of validators contract // would like this type to be []byte but the unmarshalling is not working
+	Bytecode string `json:"bytecode" toml:",omitempty"`
+	// Json ABI of the contract
+	ABI   string `json:"abi "toml:",omitempty"`
+	Users []User `json:"users"`
+}
+
+func (ac *AutonityContract) Validate() error {
+	if len(ac.Bytecode) == 0 || len(ac.ABI) == 0 {
+		return errors.New("autonity contract is empty")
+	}
+	if reflect.DeepEqual(ac.Deployer, common.Address{}) {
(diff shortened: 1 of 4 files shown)


yazzaoui:
I was thinking that it was required to have "autonityContract" mandatory because we need to remove "validators" but then I noticed that we didn't have the distinction between validators and stake-holder in the srs which is clearly confusing .. That's why prototyping is important.

@aesmonty I think we need to introduce a new user class for "validators" to make it clear that stake-holder might not validate.

aesmonty:
@yazzaoui Makes sense for me. Then, the difference would be:

- `member`: Can own stake token.
- `validator`: Can own stake token and can be selected to be part of the committee.

In A0.5 there is no committee so all `validators` are effectively part of the committee and participate in consensus voting.

core/autonity_contract_genesis.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/p2p/enode"
+	"reflect"
+)
+
+const (
+	//participant: Authorized to operate a full node, is able to join the network, is not authorized to own stake.
+	UserParticipant UserType = "participant"
+	//member: Authorized to operate a full node, is able to join the network, authorized to own stake.
+	UserMember UserType = "member"
```
yazzaoui: From https://github.com/clearmatics/doc-srs-008/issues/2 we have to rename "member" to "stakeholder"

core/autonity_contract_genesis.go
```diff
@@ -0,0 +1,115 @@
...
+// Autonity contract config. It'is used for deployment.
+type AutonityContract struct {
+	// Address of the validator who deploys contract stored in bytecode
+	Deployer common.Address `json:"deployer" toml:",omitempty"`
+	// Bytecode of validators contract // would like this type to be []byte but the unmarshalling is not working
+	Bytecode string `json:"bytecode" toml:",omitempty"`
+	// Json ABI of the contract
+	ABI                string         `json:"abi "toml:",omitempty"`
+	MinGasPrice        uint64         `json:"minGasPrice" toml:",omitempty"`
+	GovernanceOperator common.Address `json:"governanceOperator" toml:",omitempty"`
```
yazzaoui: https://github.com/clearmatics/doc-srs-008/issues/2 : need to be renamed to "Operator"

core/autonity_contract_genesis.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if u.Type == UserParticipant && u.Stake > 0 {
+		return errors.New("user.stake must be nil or equal to 0 for users of type participant")
+	}
+
+	if u.Type == UserValidator && len(u.Enode) == 0 {
+		return errors.New("if user.type is validator then user.enode must be defined")
+	}
+	if len(u.Enode) > 0 {
+		n, err := enode.ParseV4(u.Enode)
```
yazzaoui: I'm not sure if `enode.ParseV4` is compatible with enode using domain names, maybe @JekaMas can confirm?

yazzaoui: probably it should be `ParseV4WithResolve`

JekaMas: Yeap, it's better to use `ParseV4WithResolve` and write a few tests.

b00ris:
@yazzaoui @JekaMas fixed

PR #203, opened 2019-07-19
Fix precommit seal

### LINKED ISSUE #202: Fix Verify Commit and Committed Seals
### Background

A consensus message in our Tendermint implementation is formatted as shown below (inherited from IBFT):

```golang
type message struct {
	Code          uint64
	Msg           []byte
	Address       common.Address
	Signature     []byte
	CommittedSeal []byte
}
```
In the snippet above `code` refers to the `step` message. If the `step` is either `prevote` or `precommit`, the second field, `Msg` would be of type `Vote`.

```golang
type Vote struct {
	Round             *big.Int
	Height            *big.Int
	ProposedBlockHash common.Hash
}
```
`Address` is the sender address.

`Signature` is the Keccak Sig over `message` (with signature field empty).

`CommittedSeal` is the Keccak Sig over `Vote`.

Verifying commits and committed seals properly is very important for the correct implementation of a BFT algorithm, as the requirement to insert a new block is having `> 2/3` committed seals from valid validators (which must be verified).

### Problem

Currently, in TDM we do verify the `Signature` field and verify that the `CommitedSeal` is signed by a correct validator.

However, we do not verify:

1.  That the signer of `Signature` and `CommitedSeal` are the same (and they need to be the same or the `precommit` would be invalid).
2. That the content (and length) being signed in `CommitedSeal` correspond to the contents of the `Vote` struct of that message.

### Solution

1. Implement a `VerifyCommit` function in TDM to compare that both signers are the same.
2. Add a check to fulfil condition 2.

For reference, [Correctness Analysis of IBFT](https://arxiv.org/pdf/1901.07160.pdf), page 17.

aneequesafdar:
closes #202


JekaMas:
Could you merge master branch and fix linters?

aneequesafdar:
Sure

aesmonty:
Looking good!

PR #212, opened 2019-07-25
Basic WAL to restore after node crush or consensus got stuck

### LINKED ISSUE #177: Consensus should continue if 1/3N nodes go down and start again
Based on the tendermint specification there is no mention of re-sending already broadcasted messages, this is problematic if 1/3N honests node goes down without anyone's fault and come back up then the consensus should continue. However, in the current implementation where there is no re-sending of old messages, the consensus will be halted even if all of the validators are online.

One solution is for validators to send their consensus state after block synchronisation has finished, to allow for consensus to continue.

JekaMas:
closes #177


consensus/tendermint/core/core.go
```diff
@@ -298,7 +307,18 @@ func (c *core) setCore(r *big.Int, h *big.Int, lastProposer common.Address) {
 		// This is a shallow copy, should be fine for now
 		c.currentHeightOldRoundsStates[r.Int64()-1] = *c.currentRoundState
 	}
+
 	c.currentRoundState.Update(r, h)
+
+	err := c.wal.UpdateHeight(h)
+	if err != nil {
+		c.logger.Error("Starting new WAL with error", "Height", h, "Round", r, "err", err)
```
b00ris: Shouldn't we stop it if we got an error? We could get an inconsistent state.

consensus/tendermint/core/handler.go
```diff
@@ -191,6 +186,67 @@ func (c *core) handleConsensusEvents(ctx context.Context) {
...
+		c.logger.Info("WAL: cant get messages", "height", height.String(), "round", round.String(), "err", err.Error())
+		return
+	}
+
+	c.logger.Warn("WAL: broadcasting", "height", c.currentRoundState.Height().String(), "round", c.currentRoundState.Round().String(), "currentHeight", height.String(), "currentRound", round.String(), "msg", len(pastMessages))
+
+	for _, msg := range pastMessages {
+		c.logger.Debug("WAL: broadcasting message", "height", height.String(), "round", round.String(), "msg", msg)
+
+		if err = c.backend.Broadcast(ctx, c.valSet.Copy(), msg); err != nil {
```
b00ris: Do we need to send messages to other peers? Or I'm missing something?

JekaMas: Broadcast() will broadcast to all peers

consensus/tendermint/core/core.go
JekaMas: Agreed

JekaMas: Hmmm.... we never return errors in the state machine...

b00ris: Anyway, it's a separate issue. Could you create it?

JekaMas:
@yazzaoui @aneequesafdar @shapeshed The core review takes too much time. After https://github.com/clearmatics/autonity/pull/221 I'd have to rewrite this PR.

JekaMas:
@aneequesafdar why did you close this pr?

aneequesafdar:
I didn't intend to close the pr, I deleted the tendermint branch (as we no longer needed it) and then it closed every pr which was meant to be merged to tendermint. I didn't know that this would happen.

eth/config.go
```diff
@@ -29,38 +27,42 @@ import (
...
-	MinerGasCeil:   8000000,
-	MinerGasPrice:  big.NewInt(params.GWei),
-	MinerRecommit:  3 * time.Second,
-
-	TxPool: core.DefaultTxPoolConfig,
-	GPO: gasprice.Config{
-		Blocks:     20,
-		Percentile: 60,
-	},
+func DefaultConfig() Config {
```
aneequesafdar: any specific reason to change the config to a function?

JekaMas: Yeap. The data race.

local_test_network/autonity-start.sh
```diff
@@ -10,13 +10,17 @@ RPC_ADDR=$(awk 'END{print $1}' /etc/hosts)
...
+  echo "Autonity INIT $RPC_ADDR $DATADIR\autonity\chaindata"
+  $AUTONITY init --datadir $DATADIR genesis.json --verbosity 4
+fi
 
 # start the node with the keystore and nodekey
 echo "Autonity START"
 $AUTONITY \
   --datadir $DATADIR \
+  --ethash.cachedir "$DATADIR/cache" \
+  --ethash.dagdir "$DATADIR/.ethash" \
```
aneequesafdar: Why define ethhash cache and dagdir?

JekaMas: Because you want to keep this data between docker restart.

PR #221, opened 2019-07-31
Very basic tendermint tests

JekaMas:
The main idea is to start a bunch of nodes with basic Tendermint consensus and a few nodes with Tendermint, but some methods could be redefined by a programmer.

Also, the refactoring was done because of the test introduction causes many cycle dependencies, eg Backend and Core packages do not depend on Tendermint (the upper layer) and Backend doesn't depend on Core anymore.

The main file is `consensus/tendermint/tendermint_test.go`

Test cases(it's all distributed networks not simulations):
- 5 peers, for 5 blocks, 1 transaction per peer per block
- 5 peers, for 5 blocks, 100 transactions per peer per block
- 4 normal peers, 1 slow peer (50KB/sec in, 50KB/sec out), for 5 blocks, 1 transaction per peer per block
- 5 slow peers (50KB/sec in, 50KB/sec out) for 5 blocks, 1 transaction per peer per block
- 4 normal peers, 1 malicious that accepts even bad blocks, for 5 blocks, 1 transaction per peer per block
- 4 normal peers, 1 peer stops for 10 seconds and then starts again (not only stops in terms of the network but stops all services), for 10 blocks, 1 transaction per peer per block

Data races are fixed:
- https://github.com/clearmatics/autonity/issues/232
- https://github.com/clearmatics/autonity/issues/227


eth/sync_test.go
```diff
@@ -58,3 +58,72 @@ func TestFastSyncDisabling(t *testing.T) {
 		t.Fatalf("fast sync not disabled after successful synchronisation")
 	}
 }
+
+// Tests that fast sync gets disabled as soon as a real block is successfully
+// imported into the blockchain.
+func TestFastSyncDisablingMany(t *testing.T) {
```
aneequesafdar: Would you still like to keep this test case? Seems to me that this was meant for your initial testing

JekaMas: Let's keep it. It checks fast sync for many nodes and it seems a useful thing.

p2p/server.go
```diff
@@ -363,13 +372,14 @@ func (src *Server) UpdateWhitelist(enodes []*enode.Node) {
...
 			src.RemovePeer(connectedPeer.Node())
+			src.RemoveTrustedPeer(connectedPeer.Node())
 		}
 	}
 
 	// Check for peers that needs to be connected
 	for _, whitelistedEnode := range enodes {
 		found := false
-		for _, oldEnode := range src.StaticNodes {
+		for _, oldEnode := range src.TrustedNodes {
```
yazzaoui: You're changing network permissionning logic, I'm worried about that, is it really necessary?

JekaMas: Yeap, because I've faced the issue that the network is not fully connected: we have 5 nodes, however, each peer has from 1 to 3 peers. It was caused by the fact that static nodes could be disconnected if we reach connections limits. So I've decided to add white list nodes into static and trusted lists.

eth/sync_test.go
```diff
@@ -58,3 +58,72 @@ func TestFastSyncDisabling(t *testing.T) {
...
+
+	pmEmpty1, _ := newTestProtocolManagerMust(t, downloader.FastSync, 0, nil, nil, []string{p2pPeerEmpty.Info().Enode, p2pPeerFull.Info().Enode, p2pPeerEmpty1.Info().Enode})
+
+	if atomic.LoadUint32(&pmEmpty1.fastSync) == 0 {
+		t.Fatalf("fast sync disabled on pristine blockchain 1")
+	}
+
+	// Create a full protocol manager, check that fast sync gets disabled
+	pmFull, _ := newTestProtocolManagerMust(t, downloader.FastSync, 1024, nil, nil, []string{p2pPeerEmpty.Info().Enode, p2pPeerFull.Info().Enode, p2pPeerEmpty1.Info().Enode})
+
```
aneequesafdar: The test peer `p2pPeerEmpty`, is it assigned to `pmEmpty`? Seems to me that the last argument for `newTestProtocolManagerMust()` is just defining the initial whitelist, so can we create a protocol manager without an associated node, if so is this correct behaviour for network permissioning?

JekaMas: It's not assigned. It creates protocolManager that knows about the list of the peers `[]string{p2pPeerEmpty.Info().Enode, p2pPeerFull.Info().Enode, p2pPeerEmpty1.Info().Enode}`

aneequesafdar: I see it should be fine for syncing. but I'm not sure about network permissioning

JekaMas: `network permissioning` is untouched, afaik.

aneequesafdar: Yeah, that is fine, it was just strange to me to not have the protocol managers server info in the whitelist but since it's for testing it doesn't matter for the test case being tested.

p2p/server.go
yazzaoui: Ok I reviewed trusted node logic and it should be fine, however we might want to initialize `srv.TrustedNodes` to an empty array line 522, to avoid potential conflicts with a configuration file.

JekaMas: Why not to keep user's trusted list and add our trusted peers?

yazzaoui: Ok but if you do that, please do it as well with static nodes

JekaMas: Ok.

build/ci.go
```diff
@@ -327,7 +327,7 @@ func doTest(cmdline []string) {
 	// Test a single package at a time. CI builders are slow
 	// and some tests run into timeouts under load.
 	gotest := goTool("test", buildFlags(env)...)
-	gotest.Args = append(gotest.Args, "-p", "1", "-timeout", "5m")
+	gotest.Args = append(gotest.Args, "-p", "1", "-timeout", "30m", "-short")
```
aneequesafdar: Could you please give more context on how this `-short` flag work, it seems that you always set in ci.go, so the tendermint tests aren't run by travis, is this something we want?

.travis.yml
```diff
@@ -20,9 +20,11 @@ matrix:
         - sudo chown root:$USER /etc/fuse.conf
         - go run build/ci.go install
         - make lint-ci
-        - go test ./consensus/tendermint/... -race -v
+        - go test ./consensus/tendermint/... -race -v -parallel 1
```
aneequesafdar: How is running `-parallel 1` different from not calling parallel at all? And which test do you want to run in parallel?

.travis.yml
```diff
@@ -20,9 +20,11 @@ matrix:
...
         - go run build/ci.go test -coverage $TEST_PACKAGES
-        - $HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci -repotoken $COVERALLS_TOKEN
+        - travis_wait 30 go test ./consensus/test/... -v -timeout 30m -covermode=atomic -cover -coverprofile=coverage_e2e.out
+        - echo "$(tail -n +2 coverage_e2e.out)" > coverage_e2e.out && cat coverage.out coverage_e2e.out > coverage_all.out
+        - $HOME/gopath/bin/goveralls -coverprofile=coverage_all.out -service=travis-ci -repotoken $COVERALLS_TOKEN
```
aneequesafdar: general question, why don't we call `ci.go` functions through `Makefile` instead of calling all the different test manually?

core/types/bft.go
```diff
@@ -163,14 +166,8 @@ func Ecrecover(header *Header) (common.Address, error) {
...
-func PrepareExtra(extraData *[]byte, vals []common.Address) ([]byte, error) {
-	var buf bytes.Buffer
-
-	// compensate the lack bytes if header.Extra is not enough BFTExtraVanity bytes.
-	if len(*extraData) < BFTExtraVanity {
-		*extraData = append(*extraData, bytes.Repeat([]byte{0x00}, BFTExtraVanity-len(*extraData))...)
-	}
-	buf.Write((*extraData)[:BFTExtraVanity])
+func PrepareExtra(extraData []byte, vals []common.Address) ([]byte, error) {
+	extraDataCopy := append([]byte{}, extraData...)
```
aneequesafdar: Why do we copy the extraData? what made you decide that the function signature needed to be changed?

core/genesis.go
```diff
@@ -62,6 +63,8 @@ type Genesis struct {
 	Number     uint64      `json:"number"`
 	GasUsed    uint64      `json:"gasUsed"`
 	ParentHash common.Hash `json:"parentHash"`
+
+	mu sync.RWMutex
```
aneequesafdar: What was the data race causing issues?

core/genesis.go
```diff
@@ -273,18 +276,33 @@ func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
...
+		//ex, _ := types.ExtractBFTExtra(g.ExtraData)
+		//for _, val := range ex.Validators {
+		//	res += val.String() + "\n"
+		//}
+		//fmt.Println(res)
 	}
 
+	//g.mu.Unlock()
 	block := g.ToBlock(db)
+	//g.mu.Lock()
```
aneequesafdar: Do you still want to keep all of these comments?

core/genesis.go
```diff
@@ -295,23 +313,30 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
 	rawdb.WriteHeadBlockHash(db, block.Hash())
 	rawdb.WriteHeadHeaderHash(db, block.Hash())
 
-	rawdb.WriteEnodeWhitelist(db, types.NewNodes(g.Config.EnodeWhitelist, true))
+	g.Config.SortEnodeWhitelist()
```
aneequesafdar: Was the enode list not sorted before? If it was could you please point me where, please

eth/filters/filter_system.go
```diff
@@ -140,7 +140,13 @@ func NewEventSystem(mux *event.TypeMux, backend Backend, lightMode bool) *EventS
...
 	if m.txsSub == nil || m.logsSub == nil || m.rmLogsSub == nil || m.chainSub == nil ||
 		m.pendingLogSub.Closed() {
-		log.Crit("Subscribe for event system failed")
+		log.Crit("Subscribe for event system failed",
+			"txsSub", m.txsSub == nil,
+			"logsSub", m.logsSub == nil,
+			"rmLogsSub", m.rmLogsSub == nil,
+			"chainSub", m.chainSub == nil,
+			"pendingLogSub.Closed", m.pendingLogSub.Closed(),
+		)
```
aneequesafdar: Where did you want to use `NewEventSystem` that cause to extend the log messages?

eth/config.go
```diff
@@ -18,7 +18,7 @@ package eth
 
 import (
 	"github.com/clearmatics/autonity/consensus/istanbul"
-	"github.com/clearmatics/autonity/consensus/tendermint"
+	"github.com/clearmatics/autonity/consensus/tendermint/config"
```
aneequesafdar: having a `tendermintConfig` as an alias to the import would be better naming

eth/gen_config.go
```diff
@@ -3,14 +3,14 @@
 package eth
 
 import (
+	"github.com/clearmatics/autonity/consensus/tendermint/config"
```
aneequesafdar: having a tendermintConfig as an alias to the import would be better naming

miner/stress_clique.go
```diff
@@ -203,7 +204,7 @@ func makeSealer(genesis *core.Genesis) (*node.Node, error) {
 			MinerGasCeil:    genesis.GasLimit * 11 / 10,
 			MinerGasPrice:   big.NewInt(1),
 			MinerRecommit:   time.Second,
-		})
+		}, cons)
```
aneequesafdar: Why add `cons` and then not pass any value? If you look at where makeSealer is being called there is no second argument, in  Goland, there is an error when the file is opened `stress_clique.go doesn't match to the target system. File will be ignored by build tool`, maybe that's why no compilation error occurs. Preferably you should just pass `nil` and remove the second argument like `stress_ethash.go`

miner/stress_ethash.go
```diff
@@ -183,7 +183,7 @@ func makeMiner(genesis *core.Genesis) (*node.Node, error) {
 			MinerGasCeil:    genesis.GasLimit * 11 / 10,
 			MinerGasPrice:   big.NewInt(1),
 			MinerRecommit:   time.Second,
-		})
+		}, nil)
```
aneequesafdar: Same change should be applied to `stress_clique.go` unless I am mistaken

node/node.go
```diff
@@ -484,6 +484,11 @@ func (n *Node) Attach() (*rpc.Client, error) {
 	if n.server == nil {
 		return nil, ErrNodeStopped
 	}
+
+	if n.config.P2P.IsRated {
+		return rpc.DialInProcWithRate(n.inprocHandler, n.config.P2P.InRate, n.config.P2P.OutRate), nil
+	}
+
```
aneequesafdar: Could you please explain what Attach() does and why do you want to add this `DialInProcWithRate()` function?

p2p/enode/urlv4.go
```diff
@@ -236,6 +236,20 @@ func (n *Node) v4URL() string {
...
+
+	addr := net.TCPAddr{IP: ip, Port: tcp}
+	u.User = url.User(nodeid)
+	u.Host = addr.String()
+	if udp != tcp {
+		u.RawQuery = "discport=" + strconv.Itoa(udp)
+	}
+	return u.String()
+}
+
```
aneequesafdar: What was the reason to not add the new function in test/test_helper.go or why it should be here, could you please explain your thought process?

p2p/server.go
aneequesafdar: Could you please explain the trusted node and static node logic?

p2p/server.go
```diff
@@ -378,10 +388,12 @@ func (src *Server) UpdateWhitelist(enodes []*enode.Node) {
 		if !found {
 			log.Info("Connecting to newly authorized peer", "enode", whitelistedEnode.String())
 			src.AddPeer(whitelistedEnode)
+			src.AddTrustedPeer(whitelistedEnode)
 		}
 	}
 
 	src.StaticNodes = enodes
+	src.TrustedNodes = enodes
```
aneequesafdar: Would it make more sense to just use trusted node, not static nodes, seems to me that this makes more sense because whitelist is a trusted list, however, the only question is that will trusted nodes be reconnected to if they are disconnected? Having both will only cause confusion in future and add technical debt if we don’t have time we should create an issue for it.

params/config.go
```diff
@@ -174,6 +176,7 @@ type ChainConfig struct {
 	Tendermint *TendermintConfig `json:"tendermint,omitempty"`
 
 	// Network Permissioning
+	mu                sync.RWMutex
```
aneequesafdar: Which data race caused you to include a mutex in ChainConfig?

vendor/vendor.json
```diff
@@ -760,6 +760,12 @@
 			"revision": "b4690f45fa1cafc47b1c280c2e75116efe40cc13",
 			"revisionTime": "2017-02-15T08:41:58Z"
 		},
+		{
+			"checksumSHA1": "iEK5hCRfrkdc1JOJsaiWuymHmeQ=",
+			"path": "golang.org/x/sync/errgroup",
+			"revision": "56d357773e8497dfd526f0727e187720d1093757",
+			"revisionTime": "2019-04-12T04:58:03Z"
+		},
```
aneequesafdar: How do you update vendors?

consensus/istanbul/backend/engine_test.go
```diff
@@ -485,18 +485,22 @@ func TestPrepareExtra(t *testing.T) {
 		Extra: vanity,
 	}
 
-	payload, err := types.PrepareExtra(&h.Extra, validators)
+	payload, err := types.PrepareExtra(h.Extra, validators)
+	//panic(fmt.Sprint(len(payload), len(expectedResult)))
```
aneequesafdar: Please remove the commented code on line 489 if not required

consensus/tendermint/backend/backend.go
```diff
@@ -447,3 +463,20 @@ func (sb *Backend) WhiteList() []string {
...
+func (sb *Backend) GetPrivateKey() *ecdsa.PrivateKey {
+	sb.privateKeyMu.RLock()
+	defer sb.privateKeyMu.RUnlock()
+
+	pk := sb.privateKey.PublicKey
+	d := big.NewInt(0).Set(sb.privateKey.D)
+	return &ecdsa.PrivateKey{PublicKey: pk, D: d}
+}
+
+func (sb *Backend) SetPrivateKey(key *ecdsa.PrivateKey) {
```
aneequesafdar: Lines 91 and 92, should we use the newly defined setter and getter to set the privatekey and address seems we may be setting them more than necessary

consensus/tendermint/backend/backend.go
```diff
@@ -131,10 +143,12 @@ type Backend struct {
 
 // Address implements tendermint.Backend.Address
 func (sb *Backend) Address() common.Address {
+	sb.privateKeyMu.RLock()
+	defer sb.privateKeyMu.RUnlock()
```
aneequesafdar: Why use mutex for address?

consensus/tendermint/core/handler.go
```diff
@@ -86,38 +98,17 @@ func (c *core) unsubscribeEvents() {
...
 
 func (c *core) handleNewUnminedBlockEvent() {
-	defer func() {
-		if r := recover(); r != nil {
-			debug.PrintStack()
-
-			c.logger.Crit("panic in core.handleNewUnminedBlockEvent", "panic", r)
-		}
-	}()
-
```
aneequesafdar: Why delete the recover functions?

consensus/tendermint/core/handler.go
```diff
@@ -86,38 +98,17 @@ func (c *core) unsubscribeEvents() {
...
 	for e := range c.newUnminedBlockEventSub.Chan() {
-		c.logger.Debug("Started handling tendermint.NewUnminedBlockEvent")
-
-		newUnminedBlockEvent := e.Data.(tendermint.NewUnminedBlockEvent)
+		newUnminedBlockEvent := e.Data.(events.NewUnminedBlockEvent)
 		pb := &newUnminedBlockEvent.NewUnminedBlock
 		c.storeUnminedBlockMsg(pb)
-
-		c.logger.Debug("Finished handling tendermint.NewUnminedBlockEvent")
```
aneequesafdar: I thought it was useful to see how each message handling starting and finishing

consensus/tendermint/crypto/utils_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"crypto/ecdsa"
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/consensus/tendermint/config"
+	"github.com/clearmatics/autonity/consensus/tendermint/validator"
+	"github.com/clearmatics/autonity/crypto"
+	"sort"
+	"strings"
+	"testing"
+)
+
```
aneequesafdar: Are these function copy-pasted from backend_test.go or did you change some stuff? Also, it seems in the latest commits some of the renamed/moved files were introduced back after the marge from master

consensus/tendermint/validator/validator.go
```diff
@@ -18,7 +18,7 @@ package validator
 
 import (
 	"github.com/clearmatics/autonity/common"
-	"github.com/clearmatics/autonity/consensus/tendermint"
+	"github.com/clearmatics/autonity/consensus/tendermint/config"
```
aneequesafdar: in general, Is there a way to simplify the validators and are there duplicated tests in core/validator_test.go and validator/default_test.go?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,592 @@
...
+		return nil
+	}
+
+	return validatorHook
+}
+
+type hook func(block *types.Block, validator *testNode, tCase *testCase, currentTime time.Time) error
+
+func tunTest(t *testing.T, test *testCase) {
+	log.Root().SetHandler(log.LvlFilterHandler(log.LvlError, log.StreamHandler(os.Stderr, log.TerminalFormat(true))))
```
aneequesafdar: Are you changing the global logger (L254) or am I misunderstanding?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,592 @@
...
+	_ = fdlimit.Raise(2048)
```
aneequesafdar: Why do you want to raise the file descriptor limit of the process and why are you not handling the error, if it is important surely you should handle the error, right?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,592 @@
...
+	if !ok || validatorHook == nil {
+		return nil
+	}
+
+	return validatorHook
+}
+
+type hook func(block *types.Block, validator *testNode, tCase *testCase, currentTime time.Time) error
+
+func tunTest(t *testing.T, test *testCase) {
```
aneequesafdar: tunTest() is misspelt it should be changed to runTest()

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,592 @@
...
+	for i := range validators {
+		validators[i] = new(testNode)
+		validators[i].privateKey, err = crypto.GenerateKey()
+		if err != nil {
+			t.Fatal("cant make pk", err)
+		}
+	}
+
+	for i := range validators {
+		validators[i].listener, err = net.Listen("tcp", "127.0.0.1:0")
```
aneequesafdar: Is it ok to assign the same port to each validator?

build/ci.go
JekaMas: Yeap. I did it to run "long" tests using standard "go test". You can find this new command in the travis file.

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_script:
   - make test-deps
 matrix:
   include:
-    - if: branch IN (master, tendermint) OR type = pull_request
+    - if: branch = master OR type = pull_request
```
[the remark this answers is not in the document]
JekaMas: We don't have tendermint branch anyone.

.travis.yml
JekaMas: Parallel=1 is the only way to run not parallel at all.

.travis.yml
JekaMas: ci.go dont have most of go test flags, like race, timeout, coverage. I don't want to spend time supporting ci.go.

consensus/tendermint/crypto/utils_test.go
JekaMas: Circle imports. The only way yo handle it is to extract circle imports into new packages. You can try to restore old file and check the way how circle imports become.

consensus/tendermint/validator/validator.go
JekaMas: I don't know it was a bit out of the scope of the issue. Lets create a new one for it.

consensus/test/tendermint_test.go
JekaMas: But works :) thank you, will fix it.

consensus/test/tendermint_test.go
JekaMas: Yeap. To manage a few local instances at the same time. This approach is used in the go-ethereum test that I used as the base for my tests.

consensus/test/tendermint_test.go
JekaMas: Please, check the original go-ethereum test.

consensus/test/tendermint_test.go
JekaMas: It's not the same, it's "any free port" each time.

core/types/bft.go
JekaMas: To remove the data race with the slice.

eth/config.go
JekaMas: Why do we need alias if no other "config" packages are imported?

eth/filters/filter_system.go
JekaMas: Because I got 2 cases when chainSub was nil.

eth/gen_config.go
JekaMas: See above

p2p/enode/urlv4.go
JekaMas: I need that only in tests to get v4enode address from basic types without creating a peer.

p2p/server.go
JekaMas: Go-ethereum has limits to the number of connected peers. Trusted nodes can be connected to the peer even if this limits were reached.
Static peers - peers that should be always connected to the peer.

p2p/server.go
JekaMas: Nope. See above.

vendor/vendor.json
JekaMas: https://github.com/ethereum/go-ethereum/wiki/Developers'-Guide

node/node.go
JekaMas: It creates rpc from node.
I need to add new function here to create rpc limited with given in config network rates.

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,592 @@
...
+		}
+	}
+
+	for _, validator := range validators {
+		listener := validator.listener
+		validator.address = listener.Addr().String()
+		port := strings.Split(listener.Addr().String(), ":")[1]
+		validator.port, _ = strconv.Atoi(port)
+
+		validator.url = enode.V4URL(
```
aneequesafdar: Was this meant to be validator.enode?

JekaMas: Yeap, you're right.

params/config.go
JekaMas: We had data races with EnodeWhitelist, GlienickeDeployer, GlienickeBytecode and GlienickeABI. You can see new guarded methods for these fields.

consensus/tendermint/backend/backend.go
JekaMas: Because we had data race with private key and address fields both. And they depend on each other.

consensus/tendermint/backend/backend.go
JekaMas: Lines 91, 92 are in the constructor function, the object doesn't exist at this moment.

core/types/bft.go
JekaMas: Data races with fields Difficulty and ExtraData

core/genesis.go
JekaMas: Data races with fields Difficulty and ExtraData

core/genesis.go
JekaMas: Thank you, removed.

core/genesis.go
JekaMas: I don't think it's sorted at this point.

consensus/tendermint/core/types.go
```diff
@@ -0,0 +1,142 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+import (
+	"fmt"
+	"github.com/clearmatics/autonity/log"
```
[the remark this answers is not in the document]
JekaMas: Not sorted like "fmt" or "goimports"?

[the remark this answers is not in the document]
screwyprof: go imports

consensus/tendermint/crypto/utils.go
```diff
@@ -0,0 +1,28 @@
+package crypto
```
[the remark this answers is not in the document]
JekaMas: I'd find a new name for it.

consensus/tendermint/backend.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+	"github.com/clearmatics/autonity/core/types"
+	"time"
+
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/event"
+)
+
+// Backend provides application specific functions for Istanbul core
+type Backend interface {
```
aneequesafdar: Duplicated file after the merge, it was moved to core/core_backend.go

consensus/tendermint/backend/backend_test.go
```diff
@@ -71,6 +76,51 @@ func TestCheckSignature(t *testing.T) {
 	}
 }
 
+func TestCheckValidatorSignature(t *testing.T) {
```
aneequesafdar: again reintroduced after merge this was moved to tendermint/crypto/utils_test.go

consensus/tendermint/backend/backend_test.go
```diff
@@ -175,6 +225,34 @@ func generatePrivateKey() (*ecdsa.PrivateKey, error) {
 	return crypto.HexToECDSA(key)
 }
 
+func newTestValidatorSet(n int) (validator.Set, []*ecdsa.PrivateKey) {
```
aneequesafdar: same as above

consensus/tendermint/config.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2017 The go-ethereum Authors
```
aneequesafdar: again file reintroduced

consensus/tendermint/errors.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2017 The go-ethereum Authors
```
aneequesafdar: these errors were moved to various files and errors.go deleted

consensus/tendermint/types.go
```diff
@@ -0,0 +1,142 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package tendermint
+
+import (
```
aneequesafdar: this file was moved to core/types.go

consensus/tendermint/validator.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2017 The go-ethereum Authors
```
aneequesafdar: this is also reintroduced after merge it was moved to core/validator/validator_interface.go

consensus/tendermint/validator_mock.go
```diff
@@ -0,0 +1,252 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: consensus/tendermint/validator.go
+
+// Package tendermint is a generated GoMock package.
```
aneequesafdar: same as above

p2p/discv5/node_test.go
```diff
@@ -152,7 +152,7 @@ func TestParseNode(t *testing.T) {
 			if err == nil {
 				t.Errorf("test %q:\n  got nil error, expected %#q", test.rawurl, test.wantError)
 				continue
-			} else if err.Error() != test.wantError {
+			} else if !strings.Contains(err.Error(), test.wantError) {
```
aneequesafdar: was this linting error?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+	}
+
+	genesis := makeGenesis(validators)
+	for i, validator := range validators {
+		var engineConstructor func(basic consensus.Engine) consensus.Engine
+		if test.maliciousPeers != nil {
+			engineConstructor = test.maliciousPeers[i]
+		}
+
+		validator.listener.Close()
```
aneequesafdar: Why close the listener, was it only opened to assign the ports?

consensus/test/test_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+			DatabaseCache:   256,
+			DatabaseHandles: 256,
+			TxPool:          core.DefaultTxPoolConfig,
+			Tendermint:      *config.DefaultConfig(),
+		}, cons)
+	}); err != nil {
+		return nil, err
+	}
+
+	// Start the node and return if successful
```
aneequesafdar: please update comment

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+		validator.isInited = true
+	} else {
+		signer = store.Accounts()[0]
+	}
+
+	if err = store.Unlock(signer, ""); err != nil {
+		return fmt.Errorf("cant unlock: %s", err)
+	}
+
+	validator.node.ResetEventMux()
```
aneequesafdar: Why do you want to reset the EventMux?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+
+	if err := validator.node.Start(); err != nil {
+		return fmt.Errorf("cannot start a node %s", err)
+	}
+
+	// Start tracking the node and it's enode
+	validator.enode = validator.node.Server().Self()
```
aneequesafdar: Would this enode assignment be redundant if you change [this](https://github.com/clearmatics/autonity/pull/221/files#r315685331)

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+	return nil
+}
+
+func (validator *testNode) startService() error {
+	var ethereum *eth.Ethereum
+	if err := validator.node.Service(&ethereum); err != nil {
+		return fmt.Errorf("cant start a node %s", err)
+	}
+
+	for !ethereum.IsListening() {
```
aneequesafdar: Doesn't `IsListening()` always return true? If ethereum was no initialised would there not be a nil pointer error?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	if err := ethereum.StartMining(1); err != nil {
+		return fmt.Errorf("cant start mining %s", err)
+	}
+
+	for !ethereum.IsMining() {
+		time.Sleep(50 * time.Millisecond)
+	}
```
aneequesafdar: Why do you want to wait for it to start mining?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+	err = wg.Wait()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	wg = &errgroup.Group{}
+	for i, validator := range validators {
+		validator := validator
+		i := i
```
aneequesafdar: aren't these 2 lines redundant, don't they get reassigned at the beginning of the new loop

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+func sendTransactions(t *testing.T, test *testCase, validators []*testNode, txPerPeer int, errorOnTx bool) {
+	const blocksToWait = 10
+
+	txs := make(map[uint64]int) // blockNumber to count
+	txsMu := sync.Mutex{}
+
+	wg := &errgroup.Group{}
+	for index, validator := range validators {
+		index := index
+		validator := validator
```
aneequesafdar: Again these 2 lines of code seem redundant

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+		count := txs[uint64(key)]
+		fmt.Printf("Block %d has %d transactions\n", key, count)
+	}
+
+	//check that all nodes reached minimum height
+	lastBlock := uint64(keys[len(keys)-1])
+	for index, validator := range validators {
+		validatorBlock := validator.service.BlockChain().CurrentBlock().Number().Uint64()
+
+		if validatorBlock < lastBlock-blocksToWait/2 {
```
aneequesafdar: Why do you divide `blocksToWait` by 2?

consensus/test/tendermint_test.go
```diff
@@ -0,0 +1,605 @@
...
+						}
+
+						break wgLoop
+					}
+				case innerErr := <-validator.subscription.Err():
+					if innerErr != nil {
+						return fmt.Errorf("error in blockchain %q", innerErr)
+					}
+
+					time.Sleep(500 * time.Millisecond)
```
aneequesafdar: why do you want to sleep here for 0.5s?

consensus/tendermint/crypto/utils_test.go
```diff
@@ -0,0 +1,190 @@
...
+		t.Errorf("error mismatch: have %v, want %v", err, ErrUnauthorizedAddress)
+	}
+
+	emptyAddr := common.Address{}
+	if addr != emptyAddr {
+		t.Errorf("address mismatch: have %v, want %v", addr, emptyAddr)
+	}
+}
+
+func TestCheckValidatorSignatureInvalid(t *testing.T) {
```
aneequesafdar: Did you add this new test function or did you take it from elsewhere? Just asking to see if i need to review or not :)

consensus/tendermint/crypto/utils_test.go
```diff
@@ -0,0 +1,190 @@
...
+		t.Errorf("error mismatch: have %v, want %v", err, ErrUnauthorizedAddress)
+	}
+
+	emptyAddr := common.Address{}
+	if addr != emptyAddr {
+		t.Errorf("address mismatch: have %v, want %v", addr, emptyAddr)
+	}
+}
+
+func TestCheckValidatorUnauthorizedAddress(t *testing.T) {
```
aneequesafdar: same as above

consensus/tendermint/core/core.go
```diff
@@ -278,12 +283,18 @@ func (c *core) setCore(r *big.Int, h *big.Int, lastProposer common.Address) {
...
+		c.proposeTimeout = proposeTime
+	}
+	prevoteTime := newTimeout(prevote)
+	if ok := c.prevoteTimeout.set(prevoteTime); !ok {
+		c.prevoteTimeout = prevoteTime
+	}
+	precommitTime := newTimeout(precommit)
+	if ok := c.precommitTimeout.set(precommitTime); !ok {
+		c.precommitTimeout = precommitTime
+	}
```
aneequesafdar: I am really not sure whether this is the best approach to update the core timeouts, you are creating new timout only to update the old one, it might have been better to simply create a new function `update` instead of `set` which would update the timers. I think this should be done similar to how `currentRoundState` is updated. Please correct me if I am missing something.

consensus/tendermint/core/handler.go
```diff
@@ -19,14 +19,39 @@ package core
...
+	"github.com/clearmatics/autonity/consensus/tendermint/events"
+	"github.com/clearmatics/autonity/consensus/tendermint/validator"
+	"github.com/clearmatics/autonity/core/types"
+	"github.com/clearmatics/autonity/log"
 )
 
 // Start implements core.Engine.Start
-func (c *core) Start() error {
+func (c *core) Start(ctx context.Context, chain consensus.ChainReader, currentBlock func() *types.Block, hasBadBlock func(hash common.Hash) bool) error {
+	// prevent double start
```
aneequesafdar: Could you give more context on how the core is started twice?

consensus/tendermint/core/handler.go
```diff
@@ -19,14 +19,39 @@ package core
...
+	if atomic.LoadUint32(c.isStarted) == 1 {
+		return nil
+	}
+	if !atomic.CompareAndSwapUint32(c.isStarting, 0, 1) {
```
aneequesafdar: What is this doing? tried looking into the function the comment was not very helpful. What is it comparing?

consensus/tendermint/core/core.go
```diff
@@ -67,44 +68,48 @@ var (
...
 	messageEventSub         *event.TypeMuxSubscription
 	newUnminedBlockEventSub *event.TypeMuxSubscription
 	committedSub            *event.TypeMuxSubscription
 	timeoutEventSub         *event.TypeMuxSubscription
 	futureProposalTimer     *time.Timer
+	stopped                 chan struct{}
+	isStarted               *uint32
+	isStarting              *uint32
+	isStopping              *uint32
+	isStopped               *uint32
```
aneequesafdar: Why do you want to have isStarted/isStopped isStarting/isStopping separately? Isn't one sufficient?

consensus/tendermint/core/core.go
```diff
@@ -67,44 +68,48 @@ var (
...
+	return &core{
 		config:                config,
 		address:               backend.Address(),
 		logger:                log.New(),
 		backend:               backend,
-		backlogs:              make(map[tendermint.Validator]*prque.Prque),
+		backlogs:              make(map[validator.Validator]*prque.Prque),
 		pendingUnminedBlocks:  make(map[uint64]*types.Block),
 		pendingUnminedBlockCh: make(chan *types.Block),
+		stopped:               make(chan struct{}, 3),
```
aneequesafdar: Why do you have a buffered channel for 3 if you only use it for exiting 2 go routines?

consensus/tendermint/core/handler.go
```diff
@@ -86,87 +127,73 @@ func (c *core) unsubscribeEvents() {
...
-
-func (c *core) handleNewUnminedBlockEvent() {
-	defer func() {
-		if r := recover(); r != nil {
-			debug.PrintStack()
+// TODO: update all of the TypeMuxSilent to event.Feed and should not use backend.EventMux for core internal messageEventSub: backlogEvent, TimeoutEvent
 
-			c.logger.Crit("panic in core.handleNewUnminedBlockEvent", "panic", r)
+func (c *core) handleNewUnminedBlockEvent(ctx context.Context) {
+eventLoop:
```
aneequesafdar: Why do you want to name the loop? break and continue should be sufficient, right?

consensus/tendermint/core/handler.go
```diff
@@ -86,87 +127,73 @@ func (c *core) unsubscribeEvents() {
...
-			debug.PrintStack()
-
-			c.logger.Crit("panic in core.handleConsensusEvents", "panic", r)
-		}
-	}()
-
 	// Start a new round from last height + 1
 	c.startRound(ctx, common.Big0)
 
+eventLoop:
```
aneequesafdar: same as above

consensus/tendermint/core/precommit.go
```diff
@@ -130,16 +128,19 @@ func (c *core) handlePrecommit(ctx context.Context, msg *message) error {
 	return nil
 }
 
-func (c *core) verifyPrecommitCommittedSeal(m *message, precommit tendermint.Vote) error {
-	addressOfSignerOfCommittedSeal, err := types.GetSignatureAddress(PrepareCommittedSeal(precommit.ProposedBlockHash), m.CommittedSeal)
+func (c *core) verifyPrecommitCommittedSeal(addressMsg common.Address, committedSealMsg []byte, proposedBlockHash common.Hash) error {
```
aneequesafdar: Why change function signature and did you fix the seal issue that you were mentioning?

consensus/tendermint/core/precommit.go
```diff
@@ -130,16 +128,19 @@ func (c *core) handlePrecommit(ctx context.Context, msg *message) error {
...
+	addressOfSignerOfCommittedSeal, err := types.GetSignatureAddress(committedSeal, committedSealMsg)
 	if err != nil {
 		c.logger.Error("Failed to get signer address", "err", err)
 		return err
 	}
 
 	// ensure sender signed the committed seal
-	if !bytes.Equal(addressOfSignerOfCommittedSeal.Bytes(), m.Address.Bytes()) {
+	if !bytes.Equal(addressOfSignerOfCommittedSeal.Bytes(), addressMsg.Bytes()) {
+		log.Error("precommit message. seal error", "got", addressMsg.String(), "expected", addressOfSignerOfCommittedSeal.String())
```
aneequesafdar: better error message: `verify precommit seal error`?

miner/stress_clique.go
```diff
@@ -170,7 +171,7 @@ func makeGenesis(faucets []*ecdsa.PrivateKey, sealers []*ecdsa.PrivateKey) *core
 	return genesis
 }
 
-func makeSealer(genesis *core.Genesis) (*node.Node, error) {
+func makeSealer(genesis *core.Genesis, _ consensus.Engine) (*node.Node, error) {
```
aneequesafdar: no reason to have this argument if it is not used

consensus/test/tendermint_test.go
JekaMas: Without this lines you'll get incorrect index and validator values in goroutines.

JekaMas: Here you can find correct approaches and incorrect one https://play.golang.org/p/uOyyJdQT1Ep

consensus/tendermint/core/core.go
JekaMas: At the moment I want to have obvious and simple logic on the places and don't care about optimizations.

consensus/tendermint/core/core.go
JekaMas: To be more sure that it won't be blocked. I can change it to any value >=2

consensus/tendermint/core/handler.go
JekaMas: You cant break outer structure without a label. If you call break inside of for select/switch this break will break select/switch not for.

consensus/tendermint/core/precommit.go
JekaMas: The method used only 2 fields from vote message type, having less params helps to control the code and data races (and it was here).

consensus/tendermint/core/precommit.go
JekaMas: Ok.

consensus/tendermint/core/handler.go
JekaMas: If we already started so isStarted==1, than exit.
If we're starting (swap unsuccessful and returns false) than exit.
If we're not starting, so swap successful, than continue and don't forget to set isStarting=0 before exit.

consensus/tendermint/core/handler.go
JekaMas: It is started 2-3 times and stopped 2-3 times at each node.Start() and node.Stop(). I prefer to add this protection from such cases and then find the cause.

consensus/tendermint/core/core.go
JekaMas: You're updating them using = and at the same time you use them for read operations in other goroutines. It is a data race.

consensus/test/tendermint_test.go
JekaMas: Answered below

consensus/test/tendermint_test.go
JekaMas: We need completely initialized and started nodes.

consensus/test/tendermint_test.go
JekaMas: Nope, it can be false.

consensus/test/tendermint_test.go
JekaMas: I'm not sure, I understand you.

consensus/test/tendermint_test.go
JekaMas: To have clean state if .stop method was called.

consensus/test/tendermint_test.go
JekaMas: Yeap. I want the shortest time between releasing the port and node init.

p2p/discv5/node_test.go
JekaMas: As it said in the commit: on new golang versions the error message is a bit different so we have to change the if statement.

eth/config.go
aneequesafdar: ok that is fine for now just felt `config.Config` could have a better name to represent that it was tendermint's as there are loads of configs

consensus/tendermint/core/handler.go
aneequesafdar: please, could you explain why you want to delete the recover functions?

consensus/tendermint/validator/validator.go
aneequesafdar: I think we should create a bigger issue to refactor the tendermint package, so for the time being it is fine

consensus/test/tendermint_test.go
aneequesafdar: Could you please tell me where is the base test you used so that I can have a look as well

consensus/test/tendermint_test.go
aneequesafdar: same as above

consensus/test/tendermint_test.go
aneequesafdar: I see that 0 means to assign any new free port

consensus/test/tendermint_test.go
aneequesafdar: `func (s *Ethereum) IsListening() bool                  { return true } // Always listening` its never false

consensus/tendermint/core/handler.go
aneequesafdar: I see i forgot that the break will exit the switch and not the for loop which is why I had `return` before

consensus/test/tendermint_test.go
JekaMas: You're right. Fixed.

PR #222, opened 2019-08-01
Tendermint double signing and precommit issue

yazzaoui:
Fixing https://github.com/clearmatics/autonity/pull/185#discussion_r309733019 and https://github.com/clearmatics/autonity/pull/185#discussion_r309619554

### DIFF
--- a/consensus/tendermint/core/core.go
+++ b/consensus/tendermint/core/core.go
@@ -125,7 +125,7 @@ type core struct {
 	lockedValue *types.Block
 	validValue  *types.Block
 
-	currentHeightOldRoundsStates map[int64]roundState
+	currentHeightRoundsState map[int64]*roundState
 
 	proposeTimeout   timeout
 	prevoteTimeout   timeout
@@ -274,7 +274,7 @@ func (c *core) setCore(r *big.Int, h *big.Int, lastProposer common.Address) {
 
 		// Assuming that round == 0 only when the node moves to a new height
 		// Therefore, resetting round related maps
-		c.currentHeightOldRoundsStates = make(map[int64]roundState)
+		c.currentHeightRoundsState = make(map[int64]*roundState)
 		c.futureRoundsChange = make(map[int64]int64)
 	}
 	// Reset all timeouts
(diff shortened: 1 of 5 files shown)


consensus/tendermint/core/core.go
```diff
@@ -83,7 +83,7 @@ func New(backend tendermint.Backend, config *tendermint.Config) Engine {
 		pendingUnminedBlocks:  make(map[uint64]*types.Block),
 		pendingUnminedBlockCh: make(chan *types.Block),
 		valSet:                new(validatorSet),
-		futureRoundsChange:    make(map[int64]int64),
+		futureRoundsChange:    make(map[int64]map[common.Address]bool),
```
JekaMas: Why not to use struct{} instead of bool?

consensus/tendermint/core/core.go
```diff
@@ -125,14 +125,14 @@ type core struct {
 	lockedValue *types.Block
 	validValue  *types.Block
 
-	currentHeightOldRoundsStates map[int64]roundState
+	currentHeightRoundsState map[int64]*roundState
```
JekaMas: Why do we want to use reference here? As far as I know, we change `c.currentRoundState`, and it'll change every reference value.

yazzaoui: This is actually what I want, to have changes in `c.currentHeightRoundsState[currentRound]` be reflected in `c.currentRoundState`.

consensus/tendermint/core/core.go
```diff
@@ -291,13 +291,11 @@ func (c *core) setCore(r *big.Int, h *big.Int, lastProposer common.Address) {
...
-	if r.Int64() > 0 {
-		// This is a shallow copy, should be fine for now
-		c.currentHeightOldRoundsStates[r.Int64()-1] = *c.currentRoundState
-	}
+	// Add a reference of c.currentRoundState to c.currentHeightRoundsState and then update c.currentRoundState
+	// We keep old round messages in c.currentHeightRoundsState in case we need to decide for an old round proposal
+	// while future messages are sent to the backlog which are processed when the step is set to propose
+
+	c.currentHeightRoundsState[r.Int64()] = c.currentRoundState
 	c.currentRoundState.Update(r, h)
```
aneequesafdar: This code doesn't work because you are just copying the "reference" of round state, so every element in the `c.currentHeightRoundsState[r.Int64()]` has the same reference, and thus when c.currentRoundState is updated on line 299 every element in `c.currentHeightRoundsState[r.Int64()]` is also updated, therefore defeating the purpose of keeping track of round states.

The reason it was done this way was that it caused data races. That is why we update the round state and not create a new one, how you are thinking is exactly how I had it before but you need to create a new round state for it to work, otherwise, every element in the list is updated when line 299 is run.

For more information please have a look at this pull request https://github.com/clearmatics/autonity/pull/150

aneequesafdar: in particular this commit https://github.com/clearmatics/autonity/pull/150/commits/fdf744594afd9e359dd14e0d6870ab55f6c712f1

aneequesafdar:
With regards to `futureRoundChange`, it is a short term fix and you are not even implementing your suggested solution of using a `messageSet`. I would rather not have message duplication, the only reason I didn't modify backlog was that you said to change as little as possible form IBFT. But if that is no longer the case than I would rather not use `futureRoundChange` at all.

consensus/tendermint/core/core.go
JekaMas: In all of the values from current round?

aneequesafdar:
closing in favour of #225 and comment above

PR #233, opened 2019-08-19
Update Autonity Contract

yazzaoui:


b00ris:
New version with some changes and fixed tests - https://github.com/clearmatics/autonity/pull/233

PR #235, opened 2019-08-21
Autonity contract integration

b00ris:
Closes https://github.com/clearmatics/autonity/issues/191


params/config.go
```diff
@@ -198,16 +185,14 @@ func (c *CliqueConfig) String() string {
...
+//// IstanbulConfig is the consensus engine configs for Istanbul based sealing.
 type IstanbulConfig struct {
-	Epoch          uint64         `json:"epoch"`             // Epoch length to reset votes and checkpoint
-	ProposerPolicy uint64         `json:"policy"`            // The policy for proposer selection
-	Deployer       common.Address `json:"contract-deployer"` // Address of the validator who deploys contract stored in bytecode
-	Bytecode       string         `json:"bytecode"`          // Bytecode of validators contract // would like this type to be []byte but the unmarshalling is not working
-	ABI            string         `json:"abi"`               // Bytecode of validators contract // ABI of the said contract
+	Epoch                  uint64                   `json:"epoch"`  // Epoch length to reset votes and checkpoint
+	ProposerPolicy         uint64                   `json:"policy"` // The policy for proposer selection
+	AutonityContractConfig *AutonityContractGenesis `json:"autonityContract"`
```
yazzaoui: I think it's better to have `AutonityContractGenesis` in `ChainConfig` directly

params/autonity_contract.go
```diff
@@ -27,19 +28,45 @@ func (ut UserType) IsValid() bool {
...
 	// Address of the validator who deploys contract stored in bytecode
 	Deployer common.Address `json:"deployer" toml:",omitempty"`
-	// Bytecode of validators contract // would like this type to be []byte but the unmarshalling is not working
+	// Bytecode of validators contract
+	// would like this type to be []byte but the unmarshalling is not working
 	Bytecode string `json:"bytecode" toml:",omitempty"`
 	// Json ABI of the contract
 	ABI         string         `json:"abi "toml:",omitempty"`
-	MinGasPrice uint64         `json:"minGasPrice" toml:",omitempty"`
+	MinGasPrice int64          `json:"minGasPrice" toml:",omitempty"`
```
yazzaoui: We need to be sure that `MinGasPrice` is always positive

cmjc:
Looking at `autonity.sol` and:
>/*
>* RemoveValidator
>* Remove validator from validators list. function MUST be restricted to the Authority Account.
>*/
>function RemoveValidator(address _validator) public onlyValidators(msg.sender) {
>        require(validators.length > 1);

Why it is `onlyValidators(msg.sender)` and not `onlyOperator(msg.sender)`?

yazzaoui:
> Looking at `autonity.sol` and:
> 
> > /*
> > 
> > * RemoveValidator
> > * Remove validator from validators list. function MUST be restricted to the Authority Account.
> >   */
> >   function RemoveValidator(address _validator) public onlyValidators(msg.sender) {
> >   require(validators.length > 1);
> 
> Why it is `onlyValidators(msg.sender)` and not `onlyOperator(msg.sender)`?

This PR is only for making the autonity contract interacting with the native client, the latest solidity file can be found here https://github.com/clearmatics/autonity/blob/autonity_contract_updates/contracts/autonity/contract/contracts/Autonity.sol
It's still work in progress and the function's signature might change so I prefer to wait a bit until we have something stable before writing the SRS.

params/config.go
b00ris: The idea was to keep it separate from istanbul/tendermint config.

params/autonity_contract.go
b00ris: fixed

yazzaoui:
LGTM

consensus/istanbul/backend/engine_test.go
```diff
@@ -82,15 +84,23 @@ func getGenesisAndKeys(n int) (*core.Genesis, []*ecdsa.PrivateKey) {
 
 	// generate genesis block
 	genesis := core.DefaultGenesisBlock()
+	//genesis.Alloc=make(core.GenesisAlloc)
```
JekaMas: Should it be removed?

consensus/istanbul/backend/engine_test.go
```diff
@@ -82,15 +84,23 @@ func getGenesisAndKeys(n int) (*core.Genesis, []*ecdsa.PrivateKey) {
...
+	}
 	genesis.Config.Ethash = nil
 	genesis.Difficulty = defaultDifficulty
 	genesis.Nonce = emptyNonce.Uint64()
 	genesis.Mixhash = types.IstanbulDigest
 
 	AppendValidators(genesis, addrs)
+	err := genesis.Config.Istanbul.AutonityContractConfig.AddDefault().Validate()
+	if err != nil {
+		fmt.Println("consensus/istanbul/backend/engine_test.go:102 ", err)
```
JekaMas: Is it a kind of debug?

PR #240, opened 2019-08-27
Add tests for `Consensus restart after more than f nodes crash`

JekaMas:
15 skipped test cases and a bit refactoring

### DIFF
--- a/consensus/test/tendermint_test.go
+++ b/consensus/test/tendermint_test.go
@@ -32,14 +32,14 @@ func TestTendermintSuccess(t *testing.T) {
 
 	cases := []*testCase{
 		{
-			name: "no malicious",
-			numPeers: 5,
+			name:      "no malicious",
+			numPeers:  5,
 			numBlocks: 5,
 			txPerPeer: 1,
 		},
 		{
-			name: "one node - always accepts blocks",
-			numPeers: 5,
+			name:      "one node - always accepts blocks",
+			numPeers:  5,
 			numBlocks: 5,
 			txPerPeer: 1,
 			maliciousPeers: map[int]func(basic consensus.Engine) consensus.Engine{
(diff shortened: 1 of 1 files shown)


consensus/test/tendermint_test.go
```diff
@@ -598,3 +783,40 @@ func runHook(validatorHook hook, test *testCase, block *types.Block, validator *
...
+
+func hookStopNode(nodeIndex int, blockNum uint64) hook {
+	return func(block *types.Block, validator *testNode, tCase *testCase, currentTime time.Time) error {
+		if block.Number().Uint64() == blockNum {
+			err := validator.stopNode()
+			if err != nil {
+				return err
+			}
+
+			tCase.mu.Lock()
```
b00ris: It'd be better if you hide mutex manipulations inside setter.

consensus/test/tendermint_test.go
```diff
@@ -598,3 +783,40 @@ func runHook(validatorHook hook, test *testCase, block *types.Block, validator *
...
+			tCase.mu.Unlock()
+		}
+
+		return nil
+	}
+}
+
+func hookStartNode(nodeIndex int, durationAfterStop float64) hook {
+	return func(block *types.Block, validator *testNode, tCase *testCase, currentTime time.Time) error {
+		tCase.mu.RLock()
```
b00ris: Same

yazzaoui:
lgtm, are the skipped test solved by https://github.com/clearmatics/autonity/pull/212 ?

JekaMas:
@yazzaoui I've run this test only on the `master` branch. They all are failing on it. So then we can merge the PR into the master and then check if any solution works.

aneequesafdar:
I guess you will remove the skip function from the test cases once we are ready to test with basic wal or something else right?

PR #246, opened 2019-08-30
Wal

aneequesafdar:
This is bringing upto date the original implementation (#212). Not all changes were carried over because it didn't make sense to me.

The tests don't pass because every time we start core in the tests multiple times it will access the same directory for wal and return `panic: resource temporarily unavailable`. I am not sure how to fix it.


JekaMas:
We can check if wal is already started.

PR #254, opened 2019-09-03
Run one or more Tendermint consensus tests with `race` flag

JekaMas:
closes https://github.com/clearmatics/autonity/issues/252

`test-race` is introduced, however, Travis CI has memory limitations and we can't use `test-race` in the CI on current travis account.


JekaMas:
@b00ris @yazzaoui @aneequesafdar waiting for the review

PR #261, opened 2019-09-05
Add tendermint core tests (part 1)

screwyprof:


consensus/tendermint/core/backlog_test.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		proposalPayload, _ := Encode(proposal)
+
+		msg := &message{
+			Code: msgProposal,
+			Msg:  proposalPayload,
+		}
+
+		c.storeBacklog(msg, validator.New(common.HexToAddress("0x0987654321")))
```
JekaMas: Could you help me with this piece of code? Where is an asset step?

screwyprof: My bad, forgot to add the assert. Fixed.

consensus/tendermint/core/backlog_test.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		proposal := &Proposal{
+			Round:         big.NewInt(1),
+			Height:        big.NewInt(2),
+			ValidRound:    big.NewInt(1),
+			ProposalBlock: types.NewBlockWithHeader(&types.Header{}),
+		}
+
+		proposalPayload, _ := Encode(proposal)
```
b00ris: Why did you skip error check?

screwyprof: Good catch, thanks. A copy-paste from old tests. Fixed.

PR #266, opened 2019-09-06
redirect gas fee

b00ris:
closes https://github.com/clearmatics/autonity/issues/167


core/state_transition.go
```diff
@@ -222,7 +222,16 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 		}
 	}
 	st.refundGas()
-	st.state.AddBalance(st.evm.Coinbase, new(big.Int).Mul(new(big.Int).SetUint64(st.gasUsed()), st.gasPrice))
+	address:=st.evm.Coinbase
+
+	if st.evm.ChainConfig().AutonityContractConfig!=nil && st.evm.ChainConfig().Istanbul!=nil || st.evm.ChainConfig().Tendermint!=nil {
```
yazzaoui: `&&` has higher precedence over `||` 
checking only if `st.evm.ChainConfig().AutonityContractConfig!=nil` is enough

b00ris: Nice catch! I'll fix it.

params/autonity_contract.go
```diff
@@ -84,6 +85,13 @@ func (ac *AutonityContractGenesis) Validate() error {
 	return nil
 }
 
+func (ac *AutonityContractGenesis) GetContractAddress() (common.Address, error) {
+	if reflect.DeepEqual(ac.Deployer, common.Address{})==true {
```
JekaMas: I think you can use `==` operator

PR #270, opened 2019-09-09
Network economics prototype

b00ris:
Implements network economics


helloworld/genesis-ibft.json
```diff
@@ -6,21 +6,50 @@
...
-      "enode://d73b857969c86415c0c000371bcebd9ed3cca6c376032b3f65e58e9e2b79276fbc6f59eb1e22fcd6356ab95f42a666f70afd4985933bd8f3e05beb1a2bf8fdde@172.25.0.11:30303",
-      "enode://1f207dfb3bcbbd338fbc991ec13e40d204b58fe7275cea48cfeb53c2c24e1071e1b4ef2959325fe48a5893de8ff37c73a24a412f367e505e5dec832813da546a@172.25.0.12:30303",
-      "enode://e766ac390e2d99b559aef773c3656fa8d50df2310496ac26ca6c3fc84e21dabb8a0162cc8e34f938d45e0a8ed04955f8ddf1c380182f8ef17a3f08885064505f@172.25.0.13:30303",
-      "enode://438a5c2cd8fdc2ecbc508bf7362e41c0f0c3754ba1d3267127a3756324caf45e6546b02140e2144b205aeb372c96c5df9641485f721dc7c5b27eb9e35f5d887b@172.25.0.14:30303",
-      "enode://3ce6c053cb563bfd94f4e0e248510a07ccee1bc836c9784da1816dba4b10564e7be1ba42e0bd8d73c8f6274f8e9878dc13814adb381c823264265c06048b4b59@172.25.0.15:30303"
-    ],
-    "glienickeDeployer": "0x0000000000000000000000000000000000000001",
-    "glienickeBytecode": "",
-    "glienickeABI": "",
     "opennetwork": false
```
aneequesafdar: Are we still using `opennetwork`? In blockchain.go:180, the line is commented out. Also, it should be changed to `openNetwork`.

helloworld/genesis-ibft.json
```diff
@@ -6,21 +6,50 @@
...
+          "address": "0xc443C6c6AE98F5110702921138D840e77dA67702",
+          "enode": "enode://438a5c2cd8fdc2ecbc508bf7362e41c0f0c3754ba1d3267127a3756324caf45e6546b02140e2144b205aeb372c96c5df9641485f721dc7c5b27eb9e35f5d887b@172.25.0.14:30303",
+          "type": 2,
+          "stake": 1
+        },
+        {
+          "address": "0x09428e8674496e2d1e965402f33a9520c5fcbbe2",
+          "enode": "enode://3ce6c053cb563bfd94f4e0e248510a07ccee1bc836c9784da1816dba4b10564e7be1ba42e0bd8d73c8f6274f8e9878dc13814adb381c823264265c06048b4b59@172.25.0.15:30303",
+          "type": 2,
+          "stake": 1
```
aneequesafdar: could we have a better name for `type` I don't understand what is meant by just `type`? Is `stake` meant to represent the stake token and for the time being everyone has the same stake?

cmd/autonity/chaincmd.go
```diff
@@ -225,6 +226,12 @@ func initGenesis(ctx *cli.Context) error {
 	if err := json.NewDecoder(file).Decode(genesis); err != nil {
 		utils.Fatalf("invalid genesis file: %v", err)
 	}
+	if genesis.Config.AutonityContractConfig != nil {
```
aneequesafdar: What happens if the autonity contract config is nill? Does that mean we will not be bringing up the network or it will default to ethash?

cmd/autonity/chaincmd.go
```diff
@@ -225,6 +226,12 @@ func initGenesis(ctx *cli.Context) error {
...
+		if err := genesis.Config.AutonityContractConfig.AddDefault().Validate(); err != nil {
+			spew.Dump(genesis.Config.AutonityContractConfig)
```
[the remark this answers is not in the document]
b00ris: It's useful to have errored config after returning an error.

cmd/autonity/main.go
```diff
@@ -276,14 +276,14 @@ func init() {
...
-		console.Stdin.Close() // Resets terminal mode.
+		console.Stdin.Close() // nolint Resets terminal mode.
 		return nil
 	}
 }
 
 func main() {
 	if err := app.Run(os.Args); err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, err) //nolint
```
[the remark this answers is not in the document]
b00ris: Sure.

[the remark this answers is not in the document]
b00ris: Checking error here is useless but we want to have clean lint.

cmd/autonity/chaincmd.go
b00ris: We'll get an error if we run istanbul/tendermint with AutonityContractConfig==nil. But we don't need AutonityContractConfig for ethash or clique

helloworld/genesis-ibft.json
b00ris: Nice catch! I'll fix it.

helloworld/genesis-ibft.json
b00ris: > could we have a better name for type I don't understand what is meant by just type? 
I'm not sure, that i got what did you mean. Do you suggest a name for it?
user.type looks quite simple.
>  Is stake meant to represent the stake token
Yes.

internal/ethapi/api.go
```diff
@@ -540,6 +541,10 @@ func (s *PublicBlockChainAPI) BlockNumber() hexutil.Uint64 {
 	return hexutil.Uint64(header.Number.Uint64())
 }
 
+func (s *PublicBlockChainAPI) AutonityContract() *autonity.Contract {
+	return s.b.AutonityContract()
```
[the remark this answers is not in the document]
b00ris: eth/backend.go:223
eth/backend.go:252
internal/ethapi/backend.go:89
eth/backend.go:324

It's a part of eth backend.

cmd/autonity/chaincmd.go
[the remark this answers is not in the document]
b00ris: It's quite useful to have a dump of errored config.

helloworld/genesis-ibft.json
aneequesafdar: > I'm not sure, that i got what did you mean. Do you suggest a name for it?
user.type looks quite simple for me.

By type do you mean "validator", "operator" and "participant"? If so then is should be rename to `userType`

yazzaoui: I think it's okay to have `type` for the name, it's inside the users array, so should not be confusing, however the value needs to be a string now

params/autonity_contract.go
```diff
@@ -0,0 +1,148 @@
...
+		return errors.New("if user.type is validator then user.enode must be defined")
+	}
+	if len(u.Enode) > 0 {
+		_, err := enode.ParseV4WithResolve(u.Enode)
+		if err != nil {
+			return fmt.Errorf("fail to parse enode for account %v, error:%v", u.Address, err)
+		}
+
+		//todo do we need this check?
+		//if reflect.DeepEqual(u.Address, common.Address{}) {
```
yazzaoui: yeah we need this check, makes no sense if we have a user that has an enode address different of his address, was there an issue with that?

params/autonity_contract.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if reflect.DeepEqual(ac.Operator, common.Address{}) {
+		ac.Operator = DefaultGovernance
+	}
+
+	for i := range ac.Users {
+		n, err := enode.ParseV4WithResolve(ac.Users[i].Enode)
+		if reflect.DeepEqual(ac.Users[i].Address, common.Address{}) && err != nil {
+			ac.Users[i].Address = EnodeToAddress(n)
+		}
```
aneequesafdar: I am not sure what is happening here, you are creating a node object using `enode.ParseV4WithResolve(ac.Users[i].Enode)`, which creates a node struct, but if there is an error shouldn't you test it separately, instead of having && condition with `reflect.DeepEqual(ac.Users[i].Address, common.Address{})`? 

Also, if there is an error surely `n` will be nill, right? So line 74 is not really helping?


I think it should be like the following:
```go
n, err := enode.ParseV4WithResolve(ac.Users[i].Enode)
if err != nil {
	log.Error("enode parse error", "err", err, "user", i)
	return nil //or something more appropriate
}

if reflect.DeepEqual(ac.Users[i].Address, common.Address{}) {
	ac.Users[i].Address = EnodeToAddress(n)
}
```

cmd/autonity/chaincmd.go
```diff
@@ -225,6 +226,12 @@ func initGenesis(ctx *cli.Context) error {
 	if err := json.NewDecoder(file).Decode(genesis); err != nil {
 		utils.Fatalf("invalid genesis file: %v", err)
 	}
+	if genesis.Config.AutonityContractConfig != nil {
+		if err := genesis.Config.AutonityContractConfig.AddDefault().Validate(); err != nil {
```
aneequesafdar: It is a bit strange to add defaults before validating because lines autonity/params/autonity_contract.go:86-94 will not run outside of test cases. It is not something that should be changed if we don't have time for it, it just seemed bit strange.

aneequesafdar: if we keep this as it is then there is no need to make changes for this comment https://github.com/clearmatics/autonity/pull/270/files#r328184408

consensus/clique/snapshot_test.go
```diff
@@ -395,21 +395,27 @@ func TestClique(t *testing.T) {
...
+		genSpec.ExtraData = make([]byte, extraVanity+common.AddressLength*len(signers)+extraSeal)
 		for j, signer := range signers {
-			copy(genesis.ExtraData[extraVanity+j*common.AddressLength:], signer[:])
+			copy(genSpec.ExtraData[extraVanity+j*common.AddressLength:], signer[:])
 		}
-		genesis.Config = new(params.ChainConfig)
-		*genesis.Config = *params.AllCliqueProtocolChanges
-		genesis.Config.EnodeWhitelist = []string{"enode://d73b857969c86415c0c000371bcebd9ed3cca6c376032b3f65e58e9e2b79276fbc6f59eb1e22fcd6356ab95f42a666f70afd4985933bd8f3e05beb1a2bf8fdde@172.25.0.11:30303"}
+		genSpec.Config = new(params.ChainConfig)
+		*genSpec.Config = *params.AllCliqueProtocolChanges
```
aneequesafdar: simple go question: is this a shallow copy `params.AllCliqueProtocolChanges` into `genSpec.Config`?

core/state_transition.go
```diff
@@ -58,6 +59,7 @@ type StateTransition struct {
 	data       []byte
 	state      vm.StateDB
 	evm        *vm.EVM
+	autonityContract *autonity.Contract
```
aneequesafdar: Throughout the codebase you are passing AutonityContract to core.ApplyMessage and core.ApplyTransaction but it is not used anywhere other than line 123, is this meant to be a place holder for future iterations, if not then could you please explain your reasoning?

params/autonity_contract.go
b00ris: We talked about it. And we haven't found the solution. The idea was that validators can want to have separated keys for wallet and validation.

consensus/clique/snapshot_test.go
b00ris: Yes

helloworld/genesis-ibft.json
b00ris: renamed

consensus/istanbul/backend/engine.go
```diff
@@ -362,45 +366,45 @@ func (sb *Backend) FinalizeAndAssemble(chain consensus.ChainReader, header *type
 	return types.NewBlock(header, txs, nil, receipts), nil
 }
 
+type AutonityContractor interface {
```
aneequesafdar: I think this interface is not being used in the istanbul package

consensus/istanbul/backend/engine_test.go
```diff
@@ -597,3 +638,31 @@ func TestWriteCommittedSeals(t *testing.T) {
...
+	}
+	_, err = chain.GetAutonityContract().DeployAutonityContract(chain, h, sdb)
+	if err != nil {
+		t.Log(string(autonity.Sl.Output()))
+		t.Fatal(err)
+	}
+
+	res, err := chain.GetAutonityContract().ContractGetValidators(chain, h, sdb)
+	if err != nil {
+		t.Log(string(autonity.Sl.Output()))
```
aneequesafdar: What does `autonity.Sl.Output()`, I tried looking into the struct logger in the vm package but didn't understand why we need it here or what it is doing

core/state_transition.go
b00ris: No. Nice catch! I forgot to remove it after refactoring.
Removed.

accounts/abi/bind/backends/simulated.go
```diff
@@ -314,7 +314,7 @@ func (b *SimulatedBackend) callContract(ctx context.Context, call ethereum.CallM
 	vmenv := vm.NewEVM(evmContext, statedb, b.config, vm.Config{})
 	gaspool := new(core.GasPool).AddGas(math.MaxUint64)
 
-	return core.NewStateTransition(vmenv, msg, gaspool).TransitionDb()
+	return core.NewStateTransition(vmenv, msg, gaspool, b.blockchain.GetAutonityContract()).TransitionDb()
```
aesmonty: Why do you change the signature of `NewStateTransition` if the new parameter (`contract`) is not used in the function?

consensus/istanbul/backend/engine.go
```diff
@@ -362,45 +367,45 @@ func (sb *Backend) FinalizeAndAssemble(chain consensus.ChainReader, header *type
...
 
 	if header.Number.Int64() == 1 {
-		// Deploy Soma on-blockchain governance contract
-		log.Info("Soma Contract Deployer", "Address", sb.config.Deployer)
-		contractAddress, err := sb.deployContract(chain, header, state)
-		if err != nil {
-			return nil, err
+		log.Info("Autonity Contract Deployer", "Address", chain.Config().AutonityContractConfig.Deployer)
+
+		sb.blockchain.GetAutonityContract().SavedValidatorsRetriever = func(i uint64) (addresses []common.Address, e error) {
```
aesmonty: You are not catching the error in this function, I don't know if this was intended or not.

If yes, you could maybe do `sb.blockchain.GetAutonityContract().SavedValidatorsRetriever, _ = f...` for better clarity.

contracts/autonity/autonity.go
```diff
@@ -0,0 +1,391 @@
...
+
+	_, _, vmerr := evm.Call(sender, ac.Address(), input, gas, value)
+	if vmerr != nil {
+		log.Error("Error Autonity Contract callPerformRedistribution()", "err", err)
+		return vmerr
+	}
+	return nil
+}
+
+func (ac *Contract) AppplyPerformRedistribution(transactions types.Transactions, receipts types.Receipts, header *types.Header, statedb *state.StateDB) error {
```
aesmonty: Typo in the name of the function: `Appply...` -> `Apply...`

contracts/autonity/autonity.go
```diff
@@ -0,0 +1,391 @@
...
+	log.Error("AppplyPerformRedistribution", "header", header.Number.Uint64())
```
aesmonty: Why is this being logged as an `Error` and not as `Debug`?

accounts/abi/bind/backends/simulated.go
aneequesafdar: Yes you are correct if autonity contract field is removed from the state_transition.go then we should also revert function signatures for core.NewStateTransition(), core.ApplyMessage() and ApplyTransaction()

consensus/tendermint/backend/engine.go
```diff
@@ -369,72 +377,66 @@ func (sb *Backend) FinalizeAndAssemble(chain consensus.ChainReader, header *type
...
 			return nil, err
 		}
-		sb.setGlienickeContract(glienickeAddr)
 
 	} else {
-		if sb.getSomaContract() == common.HexToAddress("0000000000000000000000000000000000000000") {
-			sb.setSomaContract(crypto.CreateAddress(sb.config.Deployer, 0))
+		if sb.autonityContractAddress == common.HexToAddress("0000000000000000000000000000000000000000") {
+			sb.autonityContractAddress = crypto.CreateAddress(sb.blockchain.Config().AutonityContractConfig.Deployer, 0)
+			//sb.setGlienickeContract(crypto.CreateAddress(sb.blockchain.Config().GetGlienickeDeployer(), 0))
```
aesmonty: This comment should be removed, right?

consensus/tendermint/backend/engine.go
```diff
@@ -369,72 +377,66 @@ func (sb *Backend) FinalizeAndAssemble(chain consensus.ChainReader, header *type
...
-	defer sb.contractsMu.RUnlock()
-	return sb.glienickeContract
-}
-
-func (sb *Backend) setGlienickeContract(addr common.Address) {
-	sb.contractsMu.Lock()
-	defer sb.contractsMu.Unlock()
-	sb.glienickeContract = addr
-}
+//func (sb *Backend) getSomaContract() common.Address {
```
aesmonty: And also all these comments?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -0,0 +1,377 @@
+pragma solidity ^0.5.1;
```
aesmonty: The Autonity Contract is looking good!
The only thing I am missing is possibly a data structure like:
`mapping (address => uint256) private members_stake_amount` or something like this. I think it would be helpful when we implement `Voting Power` in client.

contracts/autonity/autonity.go
```diff
@@ -0,0 +1,391 @@
...
+	if err != nil {
+		log.Error("contractABI.Pack returns err", "err", err)
+		return common.Address{}, err
+	}
+
+	data := append(contractBytecode, constructorParams...)
+	gas := uint64(0xFFFFFFFF)
+	value := new(big.Int).SetUint64(0x00)
+
+	// Deploy the Soma validator governance contract
```
aesmonty: Soma? What's Soma? :smile:

contracts/autonity/contract/contracts/Autonity.sol
aesmonty: But this is a minor thing, don't bother in changing anything.

consensus/istanbul/backend/engine.go
b00ris: There is no function call here. It's just an assignment.

core/blockchain.go
```diff
@@ -143,6 +144,7 @@ type BlockChain struct {
 	logsFeed      event.Feed
 	blockProcFeed event.Feed
 	glienickeFeed event.Feed
```
aesmonty: Outdated reference, as `glienickeFeed` should be deprecated.

consensus/istanbul/backend/engine.go
b00ris: I might missing something. Could you explain a bit more?

accounts/abi/bind/backends/simulated.go
b00ris: I forgot to clean here. Done.

consensus/istanbul/backend/engine_test.go
b00ris: I was EVM debugging. We don't need it now. I've removed.

consensus/istanbul/backend/engine.go
b00ris: Removed.

params/autonity_contract.go
b00ris: The idea of this condition is: if we have an empty address and we can parse enode. But your code is doing the same(except return. we can't exit from the cycle, but we can use continue).

contracts/autonity/autonity.go
b00ris: Fixed

core/state_processor.go
```diff
@@ -65,8 +73,22 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	if p.config.DAOForkSupport && p.config.DAOForkBlock != nil && p.config.DAOForkBlock.Cmp(block.Number()) == 0 {
 		misc.ApplyDAOHardFork(statedb)
 	}
+
+	var contractMinGasPrice = new(big.Int)
+	if p.autonityContract != nil {
+		minGasPrice, err := p.autonityContract.GetMinimumGasPrice(block, statedb)
+		if err != nil {
```
aesmonty: Boris, what would happen if the `minGasPrice` in the contract has changed? `err` of line 80 would be `nil`, so the `contractMinGasPrice` would never be updated.

I think I am missing something though, could you explain how this works?

core/state_processor.go
```diff
@@ -65,8 +73,22 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
+			contractMinGasPrice.SetUint64(minGasPrice)
+		}
+	}
 	// Iterate over and process the individual transactions
 	for i, tx := range block.Transactions() {
+		if contractMinGasPrice.Uint64() != 0 {
+			if tx.GasPrice().Cmp(contractMinGasPrice) < 1 {
```
yazzaoui: Submitting a transaction with the minimum gas price should be good, so this needs to be `== -1`

core/state_processor.go
b00ris: Are you talking about changing it during a block processing?
If we can get minGasPrice we set contractMinGasPrice to its value.
If contractMinGasPrice is non zero we add the check.

consensus/istanbul/backend/engine.go
aesmonty: My bad, overlooked it. Thanks!

core/blockchain.go
aesmonty: don't bother with these renamings, we can create an issue after this PR gets merged

core/state_processor.go
aesmonty: Imagine that the `min_gas_price` of the Autonity contract is changed. Where in the code do we update the `minimum_gas_price` in the client?

yazzaoui: Should be `if err == nil {`

consensus/tendermint/backend/engine.go
```diff
@@ -354,6 +354,14 @@ func (sb *Backend) FinalizeAndAssemble(chain consensus.ChainReader, header *type
...
 		return nil, err
 	}
+	ac := sb.blockchain.GetAutonityContract()
+	if ac != nil && header.Number.Uint64() > 1 {
+		err = ac.ApplyPerformRedistribution(txs, receipts, header, statedb)
+		if err != nil {
+			log.Error("ApplyPerformRedistribution", "err", err.Error())
+			return nil, err
+		}
+	}
```
aneequesafdar: These changes should also be added to the finalize function above and instead of calling `ApplyPerformRedistribution` on line core/state_processor.go:101 it will automatically be called when finalize is called in core_processor.go:107. Since finalize is called in cmd/autonity/retesteth.go:238 it might be a problem to have a different state for `Finalize() and FinalizeAndAssemble()` as the only difference between them is that one returns a block and the other doesn't. If you think this is of no consequence than I think it is fine

contracts/autonity/contract/contracts/Autonity.sol
b00ris: Current user stake is in users array.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -0,0 +1,377 @@
...
+        return minGasPrice;
+    }
+
+    function checkMember(address _account) public view returns (bool) {
+        return  users[_account].addr == _account;
+    }
+
+    function performRedistribution(uint256 _amount) public onlyDeployer(msg.sender) {
+        require(address(this).balance >= _amount, "not enough funds to perform redistribution");
+        require(stakeholders.length > 0, "there must be stack holders");
```
cmjc: "there must be stack holders" - _stake_ holders

consensus/tendermint/backend/engine.go
b00ris: Yes, you are right. It'd be better to have it in Finalize. But i can't do it quite fast. I'll create a separate issue for it, because we don't have receipts in Finalize.

aneequesafdar: I just realised that the function signature is different for Finalize and FinalizeAndAssemble, I thought they were the same and thus my reasoning.

By changing the signature for Finalize we will have to do it for other algos as well, so after a brief discussion with @yazzaoui we think it should be fine for now. I would rather not make changes in ethash and clique. This actually may be a non-issue, so sorry to make it seem like it was. I will think more about it

PR #280, opened 2019-09-13
Min gas price

b00ris:
closes https://github.com/clearmatics/autonity/issues/168

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -19,6 +19,7 @@ package core
 import (
 	"errors"
 	"fmt"
+	"github.com/clearmatics/autonity/contracts/autonity"
 	"math"
 	"math/big"
 	"sort"
@@ -119,6 +120,8 @@ type blockChain interface {
 	StateAt(root common.Hash) (*state.StateDB, error)
 
 	SubscribeChainHeadEvent(ch chan<- ChainHeadEvent) event.Subscription
+	GetAutonityContract() *autonity.Contract
+	Config() *params.ChainConfig
 }
 
 // TxPoolConfig are the configuration parameters of the transaction pool.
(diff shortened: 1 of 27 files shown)


core/tx_pool.go
```diff
@@ -602,6 +605,17 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 	if err != nil {
 		return err
 	}
+
+	if pool.chain.GetAutonityContract()!=nil {
+		if gp, err:= pool.chain.GetAutonityContract().GetMinimumGasPrice(pool.chain.CurrentBlock(), pool.currentState); err==nil {
```
yazzaoui: Not sure but maybe it's more efficient if we can avoid calling evm for every transaction and save a cached value after processing a block?

b00ris: Sure. But I think it should be a separate issue. Every transaction changes ethereum state. I think we need small research on it,

yazzaoui: Ok but I think it's conflicting now, because in state_processor.go:
https://github.com/clearmatics/autonity/blob/9cc82a18bcf4e24ceac30fa77403f0ec7e8ace3c/core/state_processor.go#L79

min gas price is retrieved before processing the block and the same value is used (which is fine)
maybe we can have same logic for tx pool?

b00ris: From my point of view, we can't. validateTx method is called when we are going to add tx to the local tx pool. TxPool doesn't know about the current block number and it shouldn't. 
We could integrate a kind of caching to the autonity package. I made small research. We could make it 5 times faster.
But I think we shouldn't do it.
Let's Make It (Work->Right->Fast).

yazzaoui: Good job with testing performance and you are absolutely right with respect of making it working before doing any optimization. We'll need anyways to do performance profiling if we want to be serious about it.

yazzaoui:
LGTM

PR #281, opened 2019-09-17
Optimize CI

JekaMas:


.travis.yml
```diff
@@ -1,33 +1,75 @@
...
+        after_success:
+          - $HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci -repotoken $COVERALLS_TOKEN
+        after_failure:
+          - travis_terminate 1
+      - <<: *build
+        name: "unit tests"
+        script:
+          - go run build/ci.go test -coverage $TEST_PACKAGES
+        after_success:
+          - $HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci -repotoken $COVERALLS_TOKEN
```
aneequesafdar: What does this line do?

JekaMas: To collect test coverage from different jobs

.travis.yml
```diff
@@ -1,33 +1,75 @@
...
+        after_failure:
+          - travis_terminate 1
+      - &tendermint
+        stage: "Tests"
+        name: "TestTendermintSuccess,TestTendermintOneMalicious,TestTendermintSlowConnections,TestTendermintLongRun,TestTendermintStartStopSingleNode,TestTendermintStopUpToFNodes"
+        script:
+          - travis_wait 60 go test ./consensus/test/... -v -run='TestTendermintSuccess|TestTendermintOneMalicious|TestTendermintSlowConnections|TestTendermintLongRun|TestTendermintStartStopSingleNode|TestTendermintStopUpToFNodes' -timeout 60m -coverpkg=./... -cover -covermode=atomic -test.coverprofile=coverage.out
```
aneequesafdar: What is the difference between this line and lines 55, 59, 63, 67?

aneequesafdar: Please ignore, I thought they were the same tests

JekaMas:
Each job is run in parallel, we can have no more than 5 jobs running at the same time. So I had to split tests into chunks to have more or less the same execution time for jobs and to have 5 or less jobs per stage.

PR #286, opened 2019-09-19
add run_test.sh to run tests separately

aneequesafdar:
If this doesn't work I will try splitting up the cases further

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 build/_vendor/pkg
 coverage.out
 coverage_all.out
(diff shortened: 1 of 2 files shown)


JekaMas:
Maybe it's better to change the code?

aneequesafdar:
Ok, the problem is that when you run all of the tests together even with `history_size=7` the stack is not able to be restored. So I thought maybe by running the test separately may help.

I can close it for the time being and call each test case as `t.Run()` but that will only help if we call the test cases separately.

PR #296, opened 2019-09-20
Fix update whitelist data race

aneequesafdar:


consensus/test/tendermint_test.go
```diff
@@ -875,25 +875,6 @@ func runTest(t *testing.T, test *testCase) {
 		t.Fatal(err)
 	}
 
-	wg = &errgroup.Group{}
-	for i, validator := range validators {
-		validator := validator
-		i := i
-
-		wg.Go(func() error {
-			log.Debug("peers", "i", i,
```
JekaMas: Why have you removed this test info?

consensus/tendermint/core/core.go
```diff
@@ -302,18 +305,10 @@ func (c *core) setCore(r *big.Int, h *big.Int, lastProposer common.Address) {
...
-	}
-	prevoteTime := newTimeout(prevote)
-	if ok := c.prevoteTimeout.set(prevoteTime); !ok {
-		c.prevoteTimeout = prevoteTime
-	}
-	precommitTime := newTimeout(precommit)
-	if ok := c.precommitTimeout.set(precommitTime); !ok {
-		c.precommitTimeout = precommitTime
-	}
+	c.proposeTimeout.reset(propose)
```
JekaMas: Could you write a test with concurrent reset and get timeout usage?

aneequesafdar:
Yeah i think I am going to wait till Boris fix his dead lock since I was also getting a deadlock.

aneequesafdar:
closing in favour of #300

PR #300, opened 2019-09-24
Fix datarace

b00ris:
```
==================
WARNING: DATA RACE
Write at 0x00c01d871670 by goroutine 817:
  github.com/clearmatics/autonity/p2p.(*Server).UpdateWhitelist()
      /Users/boris/go/src/github.com/clearmatics/autonity/p2p/server.go:406 +0x772
  github.com/clearmatics/autonity/eth.(*Ethereum).glienickeEventLoop()
      /Users/boris/go/src/github.com/clearmatics/autonity/eth/backend.go:635 +0x71b

Previous read at 0x00c01d871670 by goroutine 398:
  github.com/clearmatics/autonity/consensus/test.runTest.func4()
      /Users/boris/go/src/github.com/clearmatics/autonity/consensus/test/tendermint_test.go:1091 +0xe4
  github.com/clearmatics/autonity/vendor/golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/boris/go/src/github.com/clearmatics/autonity/vendor/golang.org/x/sync/errgroup/errgroup.go:57 +0x71
```

### DIFF
--- a/consensus/test/tendermint_test.go
+++ b/consensus/test/tendermint_test.go
@@ -1086,11 +1086,13 @@ func runTest(t *testing.T, test *testCase) {
 		i := i
 
 		wg.Go(func() error {
+			validator.node.Server().Lock.Lock()
 			log.Debug("peers", "i", i,
 				"peers", len(validator.node.Server().Peers()),
 				"staticPeers", len(validator.node.Server().StaticNodes),
 				"trustedPeers", len(validator.node.Server().TrustedNodes),
 				"nodes", len(validators))
+			validator.node.Server().Lock.Unlock()
 			return nil
 		})
 	}
--- a/cmd/autonity/config.go
+++ b/cmd/autonity/config.go
@@ -179,6 +179,5 @@ func dumpConfig(ctx *cli.Context) error {
 		return err
 	}
 
-
 	return nil
 }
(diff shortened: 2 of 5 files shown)


consensus/test/tendermint_test.go
```diff
@@ -1086,11 +1086,13 @@ func runTest(t *testing.T, test *testCase) {
 		i := i
 
 		wg.Go(func() error {
+			validator.node.Server().Lock.Lock()
```
JekaMas: I think it's removed by Aneeque.

JekaMas:
https://github.com/clearmatics/autonity/pull/296/files#diff-452fb4e9f04a68a05a25df13b67e6842L878 Could you have a look at the @aneequesafdar PR ?

aneequesafdar:
So there is a data race on trusted nodes and static nodes we can do it either way but I was also getting a deadlock once I removed (https://github.com/clearmatics/autonity/pull/296/files#diff-452fb4e9f04a68a05a25df13b67e6842L878). So maybe using locks might be better?

aneequesafdar:
Since this was only logging information and while running the tests the only Error messages were printed so I thought it might be best to remove the log, however, it up to you. I didn't want to add getters and setters for trusted and static nodes, I didn't know there was a lock present for the server so it might be best to use it?

b00ris:
I added lock because i thought that it could be useful if we want to write tests on network permissioning feature. We'll need a lock for checking static peers state.

aneequesafdar:
If you think its the best way seems good to me

JekaMas:
@b00ris approved

PR #301, opened 2019-09-25
Add tendermint core tests (part 4)

screwyprof:


consensus/tendermint/core/core_backend_test.go
```diff
@@ -0,0 +1,515 @@
...
+		header := &types.Header{}
+
+		backendMock := NewMockBackend(ctrl)
+		backendMock.EXPECT().Prepare(nil, header).Return(expected)
+
+		c := &core{
+			backend: backendMock,
+		}
+
+		err := c.Prepare(nil, header)
```
aneequesafdar: should this be tested in a separate function called TestCore_Prepare?

consensus/tendermint/core/core_backend_test.go
```diff
@@ -0,0 +1,515 @@
...
+
+	t.Run("invalid params given, errors returned", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		expected := errors.New("some error")
+		header := &types.Header{}
+
+		backendMock := NewMockBackend(ctrl)
+		backendMock.EXPECT().Prepare(nil, header).Return(expected)
```
aneequesafdar: these same arguments were used for valid params above, I am not sure whether is expressed correctly.

Again it may be my misunderstanding.

consensus/tendermint/core/core_backend_test.go
```diff
@@ -0,0 +1,515 @@
...
+		c := &core{
+			backend: backendMock,
+		}
+
+		_, err := c.FinalizeAndAssemble(nil, header, nil, nil, nil, nil)
+		if err != expected {
+			t.Fatalf("Expected %v, got %v", expected, err)
+		}
+	})
+}
```
aneequesafdar: Similar to the above, is it really ok to pass mocks the same parameters and ask them to return different things? 

It just seems a bit strange for me to see that we expect from the mock different return values for the same function arguments.

consensus/tendermint/core/core_backend_test.go
```diff
@@ -0,0 +1,515 @@
...
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		header := &types.Header{}
+		block := types.NewBlockWithHeader(header)
+		expected := errors.New("some error")
+
+		backendMock := NewMockBackend(ctrl)
+		backendMock.EXPECT().FinalizeAndAssemble(nil, header, nil, nil, nil, nil).
+			Return(block, expected)
```
aneequesafdar: If there is an error which we are expecting then FinalizeAndAssemble should not be returning the `block`, right?

PR #302, opened 2019-09-27
Fix freezer goroutine leak

JekaMas:


core/rawdb/freezer.go
```diff
@@ -391,3 +414,16 @@ func (f *freezer) repair() error {
 	atomic.StoreUint64(&f.frozen, min)
 	return nil
 }
+
+func (f *freezer) sleep(d time.Duration) error {
+	t := time.NewTimer(freezerRecheckInterval)
```
yazzaoui: Shouldn't it be `t := time.NewTimer(d)` ?

core/rawdb/freezer.go
```diff
@@ -254,122 +257,142 @@ func (f *freezer) Sync() error {
 func (f *freezer) freeze(db ethdb.KeyValueStore) {
 	nfdb := &nofreezedb{KeyValueStore: db}
 
+	ticker := time.NewTicker(freezerRecheckInterval)
+	defer ticker.Stop()
+	<-ticker.C
```
yazzaoui: could you explain why waiting here was needed?

core/rawdb/freezer.go
JekaMas: You're right. Will fix.

core/rawdb/freezer.go
JekaMas: Its' the same code as it was, but I've rewrote `time.Sleep` to `time.NewTicker`

JekaMas:
Closed in favor of https://github.com/clearmatics/autonity/pull/305

PR #303, opened 2019-09-27
default go race history size

aneequesafdar:


aneequesafdar:
closing since this doesn't fix the memory issue

PR #311, opened 2019-10-10
[Test] Remove resend logic

yazzaoui:
Testing branch


JekaMas:
@yazzaoui as you can see such experiments without understanding the cause of the issue can only lead to more unstable system.

*13 tests failed*

We need to add more logs and debug. Reproduce the failture on e2e level. Grep all logs and write a "red" unit test. Then fix the test and the code.

PR #318, opened 2019-10-17
Consensus State Pull Mode Synchronization 

yazzaoui:
This PR improves greatly the stability of the consensus engine. All tests are passing with Jason's e2e test engine.

Introducing:
- Pull based consensus state synchronization with a new protocol message: new validators joining the network can now ask when they are ready to retrieve the latest consensus state. Also If a validator is stuck in a view for more than 10 secs it now asks a quorum of the network to sync its consensus state.
- Resend mechanism was removed **_for the time being_** , it will likely be reintroduced in the future when we have confirmation from researchers that we can replace Gossiping by this mechanism while still ensuring best effort broadcast.
- Consensus messages are now saved in a ring buffer even if the engine is stopped, they are processed when the engine starts back again. This lets us avoid dropping connection if a validator is not yet mining due to block synchronization.

Bug fixes:
- A critical bug has been discovered with the backlog discarding future round messages, "red" test was written and then fixed.
- Only validators are now saved in the genesis extra data.
- Improvement on the stability of local e2e tests with fixing a bug preventing to receive chain event.

Local e2e tests are still a bit unstable but they appears to be passing much more often than before, planning to continue working on that, witting more tests and fixing if possible go race warnings.


consensus/tendermint/backend/backend_resend_test.go
```diff
@@ -1,299 +0,0 @@
-package backend
-
-import (
```
JekaMas: Could you add the same tests for new feature? I'm not sure that it's a good thing to introduce a few big features without any tests.

consensus/tendermint/backend/backend_test.go
```diff
@@ -369,6 +369,9 @@ func TestSyncPeer(t *testing.T) {
 		b.SetBroadcaster(broadcaster)
 
 		b.SyncPeer(peerAddr1, messages)
+
+		wait := time.NewTimer(time.Second)
```
JekaMas: What does it waiting for?

consensus/tendermint/backend/handler.go
```diff
@@ -28,17 +27,31 @@ import (
...
 )
 
 // Protocol implements consensus.Handler.Protocol
 func (sb *Backend) Protocol() (protocolName string, extraMsgCodes uint64) {
-	return "tendermint", 1
+	return "tendermint", 2 //nolint
+}
+
+func (sb *Backend) HandleUnhandledMsgs() {
+	for unhandled := sb.pendingMessages.Dequeue(); unhandled != nil; unhandled = sb.pendingMessages.Dequeue() {
```
JekaMas: Should it run forever?

consensus/tendermint/backend/handler.go
```diff
@@ -48,7 +61,8 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg) (bool, error) {
 
 	if msg.Code == tendermintMsg {
 		if !sb.coreStarted {
-			return true, ErrStoppedEngine
+			sb.pendingMessages.Enqueue(UnhandledMsg{addr: addr, msg: msg})
```
JekaMas: This logic looks the same as it was in resend. Am I right?

consensus/tendermint/core/core.go
```diff
@@ -136,7 +138,7 @@ type core struct {
 	lockedValue *types.Block
 	validValue  *types.Block
 
-	currentHeightOldRoundsStates   map[int64]roundState
+	currentHeightOldRoundsStates   map[int64]*roundState
```
JekaMas: Why was it changed?

consensus/tendermint/core/backlog_test.go
```diff
@@ -361,8 +398,51 @@ func TestProcessBacklog(t *testing.T) {
 			backlogs:          make(map[validator.Validator]*prque.Prque),
 			currentRoundState: NewRoundState(big.NewInt(2), big.NewInt(3)),
 		}
+		c.storeBacklog(msg, val)
+		c.processBacklog()
+		c.currentRoundState = NewRoundState(big.NewInt(2), big.NewInt(4))
+		c.processBacklog()
+		timeout := time.NewTimer(2 * time.Second)
```
JekaMas: Is it reliable test?

consensus/tendermint/core/handler.go
```diff
@@ -220,6 +229,47 @@ eventLoop:
...
+			timer = time.NewTimer(10 * time.Second)
+		case ev, ok := <-c.syncEventSub.Chan():
+			if !ok {
+				return
+			}
+			event := ev.Data.(events.SyncEvent)
+			log.Info("Processing sync message", "from", event.Addr)
+			c.SyncPeer(event.Addr)
+		case <-ctx.Done():
+			return
```
JekaMas: Why do we ignore the error?

consensus/tendermint/core/unmined_block.go
```diff
@@ -81,6 +81,9 @@ func (c *core) getUnminedBlock() *types.Block {
 // return errFutureHeightMessage if the height of proposal is larger than currentRoundState height
 // return errOldHeightMessage if the height of proposal is smaller than currentRoundState height
 func (c *core) checkUnminedBlockMsg(unminedBlock *types.Block) error {
+	c.currentRoundStateMu.RLock()
```
JekaMas: Could you run e2e tests with race flag?

eth/backend.go
```diff
@@ -192,18 +191,6 @@ func New(ctx *node.ServiceContext, config *Config, cons func(basic consensus.Eng
 		eth.etherbase = crypto.PubkeyToAddress(ctx.NodeKey().PublicKey)
 	}
 
-	if h, ok := eth.engine.(consensus.Handler); ok {
```
JekaMas: I am not sure if this could break the handshake.

JekaMas:
Does all tests are passing if we run python tests for many times? As we did before (for 6-8 hours).

JekaMas:
If all tests are green and python tests was successful for many times, I'm ok with this PR.

consensus/tendermint/backend/backend_resend_test.go
yazzaoui: yes I will try tomorrow

consensus/tendermint/backend/backend_test.go
yazzaoui: `b.syncPeer` starts a goroutine for calling the peer `Send` method and if I don't wait it returns making the test fails. Maybe there is a better way to do it?

consensus/tendermint/backend/handler.go
yazzaoui: `Dequeue()` is popping a value so it should not, but I'm going to write tests to make sure.

consensus/tendermint/core/backlog_test.go
yazzaoui: same as the other issue, we are using a go routine to post the event, so I need to wait a bit after the main thread returns.
Also in those tests @screwyprof was calling `backendMock.EXPECT().Broadcast(gomock.Any(), gomock.Any(), gomock.Any()).Times(0)` but this will always return true because the core isn't started.

consensus/tendermint/core/handler.go
yazzaoui: Only thing I see I can do is to print it?

consensus/tendermint/core/unmined_block.go
yazzaoui: okay

eth/backend.go
yazzaoui: eth.protocol was no longer used since 1.9.0 upstream update.

consensus/tendermint/backend/backend_test.go
JekaMas: Could we cancel that goroutine using context?

consensus/tendermint/core/handler.go
JekaMas: Maybe we could check if ctx.Err() == context.Cancelled , and if not than log an error

JekaMas:
Data races was introduced on the PR

1.
```
WARNING: DATA RACE
Write at 0x00c0005f5830 by goroutine 539:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).setCore()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/core.go:334 +0x731
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).startRound()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/core.go:257 +0x1eb
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleCommit()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/precommit.go:164 +0x187
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleConsensusEvents()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:221 +0xecd

Previous read at 0x00c0005f5830 by goroutine 536:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).updatePendingUnminedBlocks()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/unmined_block.go:52 +0x278
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).storeUnminedBlockMsg()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/unmined_block.go:39 +0x2fb
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleNewUnminedBlockEvent()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:149 +0x5a

Goroutine 539 (running) created at:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:67 +0x82e
  github.com/clearmatics/autonity/miner.(*worker).start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/worker.go:270 +0x2e8
  github.com/clearmatics/autonity/miner.(*Miner).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/miner.go:130 +0x196

Goroutine 536 (running) created at:
  github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:64 +0x7ef
  github.com/clearmatics/autonity/miner.(*worker).start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/worker.go:270 +0x2e8
  github.com/clearmatics/autonity/miner.(*Miner).Start()
      /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/miner.go:130 +0x196
```

 2. 
```
WARNING: DATA RACE
    Read at 0x00c0716a5a88 by goroutine 121:
      bytes.(*Reader).ReadByte()
          /snap/go/current/src/bytes/reader.go:69 +0x78
      github.com/clearmatics/autonity/rlp.(*Stream).readByte()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/rlp/decode.go:1024 +0xc9
      github.com/clearmatics/autonity/rlp.(*Stream).readKind()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/rlp/decode.go:918 +0x3c
      github.com/clearmatics/autonity/rlp.(*Stream).Kind()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/rlp/decode.go:896 +0x1ab
      github.com/clearmatics/autonity/rlp.(*Stream).Bytes()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/rlp/decode.go:663 +0x3f
      github.com/clearmatics/autonity/rlp.decodeByteSlice()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/rlp/decode.go:401 +0x3c
      github.com/clearmatics/autonity/rlp.(*Stream).Decode()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/rlp/decode.go:824 +0x242
      github.com/clearmatics/autonity/p2p.Msg.Decode()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/p2p/message.go:53 +0xcf
      github.com/clearmatics/autonity/consensus/tendermint/backend.(*Backend).HandleMsg()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/backend/handler.go:69 +0x455
      github.com/clearmatics/autonity/consensus/tendermint/backend.(*Backend).HandleUnhandledMsgs()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/backend/handler.go:53 +0x256
    
    Previous write at 0x00c0716a5a88 by goroutine 385:
      bytes.(*Reader).WriteTo()
          /snap/go/current/src/bytes/reader.go:148 +0x1a0
      io.copyBuffer()
          /snap/go/current/src/io/io.go:384 +0x47e
      github.com/clearmatics/autonity/p2p.Msg.Discard()
          /snap/go/current/src/io/io.go:364 +0x86
      github.com/clearmatics/autonity/eth.(*ProtocolManager).handleMsg()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/eth/handler.go:461 +0x799b
      github.com/clearmatics/autonity/eth.(*ProtocolManager).handle()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/eth/handler.go:433 +0xd05
      github.com/clearmatics/autonity/eth.(*ProtocolManager).makeProtocol.func1()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/eth/handler.go:230 +0x276
      github.com/clearmatics/autonity/p2p.(*Peer).startProtocols.func1()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/p2p/peer.go:364 +0x8d
    
    Goroutine 121 (running) created at:
      github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Start()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:69 +0x86e
      github.com/clearmatics/autonity/miner.(*worker).start()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/worker.go:270 +0x2e8
      github.com/clearmatics/autonity/miner.(*Miner).Start()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/miner.go:130 +0x196
      github.com/clearmatics/autonity/miner.(*Miner).update()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/miner/miner.go:111 +0x4d7
    
    Goroutine 385 (running) created at:
      github.com/clearmatics/autonity/p2p.(*Peer).startProtocols()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/p2p/peer.go:363 +0x2c1
      github.com/clearmatics/autonity/p2p.(*Peer).run()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/p2p/peer.go:213 +0x164
      github.com/clearmatics/autonity/p2p.(*Server).runPeer()
          /mnt/sdb/go/src/github.com/clearmatics/autonity/p2p/server.go:1110 +0x3a1
```

yazzaoui:
> It's kinda hack

Haven't found any better way without changing code outside tests, let me know if you have a better solution.

eth/protocol.go
```diff
@@ -41,7 +41,7 @@ const protocolName = "eth"
 var ProtocolVersions = []uint{eth63}
 
 // protocolLengths are the number of implemented message corresponding to different protocol versions.
-var protocolLengths = map[uint]uint64{eth63: 18, eth62: 8}
+var protocolLengths = map[uint]uint64{eth63: 19, eth62: 8}
```
aneequesafdar: this change is unnecessary, this is only required if a new message is introduced at the protocol manager level, any new messages introduced on the consensus level doesn't affect this protocol lengths (I think?)

yazzaoui: we introduced a new message at the eth service layer

aneequesafdar: I see, I looked at the files outside of the consensus package first so I assumed that the message was introduced as another `tendermintMsg`

consensus/tendermint/backend/backend.go
```diff
@@ -165,57 +170,64 @@ func (sb *Backend) postEvent(event interface{}) {
...
-			time.Now(),
-			time.Now(),
-		})
+		ps := sb.broadcaster.FindPeers(targets)
+		count := 0
+		for addr, p := range ps {
+			//ask to quorum nodes to sync, 1 must then be honest and updated
+			if count == valSet.Quorum() {
+				break
+			}
```
aneequesafdar: Unnecessary in my opinion because we cannot assume how many peers will be online therefore in my opinion we should just ask all the peers we are connected to.

consensus/tendermint/backend/backend.go
```diff
@@ -490,22 +502,20 @@ func (sb *Backend) SyncPeer(address common.Address, messages []*tendermintCore.M
...
 
 	sb.logger.Info("Syncing", "peer", address)
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
-	defer cancel()
+	targets := map[common.Address]struct{}{address: {}}
+	ps := sb.broadcaster.FindPeers(targets)
+	p, connected := ps[address]
+	if !connected {
+		return
+	}
```
aneequesafdar: I think this check is not required because you have just connected to the peer so surely it will be always be false

consensus/tendermint/backend/handler.go
aneequesafdar: Are you not always skipping the first message in `pendingMessages`? You are calling Dequeue twice when the for loop is first called or am I misunderstanding the code here?

consensus/tendermint/backend/backend.go
yazzaoui: So here we are putting a maximum limit on the number of peers we ask for sync, this limit is independent on the number of currently connected peers. 
The rationale behind that is that we require a quorum of nodes to submit votes before moving to the next step/round and since two quorums intersect in at least one honest node then one honest node must have the latest state, therefore, we don't need to ask for more than a quorum of nodes.

consensus/test/tendermint_test.go
```diff
@@ -260,37 +260,37 @@ func TestCheckFeeRedirectionAndRedistribution(t *testing.T) {
 		fBefore := func(block *types.Block, validator *testNode, tCase *testCase, currentTime time.Time) error {
 			addr, err := validator.service.BlockChain().Config().AutonityContractConfig.GetContractAddress()
 			if err != nil {
-				t.Fatal(err)
+				return err
```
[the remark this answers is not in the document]
yazzaoui: No has nothing to do with data races. When this test was failing it was hanging the whole test suite because the others thread were stuck waiting.

consensus/tendermint/backend/handler.go
```diff
@@ -61,7 +63,13 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg) (bool, error) {
 
 	if msg.Code == tendermintMsg {
 		if !sb.coreStarted {
-			sb.pendingMessages.Enqueue(UnhandledMsg{addr: addr, msg: msg})
+			buffer := new(bytes.Buffer)
+			if _, err := io.Copy(buffer, msg.Payload); err != nil {
```
aneequesafdar: Could you please explain why you need to copy the message and what problem is this fixing?

consensus/tendermint/core/precommit.go
```diff
@@ -160,7 +160,14 @@ func (c *core) verifyPrecommitCommittedSeal(addressMsg common.Address, committed
 
 func (c *core) handleCommit(ctx context.Context) {
 	c.logger.Debug("Received a final committed proposal", "step", c.currentRoundState.Step())
-	c.startRound(ctx, common.Big0)
+	lastBlock, _ := c.backend.LastCommittedProposal()
+	height := new(big.Int).Add(lastBlock.Number(), common.Big1).Uint64()
+	if height == c.currentRoundState.Height().Uint64() {
```
aneequesafdar: Under what condition does this happen where we receive commit event for already committed block, does this happen at restarts? Also, what problem did this cause?

consensus/tendermint/core/backlog.go
```diff
@@ -134,10 +137,10 @@ func (c *core) processBacklog() {
 				continue
 			}
 			// Push back if it's a future message
-			err := c.checkMessage(round, height)
+			err := c.checkMessage(round, height, Step(msg.Code))
 			if err != nil {
-				if err == errFutureHeightMessage || err == errFutureRoundMessage {
-					logger.Debug("Stop processing backlog", "msg", msg)
+				if err == errFutureHeightMessage || err == errFutureRoundMessage || err == errFutureStepMessage {
+					logger.Debug("Stop processing backlog", "msg", msg, "err", err)
```
aneequesafdar: Could you please explain the reason for enforcing this? I think it is too strict to enforce processing messages based on the step they are in.

aneequesafdar: Could you please explain the reason for enforcing this? I think it is too strict to enforce processing messages based on the step they are in.

consensus/tendermint/backend/backend.go
yazzaoui: Yes I see, that's a valid point, also when I was designing this feature I was thinking about randomly choosing only 2 or 3 peers for this function. I guess there is room to optimize things later.

consensus/tendermint/backend/handler.go
yazzaoui: in eth/handler.go `defer msg.Discard()` is exhausting the payload io.reader

consensus/tendermint/core/precommit.go
yazzaoui: In one of the failed testcase logs it was shown that after sync when he consensus engine was started it could as well receive a commit event for the latest mined block which would cause the round to be started twice resetting the currentroundstate and losing messages.

consensus/tendermint/core/backlog.go
yazzaoui: We had a failed test that where we received a quorum of prevote at propose step and it would not trigger the prevote timer.

consensus/tendermint/backend/backend.go
```diff
@@ -165,57 +171,64 @@ func (sb *Backend) postEvent(event interface{}) {
...
 
-	peers := peersToString(m.peers)
+	if sb.broadcaster != nil && len(targets) > 0 {
+		ps := sb.broadcaster.FindPeers(targets)
+		for addr, p := range ps {
+			ms, ok := sb.recentMessages.Get(addr)
+			var m *lru.ARCCache
+			if ok {
+				m, _ = ms.(*lru.ARCCache)
+				if _, k := m.Get(hash); k {
```
b00ris: How to handle the situation if the peer's lost messages? 
Can't it get them again?

consensus/tendermint/backend/backend_test.go
```diff
@@ -619,3 +814,11 @@ func makeBlockWithoutSeal(chain *core.BlockChain, engine *Backend, parent *types
 
 	return block, nil
 }
+
+// PrepareCommittedSeal returns a committed seal for the given hash
+func PrepareCommittedSeal(hash common.Hash) []byte {
+	var buf bytes.Buffer
+	buf.Write(hash.Bytes())
+	buf.Write([]byte{byte(2)})
```
b00ris: why do we need this byte?

consensus/tendermint/backend/handler.go
```diff
@@ -17,38 +17,73 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			// nothing to do
+		}
+
+		addr := unhandled.(UnhandledMsg).addr
+		msg := unhandled.(UnhandledMsg).msg
+		_, _ = sb.HandleMsg(addr, msg)
```
b00ris: I think we need to log an error

consensus/tendermint/core/roundstate.go
```diff
@@ -121,6 +120,13 @@ func (s *roundState) Step() Step {
 	return s.step
 }
 
+func (s *roundState) State() (*big.Int, *big.Int, uint64) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.height, s.round, uint64(s.step)
```
b00ris: returning reference value from guarded method

consensus/test/tendermint_test.go
```diff
@@ -39,8 +39,6 @@ func TestTendermintSuccess(t *testing.T) {
 		t.Skip("skipping test in short mode")
 	}
 
-	defer goleak.VerifyNone(t)
```
b00ris: Don't we need goleak?

consensus/tendermint/backend/backend.go
yazzaoui: In normal operation mode, we try to avoid sending a message if we know the recipient is likely to have it already received. If the peer has lost messages and is stuck in a round, then it wlil ask for sync. The important thing to notice is that we don't check the arccache when replying to a sync query and we're sending all received messages.

consensus/tendermint/backend/backend_test.go
yazzaoui: It's better for security, to distinguish a direct signature of the block hash of a signature of the committed seal.

consensus/test/tendermint_test.go
yazzaoui: question for @JekaMas

consensus/tendermint/core/roundstate.go
yazzaoui: Yes you're right we should make a copy

consensus/test/tendermint_test.go
JekaMas: It'd better to have it.

PR #322, opened 2019-10-22
[WIP] Extend Autonity Contract RPC

aesmonty:
The goal of these changes is to make Autonity, and the Autonity contract more user-friendly.

**Changes**

- Extend Autonity Contract RPC API.
- Add relevant Web3 methods.
- Rename namespace from Tendermint to Autonity.
- Remove deprecated references to Soma or Glienicke.

### DIFF
--- a/consensus/tendermint/backend/api.go
+++ b/consensus/tendermint/backend/api.go
@@ -19,9 +19,9 @@ package backend
 import (
 	"github.com/clearmatics/autonity/common"
 	"github.com/clearmatics/autonity/consensus/tendermint/core"
-	"github.com/clearmatics/autonity/rpc"
-	"github.com/clearmatics/autonity/params"
 	blockchain "github.com/clearmatics/autonity/core"
+	"github.com/clearmatics/autonity/params"
+	"github.com/clearmatics/autonity/rpc"
 )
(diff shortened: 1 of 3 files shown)


Shirikatsu:
Don't we still want compatibility with IBFT? In the case where users start the client with IBFT enabled instead of Tendermint, does this still support that?

aesmonty:
@JekaMas We still have that functionality, under `autonity_getWhitelist`, as the Whitelist functionality is derived from the Autonity contract, not from tendermint itself.
Will update the readme!

consensus/tendermint/backend/api.go
```diff
@@ -68,3 +69,33 @@ func (api *API) GetContractABI() string {
...
+	return api.chain.Config().AutonityContractConfig.Operator
+}
+
+// Get information from Autonity Users
+func (api *API) GetUsersInfo() []params.User {
+	return api.chain.Config().AutonityContractConfig.Users
+}
+
+// Get information from Autonity Validators
+func (api *API) GetValidatorsInfo() []params.User {
```
b00ris: Why do you get users from genesis? I think we need it from the autonity contract.

PR #327, opened 2019-10-29
Sync pull msg discard fix

yazzaoui:
- Adding support for TC7 test ( this change slightly how e2e tests are being performed).
- Solving issue with io.reader msg.payload not being properly saved by the handler due to discarding.
- Add new backlog error for future step messages, as prevote and precommit messages canot be properly processed while currently being on propose step.

### DIFF
--- a/consensus/test/tendermint_test.go
+++ b/consensus/test/tendermint_test.go
@@ -1040,7 +1040,7 @@ func runTest(t *testing.T, test *testCase) {
 		t.SkipNow()
 	}
 
-	log.Root().SetHandler(log.LvlFilterHandler(log.LvlDebug, log.StreamHandler(os.Stderr, log.TerminalFormat(true))))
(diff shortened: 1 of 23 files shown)


consensus/test/tendermint_test.go
```diff
@@ -1465,11 +1492,15 @@ func sendTransactions(t *testing.T, test *testCase, validators []*testNode, txPe
...
 					}
+				case <-stopch:
+					return nil
 				case <-ctx.Done():
 					return ctx.Err()
 				}
 			}
-
+			stopOnce.Do(func() {
+				close(stopch)
```
JekaMas: Why do we need this? Could we just cancel the context?

yazzaoui: Yeah I tried actually but then it returns `ctx.Err()` and makes the test fails. We can change it to return nil but then we'll have false positives.

yazzaoui:
Just to make it clear guys, this branch is still WIP

consensus/test/tendermint_test.go
JekaMas: You can check the value of the error.

PR #335, opened 2019-11-08
NoQuorum tests cases added

screwyprof:


consensus/test/tendermint_test.go
```diff
@@ -960,6 +996,8 @@ type testCase struct {
 	stopTime             map[int]time.Time
 	genesisHook          func(g *core.Genesis) *core.Genesis
 	mu                   sync.RWMutex
+	noQuorumAfter        uint64
```
b00ris: Could you add a comment?
Is here time or blocks?

screwyprof: I guess I'll rename the variable to NoQuorumAfterBlock

PR #337, opened 2019-11-12
Fix validators list in extra

b00ris:
There were not only validators users in genesis extra field.
We need only validators.


contracts/autonity/autonity.go
```diff
@@ -89,11 +90,11 @@ func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *
 	sender := vm.AccountRef(chain.Config().AutonityContractConfig.Deployer)
 
 	//todo do we need it?
-	//validators, err = ac.SavedValidatorsRetriever(1)
-	//sort.Sort(validators)
+	//users, err = ac.SavedValidatorsRetriever(1)
```
JekaMas: Could we remove commented code?

contracts/autonity/autonity.go
```diff
@@ -102,20 +103,20 @@ func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *
 		return common.Address{}, err
 	}
 
-	ln := len(chain.Config().AutonityContractConfig.GetValidatorUsers())
-	validators := make(common.Addresses, 0, ln)
+	ln := len(chain.Config().AutonityContractConfig.Users)
```
JekaMas: Could you put here a comment why and where we should use Users and Validators?

b00ris: added. But this case is just removed useless allocation.

yazzaoui:
can you check please if https://github.com/clearmatics/autonity/pull/318/commits/3422d79a636d054ba04446881eb7e5451023e141 is not fixing this issue?

b00ris:
Yes, it's the same. But my pr has one more change and comments.

contracts/autonity/autonity.go
```diff
@@ -134,34 +135,27 @@ func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *
...
+	users := make(common.Addresses, 0, ln)
 	enodes := make([]string, 0, ln)
 	accTypes := make([]*big.Int, 0, ln)
 	participantStake := make([]*big.Int, 0, ln)
-	for _, v := range chain.Config().AutonityContractConfig.Users { // Todo: Looping through all the users here
-		validators = append(validators, v.Address) //Todo: Is this appending all the users as validators?
+	for _, v := range chain.Config().AutonityContractConfig.Users {
```
aesmonty: I think there is still a problem here. In the Extradata, there should be only validators, and here we are adding all users, right? @b00ris

PR #346, opened 2019-11-26
Network economics metrics

Jason-Zhangxin-Chen:
Metric measurement for network economics, and its unit test cases from both client side in golang and solidity side in node.js.
1. Measure the block reward distributions of each block and related stakeholders
2. Measure validators, stakeholders, participants stake, balance, commission rate
3. Measure gloable parameters: min gas price and stake supply.
4. Metrics of removed user will be removed from metric registry for memory recycle.
5. Metrics of per block will cause memory exhausted, it was discussed to apply a measurement window base on block height.


contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -136,7 +154,7 @@ contract Autonity {
             }
         }
         stakeSupply = stakeSupply.sub(u.stake);
-
+        _removeFromArray(u.addr, usersList);
```
aesmonty: Good catch :)

contracts/autonity/autonity.go
```diff
@@ -345,11 +389,26 @@ func (ac *Contract) callPerformRedistribution(state *state.StateDB, header *type
 
 	value := new(big.Int).SetUint64(0x00)
 
-	_, _, vmerr := evm.Call(sender, ac.Address(), input, gas, value)
+	ret, _, vmerr := evm.Call(sender, ac.Address(), input, gas, value)
```
aesmonty: Since now we are returning the Redistribution data here, it seems a bit of an overkill to have `PerformRedistribution`, `callPerformRedistribution` and `ApplyPerformRedistribution`. What do you think @Jason-Zhangxin-Chen ?

contracts/autonity/economic_metric.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	// clean up useless metrics if there exists.
+	em.cleanUselessMetrics(v.Accounts, height)
+}
+
+func (em *EconomicMetrics) SubmitRewardDistributionMetrics(v *RewardDistributionMetaData, height uint64) {
+	if len(v.Holders) != len(v.Rewardfractions) {
+		log.Warn("reward fractions does not distribute to all stake holder.")
+		return
```
aesmonty: Why a `return` here? Don't we want to submit the metrics even though that condition does not hold?

Jason-Zhangxin-Chen: > Why a `return` here? Don't we want to submit the metrics even though that condition does not hold?
The number of stake holders shares the rewardes should be equal to the number of fractions for them. It is a checking from client side.

contracts/autonity/economic_metric_test.go
```diff
@@ -0,0 +1,286 @@
...
+	t.Run("test generate user metrics ID", func(t *testing.T) {
+		em := &EconomicMetrics{}
+		a := common.Hex2Bytes(testAddress1)
+		address := common.BytesToAddress(a)
+		stakeID, balanceID, commissionRateID, _ := em.generateUserMetricsID(address, Participant)
+		expectedStakeID := fmt.Sprintf(UserMetricIDTemplate, address.String(), "participant", "stake")
+		expectedBalanceID := fmt.Sprintf(UserMetricIDTemplate, address.String(), "participant", "balance")
+		expectedCommissionRateID := fmt.Sprintf(UserMetricIDTemplate, address.String(), "participant", "commissionrate")
+		if stakeID != expectedStakeID || balanceID != expectedBalanceID || commissionRateID != expectedCommissionRateID {
+			t.Fatal("test cas failed.")
```
aesmonty: small typo here --> "case"

contracts/autonity/economic_metric_test.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+	})
+
+	t.Run("test generate user metrics ID with wrong role", func(t *testing.T) {
+		em := &EconomicMetrics{}
+		a := common.Hex2Bytes(testAddress1)
+		address := common.BytesToAddress(a)
+		_, _, _, err := em.generateUserMetricsID(address, 3)
+		if err == nil {
+			t.Fatal("test cas failed.")
```
aesmonty: small typo here --> "case"

contracts/autonity/economic_metric_test.go
```diff
@@ -0,0 +1,286 @@
...
+		if blockRewardMetricID != fmt.Sprintf(BlockRewardBlockMetricID, blockNumber) {
+			t.Fatal("wrong result expected.")
+		}
+	})
+
+	t.Run("test generate reward distribution metric ID", func(t *testing.T) {
+		em := &EconomicMetrics{}
+		a := common.Hex2Bytes(testAddress1)
+		address := common.BytesToAddress(a)
+		metricID := em.generateRewardDistributionMetricsID(address, Participant, uint64(2))
```
aesmonty: Question: Since `Participants` will never get `rewards`, should this case against participant type? Do the participant type need the "reward distribution metric ID"?

Jason-Zhangxin-Chen: > Question: Since `Participants` will never get `rewards`, should this case against participant type? Do the participant type need the "reward distribution metric ID"?

This test case is focus on if the metric ID generation works as expected. So it does not need to know the full logic context behind.

contracts/autonity/economic_metric_test.go
Jason-Zhangxin-Chen: > small typo here --> "case"

Thanks, Andres, it was fixed.

contracts/autonity/economic_metric_test.go
Jason-Zhangxin-Chen: > small typo here --> "case"

Thanks, Andres, it was fixed too.

contracts/autonity/autonity.go
Jason-Zhangxin-Chen: > Since now we are returning the Redistribution data here, it seems a bit of an overkill to have `PerformRedistribution`, `callPerformRedistribution` and `ApplyPerformRedistribution`. What do you think @Jason-Zhangxin-Chen ?

This part of code is under refactoring by Booris and Youssef's contract_upgrade_prototype branch, they split the distribution and finalize, now we can keep it like this, I will continue to work on their branch to refine it by respecting their refactoring ideas.

PR #348, opened 2019-12-03
Convert to eth for prometheus store

Jason-Zhangxin-Chen:
Convert Wei to ETH for account balance metrics since prometheus's 64 bits registry limits.


contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -303,7 +304,9 @@ contract Autonity {
             _user.addr.transfer(reward);
             rewardfractionlist[i] = reward;
         }
-        RewardDistributionData memory rd = RewardDistributionData(true, stakeholders, rewardfractionlist, _amount);
+        total_reward += _amount;
```
JekaMas: Is it safe operation?

contracts/autonity/economic_metric.go
```diff
@@ -87,19 +94,33 @@ type EconomicMetrics struct {
...
+
+	// balance need to be converted from wei to ETH to be saved in prometheus.
+	if isWei {
+		gauge := metrics.GetOrRegisterGaugeFloat64(name, nil)
+		val2, _ := new(big.Rat).SetFrac(value, big.NewInt(params.Ether)).Float64()
+		gauge.Update(val2)
+		return
+	}
+	// stake will be keep in raw stake.
+	gauge := metrics.GetOrRegisterGauge(name, nil)
```
JekaMas: A bit of code duplication. Could we avoid it?

contracts/autonity/contract/contracts/Autonity.sol
aesmonty: What's the purpose of this? I don't think we need that. Also, `total_reward` would eventually overflow.

contracts/autonity/economic_metric.go
```diff
@@ -79,6 +85,7 @@ type RewardDistributionMetaData struct {
 	Holders         []common.Address `abi:"stakeholders"`
 	Rewardfractions []*big.Int       `abi:"rewardfractions"`
 	Amount          *big.Int         `abi:"amount"`
+	Totalrewards    *big.Int         `abi:"totalrewards"`
```
aesmonty: We could store TotalRewards at the client level maybe, but IMO doing it at the contract is not a good idea.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: Thanks Eugene, Theorical speaking, although it is in uint256, but it would overflow finally. For any sides, client side or layer contract side, it will overflow.
So, I would like to suggest to remove this metrics from client since we have reward metrics base on per block, so the totall reward in the history could be calculated from BI system. What do you think?

Jason-Zhangxin-Chen: Thanks Guys, Theorical speaking, although it is in uint256, but it would overflow finally. For any sides, client side or layer contract side, it will overflow.
So, I would like to suggest to remove this metrics from client since we have reward metrics base on per block, so the totall reward in the history could be calculated from BI system. What do you think?

contracts/autonity/economic_metric.go
Jason-Zhangxin-Chen: In the client side level, it would over flow too, and prometheus does not have sufficient word width to save this totall rewards from genesis block in the history.
I would suggest that BI system could calculate it from the metric which measures each block's reward.

contracts/autonity/contract/contracts/Autonity.sol
JekaMas: How do you think maybe it worth it to use safe math everywhere in Autonity contract?

Jason-Zhangxin-Chen:
@JekaMas @aesmonty @b00ris, Reveiw findings were closed. Please help to approved it, thanks.

contracts/autonity/economic_metric.go
Jason-Zhangxin-Chen: It is fixed with better code in swith statement. The two method for metric record are different, one for float64, another one for int64 provided by metric framework, so it is hard to say it is duplicated code.

PR #349, opened 2019-12-04
Committee selection I

aesmonty:
- Added committee selection on the Autonity contract.
- Refactored tests for Autonity contract.
- Added tests for the committee selection.

_Note: I am leaving outside this PR the committee selection integration with the client. This will come in the next PR._


contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -288,6 +327,89 @@ contract Autonity {
...
+
+        for (uint256 i = 0;i < validators.length; i++) {
+            User memory _user = users[validators[i]];
+            validatorList[i] =_user;
+        }
+
+        // If there are more validators than seats in the committee
+        if (validatorList.length > committeeSize) {
+            // sort validators by stake in ascending order
+            sortedValidatorList = sortByStake(validatorList);
```
JekaMas: Do we need this sort? Maybe we can sort only in views?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -169,6 +175,15 @@ contract Autonity {
         emit SetMinimumGasPrice(_value);
     }
 
+    /*
+    * setCommitteeSize
+    * Set the maximum size of the commitee, restricted to the Governance Operator account
+    */
+
+    function setCommitteeSize (uint256 _size) public onlyOperator(msg.sender) {
```
JekaMas: setCommittee is only for Deployer, but setCommitteeSize is for Operators only.

contracts/autonity/contract/contracts/Autonity.sol
aesmonty: What do you mean with "only in views" @JekaMas ?

contracts/autonity/contract/contracts/Autonity.sol
aesmonty: This is intended behavior:
- The `onlyDeployer` is to be automated at the end of each block (with this implementation we would change the committee at each new height).
- The `setCommitteeSize` should not be actioned per block, only when the Operator (or the Governance members when we have implemented that) decide so.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -59,10 +61,14 @@ contract Autonity {
         UserType userType;
         uint256 stake;
         string enode;
+        // uint256 selfStake;
```
b00ris: ?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -100,7 +106,7 @@ contract Autonity {
             require(_participantAddress[i] != address(0), "Addresses must be defined");
             UserType _userType = UserType(_participantType[i]);
             address payable addr = address(uint160(_participantAddress[i]));
-            _createUser(addr, _participantEnode[i], _userType, _participantStake[i]);
+           _createUser(addr, _participantEnode[i], _userType, _participantStake[i]);
```
b00ris: ?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -236,10 +251,20 @@ contract Autonity {
...
+    * getStakeholders
+    *
+    * Returns the macro stakeholders list
+    */
+
     function getStakeholders() public view returns (address[] memory) {
         return stakeholders;
     }
 
+    function getCommittee() public view returns (User[] memory) {
```
b00ris: As i remember, it's not possible to test.(returning array of structs)

contracts/autonity/contract/contracts/Autonity.sol
b00ris: @JekaMas idea is to avoid making a lot of sStore operations.
You can make sort on getCommittee function(public view).

params/autonity_contract.go
```diff
@@ -135,7 +136,7 @@ func (u *User) Validate() error {
 			return fmt.Errorf("fail to parse enode for account %v, error:%v", u.Address, err)
 		}
 
-		//todo do we need this check?
+		//todo do we need this check? --> I think we do!
```
b00ris: we might want to allow to get reward to separate address. It should be discussed)

contracts/autonity/contract/contracts/Autonity.sol
aesmonty: It is just a reminder, as we will need it when we implement stake delegation.

PR #351, opened 2019-12-09
Malicious tests

JekaMas:
Test cases
* add an extra malicious validator address to the proposed block
* remove a validator from the proposed block
* change a normal validator address with a broken one


consensus/tendermint/core/mock_engine_modify_validator_list.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		c.T.Error("c.core.FinalizeAndAssemble returned error:", err, "Expected nil")
+	}
+
+	if header.Number.Cmp(big.NewInt(0)) == 0 {
+		// skip genesis
+		return block, nil
+	}
+
+	if c.address != header.Coinbase {
```
aesmonty: @JekaMas Could you explain to me this check, please?

JekaMas: It equals to "if the block is not proposed by me".

PR #354, opened 2019-12-13
Check for memory leaks

### LINKED ISSUE #268: Tests to check memory leaks

JekaMas:
closes #268


JekaMas:
@screwyprof @b00ris @yazzaoui @Jason-Zhangxin-Chen @aneequesafdar

JekaMas:
@yazzaoui The problem is how to distinguish a leak from normal memory growth, eg. txPool, database, blockchain database - all these things are growing with time.
I think we only can compare go-ethereum vs autonity or do a proper memory profiling.

Issue #218, opened 2019-07-29
Refactor Validator Set Logic


aesmonty:
# Problem

There seems to be a lot of code duplication while dealing with the Validator Set logic. Part of this is inherited from IBFT implementation.

In particular, the `validatorSet` type in [tendermint/core/validator_set.go](https://github.com/clearmatics/autonity/blob/tendermint/consensus/tendermint/core/validator_set.go) and the `defaultSet` type in [tendermint/validator/default.go](https://github.com/clearmatics/autonity/blob/tendermint/consensus/tendermint/validator/default.go) share most of their functionality, so having only one validator set type would be more appropriate.

# Solution

Refactor the validator logic, grouping functionality of the types mentioned.

PR #359, opened 2019-12-24
New block header structure

yazzaoui:
Working implementation of [Option 2](https://github.com/clearmatics/autonity/issues/343) with a new block header structure containing the following new fields:
```go
	Committee          Committee `json:"committee"           gencodec:"required"`
	ProposerSeal       []byte    `json:"proposerSeal"        gencodec:"required"`
	Round              *big.Int  `json:"round"               gencodec:"required"`
	CommittedSeals     [][]byte  `json:"committedSeals"      gencodec:"required"`
	PastCommittedSeals [][]byte  `json:"pastCommittedSeals"  gencodec:"required"`
```
with 
```go
type Committee []CommitteeMember
type CommitteeMember struct {
	Address     common.Address `json:"address"            gencodec:"required"`
	VotingPower *big.Int       `json:"votingPower"        gencodec:"required"`
}
```
Please note that I dropped IBFT support.

I took as well the opportunity to improve a bit the use of the validator interface ( https://github.com/clearmatics/autonity/issues/218)  by directly implementing it for the CommitteeMember object, no need to have a separate "validator" object.

Waiting for your comments.


aneequesafdar:
Can you please separate the pull request into multiple commits, so that we don't get confused between ibft removal and other renaming/removal for validator functions, this would really help. Thanks

JekaMas:
Do we sure that it's the only change we need? If we choose approach to add and change header fields than we have to introduce hard forks if we need to change or add something, eg. subcommittee - ring signature, master public key or something. What if we need additional data to select committee? The current tip10 approach is temporal as far as I know.

consensus/tendermint/validator/validator_interface.go
```diff
@@ -25,10 +27,20 @@ import (
...
 type Validator interface {
 	// Address returns address
-	Address() common.Address
+	Addr() common.Address
 
 	// String representation of Validator
 	String() string
+
+	// Return Voting Power
+	VP() *big.Int
```
JekaMas: Could we use full names?
The same for Addr()

cmd/clef/main.go
```diff
@@ -751,6 +751,11 @@ func testExternalUI(api *core.SignerAPI) {
 			[]byte("Extra data Extra data Extra data  Extra data  Extra data  Extra data  Extra data Extra data"),
 			common.HexToHash("0x0000H45H"),
 			types.BlockNonce{},
+			types.Committee{},
```
JekaMas: I think we don't need to init with zero values explicitly.

thomasmodeneis:
> Working implementation of [Option 2](https://github.com/clearmatics/autonity/issues/343) with a new block header structure containing the following new fields:
> 
> ```go
> 	Committee          Committee `json:"committee"           gencodec:"required"`
> 	ProposerSeal       []byte    `json:"proposerSeal"        gencodec:"required"`
> 	Round              *big.Int  `json:"round"               gencodec:"required"`
> 	CommittedSeals     [][]byte  `json:"committedSeals"      gencodec:"required"`
> 	PastCommittedSeals [][]byte  `json:"pastCommittedSeals"  gencodec:"required"`
> ```
> 
> with
> 
> ```go
> type Committee []CommitteeMember
> type CommitteeMember struct {
> 	Address     common.Address `json:"address"            gencodec:"required"`
> 	VotingPower *big.Int       `json:"votingPower"        gencodec:"required"`
> }
> ```
> 
> Please note that I dropped IBFT support.
> 
> I took as well the opportunity to improve a bit the use of the validator interface ( #218) by directly implementing it for the CommitteeMember object, no need to have a separate "validator" object.
> 
> Waiting for your comments.

Would be nice to know why the IBFT support was dropped, it was quite of an interesting consensus :)

Since I managed to merge this before the support was dropped, those who are interested in using IBFT still have a place to go: https://github.com/smilofoundation/go-smilo

Cheers.

Cheers.

aneequesafdar:
Hi @thomasmodeneis 

IBFT had no formal description it was only described in  [EIP 650](https://github.com/ethereum/EIPs/issues/650 making it difficult to reason about its correctness. After substantial effort in understanding the EIP through referenced work, we came to the conclusion that IBFT was breaking liveness and safety properties. 

Since IBFT seemed very close to Tendermint we decided to implement Tendermint as it was well described and reasoned about in [this paper](https://arxiv.org/abs/1807.04938). 

Consensys performed a much rigorous analysis of why IBFT is broken. You can find more details [here](https://arxiv.org/abs/1807.04938).

PR #364, opened 2019-12-30
Topology lexer

JekaMas:
This is a package to parse graphs and subgraphs from markdown files. The result graph object has 3 methods: `GetEdges()` to get all nodes connected to the node, `GetNames()` to get all nodes names specified in the markdown spec and `SetNodeName()` to set node index used in the test.

It implements the core of the spec http://mermaid-js.github.io/mermaid/#/flowchart

More examples can be seen in the tests.

Nolint:govet was added because govet cant parse the tags of the lexer builder.

part of https://github.com/clearmatics/autonity/issues/355


JekaMas:
@Jason-Zhangxin-Chen how do you think maybe it worst it to create a new repo for the tool and use it directly from Python https://hackernoon.com/extending-python-3-in-go-78f3a69552ac ?

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen how do you think maybe it worst it to create a new repo for the tool and use it directly from Python https://hackernoon.com/extending-python-3-in-go-78f3a69552ac ?

I need to check if python have similar extension already, if it could be adapted into the remote e2e test framework, the topology configuration would improve a lots. Thanks @JekaMas .

PR #368, opened 2020-01-02
Use type embedding with original Ethereum header struct

yazzaoui:
An alternate approach for defining the block header using composition with the original block header structure.


core/types/block.go
```diff
@@ -133,23 +138,7 @@ func (h *Header) Hash() common.Hash {
...
-		Difficulty  *big.Int
-		Number      *big.Int
-		GasLimit    uint64
-		GasUsed     uint64
-		Time        uint64
-		Extra       []byte
-		MixDigest   common.Hash
-		Nonce       BlockNonce
-	}{h.ParentHash, h.UncleHash, h.Coinbase, h.Root, h.TxHash, h.ReceiptHash, h.Bloom, h.Difficulty, h.Number, h.GasLimit, h.GasUsed, h.Time, h.Extra, h.MixDigest, h.Nonce})
+	return rlpHash(h.OriginalHeader)
```
aneequesafdar: This will cause issues in future when we want to include the following fields in the calculation of the hash

```go
Committee          Committee `json:"committee"           gencodec:"required"`
ProposerSeal       []byte    `json:"proposerSeal"        gencodec:"required"`
Round              *big.Int  `json:"round"               gencodec:"required"`
CommittedSeals     [][]byte  `json:"committedSeals"      gencodec:"required"`
PastCommittedSeals [][]byte  `json:"pastCommittedSeals"  gencodec:"required"`
```

Most of the tests are passing now because of this `Hash()` is only returning the hash of the `OriginalHeader`, so inside the tendermint package none of the proposal hash will be predicated on the above fields, thus we can no longer trust the committee, proposal seal round etc, as they are never included in the calculation of the hash.

JekaMas:
I think this approach should break rlp, unfortunately.

yazzaoui:
> I think this approach should break rlp, unfortunately.

yeah I thought about that but it's fine in fact.

JekaMas:
The thing I still in doubt is that any change in the signatures, commitiees, keys for committees or subcommittees will lead to hard forks because all nodes in the network have to have the same header structure. The second question is how we'll perform chain sync in case that a few fork happened? We're going to store all past header structures and have something like
if block.number < originalFork { return originalHeader}
if block.number <= firstFork {return firstForkHeader}

aneequesafdar:
For me, the same problems would exist if we had stuck with extra data, so this is not a problem for this pull request specifically. 

Also, since our first block will be the new `Header` not the original one, I don't think we will ever run a scenario where we have to separate `originalHeader` from new `Header` if a forked occurred.

JekaMas:
Are you sure that we can guarantee that the change of header structure will never happened? I'm not sure about it. For me, if something can be changed it'll be changed one day and it's better to think about it and decide how we're going to change header structure in the future and perform hard forks.

JekaMas:
If we keep extraData than anyone will be able at least download the chain and check header hashes. If we're changing header structure than we have to keep each change to make sync possible.

yazzaoui:
yeah if we decide for example to start using threshold signatures it's a valid point but I think we should be fine because we still have plenty of time before prod deployment, and in the worst case we still have extradata, just have to reintroduce the helpers.

JekaMas:
I think this should be noticed to business that we will have hard forks in that case and we should develop business process for it.

yazzaoui:
yes and not only for block header changes https://github.com/clearmatics/proj-platform-launch/issues/13

PR #369, opened 2020-01-03
Commitseal fix

yazzaoui:
Solves Committed Seals issue of https://github.com/clearmatics/autonity/issues/347


consensus/tendermint/core/core.go
```diff
@@ -243,8 +243,10 @@ func (c *core) commit() {
 			committedSeals[i] = make([]byte, types.BFTExtraSeal)
 			copy(committedSeals[i][:], v.CommittedSeal[:])
 		}
-
-		if err := c.backend.Commit(*proposal.ProposalBlock, committedSeals); err != nil {
+		h := proposal.ProposalBlock.Header()
+		h.Round = c.currentRoundState.Round()
```
yazzaoui: Would need a bit of refactoring, maybe moving that to `backend.Commit`

consensus/tendermint/core/core.go
```diff
@@ -389,9 +391,10 @@ func (c *core) Quorum(i int) bool {
 }
 
 // PrepareCommittedSeal returns a committed seal for the given hash
-func PrepareCommittedSeal(hash common.Hash) []byte {
+func PrepareCommittedSeal(hash common.Hash, round *big.Int, height *big.Int) []byte {
 	var buf bytes.Buffer
+	buf.Write(round.Bytes())
+	buf.Write(height.Bytes())
```
yazzaoui: Note that the height information is not strictly needed.

consensus/tendermint/core/precommit_test.go
```diff
@@ -285,7 +285,7 @@ func TestHandlePrecommit(t *testing.T) {
 		}
 
 		backendMock := NewMockBackend(ctrl)
-		backendMock.EXPECT().Commit(*proposal.ProposalBlock, gomock.Any()).Return(nil)
+		backendMock.EXPECT().Commit(proposal.ProposalBlock, gomock.Any(), gomock.Any()).Return(nil)
```
JekaMas: Are we waiting here 'any' or particular parameter?

yazzaoui: yes the round and height number must match with the precommited seals, will update the test.

PR #370, opened 2020-01-06
Autonity Contract Refine and Contract Upgrade Mechanism.

Jason-Zhangxin-Chen:
1. Autonity Contract could be upgraded by operator issuing a remote contract call by passing the new binary and abi spec of the autontiy contract.
2. In case of contract upgade failure, the state will be rollback to the snapshot of the former state.
3. After contract upgrade, abi spec will be persisted in level db, node will load abi spec from level db if it existed from level db, otherwise, it would load the spec from chain configuration. 
4. Refine autonity contract and unit test cases of it in node.js.
5. NOTE: we need to define a proper MAX SIZE of our contract binary since our contract bin size already exceed ETH's standard size, to reserve enough MAX SIZE for the future to avoid chain fork.


core/tx_pool.go
```diff
@@ -515,8 +515,8 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
-	// Heuristic limit, reject transactions over 32KB to prevent DOS attacks
-	if tx.Size() > 32*1024 {
+	// Heuristic limit, reject transactions over 1MB to prevent DOS attacks
+	if tx.Size() > 1024*1024 {
```
JekaMas: Could you document somewhere this kind of changes?

params/protocol_params.go
```diff
@@ -69,7 +69,7 @@ const (
 	MemoryGas        uint64 = 3     // Times the address of the (highest referenced byte in memory + 1). NOTE: referencing happens on read, write and in instructions such as RETURN and CALL.
 	TxDataNonZeroGas uint64 = 68    // Per byte of data attached to a transaction that is not equal to zero. NOTE: Not payable on data of calls between transactions.
 
-	MaxCodeSize = 24576 // Maximum bytecode to permit for a contract
+	MaxCodeSize = 2457600 // Maximum bytecode to permit for a contract
```
JekaMas: And this one.

JekaMas:
The thing I warried about is the datarace. It can only happend if we do have many writings or reading+writing to the variable. But this means that we change commin.Big0 somewhere, it can influence on different parts of the code.

Could you check it
/home/travis/gopath/src/github.com/clearmatics/autonity/accounts/abi/pack.go:76
...

github.com/clearmatics/autonity/consensus/tendermint/backend.(*Backend).getCommittee()

/home/travis/gopath/src/github.com/clearmatics/autonity/consensus/tendermint/backend/engine.go:374 +0x34f

This is the trace of where we change common.Big0

Jason-Zhangxin-Chen:
> The thing I warried about is the datarace. It can only happend if we do have many writings or reading+writing to the variable. But this means that we change commin.Big0 somewhere, it can influence on different parts of the code.
> 
> Could you check it
> /home/travis/gopath/src/github.com/clearmatics/autonity/accounts/abi/pack.go:76
> ...
> 
> github.com/clearmatics/autonity/consensus/tendermint/backend.(*Backend).getCommittee()
> 
> /home/travis/gopath/src/github.com/clearmatics/autonity/consensus/tendermint/backend/engine.go:374 +0x34f
> 
> This is the trace of where we change common.Big0

I will check it, thanks Eugene.

Jason-Zhangxin-Chen:
> > The thing I warried about is the datarace. It can only happend if we do have many writings or reading+writing to the variable. But this means that we change commin.Big0 somewhere, it can influence on different parts of the code.
> > Could you check it
> > /home/travis/gopath/src/github.com/clearmatics/autonity/accounts/abi/pack.go:76
> > ...
> > github.com/clearmatics/autonity/consensus/tendermint/backend.(*Backend).getCommittee()
> > /home/travis/gopath/src/github.com/clearmatics/autonity/consensus/tendermint/backend/engine.go:374 +0x34f
> > This is the trace of where we change common.Big0
> 
> I will check it, thanks Eugene.

Here when client side packing the parameters via contract ABI spec, the data casting was treated as data writing in our test case. So the data race was detected.
	case reflect.Ptr:
		return U256(value.Interface().(*big.Int))     // cast to appropriate representation in solidity.
	default:
		panic("abi: fatal error")

Issue #372, opened 2020-01-08
Multiply TYPE declaration for Prometheus metrics


eastata:
# Problem

Example Autonity Prometheus metrics output:
```
# TYPE chain_account_commits_count counter
chain_account_commits_count 874528

# TYPE chain_account_commits summary
chain_account_commits {quantile="0.5"} 0

# TYPE chain_account_commits summary
chain_account_commits {quantile="0.75"} 0

...
```
According to [prometheus format specification](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#comments-help-text-and-type-information)
for `# TYPE`:
> Only one TYPE line may exist for a given metric name.

It leads to error for [go github.com/prometheus/common](https://github.com/prometheus/common/blob/7680068d4af7bd5cb3043381494f956c22c034b2/expfmt/text_parse.go#L500) That used as go dependency for other software that handle this metrics, particular for [Telegraph](https://github.com/influxdata/telegraf/blob/5b92477603c8cd7ce5178f4b8b1d5a5a0c6e7330/plugins/inputs/prometheus/parser.go#L55).

# How to reproduce

* Autonity `v0.3.0`
* Run autonity with Prometheus metrics enabled

# Acceptance Criteria

- Telegraf that use prometheus/common for parsing should handle autonity metrics output as valid Prometheus format

# Future Steps

Steps which should be taken after this issue has been resolved.

yazzaoui:
Ok the issue is coming from the go-ethereum prometheus exporter, it's writing for every metric the TYPE line which is problematic for histogram metrics.

PR #373, opened 2020-01-09
Prometheus exporter: Fix TYPE output

yazzaoui:
The `TYPE` line was duplicated for histogram and timer metrics.

Issue: #372


eastata:
I have build and test autonity + autonity init from this branch and it works. Telegraf can accept this metrics

PR #380, opened 2020-01-15
[WIP] Tendermint Fixes

yazzaoui:
This is rather a large pull request with significant refactoring as it was made clear that it was needed for fixing the remaining issues we had with our implementation.

- Fixed `TotalSize()` vulnerability, where a potentially malicious validator when sending multiple votes for the same round they were counted separately.
- Change `round` type to int64, we don't need arbitrary precision for it, and because`validRound` and `lockedRound` can be negative (-1) and are of the same type. Updated block type in consequence.
- Made changes to simplify the consensus `Message` struct  with the removal of `IsValidRoundNil` and added propoer DecodeRLP and EncodeRLP.
- Created the `consensusMsg` interface to retrieve easilly height and round of a message instead of type switching.  
- Simplified `roundState` structure, removed redundant `round` and `height` field and putting them in `core` struct.
- Simplified the validator package, removal of the validatorSet interface (in core), removed useless functions, made it ready for Autonity Contract invokation. Removed validator interface because why not using directly `types.committeeMember`
- Enforcing that coinbase is miner at verification time (**WIP**)
- Removed `core.Quorum()` definition to avoid code duplication on this critical function. 
- Increased memory cache for `inmemoryAddresses` to 500.
- Renaming of various functions when it seemed to me appropriate.
- To avoid introducing race-conditions I limited to the thread running the main event loop access to the state data, however still needed a mutex to protect round, height and committee.
- Fixed remaining UPON conditions not enforced
- Various other small fixes/refactoring.

Major e2e tests are passing but there I've got still significant work to finish fixing tests.


consensus/tendermint/backend/api.go
```diff
@@ -46,10 +46,10 @@ func (api *API) GetValidatorsAtHash(hash common.Hash) ([]common.Address, error)
 		return nil, errUnknownBlock
 	}
 
-	validators := api.tendermint.Validators(header.Number.Uint64()).List()
+	validators := api.tendermint.Validators(header.Number.Uint64()).Committee()
 	addresses := make([]common.Address, len(validators))
 	for i, validator := range validators {
-		addresses[i] = validator.GetAddress()
+		addresses[i] = validator.Address
```
JekaMas: Is this a good idea to use fields directly instead having access method?

consensus/tendermint/backend/backend.go
```diff
@@ -146,17 +146,17 @@ func (sb *Backend) Address() common.Address {
...
 
-func (sb *Backend) Validators(number uint64) validator.Set {
-	validators, err := sb.retrieveSavedCommittee(number, sb.blockchain)
-	proposerPolicy := sb.config.GetProposerPolicy()
+func (sb *Backend) Validators(number uint64) committee.Set {
+	validators, err := sb.savedCommittee(number, sb.blockchain)
 	if err != nil {
-		return validator.NewSet(nil, proposerPolicy)
+		sb.logger.Error("could not retrieve saved committee", "height", number, "err", err)
+		return nil
```
JekaMas: Why we shouldn't throw an error here?

consensus/tendermint/backend/engine.go
```diff
@@ -98,7 +100,10 @@ func (sb *Backend) verifyHeader(chain consensus.ChainReader, header *types.Heade
 	if header.Number == nil {
 		return errUnknownBlock
 	}
-
+	// TODO : change it to use the const defined in core (refactoring needed)
+	if header.Round > 99 {
```
JekaMas: Could you make it as a package layer constant with a proper comment?

consensus/tendermint/backend/engine.go
```diff
@@ -207,36 +212,36 @@ func (sb *Backend) verifySigner(chain consensus.ChainReader, header *types.Heade
...
 	// The length of Committed seals should be larger than 0
 	if len(header.CommittedSeals) == 0 {
 		return types.ErrEmptyCommittedSeals
 	}
 
-	// Check whether the committed seals are generated by parent's validators
+	// Check whether the committed seals are generated by parent's committee
 	validSeal := 0
-	proposalSeal := tendermintCore.PrepareCommittedSeal(header.Hash(), header.Round, header.Number)
+	proposalSeal := tendermintCore.PrepareCommittedSeal(header.Hash(), int64(header.Round), header.Number)
```
JekaMas: Should we change Header and use int64 as a round?

consensus/tendermint/backend/engine.go
```diff
@@ -548,41 +549,46 @@ func (sb *Backend) Close() error {
 	return nil
 }
 
-// retrieve list of validators for the block at height passed as parameter
-func (sb *Backend) retrieveSavedCommittee(number uint64, chain consensus.ChainReader) (types.Committee, error) {
+// retrieve list of committee for the block at height passed as parameter
+func (sb *Backend) savedCommittee(number uint64, chain consensus.ChainReader) (committee.Set, error) {
+	var lastMiner common.Address
```
JekaMas: LastProposer?

consensus/tendermint/committee/default.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (set *defaultSet) GetProposer(round int64) types.CommitteeMember {
+	set.mu.Lock()
+	defer set.mu.Unlock()
+	_, ok := set.proposer[round]
+	if !ok {
+		set.proposer[round] = set.selector(set, set.lastProposer, round)
+	}
+
+	return set.proposer[round]
```
JekaMas: Maybe
```
v, ok := set.proposer[round]
	if !ok {
		v = set.selector(set, set.lastProposer, round)
                set.proposer[round] = v
	}

 	return v
```

consensus/tendermint/core/backlog.go
```diff
@@ -158,9 +131,9 @@ func (c *core) processBacklog() {
 	}
 }
 
-func toPriority(msgCode uint64, r *big.Int, h *big.Int) float32 {
-	// FIXME: round will be reset as 0 while new height
+func toPriority(msgCode uint64, r int64, h *big.Int) float32 {
+	// TODO check for overflows!!
```
JekaMas: Could we use big.Ints or big.Rat here?

consensus/tendermint/core/consensus_msg.go
```diff
@@ -18,120 +18,143 @@ package core
...
+	} else {
+		validRound = int64(proposal.ValidRound)
 	}
 
-	p.Round = proposal.Round
-	p.Height = proposal.Height
-	p.ValidRound = proposal.ValidRound
-	p.IsValidRoundNil = proposal.IsValidRoundNil
-	p.ProposalBlock = proposal.ProposalBlock
+	if !(validRound <= maxRound && proposal.Round <= maxRound) {
```
JekaMas: validRound < maxRound || proposal.Round > maxRound ?

consensus/tendermint/core/core.go
```diff
@@ -71,31 +71,37 @@ var (
 	errMovedToNewRound = errors.New("timer expired and new round started")
 )
 
+const (
+	maxRound = 99 // consequence of backlog priority
```
JekaMas: The same constant by value was used in another package. It seems that we need only one public constant.

consensus/tendermint/core/message.go
```diff
@@ -131,13 +139,53 @@ func (m *Message) PayloadNoSig() ([]byte, error) {
...
+	//Decode is responsible to rlp-decode m.Msg. It is meant to only perform the actual decoding once,
+	//saving a cached value in m.decodedMsg.
+
+	rval := reflect.ValueOf(val)
+	if rval.Kind() != reflect.Ptr {
+		return errors.New("decode arg must be a pointer")
+	}
+
+	// check if we already have a cached value decoded
+	if m.decodedMsg != nil {
```
JekaMas: Looks like premature optimization.

consensus/tendermint/core/round_messages.go
```diff
@@ -0,0 +1,183 @@
...
+
+type messagesMap struct {
+	internal map[int64]*roundMessages
+	mu       sync.RWMutex
+}
+
+func newMessagesMap() messagesMap {
+	return messagesMap{
+		internal: make(map[int64]*roundMessages),
+		mu:       sync.RWMutex{},
```
JekaMas: You don't have to init it.

consensus/tendermint/core/round_messages.go
```diff
@@ -0,0 +1,183 @@
...
+func (s *roundMessages) SetProposal(proposal *Proposal, msg *Message, verified bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.proposalMsg = msg
+	s.verifiedProposal = verified
+	s.proposal = proposal
+}
+
+func (s *roundMessages) PrevotesCount(hash common.Hash) int {
+	return s.prevotes.VoteCount(hash)
```
JekaMas: Should this methods use the mutex too?

consensus/tendermint/core/validator_set_test.go
```diff
@@ -1,398 +0,0 @@
...
-	"reflect"
-	"testing"
-
-	"github.com/clearmatics/autonity/common"
-	"github.com/clearmatics/autonity/consensus/tendermint/config"
-	"github.com/clearmatics/autonity/consensus/tendermint/validator"
-	"github.com/golang/mock/gomock"
-)
-
-func TestValidatorSetSetEmpty(t *testing.T) {
```
JekaMas: Could you mark that we need the same tests to test new types and functionality?

consensus/tendermint/crypto/check_signature.go
```diff
@@ -20,9 +20,10 @@ func CheckValidatorSignature(valSet validator.Set, data []byte, sig []byte) (com
 	}
 
 	// 2. Check validator
-	if _, val := valSet.GetByAddress(signer); val != nil {
-		return val.GetAddress(), nil
+	idx, val := valSet.GetByAddress(signer)
+	if idx == -1 {
```
JekaMas: Is it a good thing that we check for this magic value at a quite a few places?

consensus/tendermint/backend/api.go
yazzaoui: The reason is that I didn't want to create a new interface and those fields were already public.

consensus/tendermint/backend/backend.go
yazzaoui: It was the original function signature but I can change it

consensus/tendermint/backend/engine.go
yazzaoui: yes

consensus/tendermint/backend/engine.go
yazzaoui: I think RLP doesn't support int64 so I have to define custom encoding function

consensus/tendermint/backend/engine.go
yazzaoui: Ok If you prefer

consensus/tendermint/core/message.go
yazzaoui: can be discussed

consensus/tendermint/core/round_messages.go
yazzaoui: yes

consensus/tendermint/core/round_messages.go
yazzaoui: prevotes.VoteCount is protected

consensus/tendermint/crypto/check_signature.go
yazzaoui: yeah you're right I'll add an error

consensus/tendermint/core/round_messages.go
JekaMas: But prevotes can be changed in SetProposals() and this is a data race.

consensus/tendermint/backend/api.go
JekaMas: For me it's better to use methods that if guarded if needed than fields directly.

consensus/tendermint/backend/engine.go
aneequesafdar: I am not sure about picking an arbitrary number as the max round is the best way to limit denial of service attack, also since verify proposal will only be run after the proposal message is checked in the handler I don't think there is a need to perform this check here. When verifyHeader is run outside of the consensus package it is to check a block on which we have already reached consensus on, and since in the consensus layer we won't allow for a block with round more than max round, therefore, this check, in my opinion, is not required.

Also, a suggestion on max round, I think the max round should be a function of the size of the committee set, rounds required to reach timeout greater the GST (again an estimate based on empirical analysis)/average rounds needed to reach consensus, round required to reach honest proposer in the committee and the current round. This would drastically reduce the rounds malicious users can use to send bad messages.

yazzaoui: For the time being, I prefer to leave this check here even though like you said it's probably not needed, but I don't think it's a bad thing to be defensive here, it doesn't cost much and might help to catch some bugs.

For the value 99 agree we have to think about a better way, I just made it explicit that we were already limiting the number of rounds to this value due to the backlog priority queue.

consensus/tendermint/backend/engine.go
yazzaoui: Better to keep round in header as uint, we need int64 only at consensus layer because of the conventions for lockedround and validround that could be equal to -1. Outside of consensus it doesn't make sense.

consensus/tendermint/core/round_messages.go
yazzaoui: ` SetProposal()` is not changing prevotes

consensus/tendermint/core/consensus_msg.go
yazzaoui: Don't understand the point, `validRound < maxRound` is actually correct?

consensus/tendermint/core/backlog.go
yazzaoui: I'll create a separate issue for this one it's a bit more complicated than it seems

consensus/tendermint/core/core.go
yazzaoui: couldn't find it

consensus/tendermint/core/validator_set_test.go
yazzaoui: yes I'm finishing fixing them

consensus/tendermint/committee/default.go
```diff
@@ -0,0 +1,124 @@
...
+	members      types.Committee
+	policy       config.ProposerPolicy
+	lastProposer common.Address
+	selector     ProposalSelector
+
+	mu       sync.RWMutex                    // members doesn't need to be protected as it is read-only
+	proposer map[int64]types.CommitteeMember // cached computed values
+}
+
+func NewSet(members types.Committee, policy config.ProposerPolicy, lastProposer common.Address) *defaultSet {
```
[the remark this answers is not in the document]
yazzaoui: Ok I've added an error

consensus/tendermint/core/backlog.go
```diff
@@ -158,9 +131,9 @@ func (c *core) processBacklog() {
 	}
 }
 
-func toPriority(msgCode uint64, r *big.Int, h *big.Int) float32 {
-	// FIXME: round will be reset as 0 while new height
+func toPriority(msgCode uint64, r int64, h *big.Int) float32 {
```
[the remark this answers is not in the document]
yazzaoui: I'll create a separate issue for this one

consensus/tendermint/backend/api.go
yazzaoui: We'll change that to use methods the day it's needed

consensus/tendermint/backend/engine.go
```diff
@@ -207,36 +212,36 @@ func (sb *Backend) verifySigner(chain consensus.ChainReader, header *types.Heade
 
 	// Signer should be in the validator set of previous block's extraData.
 
-	for i := range committee {
-		if committee[i].Address == signer {
+	for _, committeeMember := range committee.Committee() {
```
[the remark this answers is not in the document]
yazzaoui: `Committee()` can't be nil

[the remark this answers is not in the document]
yazzaoui: we check that the committee is not nil with the error

JekaMas:
I'm a bit worry that code coverage of tendermint package decreased https://codecov.io/gh/clearmatics/autonity/compare/c545125a686ad62c555abf6294aaac576a227bd1...abfb84d188a9f20ce0aaaaa13987028d31d9ed8c/tree/consensus/tendermint/core

Especially
unmined_block.go -20.85%
message_set.go -20.45%
precommit.g9 -11.39%
prevote.go  -13.78%
propose.go -14.96%
message.go -9.73%

yazzaoui:
yes I'm adding new tests

consensus/tendermint/committee/default.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (set *defaultSet) Copy() Set {
+	newSet, _ := NewSet(copyMembers(set.members), set.policy, set.lastProposer)
+	return newSet
+}
+
+func (set *defaultSet) F() int { return int(math.Ceil(float64(set.Size())/3)) - 1 }
+
+func (set *defaultSet) Quorum() int { return int(math.Ceil((2 * float64(set.Size())) / 3.)) }
```
piersy: This quorum size calculation is wrong, since it can return exactly 2/3 of the set size where it should be returning a number > 2/3 of the set size.

For example given a set of size 6 it returns 4 instead of 5.

```suggestion
func (set *defaultSet) Quorum() int { return int(2.0/3.0*float64(set.Size()))+1 }
```

consensus/tendermint/committee/default.go
```diff
@@ -0,0 +1,132 @@
...
+	curProposer := set.GetProposer(round)
+	return reflect.DeepEqual(curProposer, val)
+}
+
+func (set *defaultSet) Copy() Set {
+	newSet, _ := NewSet(copyMembers(set.members), set.policy, set.lastProposer)
+	return newSet
+}
+
+func (set *defaultSet) F() int { return int(math.Ceil(float64(set.Size())/3)) - 1 }
```
piersy: We can de-duplicate the logic of calculating the Quorum size/Tolerable failures by reusing `set.Quorum` here.

```suggestion
func (set *defaultSet) F() int { return set.Size()-set.Quorum() }
```

consensus/tendermint/committee/default.go
yazzaoui: This formula is correct,  `Q > 2N/3` is actually not always true, you can find the formal demonstration somewhere in this paper https://arxiv.org/pdf/1901.07160.pdf

if you want to convince yourself, for a network of 6 validators we  can only support 1 faulty node.

1.` N-f >= Q,` there must be at least a quorum of correct processes to ensure liveness, 6 - 1 = 5 >= 4 
2. `2 Q - N > f`, the intersection of two quorums must be greater than the number of faulty nodes.  2 x 4 - 6 = 2 > 1

consensus/tendermint/committee/default.go
yazzaoui: F is not always equal to N - Q

yazzaoui:
the PR seems ready to me, we have a separate task to add more tests anyways

PR #367, opened 2020-01-02
(WIP) Merge ethereum upstream

screwyprof:

PR #381, opened 2020-01-16
Hide tendermint specific Header fields on transport layer

### LINKED ISSUE #371: Keep header structure changed but encode all the changes into extradata field
It's about custom encoder and decoder maybe with additional tag for the structure. It'll help us to keep all the changes and don't have to perform hard forks if header is changed.

JekaMas:
Use custom RLP encoder and decoder to hide actual changes of Header and keep using extraData field as a container.

closes #371


aneequesafdar:
I think the pr looks good, however, I feel you are going to run into the same problem with test that Youssef and I did since the header hashes will be different with the new header, so I am not too sure how that will be handled.

Also, I am not sure about the benefit of this approach, can you either schedule a call or do a small write up to explain the benefit of this approach compared to Youssef.

JekaMas:
The main benefit is that Header structure is untouched in terms of p2p communication. So we're still compatible with Ethereum tooling and in case of changing header we don't change anything on network and protocol layers.
We even can decode old headers with new versions if new fields were added. And this can be done without hard forks.

JekaMas:
@aneequesafdar just to clarify the PR, it's not an alternative approach, it's an iteration. All the changes in Header are still on the place. So we as developers will work with modified Header, but p2p communication will see old Headers with extraData

JekaMas:
@aneequesafdar @b00ris @yazzaoui The PR fixes all the issues in #367 . so looking forward for the review.

core/types/bft_test.go
```diff
@@ -56,83 +56,75 @@ func TestHeaderHash(t *testing.T) {
...
 		// BFT header tests
 		{
-			Header{OriginalHeader: PosHeader}, // test 7
+			PosHeader, // test 3
 			posHeaderHash,
 		},
 		{
-			Header{OriginalHeader: PosHeader, CommittedSeals: [][]byte{common.Hex2Bytes("0xfacebooc"), common.Hex2Bytes("0xbabababa")}},
+			setExtra(PosHeader, headerExtra{
+				CommittedSeals: [][]byte{common.Hex2Bytes("0xfacebooc"), common.Hex2Bytes("0xbabababa")},
```
donmiguelenriquez, line 74: free marketing :)

PR #382, opened 2020-01-16
Tendermint bft delta

aneequesafdar:


JekaMas:
So now developer should know that empty hash is a nil vote and we should document it somehow. Maybe it's a less strong abstraction than having 2 methods that hide implementation inside?

aneequesafdar:
Yeah I felt it was better to just add a vote regardless of the hash and message set will handle nil votes from non nil votes. Also, I think it is intuitive to have nil block as empty hash.

JekaMas:
I'm not sure that it's a better interface. However it's all up to you.

aneequesafdar:
closing because outdated

Issue #263, opened 2019-09-05
Ethereum protocol manager return wrong p2p protocol name


yazzaoui:
# Problem

With the 1.9.1 merge there is a small issue with the eth protocol manager new `makeProtocol()` function not calling consensus engine `Protocol()` causing to return the string `eth` instead of `tendermint` .
Also the `protocol` field of the `Ethereum` object in `eth/backend.go` appears now as useless.

PR #384, opened 2020-01-20
P2p handshaking consensus protocol name

Jason-Zhangxin-Chen:
Ethereum protocol manager return wrong p2p protocol name #263
https://github.com/clearmatics/autonity/issues/263

Get protocol name from backend engine interface.


eth/handler.go
```diff
@@ -217,8 +217,15 @@ func (pm *ProtocolManager) makeProtocol(version uint) p2p.Protocol {
 		panic("makeProtocol for unknown version")
 	}
 
+	protoName := protocolName
+	// get consensus protocol from backend engine.
+	if handler, ok := pm.engine.(consensus.Handler); ok {
+		protoName, _ = handler.Protocol()
+		log.Info("Get consensus protocol ", "name: ", protoName)
```
b00ris: I think Debug level is more suitable here.

PR #392, opened 2020-01-24
Autonity contract upgrade test

b00ris:
closes https://github.com/clearmatics/autonity/issues/378


b00ris:
Something was wrong with CI

PR #411, opened 2020-02-07
Remove open network flag

### LINKED ISSUE #406: Remove open network flag

JekaMas:
closes #406

### DIFF
--- a/cmd/autonity/chaincmd.go
+++ b/cmd/autonity/chaincmd.go
@@ -211,6 +211,9 @@ func initGenesis(ctx *cli.Context) error {
 		utils.Fatalf("invalid genesis file: %v", err)
 	}
 	// Make AutonityContract and Tendermint consensus mandatory for the time being.
+	if genesis.Config == nil {
+		utils.Fatalf("No Autonity Contract and Tendermint configs section in genesis")
+	}
 	if genesis.Config.AutonityContractConfig == nil {
 		utils.Fatalf("No Autonity Contract config section in genesis")
 	}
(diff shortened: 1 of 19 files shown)


JekaMas:
@yazzaoui It is not an easy one issue. https://travis-ci.org/clearmatics/autonity/jobs/647690584 a lot tests assumes that the network is open.

yazzaoui:
yeah well, I'm not surprised, how I see it we've got 3 options:

1 - Officially support open-network mode and make sure it works correctly, write tests
2 - Keep it like it is, but remove the command line flag 
3 - Fix the failing tests

What do you think ?

JekaMas:
3. Definitely.

JekaMas:
@b00ris @aneequesafdar @screwyprof please review.

cmd/autonity/chaincmd.go
```diff
@@ -211,6 +211,9 @@ func initGenesis(ctx *cli.Context) error {
 		utils.Fatalf("invalid genesis file: %v", err)
 	}
 	// Make AutonityContract and Tendermint consensus mandatory for the time being.
+	if genesis.Config == nil {
+		utils.Fatalf("No Autonity Contract and Tendermint configs section in genesis")
```
aneequesafdar: I am not sure if this the correct error to return if config is nil. Please update the comment

JekaMas: I think it is a correct error because the next thing we check is
```
if genesis.Config.AutonityContractConfig == nil {
		utils.Fatalf("No Autonity Contract config section in genesis")
}
```

and then the same for tendermint config.

PR #419, opened 2020-02-13
Contract binding

Jason-Zhangxin-Chen:
Bind JS contract object with the client's internal contract on autonity client console.
1. Binding happens on console initialization with RPC calls.
2. Remove wrong RPC extends in the web3.extends.
3. Since contract is upgradable, RPC getContractABI now get abi from leveldb or conf rather than to get it from conf.
4. In the console welcome MSG, there will be a "The backend contract binds to JS object: autonity!" 
Type `autontiy` and click 'Enter' then console will display all the methods of autontiy contract. 
For example: type: `autontiy.getVersion()` will return the version of the autontiy contract.


contracts/autonity/autonity.go
```diff
@@ -298,3 +298,16 @@ func (ac *Contract) upgradeAbiCache(newAbi string) error {
 	ac.contractABI = &newABI
 	return nil
 }
+
+func (ac *Contract) RPCGetContractABI() string {
```
yazzaoui: Why this `RPC` prefix?

console/console.go
```diff
@@ -220,9 +220,24 @@ func (c *Console) init(preload []string) error {
...
+		if (typeof(tendermint) != 'undefined') {
+			var abi = JSON.parse(tendermint.getContractABI());
+			var addr = tendermint.getContractAddress();
+			var autonity = eth.contract(abi).at(addr);
+		}
+	`); err != nil {
+		fmt.Fprintln(c.printer, "contract binding failed: ", err)
+	}
+}
+
```
aneequesafdar: Can you please explain the purpose of this function? Is it so that when the user attaches to a javascript console they have direct access to `abi`, `addr` and `autonity` without the need to call the tendermint api functions? If so can you please rename the variables to `autonity_contract_abi`, `autonity_contract_addr` and `autonity_contract` (or use camel case) to make the variable names more expressive

Jason-Zhangxin-Chen: Resolved. Thank you, Aneeque.

yazzaoui:
closing https://github.com/clearmatics/autonity/issues/402

PR #426, opened 2020-02-20
Fix DNS resolve

### LINKED ISSUE #405: DNS resolve approach
Design/ define a better protocol for DNS resolving

screwyprof:
closes #405

### DIFF
--- a/cmd/autonity/main.go
+++ b/cmd/autonity/main.go
@@ -19,6 +19,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/clearmatics/autonity/p2p/enode"
 	"math"
 	"os"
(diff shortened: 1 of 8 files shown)


yazzaoui:
Can we have a small description of the change you brought, for us and documentation? I think DevOps would appreciate understanding the new mechanism also.

screwyprof:
> Can we have a small description of the change you brought, for us and documentation? I think DevOps would appreciate understanding the new mechanism also.

Within this task I implemented an asynchronous enode DNS resolution. The resolution happens once per the given interval (10 seconds by default). Resolved enodes remain in cache for 10 minutes. enode package was refactored to support resolving.

PR #457, opened 2020-03-05
Add make target to generate abi and bytecode

piersy:
This make target automatically generates go source files containing the bytecode and abi of the current Autonity.sol contract. I've set up the makefile so that building autonity depends on this.

Ideally I'd like to ensure all build and test targets depend on this but looking through the Makefile it looks like there is some stuff that could be removed (android and ios builds?) and thought maybe we could address that first.


Makefile
```diff
@@ -15,11 +15,39 @@ ifeq ($(LATEST_COMMIT),)
...
+.make.embed-autonity-contract: AUTONITY_CONTRACT_PATH = contracts/autonity/contract/contracts/Autonity.sol
+.make.embed-autonity-contract: GENERATED_DIR = common/acdefault/generated
+.make.embed-autonity-contract: $(AUTONITY_CONTRACT_PATH)
+	mkdir $(GENERATED_DIR)
+	solc --overwrite --abi --bin -o $(GENERATED_DIR) $(AUTONITY_CONTRACT_PATH)
+
+	@echo Generating $(GENERATED_DIR)/bytecode.go
+	@> $(GENERATED_DIR)/bytecode.go
+	@echo 'package generated' >> $(GENERATED_DIR)/bytecode.go
+	@echo -n 'const Bytecode = "' >> $(GENERATED_DIR)/bytecode.go
```
yazzaoui: Wouldn't it be better to use go generate instead?

piersy: I didn't realise we were already using go generate, but given that we are, it seems like a sensible choice.

piersy:
I'm closing this pull request because I don't think the automatic abi and bytecode generation justify the amount of additions to .github/workflows/go.yml. I will come up with another approach.

PR #459, opened 2020-03-06
Resolve e-nodes in e2e tests

screwyprof:


consensus/test/testcase_test.go
```diff
@@ -296,6 +306,22 @@ func runTest(t *testing.T, test *testCase) {
 	}
 }
 
+func TestName(t *testing.T) {
+	enode.SetResolveFunc(func(host string) (ips []net.IP, e error) {
```
b00ris: Looks as debug case. Just rename or remove

Issue #461, opened 2020-03-09
Create a plan for updating the Autonity documentation


piersy:
# Problem

We don't know have a clear idea of which areas of the [autonity docs](https://github.com/clearmatics/autonity-docs-v2) are are lacking.

We don't know the relative importance of different areas of the documentation, so we cannot prioritise documentation.

We don't have a process for creating documentation.
 
# Acceptance Criteria

Write up a plan discussing how to address these issues.

SherifSK:
My understanding is that @maxcroasdale started this effort with the Engineering team members a while back using Hyperledger Besu as a starting point i.e. a table of content.

Hyperledger Besu docs has the following table of content (ToC):

1. How to
2. Tutorials
3. Concepts
4. Reference (developer docs)

(Hyperledger team apparently have read this great [article](https://www.divio.com/blog/documentation/) and following its breakdown of documentation)

I think we should differentiate between:
- User docs (items 1, 2 and 3 in the above ToC)
- Developer docs (item 4)

I think User Docs currently has a higher priority to address the requirements for Q1 and Q2 releases (focus on System X). Based on this understanding (which I am happy to change or have it challenged), I directed @maxcroasdale to use for the How To section of the docs the how-to guides (as a base) that were developed/undergoing development by Ben and Frank as part of their SystemX work (specifically the how-to guides developed early on as they were only using the AFNC i.e. no System Console).

IMHO, the order of priority of the above ToC should be:

1. How to
2. Concepts
3. Tutorials
4. Reference

However, despite all of the above efforts I think the right approach to designing a successful documentation experience is to start from a UX perspective by creating user personas and then creating user stories/journey/epics and try to understand properly what should be included (and why) and how i.e. information design. The question is bandwidth vs. time available. I think this should be discussed with @yazzaoui.

saronimo:
I;m pleased to see this is being dedicated time, but would echo @SherifSK in saying that a lot of work has already gone into thinking about this and what we really need now in content from the team who built autonity! 💪 

What can we do to help?

piersy:
@SherifSK @saronimo Thanks for the feedback, lets schedule a meeting where you could bring me up to speed with the current situation from your perspective.

saronimo:
Ok please set something up later as this is a very high priority given our timelines and please invite @maxcroasdale as well

piersy:
@saronimo @SherifSK @maxcroasdale As per our discussion:

[...]

PR #469, opened 2020-03-13
Add framework to generate user facing docs

piersy:
This pr adds an empty sphinx framework with a Makefile that builds and serves a site with the docs.

It would be good if reviewers could, verify that running `make install-prerequisites-<linux/mac>` and  `make serve`  inside the `docs` directory works for them.

This is phase 1 of the plan emerging from #461


JekaMas:
I do have an idea about codecov - https://github.com/clearmatics/autonity/blob/develop/codecov.yml let's add yhe rule to ignore all non ".go" files.

piersy:
@JekaMas thanks for the suggestion, I don't think we can add negative rules to ignore see the [ codecove docs](https://docs.codecov.io/docs/ignoring-paths), but I could just add our `docs` dir to ignored paths. Which I will try now

PR #455, opened 2020-03-05
[wip] Connection to non-validator nodes fixes

b00ris:
closes https://github.com/clearmatics/autonity/issues/431

Issue #431, opened 2020-02-21
Stakeholder/Participant get banned. Node tries to participate in consensus


b00ris:
# Problem
Reproduced case - TestTendermintAddStakeholdersToTopologySuccess and connected ones: TestTendermintStarSuccess TestTendermintBusSuccess TestTendermintChangeTopologyFromStarToBusSuccess 
TestTendermintChangeTopologyFromBusToStarSuccess
TestTendermintAddValidatorsToTopologySuccess TestTendermintAddParticipantsToTopologySuccess TestTendermintAddStakeholdersToTopologySuccess

https://github.com/clearmatics/autonity/pull/455#issuecomment-601114003

```
ERROR[02-21|13:12:23.035] Failed to decode message from payload    addr=0x642d153F77B014aC4B17dFbBE4068D4f382eBabB err="unauthorized address"
ERROR[02-21|13:12:49.042] error validator errUnauthorized          addr=0x642d153F77B014aC4B17dFbBE4068D4f382eBabB addr=0x642d153F77B014aC4B17dFbBE4068D4f382eBabB
```
After it Stakeholder node is dropped.

# Acceptance Criteria
TestTendermintAddStakeholdersToTopologySuccess don't fail.
Stakeholder node can be connected to validator

Jason-Zhangxin-Chen:
Been fixed by #661

PR #470, opened 2020-03-16
Simplify CI

### LINKED ISSUE #462: Remove tendermint-success-test CI job
# Problem

`tendermint-success-test` runs `TestTendermintSuccess` with `-v` and `-timeout 40m` but `TestTendermintSuccess` is already being run with the same flags by `start_stop_tests`

# Implementation

Delete the job.

# Acceptance Criteria

`TestTendermintSuccess` is run once with flags `-v` and `-timeout 40m` by the CI.

### LINKED ISSUE #463: TestTendermintAddConnectionToTopologySuccess is failing
# Problem

`TestTendermintAddConnectionToTopologySuccess` is failing on the tip of develop 51480efb4, the test is not currently being run by CI.

# Acceptance Criteria

The test should pass

### LINKED ISSUE #464: Add missing tests to CI
# Problem

The following tests are not being run by CI.

```
TestContractUpgradeSeveralUpgrades
TestContractUpgradeSeveralUpgradesOnBusTopology
TestContractUpgradeSeveralUpgradesOnStarTopology
TestContractUpgrade_Success
TestRemoveFromValidatorsList
TestTendermintAddConnectionToTopologySuccess
TestTendermintAddValidatorsToTopologySuccess
TestTendermintBusSuccess
TestTendermintChangeTopologyFromBusToStarSuccess
TestTendermintChangeTopologyFromStarToBusSuccess
TestTendermintNoQuorum
TestTendermintStarOverParticipantSuccess
TestTendermintStarSuccess
TestTendermintTC7
```

TestTendermintTC7 takes an incredibly long time to run.
`TestTendermintAddConnectionToTopologySuccess` is currently failing see #463 and is out of scope for this ticket.

The rest pass if run individually.

```
--- PASS: TestContractUpgradeSeveralUpgrades (24.87s)
--- PASS: TestContractUpgradeSeveralUpgradesOnBusTopology (25.00s)
--- PASS: TestContractUpgradeSeveralUpgradesOnStarTopology (25.00s)
--- PASS: TestContractUpgradeSeveralUpgradesOnBusTopology (24.53s)
--- PASS: TestContractUpgradeSeveralUpgradesOnStarTopology (25.58s)
--- PASS: TestContractUpgrade_Success (15.56s)
--- PASS: TestRemoveFromValidatorsList (15.40s)
--- PASS: TestTendermintAddValidatorsToTopologySuccess (56.54s)
    --- PASS: TestTendermintAddValidatorsToTopologySuccess/test_case_no_malicious (56.54s)
--- PASS: TestTendermintBusSuccess (9.55s)
    --- PASS: TestTendermintBusSuccess/test_case_no_malicious (9.55s)
--- PASS: TestTendermintChangeTopologyFromBusToStarSuccess (56.60s)
    --- PASS: TestTendermintChangeTopologyFromBusToStarSuccess/test_case_no_malicious (56.60s)
--- PASS: TestTendermintChangeTopologyFromStarToBusSuccess (56.66s)
    --- PASS: TestTendermintChangeTopologyFromStarToBusSuccess/test_case_no_malicious (56.66s)
--- PASS: TestTendermintNoQuorum (14.01s)
    --- PASS: TestTendermintNoQuorum/test_case_2_validators,_one_goes_down_after_block_3 (6.48s)
    --- PASS: TestTendermintNoQuorum/test_case_3_validators,_two_go_down_after_block_3 (7.53s)
--- PASS: TestTendermintStarOverParticipantSuccess (9.99s)
    --- PASS: TestTendermintStarOverParticipantSuccess/test_case_no_malicious (9.99s)
--- PASS: TestTendermintStarSuccess (9.53s)
    --- PASS: TestTendermintStarSuccess/test_case_no_malicious (9.53s)
```

 Although if run together some fail and they all fail due to the following error.
`leaks.go:78: found unexpected goroutines:`

```
--- PASS: TestRemoveFromValidatorsList (14.82s)
--- PASS: TestContractUpgrade_Success (15.00s)
--- PASS: TestContractUpgradeSeveralUpgrades (25.00s)
--- PASS: TestContractUpgradeSeveralUpgradesOnBusTopology (24.99s)
--- PASS: TestContractUpgradeSeveralUpgradesOnStarTopology (26.03s)
--- PASS: TestTendermintNoQuorum (12.97s)
    --- PASS: TestTendermintNoQuorum/test_case_2_validators,_one_goes_down_after_block_3 (4.96s)
    --- PASS: TestTendermintNoQuorum/test_case_3_validators,_two_go_down_after_block_3 (8.01s)
--- FAIL: TestTendermintStarSuccess (10.48s)
    --- FAIL: TestTendermintStarSuccess/test_case_no_malicious (10.48s)
--- FAIL: TestTendermintStarOverParticipantSuccess (11.01s)
    --- FAIL: TestTendermintStarOverParticipantSuccess/test_case_no_malicious (11.01s)
--- FAIL: TestTendermintBusSuccess (10.97s)
    --- FAIL: TestTendermintBusSuccess/test_case_no_malicious (10.96s)
--- FAIL: TestTendermintChangeTopologyFromBusToStarSuccess (59.01s)
    --- FAIL: TestTendermintChangeTopologyFromBusToStarSuccess/test_case_no_malicious (59.01s)
--- FAIL: TestTendermintChangeTopologyFromStarToBusSuccess (58.97s)
    --- FAIL: TestTendermintChangeTopologyFromStarToBusSuccess/test_case_no_malicious (58.97s)
--- FAIL: TestTendermintAddValidatorsToTopologySuccess (53.01s)
    --- FAIL: TestTendermintAddValidatorsToTopologySuccess/test_case_no_malicious (53.01s)
``` 

# Acceptance Criteria
The following tests should not fail if run together
```
TestTendermintStarSuccess
TestTendermintStarOverParticipantSuccess
TestTendermintBusSuccess
TestTendermintChangeTopologyFromBusToStarSuccess
TestTendermintChangeTopologyFromStarToBusSuccess
TestTendermintAddValidatorsToTopologySuccess
```

The following tests should be added to CI
```
TestContractUpgradeSeveralUpgrades
TestContractUpgradeSeveralUpgradesOnBusTopology
TestContractUpgradeSeveralUpgradesOnStarTopology
TestContractUpgrade_Success
TestRemoveFromValidatorsList
TestTendermintAddValidatorsToTopologySuccess
TestTendermintBusSuccess
TestTendermintChangeTopologyFromBusToStarSuccess
TestTendermintChangeTopologyFromStarToBusSuccess
TestTendermintNoQuorum
TestTendermintStarOverParticipantSuccess
TestTendermintStarSuccess
```

### LINKED ISSUE #465: Simplifying CI config
# Problem

The github actions CI config is quite fragmented because there are many jobs which causes it to be very verbose and have a lot of duplication. This leads to hard to understand CI config, which in turn leads to mistakes. For example repeated tests #462 or missing tests #464.

The fragmentation also makes updating things that are common between jobs a lot of work. See https://github.com/clearmatics/autonity/pull/457/files#diff-25a57620242e0929b57dd916b7894b32 for the changes required to install a binary and generate contract source. This pull request shows how a simple change ends up as total overkill in the CI config. The result being that the CI as it stands can prevent us from achieving the desired processes.

# Implementation

The large number of jobs is due to the fact that we can run tests quicker by having many jobs which parallelise the work.

A possible quick fix is that there are a number of short testing jobs that could be combined and still execute in less time than the main test job.

Better than that though would be to tackle the source of the problem, which is slow tests. If we can speed up our tests sufficiently we should be able to put all tests into a single job, ideally we could get build and test into a single job.  

# Acceptance Criteria

Reduce the size of the ci yaml config.

# Notes
Probably speeding up tests deserves its own ticket and should be done first.

screwyprof:
fixes #462 
fixes #463 
fixes #464 
fixes #465

The following tests are disabled in code and should be turned on when #455 is ready:
-  TestTendermintChangeTopologyFromBusToStarSuccess,
-  TestTendermintChangeTopologyFromStarToBusSuccess,
-  TestTendermintAddConnectionToTopologySuccess,
-  TestTendermintAddValidatorsToTopologySuccess,
-  TestTendermintAddParticipantsToTopologySuccess,
-  TestTendermintAddStakeholdersToTopologySuccess


piersy:
@screwyprof This looks like a big improvement :+1: 

I have some questions/suggestions.

`tests` and  `tests-tendermint` look like they are both running the tests under `/consensus/tendermint` with coverage. Can we remove `tests-tendermint`? If not can we document them to clarify their purpose, or change their names to be more descriptive?

Since `tendermint_e2e_tests`, `race_tests` and `contract_tests` are all executing in under a minute, could we combine these into one job (eg `tests_extended`) and run them as separate steps? I don't see this slowing people down since they still have to wait for `tests`.

In some places we run things through `build/ci.go` and in others we don't, I find this quite confusing since the ci adds some flags and you need to know what flags it will add to understand the meaning of running it. Could we standardise on just calling go directly, or I think even better than that, could we instead only call make targets, which will make it easy for us to locally run the steps taken by our CI. I see that it always adds the git commit and timestamp we could do this easily with a small script instead.

Also I wouldn't say that this fixes #464 since there are still tests mentioned in that ticket that are not being run by CI. Could we update #464 to make reference to #455 with the list of tests that can be enabled when #455 is complete.

screwyprof:
@piersy 
1. A nice idea to exclude tests-tendermint from tests. I've implemented it, could you review?
2. I don't see any benefits in combining some tests, I'd rather split them in smaller pieces. They all run really fast as separate jobs now.
3. After excluding tests-tendermint from tests we only use ci.go for go run build/ci.go install
4. All tests are in the CI now

piersy:
Good morning @screwyprof 

1. I see you have filtered the tendermint tests from `tests` but it's not clear to me why we have 2 jobs for these tests, can we add a comment to the jobs to explain why tests under `/consensus/tendermint` are not run as part of `tests` and instead run in their own job?

2. The benefit in combining tests is reduced duplication in the workflows. Currently the following block is duplicated 6 times.

```

**Comment by piersy** [comment]
Good morning @screwyprof 

1. I see you have filtered the tendermint tests from `tests` but it's not clear to me why we have 2 jobs for these tests, can we add a comment to the jobs to explain why tests under `/consensus/tendermint` are not run as part of `tests` and instead run in their own job?

2. The benefit in combining tests is reduced duplication in the workflows and hence improved readability. Currently the following block is duplicated 6 times.

```
      - name: Set up Go 1.13
        uses: actions/setup-go@v1
        with:
          go-version: 1.13
        id: go

      - name: Set GOPATH
        # temporary fix
        # see https://github.com/actions/setup-go/issues/14
        run: |
          echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
          echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
        shell: bash

      - name: Check out code into the Go module directory
        uses: actions/checkout@v2
```

I don't see that it is very beneficial to have these jobs execute so quickly, given that you would need to wait another ~10 minutes for `tests` to complete before you know that a PR is in good shape. What are the benefits you perceive from having these jobs execute so fast?

3. I think it's an improvement to not use `ci.go` for tests, but why stop there? It obscures what is actually happening and I don't feel that its adding enough value to justify that extra complexity.

To be clear, all I can see the `ci.go` doing for us is adding the following flags to the build.
```
-ldflags -X main.gitCommit=e3f963b9090bf75368e01b25e2fa62ad5a40baf5 -X main.gitDate=20200316
```
Which I don't think is justified by the roughly 2000 lines of code in the build system.

4. Good stuff :+1:

**Comment by screwyprof** [comment]
@piersy 
Thanks for the feedback! 

We don't have any test duplication at the moment. We need separate test jobs to calculate overall code coverage and `tendermint` coverage in particular. When it comes to combining tests, I see no actual point here at the moment. The bottleneck now is not the test bootstrap, but in some particular tests (about 2-3) which run really slow. If we should want to make the bootstrap faster, we may add go mod dependencies caching.

I see your point about duplication code. I'll think over it.

Thank you!

**Comment by screwyprof** [comment]
I was thinking about the code duplication the CI config file when I recalled that YML supports anchors to avoid duplication, but unfortunately this feature doesn't work for github actions :(
https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336

Having said that, I don't think that fighting the dups in yml makes a lot of sense, providing that we would sacrifice CI speed.

**Comment by piersy** [comment]
@screwyprof  Yeah, its a real shame that anchors are not supported, that was one of my first thoughts.

With regards to duplication and the number of jobs, you talk about sacrificing speed. From my perspective when I submit a PR, I'm waiting for that signal that the whole CI has passed, so I'll be waiting 11 minutes for that. Having tests that complete in 1 minute doesn't change the fact that I still have to wait 11 minutes to know nothing is broken. So for me as long as no other jobs take longer than 11 minutes I wouldn't consider speed to have been sacrificed. So given that combining the faster jobs wouldn't sacrifice speed from my perspective, it would make sense to cut down on some of the duplication.

I'd just like to understand why it's important for you to have these jobs run so much faster than `test`?

-----------------------------------------------------------------------------

Did you have any thoughts on removing use of `ci.go` entirely?
 
We could replace `go run build/ci.go install` with
```
go install -ldflags "-X main.gitCommit=$(./commit.sh) -X main.gitDate=$(./commit_date.sh)" -v ./...
```
And the following scripts
`commit.sh:`
```
#!/bin/bash
git rev-parse HEAD
```
`commit_date.sh:`
```
#!/bin/bash
git show -s  --date=format:'%Y%m%d' --format=%cd HEAD
```

**Comment by piersy** [comment]
@screwyprof I just noticed, all the checks seem to be running twice!

File: .github/workflows/go.yml [hunk 1]
```diff
@@ -79,17 +90,51 @@ jobs:
...
+        env:
+          cache-name: cache-golang-modules
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-
+
       - name: Install
         run: go run build/ci.go install
```
piersy: I don't think we need to run this to be able to run the tests.

screwyprof:
@piersy 
Starting all the tests in parallel it is possible to provide a fail fast result for the developers. It means that we don't need to wait for all the test to finish if some of them have already failed.

I added some caching to make tests start faster. Removing ci.go from the pipeline is out of scope of this task. It can be done in a separate ticket. I would also add that we can neglect ldflags for go install to make it even simpler.

codecov.yml
```diff
@@ -2,8 +2,12 @@ ignore:
   - "**/*_mock.go"  # ignore mock files
   - "**/*_gen.go"  # ignore generated client
   - "**/mock*.go"
-  - "docs"
+  - "**/*.^go"
+  - "**/docs/**/*"
```
piersy: I ignored docs because I didn't realise you could use the negative marcher in the ignore patterns, but now that we ignore anything that is not a `.go` file we can get rid of the rule to ignore docs.

consensus/test/topology_test.go
```diff
@@ -114,6 +114,8 @@ func TestTendermintBusSuccess(t *testing.T) {
 }
 
 func TestTendermintChangeTopologyFromBusToStarSuccess(t *testing.T) {
+	t.Skip("Topology tests are not stable")
```
piersy: Is there a ticket to fix these tests that are not stable?

screwyprof:
> I just noticed, all the checks seem to be running twice!
Looks like a github actions's glitch. If you open the workflow you'll see that the even "on push" was triggered twice

screwyprof:
> Is there a ticket to fix these tests that are not stable?
@b00ris is on it

codecov.yml
screwyprof: fixed

consensus/test/topology_test.go
screwyprof: @b00ris is on it

piersy: I don't actually see those tests mentioned in the requirements of any tickets. I'd want to avoid the situation where those tests are ignored and forgotten. Forgotten tests was the reason I created #464 in the first place.

So I'm wondering what exactly is the chain of events that will ensure that those ignored tests are not forgotten?

piersy:
@screwyprof 
>Starting all the tests in parallel it is possible to provide a fail fast result for the developers. It means that we don't need to wait for all the test to finish if some of them have already failed.

So that makes sense to me, but the decision seems arbitrary, why single out only one package to be run in a separate job and fail fast? 

If we are valuing the fail fast feedback over simplicity of CI config, why not create a separate job for each package? Then we will have much faster feedback across the board.

So my question would be, what is the process that we follow to draw the line between separating a package to run in its own job (for the purposes of fail fast feedback) vs keeping a package in the `tests` job (resulting in simplified CI config)?

Before I submit a PR, I generally try to run all the tests locally, so I will be expecting the tests to pass. For that reason I'm unlikely to keep checking my pull request for the first few minutes and instead I'll check back in ~15 minutes to see that all went well. So I don't see the fail fast feedback being particularly useful for me.

@clearmatics/autonity What are your thoughts on having certain parts of the CI fail much faster than the time taken for CI as a whole to complete?

JekaMas:
"If we are valuing the fail fast feedback over simplicity of CI config, why not create a separate job for each package? Then we will have much faster feedback across the board."
Not really for units, we'll spend most of the time in bootstrapping of the job. Moreover, github actions have a limit of parallel jobs https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions  and we'll reach this limit.
For me, we should split all tests into nearly equal parallel jobs. We can threat e2e tests as equal in terms of time.

piersy:
>For me, we should split all tests into nearly equal parallel jobs. We can threat e2e tests as equal in terms of time

@JekaMas Are you saying that we should divide all non e2e tests into equally sized groups and run each group as a separate job, and that we could consider a single e2e test to be equal in terms of time to a group of non e2e tests?

JekaMas:
For me, 1 e2e test is a good enough group. I don't see reasons for more complex solutions.

piersy:
@JekaMas Could you just clarify what you mean by
>we should split all tests into nearly equal parallel jobs

JekaMas:
We have 36 jobs. Slow ones are:
* tests 13mins
* TestTendermintStartStopAllNodes 11mins
* TestTendermintStartStopFPlusTwoNodes 8mins
* TestTendermintStartStopFPlusOneNodes 6mins
* TestTendermintStartStopFNodes 6mins
* TestTendermintStartStopSingleNode 4mins
* TestTendermintStopUpToFNodes 3mins

in total 7 slow ones.

The rest of the jobs (29) take 1-1.5 mins each.

So we have 29 pretty much equal jobs and 7 not. We can split this 7 even further: e2e tests can be split into groups of subtests, units can be split into 4 parts.

However, I think that we can do it as a new iteration of improvements.

piersy:
@JekaMas Thanks for the clarification, I agree that that work would be better tackled under a new ticket.

JekaMas:
Guys, this pr block all the rest prs. Let's review it and merge or decide what we want to change.

consensus/test/topology_test.go
JekaMas: I've discussed this with @b00ris https://github.com/clearmatics/autonity/issues/431
I think https://github.com/clearmatics/autonity/issues/464 is a duplicate of #431 and should be closed.

piersy: #464 references TestTendermintChangeTopologyFromBusToStarSuccess which isn't referenced in #431, if we want to close #464 lets add TestTendermintChangeTopologyFromBusToStarSuccess to the list in #431

JekaMas: Just add this test to the #431

piersy: Sure, will do

PR #471, opened 2020-03-16
Eth_gasPrice API returns minGasPrice from the Autonity contract

### LINKED ISSUE #436: Eth_gasPrice API return minGasPrice
# Problem

Even though the Autonity Contract is offering an API to retrieve minGasPrice, users are getting confused as the default`eth_gasPrice` API is returning the default estimated gas price using the oracle logic instead of `minGasPrice` defined in the Autonity Contract. 

# Acceptance Criteria

Return Autonity Contract `minGasPrice` when calling `eth_gasPrice`

JekaMas:
closes #436


core/blockchain.go
```diff
@@ -2294,6 +2294,36 @@ func (bc *BlockChain) GetKeyValue(key []byte) ([]byte, error) {
...
+	var block *types.Block
+	if len(blockNumber) > 0 {
+		block = bc.GetBlockByNumber(blockNumber[0])
+	} else {
+		block = bc.CurrentBlock()
+	}
+
+	parent := bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
+
+	statedb, err := state.New(parent.Root, bc.stateCache)
```
yazzaoui: Why taking parent's state and not the current state?

JekaMas: A good point. I've fixed it and use current block now.

core/blockchain.go
```diff
@@ -2294,6 +2294,34 @@ func (bc *BlockChain) GetKeyValue(key []byte) ([]byte, error) {
...
 }
 
+func (bc *BlockChain) GetMinGasPrice(blockNumber ...uint64) (*big.Int, error) {
+	if bc.autonityContract == nil {
+		return nil, errors.New("the autonity contract is not specified")
+	}
+
+	var block *types.Block
+	if len(blockNumber) > 0 {
+		block = bc.GetBlockByNumber(blockNumber[0])
```
aneequesafdar: I think there should be a nil check to make sure that if the user asks for a block number much higher than the block mined.

JekaMas: Thank you, Aneeque! I've added this check/

Issue #326, opened 2019-10-28
Tendermint core-backend re-architecture


yazzaoui:
## Tendermint core-backend re-architecture

- Backend should have a reference of core instead of the opposite. All functions called from the rest of the codebase should be directed to backend implementing `Engine`.
If this architecture is difficult to realise then I suggest merging core and backend and creating a single package.

- Drop complete support from clique and ethash: This system is designed around the BFT consensus engines at its core and statements like `if h, ok: = w.engine.(consensus.Handler);` that clutter the rest of the codebase have no reason to exists. As a consequence, we can merge the `BFT` and `Handler` interfaces with `Engine` .

- This task is now high priority, it will need to be done one day so better to do it now.

JekaMas:
Maybe we should do it another way: write down all the functionality, group it, write schemes and then do refactoring? Maybe we need to rewrite everything as a stateless state machine for further verification.

For me, it won't change much if we do the refactoring as it's described in the issue

PR #472, opened 2020-03-16
Tendermint package refactor

aneequesafdar:
- core is private and is only available to backend package, everyone outside of the tendermint package is only aware of backend through the consensus.Engine
- backend/backend.go has a reference to core interface
- core no longer has reference to backend struct, rather through an interface
- CreateConsensusEngine no longer takes `cons` and `backs` as an argument (since it was meant only for malicious tests), now only takes cons to modify the engine interface.
- E2E tests are only aware of the backend now
- Committee set refactor added tests for F() and Quorum sizes

The bus topology test cases which fail are:
[1585037042.0052538.zip](https://github.com/clearmatics/autonity/files/4438944/1585037042.0052538.zip)
[1585064750.374162.zip](https://github.com/clearmatics/autonity/files/4438945/1585064750.374162.zip)
[1585077799.096365.zip](https://github.com/clearmatics/autonity/files/4438946/1585077799.096365.zip)
[1585090856.1116016.zip](https://github.com/clearmatics/autonity/files/4438948/1585090856.1116016.zip)


JekaMas:
Could you fix the tests and linters, it looks like they are broken?

aneequesafdar:
Yes, I am working on it, I will need some help.

aneequesafdar:
fixes first bullet point of #326

consensus/tendermint/core/handler_test.go
```diff
@@ -163,46 +164,20 @@ func TestCoreStopDoesntPanic(t *testing.T) {
...
+	c.stopped <- struct{}{}
+	c.stopped <- struct{}{}
+	c.stopped <- struct{}{}
 
 	if err := c.Stop(); err != nil {
 		t.Fatal(err)
 	}
 }
-
-func TestCoreMultipleConcurrentStopsDontPanic(t *testing.T) {
```
JekaMas: Why have you removed this test?

consensus/tendermint/core/handler.go
```diff
@@ -69,22 +49,6 @@ func (c *core) Start(ctx context.Context, chain consensus.ChainReader, currentBl
 
 // Stop implements core.Engine.Stop
 func (c *core) Stop() error {
-	// prevent double stop
```
JekaMas: Are you sure that this can be removed?

consensus/tendermint/core/handler.go
```diff
@@ -19,38 +19,18 @@ package core
...
 	"github.com/clearmatics/autonity/common"
-	"github.com/clearmatics/autonity/consensus"
 	"github.com/clearmatics/autonity/consensus/tendermint/crypto"
 	"github.com/clearmatics/autonity/consensus/tendermint/events"
 	"github.com/clearmatics/autonity/core/types"
 )
 
 // Start implements core.Engine.Start
-func (c *core) Start(ctx context.Context, chain consensus.ChainReader, currentBlock func() *types.Block, hasBadBlock func(hash common.Hash) bool) error {
-	// prevent double start
```
JekaMas: Do you have a test for double start?

consensus/tendermint/core/handler_test.go
```diff
@@ -163,46 +164,20 @@ func TestCoreStopDoesntPanic(t *testing.T) {
...
+	backendMock.EXPECT().Subscribe(gomock.Any()).Return(sub).MaxTimes(5)
 
-	c := New(backendMock, nil)
-	if err := c.Stop(); err != nil {
-		t.Fatal(err)
-	}
+	c := New(backendMock)
+	_, c.cancel = context.WithCancel(context.Background())
+	c.subscribeEvents()
+	c.stopped <- struct{}{}
```
JekaMas: Why this is called in the test?

consensus/tendermint/core/handler_test.go
aneequesafdar: I have moved it to backend/enginge_test.go, since now backend's start and stop functions are called multiple time, core no longer implements consensus.Engine. I also wrote a few more test cases for it.

consensus/tendermint/core/handler.go
aneequesafdar: Yes because core no longer implements consensus.Engine and this start and stop function are called in backend start and close function respectively. If there is a double start or stop it will be for backend, for which I have moved the tests and written a few more cases.

consensus/tendermint/core/handler.go
aneequesafdar: yes please check engine_test.go

consensus/tendermint/core/handler_test.go
aneequesafdar: Not sure what you mean? We are not calling start before stop, hence `c.stopped <- struct{}{}` need to be called otherwise there is a deadlock

consensus/tendermint/backend/api_test.go
```diff
@@ -25,8 +25,10 @@ func TestGetCommittee(t *testing.T) {
 	}
 
 	want := types.Committee{val}
-	committeeSet := committee.NewMockSet(ctrl)
-	committeeSet.EXPECT().Committee().Return(types.Committee{val})
```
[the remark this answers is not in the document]
aneequesafdar: As far as I understood, we should only use mock whether the real dependency takes too much time or cause a lot of other set-up which is not desired for tests. Also it was strange for me to see that some test files in the same package were using actual validator/committee set (backend_test.go), while the api_test.go was using mocks, so I decided to stay consistent and not use mocks.

consensus/tendermint/core/precommit_test.go
```diff
@@ -474,8 +474,12 @@ func TestHandleCommit(t *testing.T) {
 	backendMock := NewMockBackend(ctrl)
 	backendMock.EXPECT().LastCommittedProposal().MinTimes(1).Return(block, addr)
 
-	committeeSet := committee.NewMockSet(ctrl)
-	committeeSet.EXPECT().IsProposer(gomock.Any(), addr).Return(false)
+	testCommittee, _ := generateCommittee(3)
```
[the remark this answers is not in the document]
aneequesafdar: I think the test logic remains the same (otherwise test would fail, which it did when I was changing it) because we want to have an address for round 2 who is not the proposer, therefore we are setting the committee set in such a way that for round 2 isProposer will be false.

consensus/tendermint/core/propose_test.go
```diff
@@ -193,11 +205,14 @@ func TestHandleProposal(t *testing.T) {
 			Signature:     []byte{0x1},
 		}
```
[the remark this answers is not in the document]
aneequesafdar: I don't think it does (same reasoning as above). Also it was strange to see both actual validator set and mock validator set used in the same package tests, we should use one or the other

consensus/tendermint/core/precommit_test.go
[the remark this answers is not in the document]
JekaMas: It can test something different and be a 'green' one.
Could you log consensus from develop branch and from refactoring and compare?

consensus/tendermint/core/propose_test.go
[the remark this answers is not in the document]
JekaMas: We run a few participants so it's ok to use normal consensus and mocked one.

I'd prefer to keep test logic as it was.

consensus/tendermint/backend/api_test.go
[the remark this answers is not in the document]
JekaMas: Nope, the mock is also needed to check that the call was done with particular parameter. At the moment refactored test doesn't do this checks.

consensus/tendermint/core/handler_test.go
JekaMas: I've tried TestCoreMultipleConcurrentStopsDontPanic on the PR and I got a panic
```
=== RUN   TestCoreMultipleConcurrentStopsDontPanic
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8c6e60]

goroutine 16 [running]:
github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Stop(0xc00013b900, 0x0, 0x0)
	/mnt/sdb/go/src/github.com/clearmatics/autonity/consensus/tendermint/core/handler.go:58 +0x140
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc000232300, 0xc0001d94d0)
	/mnt/sdb/go/pkg/mod/golang.org/x/sync@v0.0.0-20190911185100-cd5d95a43a6e/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
	/mnt/sdb/go/pkg/mod/golang.org/x/sync@v0.0.0-20190911185100-cd5d95a43a6e/errgroup/errgroup.go:54 +0x66

Process finished with exit code 1
```

JekaMas: I think we need to fix the panic and restore tests for multiple starts and stops.

aneequesafdar:
closing in favour of #512

PR #478, opened 2020-03-23
Permission tests with different roles

JekaMas:
Tests that users not listed in the Autonity contract can't get blocks or transactions. Test with different topologies and fraction of external(not from the contract) users.

closes https://github.com/clearmatics/autonity/issues/432


consensus/test/topology_roles_test.go
```diff
@@ -0,0 +1,561 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/clearmatics/autonity/common/graph"
+)
+
+func TestTendermintExternalUser(t *testing.T) {
```
piersy: I can't actually understand what this test is testing, it would be good if you could provide an explanatory comment.

In fact I can't actually understand what any of these tests are meant to be testing. Do we have an explanation of the vocabulary used in test names?

By way of example let me break down a test name `TestTendermintBusSuccessWithExternalUser`, what can I infer from this name?

Its testing the tendermint protocol with some configuration that we refer to as 'Bus' with an external user, and we expect it to succeed. 

I don't know what bus is but that is something that I can find out.

More importantly though, to be able to asses this test case, I need to understand what exactly success is, how do we check that? Its also not clear to me in what way is an external user being integrated into this test case, how is the external user connected to the network what is the external user doing?

JekaMas: Does it makes sense to put this extra information to the test? It tests topology and it assumes that external user is familiar with different network topologies.

Could you clarify who is an external user? If it's an autonity developer than all of us have to dig into golang geth and autonity code and iteratively be more and more familiar with it. If it's a non-developer than why he or she reading the source code?

* More information about bus network topology you can find on wiki https://en.m.wikipedia.org/wiki/Bus_network

consensus/test/topology_roles_test.go
```diff
@@ -0,0 +1,561 @@
...
+
+	for _, testCase := range cases {
+		testCase := testCase
+		t.Run(fmt.Sprintf("test case %s", testCase.name), func(t *testing.T) {
+			runTest(t, testCase)
+		})
+	}
+}
+
+func TestTendermintMajorityExternalUsers(t *testing.T) {
```
piersy: This test fails for me
```
$ go test -v -run TestTendermintMajorityExternalUsers/test_case_no_malicious
=== RUN   TestTendermintMajorityExternalUsers
=== RUN   TestTendermintMajorityExternalUsers/test_case_no_malicious
signal: killed
FAIL	github.com/clearmatics/autonity/consensus/test	3.702s
```

consensus/test/topology_roles_test.go
```diff
@@ -0,0 +1,561 @@
...
+
+	for _, testCase := range cases {
+		testCase := testCase
+		t.Run(fmt.Sprintf("test case %s", testCase.name), func(t *testing.T) {
+			runTest(t, testCase)
+		})
+	}
+}
+
+func TestTendermintMajorityExternalUsersFullyConnected(t *testing.T) {
```
piersy: This test also fails for me
```
$ go test -v -run  TestTendermintMajorityExternalUsersFullyConnected
=== RUN   TestTendermintMajorityExternalUsersFullyConnected
=== RUN   TestTendermintMajorityExternalUsersFullyConnected/test_case_no_malicious
signal: killed
FAIL	github.com/clearmatics/autonity/consensus/test	3.534s
```

JekaMas: "signal: killed" - it's like ctrl+C or OOM. Do you have enough resources to run 6 Autonity nodes? Have you interrupted the test?

piersy: Sure enough, that was the case. I have a very small swap partition, I'm going to see if I can resize it.

Could you let me know how much memory and swap you have?

JekaMas: 64 Gb :)

consensus/test/topology_roles_test.go
piersy: >Does it makes sense to put this extra information to the test?

I think having a natural language explanation of what a test (or any code) does is important.

Otherwise the reader has no way to verify that the test is checking the correct thing.

The test may pass incorrectly because it is not checking the correct conditions. The reader has no way to determine this, because they did not know the intent of the test, the only thing letting them know the intent of the test is the test itself, and since the test passes they can assume it is correct and that the conditions checked in the test were the intended conditions.

I'm still not clear on what exactly "success" means, is it that no nodes crash, that a block is mined, that everyone mines x blocks, that the external user is not let in?

With respect to my last two questions, I think maybe the problem could be I don't know what an external user is. What does it mean to be an external user, at least having a better understanding of that could be a starting point.

JekaMas: Shall I replace `Success` in test name with `ChecksThatNonOfTheBlockchainBlocksOrTransactionReceivedByAnUnknownUserButAllOfThemReceivedByNetworkParticipants`? I'm not very sure...

piersy: Haha :) that is a bit of a mouthful!

But we could add a bit of package doc that explains the concept of external user (i.e not a participant). And that tests involving external users validate that the network proceeds as would be expected without the presence of an external user and also that the external user receives no blocks or transactions.

consensus/test/topology_roles_test.go
piersy: Thanks :+1:

eth/handler.go
```diff
@@ -126,6 +128,7 @@ func NewProtocolManager(config *params.ChainConfig, checkpoint *params.TrustedCh
 		quitSync:    make(chan struct{}),
 		engine:      engine,
 		whitelistCh: make(chan core.WhitelistEvent, 64),
+		pub:         pub,
```
yazzaoui: why the public key and not the nodeId ?

consensus/test/topology_roles_test.go
piersy: Actually, probably just having a file level comment instead of adding to package doc would be better. I say this because the package doc should be contained in a single file and so we wouldn't be able to keep the explanations for all the different tests in the same files as the tests.

JekaMas: However, a term `success` can be different depending on the context of different tests.

piersy: In that case we should at least explain what success means at the test level, but we could keep the description of external user in one place in a file level comment.

I've just realised that these test names are actually just a bunch of existing tests with `WithExternalUser` appended to their names. I'm assuming that we are expecting these tests to have the same outcome as their counterparts without the `WithExternalUser` suffix.

So we could just explain that and the meaning of external user in a file level comment. 

I am still confused though about what `TestTendermintExternalUser` is testing though?

PR #481, opened 2020-03-24
Add make target to embed autonity contract

### LINKED ISSUE #456: Automate contract bytecode and abi update
# Problem

Currently this task has to be done manually, which is time consuming and error prone.

# Acceptance Criteria

When building the client the contract bytecode and abi corresponding to the current version of the autonity contract should be baked into the go binary without need for manual intervention.

piersy:
The new make target generates the autonity abi and bytecode and embeds them in the
codebase.

Fixes #456 

The abi and bytecode are no longer committed to git. This helps ensure that we run tests with the correct abi and bytecode, it removes the burden of having to remember to do this step manually after updating the contract. 

It also helps provide consistency by enforcing that a single solc version is used to perform the generation.


Makefile
```diff
@@ -15,12 +15,44 @@ ifeq ($(LATEST_COMMIT),)
...
-all:
+embed-autonity-contract: $(GENERATED_BYTECODE) $(GENERATED_ABI)
+
+# NOTE previously we were using
+# https://github.com/ethereum/solidity/releases/download/v0.5.1/solc-static-linux
+# this binary does not produce the same bytecode as the ethereum/solc:0.5.1
+# docker image. This was causing tests to fail.
+$(GENERATED_BYTECODE) $(GENERATED_ABI): $(AUTONITY_CONTRACT_DIR)/$(AUTONITY_CONTRACT)
+	mkdir -p $(GENERATED_CONTRACT_DIR)
+	docker run --rm -v $(CURDIR)/$(AUTONITY_CONTRACT_DIR):/contracts -v $(CURDIR)/$(GENERATED_CONTRACT_DIR):/output ethereum/solc:0.5.1 --overwrite --abi --bin -o /output /contracts/$(AUTONITY_CONTRACT)
```
aneequesafdar, line 37: can you please explain what each of the arguments are doing here?

Makefile
```diff
@@ -15,12 +15,44 @@ ifeq ($(LATEST_COMMIT),)
...
+
+	@echo Generating $(GENERATED_BYTECODE)
+	@> $(GENERATED_BYTECODE)
```
aneequesafdar, line 40: Can you please explain what does `@>` mean?

Makefile
```diff
@@ -15,12 +15,44 @@ ifeq ($(LATEST_COMMIT),)
...
+
+	@echo Generating $(GENERATED_ABI)
+	@> $(GENERATED_ABI)
+	@echo 'package generated' >> $(GENERATED_ABI)
+	@echo -n 'const Abi = `' >> $(GENERATED_ABI)
+	@cat  $(GENERATED_CONTRACT_DIR)/Autonity.abi | json_pp  >> $(GENERATED_ABI)
+	@echo '`' >> $(GENERATED_ABI)
+	@gofmt -s -w $(GENERATED_ABI)
+
+all: embed-autonity-contract
```
aneequesafdar, line 55: What does it mean to have `embed-autointy-contract` after `all` target?

Makefile
piersy: Hi @aneequesafdar.

This line is executing solc from within a docker image on the autonity contract and outputting the contract bytecode and abi. Using docker means that this can be run on both mac and linux, otherwise we would need to detect the OS in make and pull a relevant version of solc.

The `$(...)` Notation is just make's syntax for variable expansion, you can see the variables defined above, I think they are fairly self explanatory. Except for `CURDIR` which is set by make and is described as follows [here](https://www.gnu.org/software/make/manual/html_node/Quick-Reference.html).

```
CURDIR
Set to the absolute pathname of the current working directory (after all -C options are processed, if any). Setting this variable has no effect on the operation of make.
See Recursive Use of make.
```

The command as a whole is a docker run command (documented [here](https://docs.docker.com/engine/reference/commandline/run/)) which takes the following form
```
docker run [OPTIONS] IMAGE [COMMAND] [ARG...]
```

I'm making use of the following arguments to run which are
```
--rm | Automatically remove the container when it exits
--volume , -v | Bind mount a volume
```

When you run a docker image docker makes it into a container, we could think of that as an instance of the image, using `--rm` ensures that docker doesn't keep the dead container after solc has finished running, it's not critical to the command functioning correctly but it stops us having a buildup of  dead containers, which consume disk space.

Mounting a volume in docker is a way to make a file or directory on your local machine available inside the docker container.

The form for mounting a container is `-v <local_machine_absolute_path>:<docker_container_absolute_path>`

In this case I'm mounting directories for the input and output of the solc command.

Following the mounting arguments we have the docker image name `ethereum/solc:0.5.1` which is just an empty docker image with solc v0.5.1 inside.

If you look [here](https://hub.docker.com/layers/ethereum/solc/0.5.1/images/sha256-4ec7958dfb2a4ec9842c5aea23ded91a9ece301f6b71cd671af29797da0bc267?context=explore) you can see the image history and it can be seen that the last command to be run was.

```
 ENTRYPOINT ["/usr/bin/solc"]
```

This sets up the docker image to execute that whenever it is run, meaning that you only specify the arguments to the command. I don't like using ENTRYPOINT because it makes understanding the docker commands harder, but I can't change the image retrospectively. We could make our own but seems like overkill.

So immediately after `ethereum/solc:0.5.1` we specify the arguments to solc, which I guess you are familiar with. I'm just specifying the input and output with respect to the mounted directories.

The outcome of running the command is that we end up with the contract abi and bytecode in `$(CURDIR)/$(GENERATED_CONTRACT_DIR)`

Having explained that I realise that there is a lot going on there, so this probably warrants some more explanation in the Makefile.

I think what I have just explained is probably too detailed, but I would appreciate you letting me know what you think is a reasonable level of explanation for the Makefile.

Makefile
piersy: The `@` is a make thing, without the `@` make outputs the line to stdout before executing it.

Makefile
piersy: The elements that follow the target are referred to as prerequisites. They can be files, targets or files that are also targets.  In this case this means ensure you have embedded the autonity contract before you try to build anything in the all target.

That is a somewhat simplified explanation, prerequisites are a fundamental pillar of make and I encourage you to check out [this](https://blog.mindlessness.life/2019/11/17/the-language-agnostic-all-purpose-incredible-makefile.html) blogpost which I found very helpful in establishing my understanding of make.

The structure of a rule is eplained [here](https://www.gnu.org/software/make/manual/html_node/Rule-Example.html#Rule-Example) in the make manual, which I think is pretty well written.

Makefile
aneequesafdar: Thank you very much for such a detailed explanation @piersy, it has been very helpful for me to understand!

Makefile
aneequesafdar: I see, so `@`is just to hide the command from the prompt

Makefile
aneequesafdar: Thanks for the link, I will have a read

Makefile
piersy: np :)

Makefile
piersy: Yes that is all it does.

PR #482, opened 2020-03-24
Add make target to embed autonity contract (go:generate impl)

### LINKED ISSUE #456: Automate contract bytecode and abi update

piersy:
This provides the same functionality as #481 but the generation is done with go:generate.

Fixes #456 

I think this solution is slightly inferior to the one proposed in #481 for the following reasons.

* Knowledge of the contract path and the paths to generate is duplicated between the Makefile and where go:generate is happening.
* The generation commands became more fiddly because we need to invoke bash explicitly with a quoted command which gives rise to the need to escape inner quotes.
* Its slower (20s as opposed to 8s on CI) because of the need to pull modules for go.

I feel like this is not the intended way to use go:generate, most go generate lines I see in this repo are one line commands that do something with a type, this by contrast seems like shoehorning a script into go:generate.


JekaMas:
I'd rather stay with `go generate`, if we want to improve something we can write our generator and encapsulate all the bash, docker commands there. Like a simple template + a bit of generation magic.

piersy:
@JekaMas Why would you rather stay with go generate?

JekaMas:
It's easy. We already use go generate to generate some code and even after this PR we'll still use it. However, this PR introduces a new instrument for the same purposes. So we'll have 2 instruments in different parts of the code. Also, it's unclear what benefits of the script are.

piersy:
@JekaMas 
Could you elaborate on what you mean by "easy"? I would consider both this and #481 equally easy since you just run `make embed-autonity-contract`.

> We already use go generate to generate some code and even after this PR we'll still use it

Sure, I'm not suggesting that we should change that.

> However, this PR introduces a new instrument for the same purposes.

Do you mean the script?

>  Also, it's unclear what benefits of the script are

If you are talking about this
```
//go:generate mkdir -p generated
//go:generate bash -c "docker run --rm -v $(pwd)/../../contracts/autonity/contract/contracts:/contracts -v $(pwd)/generated:/output ethereum/solc:0.5.1 --overwrite --abi --bin -o /output /contracts/Autonity.sol"

//go:generate echo Generating generated/bytecode.go
//go:generate bash -c "echo 'package generated' > generated/bytecode.go"
//go:generate bash -c "echo '// Code generated DO NOT EDIT.' >> generated/bytecode.go"
//go:generate bash -c "echo -n 'const Bytecode = \"' >> generated/bytecode.go"
//go:generate bash -c "cat  generated/Autonity.bin >> generated/bytecode.go"
//go:generate bash -c "echo '\"' >> generated/bytecode.go"
//go:generate gofmt -s -w generated/bytecode.go

//go:generate echo Generating generated/abi.go
//go:generate bash -c "echo 'package generated' > generated/abi.go"
//go:generate bash -c "echo '// Code generated DO NOT EDIT.' >> generated/abi.go"
//go:generate bash -c "echo -n 'const Abi = `' >> generated/abi.go"
//go:generate bash -c "cat  generated/Autonity.abi | json_pp  >> generated/abi.go"
//go:generate bash -c "echo '`' >> generated/abi.go"
//go:generate gofmt -s -w generated/abi.go
```
The script is what generates the code, without it you would not have any code generation.

piersy:
@JekaMas did you have any thoughts on my response?

JekaMas:
"script" I mean a small makefile command rather than migrating go generate commands to makefile.

For me the PR complicates the code, it's better to stay with go generate.

piersy:
It looks like people (including @JekaMas ) approve the alternate PR so I'm closing this one.

PR #484, opened 2020-03-25
Fix resolver

### LINKED ISSUE #451: Consensus is stuck when an invalid enode is added.
## How to reproduce
1. `git checkout develop`
2. `make autonity`
3. Deploy a network of 4 nodes:
`./main.py /home/ya/dev/autonity/build/bin/autonity -n 4`
4. Make sure mining works and new blocks are minted.
5. Add a random node via console:
`autonity.addParticipant("0xb605e98649Fb40256d73f6C321cFa399BB6534CD", "enode://e7507c6c67d1307010e7ffb07a572b868b5dc65bc264ccbdcd94c0ff8d9023c25eb090b1dd5ef9948bb9847c506452c204b37187aa4954cbc06592a8358abb94@127.0.0.1:8527", {from: eth.accounts[0], gasPrice:500000})`

## Current behaviour
Consensus is stuck.  The invalid nodes are being parsed many times (the attempts messages are duplicated)

## Expected behaviour
If a new validator is being added and parsed unsuccessfully, the consensus should  be still functioning given that a quorum (2/3 of the network) is still well-behaving.

screwyprof:
This PR fixes #451 by adding error handling for cases when an invalid e-node is added. The lack of this led to a long resolution cycle which drastically slowed down the consensus. (One block per 3 minutes).

### DIFF
--- a/consensus/test/autonity_contract_test.go
+++ b/consensus/test/autonity_contract_test.go
@@ -9,7 +9,9 @@ import (
 	"github.com/clearmatics/autonity/common/keygenerator"
 	"github.com/clearmatics/autonity/common/math"
 	"github.com/clearmatics/autonity/log"
+	"github.com/clearmatics/autonity/p2p/enode"
 	"math/big"
+	"net"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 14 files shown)


piersy:
@screwyprof It would be really useful if you could provide a summary of what the problem was and how this pull request fixes it. "Fix resolver" doesn't mean anything to me, so its very hard to review the changes without even having an idea of what effect the changes are meant to have.

screwyprof:
@piersy I've updated the PR description. All the details can be found in the code. The test cases highlight the problem.

piersy:
@screwyprof Great, thanks :+1:

consensus/test/autonity_contract_test.go
```diff
@@ -340,6 +342,159 @@ func TestRemoveFromValidatorsList(t *testing.T) {
 	runTest(t, testCase)
 }
 
+func TestAddIncorrectStakeholdersToList(t *testing.T) {
```
piersy: It would be helpful to have a natural language explanation of what this test does. So the reader could know what constitutes an incorrect stakeholder (I feel like that could be so many things e.g. a stakeholder with a corrupted enode). Also to know what is the expected outcome of adding an incorrect stakeholder to the list?

JekaMas: It's a tradeoff between having test names > 200 characters and reading some of the test's code. I'm not sure where is the balance and how expressed and detail test names should be.

I think it should be decided by the whole team.
@aneequesafdar @screwyprof @b00ris @yazzaoui @Jason-Zhangxin-Chen

yazzaoui: What about writing a comment in the code explaining what's going on if someone asks for it?

p2p/enode/resolver.go
```diff
@@ -148,7 +163,7 @@ func (rs *resolveSet) Get(enodeStr string) (*Node, error) {
 			rs.addNoLock(enodeStr)
 		}
 		rs.Unlock()
-		node, err = rs.ParseV4WithResolveMaxTry(enodeStr, rs.maxTries, rs.delayBetweenTries)
+		node, err = rs.ParseV4WithResolveMaxTry(enodeStr, 1, rs.delayBetweenTries)
```
yazzaoui: would have been nice to keep trying to resolve in the background, but don't think it's easily achievable,

JekaMas: Maybe I can add a kind of cache.

consensus/test/autonity_contract_test.go
JekaMas: It's ok after someone has read the code :)

p2p/enode/resolver.go
JekaMas: I've written @screwyprof my idea of `bad enodes cache` and he's going to all a commit

consensus/test/autonity_contract_test.go
piersy: > It's ok after someone has read the code :)

Unless of course the test is broken and the code is not checking what it was intended to be checked. An over trivialised example that I think gets the point across.

```go
func TestGeneratedPasswordNotInvalid(t *testing.T) {
	password := GeneratePassword()

	if len(password) < 14 {
		t.Fail()
	}
}
```

That test looks fine to any reader, but what if the definition of valid password is that in fact passwords must be at least 14 characters ling and contain a lowercase letter, capital letter and a number.

JekaMas: I can't find the test you've mentioned. However, for me, it's obvious that a correct password >= 14 characters long. I don't see any problems. If you want to comment or document it, you always can do it.

PR #486, opened 2020-03-26
Consensus Voting Power integration

yazzaoui:
closes https://github.com/clearmatics/autonity/issues/409

### Autonity Contract:
- upgraded to use solidity 0.6.4
- Improved consistency with Autonity contract variable naming conventions and other minor things.
- Committee set calculation now done and returned at finalize step to reduce evm invocations.
- Fee reward distribution using standard evm logging facilities instead of layer1 tight coupling, the rationale was to keep the mechanism as generic as possible.

### Layer 1:

- Refactoring of finalize, finalizeAndAssemble and proposal verification to reduce code redundancy. Changed finalize function signature in this optic. 
- Integration of consensus messages weighted by voting power.
- Removed fee redistribution metrics
- Added a receipt for the Autonity Contract finalize evm call, in order to support logging.
E.g. To measure now fee redistribution or any other event thrown during the finalize call, one can do the following using the REPL console:
```
autonity.BlockReward().watch(function(error, result){
    if (!error)
        console.log(result.args._args);
});
```
or use the `eth_getLogs` RPC API. 

Every receipt needs to have an associated transaction hash. Since the Autonity Contract `finalize()` call is not creating a transaction, the transaction hash is thus being manually generated by using the block number 32 byte big-endian representation and replacing the first byte by 0xAC. This hash will be unique and easily recognizable. 
Other standard evm monitoring tools should be as well supported.

- Fixed issue with the block hash computation accounting the extra data field
- Enforcement of the header coinbase field to be equal to the original block proposer.


consensus/tendermint/backend/backend.go
```diff
@@ -183,15 +183,20 @@ func (sb *Backend) AskSync(valSet committee.Set) {
...
+			if count >= valSet.Quorum() {
 				break
 			}
 			sb.logger.Info("Asking sync to", "addr", addr)
 			go p.Send(tendermintSyncMsg, []byte{}) //nolint
-			count++
+
+			_, member, err := valSet.GetByAddress(addr)
+			if err != nil {
+				sb.logger.Error("could not retrieve member from address")
```
JekaMas: Maybe after logging we should skip this iteration with continue?

common/types.go
```diff
@@ -64,6 +64,13 @@ func BytesToHash(b []byte) Hash {
 // If b is larger than len(h), b will be cropped from the left.
 func BigToHash(b *big.Int) Hash { return BytesToHash(b.Bytes()) }
 
+// ACHash returns the Autonity Contract Finalize transaction Hash.
+func ACHash(b *big.Int) Hash {
```
JekaMas: Could you give a bit more thoughts why do we need this hash?

contracts/autonity/calls.go
```diff
@@ -197,16 +198,19 @@ func (ac *Contract) callGetMinimumGasPrice(state *state.StateDB, header *types.H
...
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
-
+	sort.Sort(committee)
 	// submit the final reward distribution metrics.
-	ac.metrics.SubmitRewardDistributionMetrics(&v, header.Number.Uint64())
-	return v.Result, nil
+	//ac.metrics.SubmitRewardDistributionMetrics(&v, header.Number.Uint64())
```
JekaMas: Should it be removed?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -578,14 +552,50 @@ contract Autonity {
...
+    */
+    function _sortByStake(User[] memory _validators) internal pure{
+        _structQuickSort(_validators, int(0), int(_validators.length - 1));
+    }
+
+    /*
+    * structQuickSort
+    * QuickSort algorithm sorting in ascending order by stake
+    */
+    function _structQuickSort(User[] memory _users, int _low, int _high) internal pure {
```
JekaMas: Could you create an issue to cover this changes by new contract tests?

consensus/tendermint/backend/backend.go
yazzaoui: Normally this error should never happen, I think we can even panic here...

JekaMas: Everything could happen :) I'd rather return an error with all the debug information that could be useful for us. If it's a practically unreal condition than if it happened then we'll have to debug it somehow.

contracts/autonity/contract/contracts/Autonity.sol
yazzaoui: There were already tests that Andres wrote for this logic, they are still valid I did just updated them.

common/types.go
yazzaoui: Sure, I've added an explanation in the PR description.

screwyprof:
LGTM

consensus/tendermint/backend/backend.go
```diff
@@ -327,64 +332,37 @@ func (sb *Backend) VerifyProposal(proposal types.Block) (time.Duration, error) {
...
 		}
 
-		if proposalNumber > 1 {
-			validators, err = sb.blockchain.GetAutonityContract().GetCommittee(sb.blockchain, header, state)
-			if err != nil {
-				return 0, err
-			}
-		} else {
-			// genesis block and block #1 have the same committee
-			s, error := sb.savedCommittee(1, sb.blockchain) //genesis block and block #1 have the same committee
```
aneequesafdar, line 359: As far as I understand at proposal verification we are no longer using block header to retrieve the committee and instead use finalize to call the contract to get the committee, why did you decide to no longer use saved committee anymore? Also could you please remind me where the committe is set for the header?

core/types/block.go
```diff
@@ -97,7 +97,7 @@ type Header struct {
 }
 
 type CommitteeMember struct {
-	Address     common.Address `json:"address"            gencodec:"required"`
+	Address     common.Address `json:"address"            gencodec:"required"       abi:"addr"`
```
aneequesafdar, line 100: What is this `abi` for?

core/chain_makers.go
```diff
@@ -262,7 +262,7 @@ func makeHeader(chain consensus.ChainReader, parent *types.Block, state *state.S
 		Number:    new(big.Int).Add(parent.Number(), common.Big1),
 		Time:      time,
 		Round:     0,
-		MixDigest: types.BFTDigest,
+		MixDigest: common.HexToHash("0x1234"),
```
aneequesafdar, line 265: Why chain the digest to use a dummy header hash?

core/state_processor.go
```diff
@@ -97,15 +97,16 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
-	if p.autonityContract != nil {
-		err := p.autonityContract.ApplyFinalize(block.Transactions(), receipts, block.Header(), statedb)
-		if err != nil {
-			return nil, nil, 0, err
-		}
-	}
 	// Finalize the block, applying any consensus engine specific extras (e.g. block rewards)
-	p.engine.Finalize(p.bc, header, statedb, block.Transactions(), block.Uncles())
-
+	statedb.Prepare(common.ACHash(block.Number()), block.Hash(), len(block.Transactions()))
```
aneequesafdar, line 101: Similar to the above could you please explain this `statedb.Prepare`

core/types/bft_test.go
```diff
@@ -44,7 +44,7 @@ func TestHeaderHash(t *testing.T) {
 	PosHeader.MixDigest = BFTDigest
 
 	originalHeaderHash := common.HexToHash("0x44381ab449d77774874aca34634cb53bc21bd22aef2d3d4cf40e51176cb585ec")
-	posHeaderHash := common.HexToHash("0xfe5aad27871b04ad6c0815c39a371b369e3537f7f117342181eefbf81ba7a686")
+	posHeaderHash := common.HexToHash("0x9acfd617b26d948236857039f3e63d48f23cfc6c8693c7a564ce1e67df0f587c")
```
aneequesafdar, line 47: how did you calculate the new header hash?

miner/worker.go
```diff
@@ -995,7 +995,7 @@ func (w *worker) commit(uncles []*types.Header, interval func(), update bool, st
 		*receipts[i] = *l
 	}
 	s := w.current.state.Copy()
-	block, err := w.engine.FinalizeAndAssemble(w.chain, w.current.header, s, w.current.txs, uncles, w.current.receipts)
+	block, err := w.engine.FinalizeAndAssemble(w.chain, w.current.header, s, w.current.txs, uncles, &receipts)
```
aneequesafdar, line 998: can you please explain why we are changing this?

contracts/autonity/calls.go
Jason-Zhangxin-Chen: @yazzaoui, It shouldn't be removed since it measures those reward distribution metris per block, and it also cleanup those metrics out of the height window to limite the memory consumed by this measurement of per block.

consensus/consensus.go
```diff
@@ -90,15 +90,15 @@ type Engine interface {
...
+		uncles []*types.Header, receipts []*types.Receipt) (types.Committee, *types.Receipt, error)
 
 	// FinalizeAndAssemble runs any post-transaction state modifications (e.g. block
 	// rewards) and assembles the final block.
 	//
 	// Note: The block header and state database might be updated to reflect any
 	// consensus rules that happen at finalization (e.g. block rewards).
 	FinalizeAndAssemble(chain ChainReader, header *types.Header, state *state.StateDB, txs []*types.Transaction,
-		uncles []*types.Header, receipts []*types.Receipt) (*types.Block, error)
+		uncles []*types.Header, receipts *[]*types.Receipt) (*types.Block, error)
```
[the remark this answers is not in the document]
yazzaoui: >slices are passed around by references anyway

the underlying array yes but not the length and capacity information, see https://golang.org/src/runtime/slice.go, so appending in `FinalizeAndAssemble` wouldn't be reflected at the call site.

common/types.go
```diff
@@ -64,6 +64,13 @@ func BytesToHash(b []byte) Hash {
...
+	hash := BigToHash(b)
```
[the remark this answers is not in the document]
yazzaoui: `BigToHash` is not hashing but converting the block number into it's big-endian representation. We would have `0xAC000....0000003` for the tx hash for the 3rd block finalize call which is unique and could be easily generated by some 3rd party tool.

consensus/tendermint/backend/backend.go
```diff
@@ -327,64 +332,37 @@ func (sb *Backend) VerifyProposal(proposal types.Block) (time.Duration, error) {
...
-				return 0, err
-			}
-		} else if proposalNumber > 1 {
-			err = sb.blockchain.GetAutonityContract().ApplyFinalize(block.Transactions(), receipts, block.Header(), state)
-			if err != nil {
-				return 0, err
-			}
-		}
-
+		state.Prepare(common.ACHash(block.Number()), block.Hash(), len(block.Transactions()))
```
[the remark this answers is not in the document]
yazzaoui: These informations are needed for the `LOG` EVM opcode.

consensus/tendermint/backend/backend.go
yazzaoui: Here we verify if the block header committee field was correctly generated. Important to notice that this committee field is representing the next height consensus committee and this has to be retrieved by calling the finalize Autonity Contract function. I refactored `sb.finalize` to return the proper data.
This field is being set in the finalizeAndAssemble call. 

Actually it's an important and difficult part of the code, prone to error, I really encourage you to spend time studying your 
 questions.. error could have been made and not be captured by our test suite..

core/types/block.go
yazzaoui: The abi encoder/decoder used to pass and retrieve data to the autonity contract need to know how to match the golang structure with the solidity ones. By default it perform the mapping with the same names but if you're using a different  then you need to specify it.
(`address` is a reserved keyword in solidity.)

core/chain_makers.go
yazzaoui: Important was to avoid using ` types.BFTDigest` as this is how we distinguish a PoS block header over an original one. The hashing function is dependent on this digest and so are the upstrea, tests with hardcoded hashes.

core/state_processor.go
yazzaoui: same answer as above

miner/worker.go
yazzaoui: The finalize call receipt needs to go in the `receipts` slice as `w.current.receipts` is a cache that could be being reused to add more transaction receipts.

consensus/tendermint/committee/default.go
```diff
@@ -125,8 +131,8 @@ func (set *defaultSet) Copy() Set {
 	return newSet
 }
 
-func (set *defaultSet) F() int { return int(math.Ceil(float64(set.Size())/3)) - 1 }
+func (set *defaultSet) F() uint64 { return uint64(math.Ceil(float64(set.totalPower)/3)) - 1 }
```
Jason-Zhangxin-Chen, line 134: If the totalPower equals 0, it would stucked.
Shall we consider if totalPower is 0, then we fall back to former round robin, and make the F() and Quelun() by node numbers.

consensus/consensus.go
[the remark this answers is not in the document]
aneequesafdar: I am not sure what you mean by

 > but not the length and capacity information

Please see [this go playground](https://play.golang.org/p/kdUIMFlwArx) where the len and cap is preserved while passing slices. Although if you were to crop the slice from the head (`slice[2:]`), then the pointer will move to the 2nd element of the underlying array and the capacity will change from to `cap = cap -  2 `

In case of append, you are correct that the append will create a new underlying array for a slice if `len(slice) == cap(slice)`, however, before passing the receipts slice you can just check `cap(receipts) - len(receipts) == 0`, then we know for sure a new underlying slice will be created, otherwise, since we are only adding one more element, we don't really need to pass around pointers of pointers, go lang was meant to make it pointer stuff implicit, unlike C and C++.

common/types.go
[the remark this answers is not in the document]
aneequesafdar: I see thanks.

PR #489, opened 2020-03-27
develop branch CI workflow

JekaMas:
closes https://github.com/clearmatics/autonity/issues/488


.github/workflows/develop.yml
```diff
@@ -0,0 +1,48 @@
+name: develop
+on: [push]
+jobs:
+  bootstrap-checkout:
```
piersy: We could remove this step, since we are only using the cache one time

JekaMas: Not really, it's different workflow.

piersy: Apologies maybe I wasn't clear.

What I mean't was we could just perform the checkout in the `long-run_tests` job and then there would be no need to have the `bootstrap-checkout` job.

JekaMas: I tried to use cache from different workflow and it doesn't work unfortunately.

piersy: Aplogies, that is not what I'm suggesting.

What I'm saying is that caching in general makes sense when you access the cache more than once. In this workflow we only have one cache access by `long-run_tests` so we don't derive any benefit from having a cache.

Roughly speaking these are the steps we are performing with the cache.

* checkout the code (`bootstrap-checkout`)
* send it to the cache (`bootstrap-checkout`)
* get the code back from the cache  (`long-run_tests`)

But we could just avoid using the cache entirely and do the following
* checkout the code (`long-run_tests`)

Then we don't have any need for a cache and hence no need for a `bootstrap-checkout` job.

piersy: @JekaMas Was my explanation clear?

JekaMas: Noticed and fixed

Issue #490, opened 2020-03-30
Add initial documentation


piersy:
# Problem

Following on from #461 this is a ticket that covers adding initial documentation for autonity. The plan was to have myself and @maxcroasdale develop documentation for a few of the required documentation headings, so that we have enough documentation to set the style/tone/approach such that subsequent contributors have enough examples of existing documentation that they can easily match the existing style/tone/approach.

As such we will add documentation for a subset of the agreed upon documentation headings for the next release.
```
How To -> Run a node (covers build, install, config, running)
How To -> Run a network (covers configuring operator, participants, validators)
How To -> Access JSON-RPC API
```

# Acceptance Criteria

Documentation for the aforementioned headings should be available in the `docs` folder.

JekaMas:
Shall it be in this repo or in a separate one?

piersy:
@JekaMas The docs are going to be held within this repo, under the docs dir.

JekaMas:
I'm thinking about the license. Wouldn't it be more suitable for us to create a new repo with something non-gpl and keep docs there?

@yazzaoui

yazzaoui:
I think we can always specify in the README which part of the repo is under our own copyright?

JekaMas:
We can't. There's the license for the whole repo.

piersy:
@JekaMas Could you explain why you think the licence could pose a problem with the docs?

JekaMas:
We'll not be able to move docs to a private repo. For me, we should use private repos everywhere where it's possible just to reduce our future work with migration to a new license.

piersy:
@JekaMas @yazzaoui Can we just license the docs folder differently, we could put an MIT licence in it?

JekaMas:
As far as I know, it doesn't work that way. You cant put different license into the folder. Moreover, the docs folder doesn't have there own license by the time and we can't just introduce a new one different from gnuv3. We even can't use anything that is in docs in a private directory.

JekaMas:
I'd suggest to be on a safe side and create a new repo. We don't have enough experience in licensing so it's easy to do something wrong.

saronimo:
I could see this being an issue if we can't use any of the content in a private file as we wouldn't be able to use to core how-to docs. That said, the list of topics outline by @piersy in #461 should be the basis for the oss docs anyway.

@piersy what are the implications of having the docs in a separate repo?

piersy:
@saronimo It makes it much harder to keep the code and documentation in sync.

[...]

PR #491, opened 2020-03-30
Add documentation overview of Autonity

piersy:
Adds an overview of the autonity project, this will be the landing page for the documentation.

A part of #490


JekaMas:
https://github.com/clearmatics/autonity/issues/490#issuecomment-606099457 we should decide something about the license

Issue #487, opened 2020-03-26
Validate enodes on Autonity contract update


JekaMas:
https://github.com/clearmatics/autonity/pull/484 showed us that we need to validate somehow if a user's data is a correct enode or not.

We can:
1. Use something like regex validator https://github.com/gnidan/solregex
2. Add a precompiled contract https://medium.com/@rbkhmrcr/precompiles-solidity-e5d29bd428c4 and validate enode on Autonity side.

For me, the second is better because we could not only validate format but also validate if the address, pubkey and hash are correct in terms of cryptography which could be difficult of not impossible at all in Solidity code.

piersy:
@JekaMas This ticket doesn't let me know what the problem is. There is the reference to #484 but there is no mention in that or the associated ticket about validating enodes.

Could you update the ticket explaining why we need to validate an enode and what is meant by validating an enode?

yazzaoui:
I guess it's about verifying if the supplied enode has a correct format, see : https://github.com/ethereum/wiki/wiki/enode-url-format .

I think the priority is to make sure layer 1 is resilient enough and can support bad enodes, treat them as error cases, and then the verification wouldn't be needed but nice to have.

JekaMas:
I think I can add verification in 3-4 days. It's up to you to decide.

piersy:
@yazzaoui @JekaMas 
I'm still not clear on why we would need to validate an enode. #484 was about fixing consensus when an invalid enode is added, and it has been merged. So I'm wondering what other problems now require us to validate enodes?

JekaMas:
It always better to check and validate then deal with incorrect data. That's why things like debit card number are validated on each application level: frontend, backend, db.

piersy:
> It always better to check and validate then deal with incorrect data.

Ahh, so this would be another way of solving #451 right?

JekaMas:
I think so. Also this solution will provide to the client a meaningful error and cancel the transaction that will return to the user some gas also.

piersy:
@JekaMas So will we revert the fix from #484 when this ticket is done, so that we have just one place in the code that will deal with enforcing enode correctness?

JekaMas:
If it completely covers all the cases from #484, can be implemented, uses reasonable amount of gas than I think so.

yazzaoui:
[...]

PR #492, opened 2020-03-30
Precompiled contract for enode checking

JekaMas:
shown the idea of #487


yazzaoui:
Yes looking good with a separate library contract with the assembly call.

contracts/autonity/contract/contracts/Precompiled.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.5.0;
+
+library Precompiled {
+    function enodeCheck(string enode) internal pure returns (uint[2] memory p) {
+        assembly {
+            if iszero(staticcall(gas, 0xff, input, 0x80, p, 0x40)) {
```
piersy: It would be nice to have a comment here explaining exactly what this line is doing

JekaMas: I can provide a link to the doc. Shall we double the documentation in the comments? May not really.

piersy: @JekaMas  A link would be good.

 I'm not sure what you mean about doubling the documentation in the comments?

JekaMas: A bit strange to provide links to solidity docs in solidity code. Shall we do it?

piersy: @JekaMas I am confused, you suggested the link. 

I think it is just hard to parse when reading, potentially a comment could make that much simpler.

I don't understand how we came up with the contract address of `0xff` or why it returns `uint[2]` or why we are limiting the input to a length of 128 when enode strings are longer. Maybe this is just really obvious to everyone else, but not to me.

JekaMas: Unfortunately, it's all about solidity and evm they are complex and complicated. And it's the only way to call precomputed contract (you can read about it with examples in the link above).

piersy: @JekaMas Which link?

JekaMas:
It's mentioned in the issue https://medium.com/@rbkhmrcr/precompiles-solidity-e5d29bd428c4

piersy:
@JekaMas Thanks, I will take a look when I get a chance.

I think it could be nice to include that link in a comment, just so that for people unfamiliar with precompiled contracts in solidity they have the guide that, I guess you followed, right there.

JekaMas:
If we include any links to the comments to the code then we have to support(check that they are still available) them somehow. It's not obvious to me how can we do that.

piersy:
@JekaMas I agree it would be nice if we could have a way of checking the links validity automatically.

But even if the link went dead it can be useful while it is still alive, eventually someone will notice the link is not working  and we can update it.

Once the link goes dead it's no worse than the current state of not having any link at all.

JekaMas:
Can't agree. Shall we provide links on how golang works or evm?

contracts/autonity/contract/contracts/Precompiled.sol
JekaMas: The link is added to the code

.github/workflows/go.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         with:
           path: /home/runner/work/autonity/autonity/
           key: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
-          restore-keys: ${{ runner.os }}-aut-${{ env.cache-name }}-
+          restore-keys: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
```
piersy, line 15: I don't think, having a restore key that is the same as the key provides any value. Restore keys are meant as a backup in case the cache key cannot be found

.github/workflows/develop.yml
```diff
@@ -1,5 +1,8 @@
 name: develop
-on: [push]
+on:
+  release:
```
piersy, line 3: Just wondering why you changed this. It seems unrelated to this PR, but also won't this stop us getting feedback on whether TestTendermintTC7 untill it is too late (I.E the release has already been made)?

.github/workflows/go.yml
JekaMas: Interesting, I've just implemented your idea about it. Why does it now not so good?

.github/workflows/develop.yml
JekaMas: I think this is the cause of flaky tests. After updating branch with recent develop, CI runs both workflows, for usual PR and for release. This took > 40m. However, using this flag everything works as expected.
So we don't have flaky tests, but we had too long CI by mistake @yazzaoui

.github/workflows/go.yml
piersy: Hey, I'm not sure what you mean by

>Why does it now not so good?

Issue #476, opened 2020-03-20
Update, Delegate, and Reclaim Functions


differentialderek:
# Problem
In relation to issue #458, implement the `updateUserSelfStake()`, `delegateStake()`, and `reclaimStake()` functions from the spec in issue #460.

# Implementation
These functions implemented, with tests, that match the spec.

# Acceptance Criteria
Functions must match the spec.

# Notes

# Future Steps

Implement `bon()` and `unbond()` functions from the spec.

Issue #460, opened 2020-03-09
Spec for Delegation and Bonding


differentialderek:
# Problem

In relation to Issue #458, we need to first make a spec for stake delegation delegation and bonding.

# Implementation

Define the spec for the functions `delegateStake()` and `reclaimStake()`, including the process for integration with things such as (un)bonding.

# Acceptance Criteria

Must include:
* `delegateStake()` function
* `reclaimStake()` function
* Checks relating to:
    * (un)bonding
    * committee/proposer selection mechanisms
    * `changeUserType()`

# Notes

None

# Future Steps

Continue with the next step of issue #458 (implementation).

differentialderek:
# Spec for Delegation and Bonding
## New Functions
This PR will include five basic functions.
* `updateUserSelfStake()`
* `delegateStake()`
* `reclaimStake()`
* `bond()`
* `unbond()`

Each of these functions can only be executed (1) by the authority account, and (2) at the end of a bonding period, given in the public variable `bondingPeriod`.

### `updateUserSelfStake(_user : User, _selfStakeDifferential : uint256)`
* Takes a stakeholder or validator who wants to sell or buy stake. A negative
value of `_selfStakeDifferential` indicates the user wants to sell stake equal to
the absolute value of `_selfStakeDifferential`. A positive `_selfStakeDifferential`
indicates that the user wants to buy stake equal to the value of
`_selfStakeDifferential`.
* Checks `_user.userType == Stakeholder | Validator`.
* If `_selfStakeDifferential < 0`:
    * Checks `(-1) * _selfStakeDifferential < _user.stake`.
    * Updates `_user.selfStake = _user.selfStake + _selfStakeDifferential` to
    reflect the sell.
    * Transfers funds to `_user` from the authority account.
* If `_selfStakeDifferential >= 0`:
    * Checks `_selfStakeDifferential < userFunds`, where `userFunds` is the amount
    of tokens owned by `_user`.
    * Executes a transfer from `_user` to the authority account equal to
    `_selfStakeDifferential`.
    * Updates `_user.selfStake = _user.selfStake + _selfStakeDifferential`
    reflecting the purchase.

### `delegateStake(_user : User, _stakeDelegate : User, _delegatedStake : uint256)`
* Takes a user, the intended stake delegate, and the amount of stake the user
wishes to delegate. The user can only delegate undelegated stake, *i.e.* stake
stored by the variable `_user.selfStake`.
* Checks that both `_user.userType == stakeholder | validator` and
[...]

PR #493, opened 2020-03-31
(Un)bonding and Stake Delegation.

differentialderek:
This PR is for issues #476 and #460.

It introduces the following functions:
- `selfStake()`
- `delegateStake()`
- `reclaimStake()`
- `bond()`
- `unbond()`
- `_isInArray()`

It also introduces two contract states:
- `delegatedStake` of type `mapping (address => mapping (address => uint256))`
- `delegatedList` of type `mapping (address => mapping (bool => address[]))`

These function as follows:
- `delegatedStake[user1][user2]` is equal to the amount of stake that `user1` has delegated to `user2`.
- `delegatedList[user][true]` is a list of users that `user` has delegated stake to.
- `delegatedList[user][false]` is a list of users that have delegated stake to `user`.

Both of these contract states are used to keep track of who has delegated stake to whom, and how much.

### DIFF
--- a/contracts/autonity/contract/contracts/Autonity.sol
+++ b/contracts/autonity/contract/contracts/Autonity.sol
@@ -29,8 +29,6 @@ contract Autonity {
         UserType userType;
         uint256 stake;
         string enode;
-        // uint256 selfStake;
-        // uint256 delegatedStake;
         uint256 commissionRate; // rate must be by default 0 and must remain unchanged if not updated.
     }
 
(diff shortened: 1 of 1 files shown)


contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -185,6 +189,137 @@ contract Autonity {
         emit RemoveUser(_address, u.userType);
     }
 
+    /*
+    * selfStake
+    * returns the user's selfStake, its stake minus that stake delegated to it
+    */
+    function selfStake(address _user) public returns(uint256) {
```
JekaMas, line 196: Could it we `view` access?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -616,6 +752,16 @@ contract Autonity {
         }
     }
 
+    function _isInArray(address _address, address[] storage _array) internal returns(bool) {
```
JekaMas, line 755: Is it `pure` function?

Issue #494, opened 2020-03-31
Fix flaky eth/downloader tests


piersy:
# Problem

The following tests fail randomly.

TestBoundedHeavyForkedSync63Fast
TestBoundedHeavyForkedSync64Fast

We can see one of them failing [here](https://github.com/clearmatics/autonity/pull/491/checks?check_run_id=546047407) for a pull request that only added documentation.

I can see them both failing multiple times if i run
```
 go test -run "TestBoundedHeavyForkedSync63Fast|TestBoundedHeavyForkedSync64Fast" --count 100
```
From the `eth/downloader` dir.

# Acceptance Criteria

We should be able to run

```
go test -run "TestBoundedHeavyForkedSync63Fast|TestBoundedHeavyForkedSync64Fast" --count 500
```

Without any failures if those tests are passing on geth upstream. If not raise an issue on the go-ethereum repo.

yazzaoui:
TestBoundedHeavyForkedSync63Fast and TestBoundedHeavyForkedSync64Fast are depreciated, closing.

PR #495, opened 2020-03-31
Skip flaky tests

piersy:
This pull request simply skips these tests, until we have time to investigate why they are failing #494 #496 

#491 is currently failing due to the flaky tests and so needs this to be merged so that it can be merged.

### DIFF
--- a/consensus/test/topology_test.go
+++ b/consensus/test/topology_test.go
@@ -43,6 +43,7 @@ func TestTendermintStarSuccess(t *testing.T) {
 }
 
 func TestTendermintStarOverParticipantSuccess(t *testing.T) {
+	t.Skip("Skipping flaky test https://github.com/clearmatics/autonity/issues/496")
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}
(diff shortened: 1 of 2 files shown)


JekaMas:
Are we going to merge upstream? It could be that the test is fixed in upstream already.

piersy:
@JekaMas Good point! I just checked geth master and I'm still seeing the same failures.

JekaMas:
Maybe it's better to fix the test?

piersy:
@JekaMas I'm not sure. I guess we wan't to avoid blocking lots of prs, but whether that will happen depends on how long it takes to fix the flaky tests and how often they flake out, which Is not something we can predict.

I can't imagine myself fixing these tests quickly, since I'm not familiar with the code at all.

Does anybody else have any thoughts on how to handle this? @clearmatics/autonity

JekaMas:
Maybe it's a good issue to start? I'd prefer having the tests fixed rather than skipped.

piersy:
@JekaMas How do we deal with PRs that are blocked by flaky tests?

JekaMas:
I don't think that someone is blocked. Each week we merge finished PRs.
It's annoying to restart jobs, I agree. However, it's better to fix tests than skip them.

piersy:
@JekaMas I didn't realise we could restart jobs i looked [here](https://github.community/t5/GitHub-Actions/re-run-jobs/td-p/30476) and thought it was not possible.

Could you let me know how I can restart a job?

JekaMas:
For example https://github.com/clearmatics/autonity/runs/551253133
on the right side, you can find "re-run jobs" button

piersy:
@JekaMas Thanks! I will re-run and close this.

Issue #434, opened 2020-02-21
Set up automated performance report generator


yazzaoui:
# Problem

We need to start thinking about how to measure the performance of our system as we cannot allow performance regressions to go unnoticed or unresolved during the development cycle.

# Implementation

One idea is to extend python's e2e tests with benchmarking tests that could output a list of key metrics. 

Consider whether we can run this on every release and track the metrics in the release notes.

# Acceptance Criteria

- Define a set of key metrics
- Propose a way on how to track regressions.

piersy:
Some work has been done on documenting the regression testing metrics here #503, the pr has been closed for now.

yazzaoui:
Old issue, it's a good idea but no longer a priority, closing for now.

JekaMas:
It's alive! OMG!

yazzaoui:
Heya @JekaMas good to hear from you, our mainnet is (finally) coming soon :))

PR #503, opened 2020-04-06
Regression testing docs

screwyprof:


docs/performance/metrics.md
```diff
@@ -0,0 +1,42 @@
+# Performance regression testing
+Once per releace we'd like to measure performance metrics on different levels.
+## Metrics to control
+### Network level
```
piersy, line 4: I think the term "Network level" might be misleading here, or I think more description is needed about how the metrics are calculated.

I would consider the consensus_block_interval at the network level to be the time taken for the network to move from block n to block n+1.

It's not clear how that would be calculated though.

It could be time from the first node to move to block n to the last node to move to block n+1. So the entire period where at least one node is working on block n.

Or it could be the average of all the node level consensus_block_interval values.

Or "Network level" doesn't really mean network level, and you are just capturing metrics at the node level.

docs/performance/metrics.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Test cases
+* Python tests
+* Melicious tests:
+    * 1/3 validators offline
+    * melicious validators try to send multiple votes for different or the same blocks: 1 validator, 5%, 10%, 1/3 of validators count
+    * transaction spam attack: 10, 100, 1000 times traffic comparing to normal
+    * slow nodes: 10, 100, 1000 less network bandwidth than normal nodes
+
+## Resources
```
piersy, line 41: Could you elaborate, its not clear to me what exactly is meant by using go-ethereum and tendermint metrics as a resource? Are there pages for each of those metrics that you could link to?

docs/performance/metrics.md
```diff
@@ -0,0 +1,42 @@
+# Performance regression testing
+Once per releace we'd like to measure performance metrics on different levels.
```
piersy, line 2:
```suggestion
Once per release we'd like to measure performance metrics on different levels.
```

There are a couple more typos, could you just run this doc through a spellchecker to catch them.

docs/performance/metrics.md
JekaMas: 'Network' as it used in distributed systems. Distributed network.

Term "consensus" would be not vert exact and strict so "network" was used.

JekaMas:
"Looking at this doc, I don't think it is relevant for end users."
I think "end users" are we, developers who would write tests and etc. I think we were thinking about different tasks. This issue is not about user documentation but about what and how developer should measure to check if performance regression happened.

docs/performance/metrics.md
JekaMas: We just opened tendermint and go-ethereum source codes and started finding out all suitable metrics.
So it's just https://github.com/tendermint/tendermint and https://github.com/ethereum/go-ethereum

docs/performance/metrics.md
piersy: I think maybe the problem is the use of the term level.

In the case of consensus_block_interval what constitutes a datapoint in the histogram?
Is it:
1. The block interval seen by a single node
1. The block interval seen by the network as a whole

docs/performance/metrics.md
piersy: It's not clear how you decided on what metrics to use from those 2 projects, because what is meant by "Basic" is not defined.

If someone else were to get the "Basic" metrics from these 2 projects they might end up with a completely different set of metrics.

I'll add a suggestion.

docs/performance/metrics.md
```diff
@@ -0,0 +1,42 @@
...
+## Test cases
+* Python tests
+* Melicious tests:
+    * 1/3 validators offline
+    * melicious validators try to send multiple votes for different or the same blocks: 1 validator, 5%, 10%, 1/3 of validators count
+    * transaction spam attack: 10, 100, 1000 times traffic comparing to normal
+    * slow nodes: 10, 100, 1000 less network bandwidth than normal nodes
+
+## Resources
+Basic go-ethereum and tendermint metrics was used.
```
piersy, line 42:
```suggestion
We used the go-ethereum and tendermint projects as a source of inspiration for suitable metrics.
```

piersy:
@JekaMas 


>I think "end users" are we, developers who would write tests and etc. I think we were thinking about different tasks. This issue is not about user documentation but about what and how developer should measure to check if performance regression happened.

In the case of this documentation I agree that the project maintainers are the "end users", my point is just that it doesn't belong in the documentation site, since that is targeted to users that are not project maintainers.

docs/performance/metrics.md
piersy: @JekaMas @screwyprof bump

piersy:
Closing until we start work on #434

PR #508, opened 2020-04-08
Proposer election integration

Jason-Zhangxin-Chen:
The PR contains: proposer election weighted on voting power.
Base on Youssef's voting power integration, extend the election algorithm with weighted round robin base on committee members voting power.


consensus/tendermint/backend/backend_test.go
```diff
@@ -681,7 +681,7 @@ func newBlockChain(n int) (*core.BlockChain, *Backend) {
 	if err != nil || validators.Size() == 0 {
 		panic("failed to get committee")
 	}
-	proposerAddr := validators.GetProposer(0).Address
+	proposerAddr := validators.GetProposer(0, big.NewInt(0)).Address
```
JekaMas, line 684: It's an often pattern so we have `common.Big0`

consensus/tendermint/committee/weighted_round_robin_test.go
```diff
@@ -0,0 +1,203 @@
...
+	validator1 = "70524d664ffe731100208a0154e556f9bb679ae6"
+	validator2 = "b37866a925bccd69cfa98d43b510f1d23d78a851"
+	validator3 = "70524d664ffe731100208a0154e556f9bb679ae0"
+	validator4 = "b37866a925bccd69cfa98d43b510f1d23d78a850"
+	validator5 = "70524d664ffe731100208a0154e556f9bb679ae1"
+	validator6 = "b37866a925bccd69cfa98d43b510f1d23d78a853"
+)
+
+func TestWeightedRoundRobinProposer(t *testing.T) {
+	testZeroVotingPower(t)
```
JekaMas, line 24: I think we can run them at the same time as subtests with `t.Run`

consensus/tendermint/committee/weighted_round_robin.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/core/types"
+	"github.com/clearmatics/autonity/crypto"
+	"math/big"
+)
+
+// weightedRoundRobinProposer, it distribute the height+round into a 256bits value space by Keccak256, then schedule
+// the proposer weighted by voting power.
+func weightedRoundRobinProposer(valSet Set, proposer common.Address, round int64, height *big.Int) types.CommitteeMember {
```
yazzaoui, line 12: Sorry If I made it unclear, the goal is to call the Autonity Contract to retrieve the proposer in order to avoid the logic to be put in the client.

Jason-Zhangxin-Chen: It was implemented in contract side, but after I have your voting_power integration, I thought that I can extend the selector with a new one. 
Okay, I can reuse the former implementation which keep it in contract side.

Jason-Zhangxin-Chen:
Close this #508, it is implemetned on Layer2 now: https://github.com/clearmatics/autonity/pull/511

PR #510, opened 2020-04-15
Remove genesis committee field

piersy:
The `Committee` field in the genesis file was a duplication of the information stored in the Users field.

This pull request removes it and instead derives the committee from the users field at the point it is needed.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -271,6 +271,9 @@ func (g *Genesis) ToBlock(db ethdb.Database) (*types.Block, error) {
 
 	var committee types.Committee
 	if g.Config.AutonityContractConfig != nil {
+		if g.Difficulty.Cmp(big.NewInt(1)) != 0 {
+			return nil, fmt.Errorf("autonity requires genesis to have a difficulty of 1, instead got %v", g.Difficulty)
+		}
 		var err error
 		committee, err = extractCommittee(g.Config.AutonityContractConfig.Users)
 		if err != nil {
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -63,9 +63,6 @@ type Genesis struct {
 	GasUsed    uint64      `json:"gasUsed"`
 	ParentHash common.Hash `json:"parentHash"`
 
-	//PoS Fields
-	Committee types.Committee `json:"committee"           gencodec:"required"`
-
 	mu sync.RWMutex
 }
 
@@ -104,7 +101,6 @@ type genesisSpecMarshaling struct {
 	Difficulty *math.HexOrDecimal256
 	Alloc      map[common.UnprefixedAddress]GenesisAccount
 
-	Committee    []common.UnprefixedAddress
 	VotingPowers []math.HexOrDecimal64
 }
 
@@ -191,7 +187,11 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, genesis *Genesis, override
 			genesis = DefaultGenesisBlock()
 		}
 		// Ensure the stored genesis matches with the given one.
-		hash := genesis.ToBlock(nil).Hash()
+		b, err := genesis.ToBlock(nil)
+		if err != nil {
+			return nil, common.Hash{}, err
+		}
+		hash := b.Hash()
 		if hash != stored {
 			return genesis.Config, hash, &GenesisMismatchError{stored, hash}
 		}
@@ -204,7 +204,11 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, genesis *Genesis, override
 
 	// Check whether the genesis block is already written.
 	if genesis != nil {
-		hash := genesis.ToBlock(nil).Hash()
+		b, err := genesis.ToBlock(nil)
+		if err != nil {
+			return nil, common.Hash{}, err
+		}
+		hash := b.Hash()
 		if hash != stored {
 			return genesis.Config, hash, &GenesisMismatchError{stored, hash}
 		}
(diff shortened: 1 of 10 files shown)


core/genesis.go
```diff
@@ -371,21 +384,13 @@ func (g *Genesis) SetBFT() error {
...
 	}
 
 	sort.Sort(committee)
-	g.Committee = committee
-
 	log.Info("starting BFT consensus", "validators", committee)
-
-	// we have to use '1' to have TD == BlockNumber for xBFT consensus
-	g.mu.Lock()
-	g.Difficulty = big.NewInt(1)
```
JekaMas, line 385: Why this line has been removed?

ethclient/ethclient_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/clearmatics/autonity"
+	ethereum "github.com/clearmatics/autonity"
```
JekaMas, line 28: Why do we need alias?

piersy: That is the outcome of running goimports on the file. I guess it does that for readability, because otherwise someone looking at the file might wonder where the ethereum package is being imported from.

core/genesis.go
piersy: I don't think we need to set the difficulty, all the tests seem to be passing without this line, so I guess it was not required.

ethclient/ethclient_test.go
JekaMas: Let's remove it if we don't have packages imported with the same name.

core/genesis.go
JekaMas: It was a kind of protection. Without this a user can run our consensus with different Difficulty and the result is unpredictable. If all tests are green this means that we need a new test.

aneequesafdar: I don't think we should be removing it, we need it to determine the height the peer is at for network permissioning. Also, I am not really sure about the repercussion of this, I know that this difficulty filed is used heavily for blockchain reorgs, so I wouldn't mess with it for the time being.

ethclient/ethclient_test.go
aneequesafdar: I agree with @JekaMas

core/genesis.go
piersy: Can you suggest what the new testcase should be, and I will add it.

ethclient/ethclient_test.go
piersy: Sure

core/genesis.go
piersy: @aneequesafdar 

> I don't think we should be removing it, we need it to determine the height the peer is at for network permissioning

Could you explain how the height is related to the difficulty?


>this difficulty filed is used heavily for blockchain reorgs

I'm not familiar with the terminology (blockchian reorg) can you explain how the difficulty is used for a reorg?

piersy: @aneequesafdar bumping

aneequesafdar: In ethash total difficulty (sum of difficulty from gensis to now) is used to determine the longest chain using the for fork resolution, therefore, the difficulty (is meant to be a measure of how much has been performed as far as I can remeber) in ethereum can change, which has an impact on how the longest chain is chosen.

In our case, we keep it constant (1), and if each block has the same difficulty then the total difficulty is equal to block height (+/-1, I can't exactly remember). Since we still use ethereum blockchain by keeping difficulty constant we ensure that no fork resolution (blockchain reorganisation (reorgs)) will need to take place.

piersy: Thanks for the explanation @aneequesafdar.

I checked out how [ethereum calculates](https://ethereum.stackexchange.com/a/13750/58673) total difficulty and it is just a sum of the block difficulties.

I would imagine that we would need to remove the possibility that a reorg could happen. It doesn't make sense to me that it could be allowed due to the fact that Autonity provides deterministic finality. Otherwise it seems to me that we run the risk of having clients switch to a different chain with a lower or equal difficulty?

With regards to this specific change (removing the setting of difficulty in the genesis), I don't see that it is a problem since we hardcode the difficulty of block headers to 1 in the tendermint backend.

https://github.com/clearmatics/autonity/blob/17d0e50295eee2aee2c5609b185722f0e259b002/consensus/tendermint/backend/engine.go#L310

yazzaoui: we need to keep the genesis block initial difficulty equal to 1 as we rely on the chain height being equal to the total difficulty + 1  for the network permissionning logic : https://github.com/clearmatics/autonity/blob/17d0e50295eee2aee2c5609b185722f0e259b002/eth/handler.go#L438

piersy: @yazzaoui I had a look in the yellow paper, I couldn't really see anything relating to the total difficulty.

I'm wondering though, what is to stop a peer faking a high td?

eth/helper_test.go
```diff
@@ -64,7 +64,11 @@ func newTestProtocolManager(mode downloader.SyncMode, blocks int, generator func
 		}
 	)
 	gspec.Config.AutonityContractConfig = &params.AutonityContractGenesis{
-		Users: []params.User{},
+		Users: []params.User{{
```
aneequesafdar, line 67: why do we need to add a dummy user?

piersy: I changed the logic slightly,

If you look here - https://github.com/clearmatics/autonity/pull/510/files/e692cb95c95e8ced0c596b22b85fb071c9924da2#diff-3e8301305b1dc6d3f96fbe3429cda929L325

The old logic was to assume that if `g.Config.Tendermint` was set then `g.Config.AutonityContractConfig` would be set and would be inspected to find the committee. I changed that to instead check that `g.Config.AutonityContractConfig` is set before trying to find the committee.

As a consequence the test `TestStatusMsgErrors63` began to fail because it was not setting `g.Config.Tendermint` but was setting `g.Config.AutonityContractConfig` and the assumption is that if you set `g.Config.AutonityContractConfig` you need to specify at least one validator, but it wasn't, so I added a "dummy validator".

Looking at this more closely, it seems to me that Tendermint and AutonityContractConfig should really be merged, since they are 2 parts of the same thing.

core/genesis.go
yazzaoui: @piersy Sounds good to me!

eth/helper_test.go
piersy: I made a ticket to consider merging #546

piersy:
@yazzaoui This is a situation where I would like to merge, there are 2 quite distinct commits, squashing and merging would lose that distinction and re-basing would cause us to lose any reference to this PR from these commits.

PR #511, opened 2020-04-15
Layer2 proposer election weighted by voting power.

Jason-Zhangxin-Chen:
Implement the proposer election weighted by voting power at autonity contract side (layer2).
1. Autonity contract side, getProposer(height, round) function was added.
2. Add contract unit testcases to cover getProposer() function.
3. Layer1 tendermint engine side, adapt the getProposer(height, round) of autonity contract.
python e2e testing on this integration will be run too.


contracts/autonity/calls.go
```diff
@@ -191,6 +191,18 @@ func (ac *Contract) callGetMinimumGasPrice(state *state.StateDB, header *types.H
 	return minGasPrice.Uint64(), nil
 }
 
+func (ac *Contract) callGetProposerFromL2(state *state.StateDB, header *types.Header, height uint64, round int64) common.Address {
+	var proposer common.Address
+	_height := new(big.Int).SetUint64(height)
```
JekaMas, line 196: Solidity style? :)

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -440,6 +440,48 @@ contract Autonity {
...
+    */
+    function getProposer(uint256 height, uint256 round) public view returns(address) {
+        // calculate total voting power from current committee.
+        uint256 total_voting_power = 0;
+        for (uint256 i = 0; i < committee.length; i++) {
+            total_voting_power += committee[i].votingPower;
+        }
+
+        // if no committee set, return operator address.
+        if (committee.length == 0) {
```
JekaMas, line 456: Could we do it at the beginning of the function?

piersy:
@Jason-Zhangxin-Chen I wouldn't consider this layer 2, I would consider this is layer 1.

If you look at the definition of [layer 2 here](https://www.binance.vision/glossary/layer-2) its clear that layer2 is about solutions that operate off the main chain (basically state channels and side chains) by contrast this PR is about selecting the proposer for the main chain consensus protocol and is clearly layer1, since without a proposer the main chain protocol cannot make progress.

consensus/tendermint/core/core_backend.go
```diff
@@ -139,6 +139,9 @@ type Backend interface {
 	// GetProposer returns the proposer of the given block height
 	GetProposer(number uint64) common.Address
 
+	// GetProposerFromL2 returns the proposer of the given block height and round from autontiy contract.
+	GetProposerFromL2(height uint64, round int64) common.Address
```
piersy, line 143: I'm guessing that `L2` is shorthand for layer 2. Further to my previous comment about this not looking like a layer 2 solution, we shouldn't be naming this layer 2.

On another note, why isn't this just an update to `GetProposer` why is this a new function?

Jason-Zhangxin-Chen: > I'm guessing that `L2` is shorthand for layer 2. Further to my previous comment about this not looking like a layer 2 solution, we shouldn't be naming this layer 2.
> Yes, L2 is shorthand for layer 2. Since Layer 2 haven't had very clear definition or unified understanding, I change it into GetProposerFromAC where AC is Autonity Contract.
> On another note, why isn't this just an update to `GetProposer` why is this a new function?
We extend the proposer election with extra option: stake weighted round robin which is implemented in autonity contract side, the height of the blockchain should be involved than the one of pure round-robin, since the voting round in the consensus engine usually have the range: [0, 3] to complete the commit, and this range would always pickup limited set of node to be proposer, when height of chain introduced, it could distribute the sead in a better distributed manner as the fairness of stake weighted election.

consensus/tendermint/committee/committee_interface.go
```diff
@@ -34,6 +34,8 @@ type Set interface {
 	GetProposer(round int64) types.CommitteeMember
 	// Check whether the validator with given address is the round proposer
 	IsProposer(round int64, address common.Address) bool
+	// Check if it's using PoS proposer election algorithm.
+	IsPoS() bool
```
piersy, line 38: I think exposing this is a mistake.

Rather than adding an `IsPos` function, since this is an interface we can actually provide a separate implementation that  behaves correctly when GetProposer is called in a PoS context.

Have you heard of the 'Tell don't Ask' principle?

It is summarised well by [Martin Fowler](https://martinfowler.com/bliki/TellDontAsk.html)
>Tell-Don't-Ask is a principle that helps people remember that object-orientation is about bundling data with the functions that operate on that data. It reminds us that rather than asking an object for data and acting on that data, we should instead tell an object what to do. This encourages to move behavior into an object to go with the data.

Adding functions such as `IsXXX` is often a violation of the 'Tell don't ask' principle, and can lead duplication and  leakage of implementation details. The addition of `IsPos` means that we are no longer telling the `committee.Set` to provide a proposer, instead users of the `committee.Set` have to interrogate the internal state of `committee.Set` to understand how they should use it. This leads to duplication of the introspection logic.

We can see the introspection logic below, repeated 3 times in this PR, we should be able to encapsulate this condition inside a  `committee.Set` implementation and remove any duplication. But more important than removing the duplication here is that it removes the burden from the users of `committee.Set`, they no longer have to worry or even know if the set is PoS or not, they also do not have to have any special understanding of what can and can't be done in different blocks.
```
	if !c.CommitteeSet().IsPoS() || height.Uint64() == 1 {
```

I have a pretty clear Idea of how this could be done, so I was thinking I could make a PR on top of this one to demonstrate what I'm thinking.

What do you think?

Jason-Zhangxin-Chen:
Close this PR by using Piers' branch: encapsulate_pos_wrr_in_committee_set

PR #512, opened 2020-04-15
Tdm refactor2

aneequesafdar:
creating this pull request to see where #472 started to break


consensus/tendermint/backend/api.go
```diff
@@ -19,15 +19,15 @@ package backend
 import (
 	"github.com/clearmatics/autonity/common"
 	"github.com/clearmatics/autonity/consensus"
-	"github.com/clearmatics/autonity/consensus/tendermint/core"
+	tendermintCore "github.com/clearmatics/autonity/consensus/tendermint/core"
```
JekaMas, line 22: Why do you need the alias here?

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,24 +458,286 @@ func TestClose(t *testing.T) {
...
+
+		var sawNil bool
+
+		// Want nil once and ErrStoppedEngine 9 times
+		for e := range errC {
+			if e != ErrStoppedEngine {
+				if e == nil {
+					if sawNil {
+						t.Fatalf("<nil> returned more than once, b.Close() should have only returned nil the first time it was closed")
+					} else {
```
JekaMas, line 547: Could we get rid of this if-else and keep simple if statement?

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,24 +458,286 @@ func TestClose(t *testing.T) {
...
+		close(errC)
+
+		assertNotCoreStarted(t, b)
+
+		var sawNil bool
+
+		// Want nil once and ErrStoppedEngine 9 times
+		for e := range errC {
+			if e != ErrStoppedEngine {
+				if e == nil {
```
JekaMas, line 544: if e != SomeError && e == nil
I think it's just if e == nil

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,24 +458,286 @@ func TestClose(t *testing.T) {
...
+	tendermintC := tendermintCore.NewMockTendermint(ctrl)
+	tendermintC.EXPECT().Start(ctx).Return(nil).MaxTimes(5)
+	tendermintC.EXPECT().Stop().MaxTimes(5)
+
+	b := &Backend{
+		core:        tendermintC,
+		coreStarted: false,
+	}
+
+	err := b.Start(ctx, &core.BlockChain{}, func() *types.Block { return &types.Block{} }, func(hash common.Hash) bool { return false })
```
JekaMas, line 688: Could you use for loop here?

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+	"github.com/clearmatics/autonity/crypto"
+)
+
+var maxRange = 100
+
+func TestNewSet(t *testing.T) {
+	var committeeSetSizes = []int64{1, 2, 10, 100}
+	var assertSet = func(t *testing.T, n int64) {
+		t.Helper()
```
JekaMas, line 39: I'm not sure if it should be marked as Helper().

consensus/tendermint/backend/api.go
aneequesafdar: I don't remember, I will change it

consensus/tendermint/backend/engine_test.go
aneequesafdar: Ok I will try to simplify it

consensus/tendermint/backend/engine_test.go
aneequesafdar: same as above

consensus/tendermint/committee/committee_set_test.go
aneequesafdar: Can you please explain why? I thought we should since its a helper function

consensus/tendermint/backend/api.go
aneequesafdar: because `"github.com/clearmatics/autonity/core"` is also being imported at line33

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,390 @@
+// Copyright 2017 The go-ethereum Authors
```
piersy, line 1: I'm seeing this copyright in a few files, this looks like one of our files though, I cant find any reference to committee in go-ethereum master, except in `signer/fourbyte/` which seems to be some sort of database.

aneequesafdar: Please use this pull request to compare https://github.com/ethereum/go-ethereum/pull/14674. I don't really know how licenses work, also at what point do you call refactoring your own work and it doesn't need a license anymore

consensus/tendermint/committee/committee_set_test.go
JekaMas: It'll suppress the line of the code where an error happened in assertSet, however, assertSet is quite big and it might be that we'd like to see where the error happened.

consensus/tendermint/committee/committee_set.go
```diff
@@ -0,0 +1,161 @@
...
+	v, ok := set.allProposers[round]
+	if !ok {
+		v = set.getNextProposer(round)
+		set.allProposers[round] = v
+	}
+
+	return v
+}
+
+func (set *Set) IsProposer(round int64, address common.Address) bool {
```
screwyprof, line 118: Data race possible, mutex should be used.

consensus/tendermint/committee/committee_set_test.go
piersy: I would say at the point that there is nothing left of the original.

I mentioned this because these files look brand new. I had a look in the link you provided and I couldn't see any connection between this file and what is in the istanbul branch, so my inclination would be to remove the copywrite notice. You I guess are better positioned to know if this has any remnants of the istanbul code.

From another perspective I think it is a bit misleading to have that notice if none of that code was actually written by anyone from the go-ethereum team.

If you imagine a member of the team that wrote the istanbul branch looking at this, would they consider it their code?

consensus/tendermint/committee/committee_set.go
aneequesafdar: Since GetByAddress() and GetProposer() use mutex, I don't think there is a need to call mutex here as well otherwise we would have lock twice in a row

consensus/tendermint/committee/committee_set_test.go
aneequesafdar: Ok I will remove the copyright notice

consensus/tendermint/committee/committee_set.go
```diff
@@ -0,0 +1,161 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
```
piersy, line 15: So if we removed the copyright notice from `committee_set_test.go` by the same rationale we should remove this one too.

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+		errC := make(chan error, stop)
+
+		for i := 0; i < stop; i++ {
+			wg.Add(1)
+
+			go func(wg *sync.WaitGroup) {
+				defer wg.Done()
+
+				err := b.Close()
+				errC <- err
```
piersy, line 529:
```suggestion
				errC <- b.Close()
```

There's no need to create a variable here.

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+		errC := make(chan error, stop)
+
+		for i := 0; i < stop; i++ {
+			wg.Add(1)
+
+			go func(wg *sync.WaitGroup) {
+				defer wg.Done()
+
+				err := b.Start(ctx, &core.BlockChain{}, func() *types.Block { return &types.Block{} }, func(hash common.Hash) bool { return false })
+				errC <- err
```
piersy, line 630:
```suggestion
				errC <- b.Start(ctx, &core.BlockChain{}, func() *types.Block { return &types.Block{} }, func(hash common.Hash) bool { return false })
```

There's no need to create a variable here.

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+		t.Fatalf("expected %v, got %v", expected, got)
+	}
+}
+
+func assertNilError(t *testing.T, err error) {
+	t.Helper()
+	if err != nil {
+		t.Fatalf("expected <nil>, got %v", err)
+	}
+}
```
piersy, line 694: Are you familiar with either of `github.com/stretchr/testify/(assert|require)` packages?

They provide this sort of functionality and a bunch more.

I think it would be good to begin using them, since they provides a consistent set of asserts that we can use in all our tests, otherwise we end up duplicating this kind of functionality between test packages. I can see that this function is actually duplicated in `committee_set_test.go`

consensus/tendermint/core/handler.go
```diff
@@ -69,22 +49,6 @@ func (c *core) Start(ctx context.Context, chain consensus.ChainReader, currentBl
 
 // Stop implements core.Engine.Stop
 func (c *core) Stop() error {
```
piersy, line 51: This can no longer return an error, so could we remove the returned error from its  method signature?

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+
+		b := &Backend{
+			core:        tendermintC,
+			coreStarted: true,
+			stopped:     make(chan struct{}),
 		}
+
+		err := b.Close()
+		assertNotCoreStarted(t, b)
+		assertNilError(t, err)
```
piersy, line 498: These two statements look the wrong way around. `Close` returning an error indicates that something went wrong, at that point we shouldn't make any assumptions about what the effect the method call might have had. Here we are asserting that core was stopped irrespective of whether the method call succeeded.

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+		}
+
+		var wg sync.WaitGroup
+		stop := 10
+		errC := make(chan error, stop)
+
+		for i := 0; i < stop; i++ {
+			wg.Add(1)
+
+			go func(wg *sync.WaitGroup) {
```
piersy, line 525: Why are we calling stop from separate goroutines? Why not just call stop directly in a loop? Similar to the approach used in "engine is running, stopped twice".

consensus/tendermint/committee/committee_set.go
aneequesafdar: I rename the file from [default.go](https://github.com/ethereum/go-ethereum/pull/14674/files#diff-24ddffd756230bef41afb61f1efee9c8) -> committee_set.go and I don't think I have significantly changed the file to warrant the removal of copyright, whereas for the committee_set_test.go, the file was heavily changed.

consensus/tendermint/backend/engine_test.go
aneequesafdar: Ok I will update

consensus/tendermint/backend/engine_test.go
aneequesafdar: Ok I will update

consensus/tendermint/backend/engine_test.go
aneequesafdar: No... but will have a look and start using it since you recommend it

consensus/tendermint/core/handler.go
aneequesafdar: ok I will update

consensus/tendermint/backend/engine_test.go
aneequesafdar: This is the expected behaviour since the engine is running when we call close, and it manages to close correctly then I am not expecting an error thus `assertNilError` check.

 The lines below call close on the already closed engine, so I expect an error to be returned thus `assertError` check.

consensus/tendermint/backend/engine_test.go
aneequesafdar: Because we want to be sure that in a concurrent environment it behaves correctly, and the test describes the expected behaviour

consensus/tendermint/backend/engine_test.go
piersy: Sorry I wasn't clear. I can see that we are checking expected behaviour. Im just suggesting re-ordering these 2 lines. like so. And performing the same re-ordering on the 2 lines below.
```suggestion
		assertNilError(t, err)
		assertNotCoreStarted(t, b)
```

In this way if `Close` unexpectedly returns an error we will fail the test for that reason, whereas currently the test would fail because core would not be started but it would not report the error which would be the root cause.

consensus/tendermint/backend/engine_test.go
piersy: I will suggest a new name for the test then.

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+		err := b.Close()
+		assertNotCoreStarted(t, b)
+		assertNilError(t, err)
+
+		err = b.Close()
+		assertNotCoreStarted(t, b)
+		assertError(t, ErrStoppedEngine, err)
+	})
+
+	t.Run("engine is running, stopped multiple times", func(t *testing.T) {
```
piersy, line 505:
```suggestion
	t.Run("engine is running, stopped from multiple goroutines", func(t *testing.T) {
```

consensus/tendermint/backend/engine_test.go
piersy: I'll just clarify the difference between the two packages here. Both contain the same named functions

* assert - failing an assert marks the test as failed but continues to execute the test. This is useful, for example if you are checking the response from a service and you want to assert several things on it. Using assert here you get to see all the things that were wrong with the response in one test run.

* require - failing a require marks the test as failed and stops execution immediately. I tend to use this for prerequisites of a test. For example when firing up a server to test against, if I can't run the server, there is no point trying to run the tests.

consensus/tendermint/backend/api.go
```diff
@@ -27,7 +27,7 @@ import (
 // API is a user facing RPC API to dump BFT state
 type API struct {
 	chain      consensus.ChainReader
-	tendermint core.Backend
+	tendermint core.Backend //TODO: This is only for testing purposes, should be changed to *Backend but that would mean to fix the api tests
```
piersy, line 30: I find TODO's tend to get forgotten if not linked to a ticket, what do you think about making a ticket for this work and adding a link to the ticket in the TODO?

consensus/tendermint/backend/engine.go
```diff
@@ -520,27 +521,36 @@ func (sb *Backend) Start(ctx context.Context, chain consensus.ChainReader, curre
...
 
+	if err := sb.core.Start(ctx); err != nil {
+		return err
+	}
 	sb.coreStarted = true
 
 	return nil
 }
 
 // Stop implements consensus.tendermint.Stop
```
piersy, line 532:
```suggestion
```

This comment looks defunct

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,374 @@
...
+			t.Fatalf("committee memebers are not set properly, expected: %v and got: %v", committeeMembers, set.members)
+		}
+
+		if !reflect.DeepEqual(allProposers, set.allProposers) {
+			t.Fatalf("initial round allProposers not set properly, expected: %v and got: %v ", allProposers, set.allProposers)
+		}
+
+		if totalPower != set.totalPower {
+			t.Fatalf("totalPower not calculated properly, expected: %v and got: %v ", totalPower, set.totalPower)
+		}
```
piersy, line 65: It looks like this test is essentially duplicating the functionality of `NewSet` in the test.

This type of testing tends to not be particularly useful because:

* If there were errors in the code under test, they are likely to be duplicated in the test code that recreates the test functionality outside of the test, or the test itself uses the code under test to create the test state.
* Any change to the implementation is very likely to break the test (the test is brittle), if the test is duplicating the functionality under test. Or if the test is using the code under test to generate the test state, then changes to that code that could cause errors go undetected and you are not really testing anything.
* Finally, because we are only checking that the private fields of the committee set have certain values, we haven't actually checked that any methods of CommitteeSet are functioning correctly.

I can see that this test uses a mixture of duplicating logic used in the method under test and also using the code under test to generate the test state.


This snippet duplicates the logic to add one to the `roundRobinOffset` but also makes use of `getMemberIndex`
```go
	// calculate offset for round robin selection of next proposer
	committee.roundRobinOffset = getMemberIndex(committee.members, lastBlockProposer)
	if len(members) > 1 {
		committee.roundRobinOffset++
	}
```



A better approach is to create your object under test and then verify through some other mechanism that it is functioning correctly.

So for example we could create a ComitteeSet and call GetProposer hundreds of times with increasing round numbers and validate that every member of the set is chosen to be proposer the same number of times(possibily within some tolerance)

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
 
 		err := b.Close()
-		if err != ErrStoppedEngine {
-			t.Fatalf("expected %v, got %v", ErrStoppedEngine, err)
-		}
+		assertNotCoreStarted(t, b)
+		assertError(t, ErrStoppedEngine, err)
 	})
 
 	t.Run("engine is running, no errors", func(t *testing.T) {
```
piersy, line 465: I think this test is badly named, it doesn't appear to be checking that anything is running.

consensus/tendermint/backend/engine_test.go
```diff
@@ -453,22 +458,253 @@ func TestClose(t *testing.T) {
...
+				sawNil = true
+			} else if e != ErrStoppedEngine {
+				assertError(t, ErrStoppedEngine, e)
+			}
+		}
+	})
+}
+
+func TestStart(t *testing.T) {
+	t.Run("engine is not running, no errors", func(t *testing.T) {
```
piersy, line 556: Ahh I think this name got confused with "engine is running, no errors"

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,374 @@
...
+		assertNilError(t, err)
+
+		gotCommittee := set.Committee()
+
+		if !reflect.DeepEqual(committeeMembers, gotCommittee) {
+			t.Fatalf("expected committee: %v and got: %v", committeeMembers, gotCommittee)
+		}
+	}
+
+	for _, size := range committeeSetSizes {
```
piersy, line 124: This seems to be a fairly straightforward test, if I understand it correctly, it checks that if we create a committee set with some members as a parameter when we  call `Committee` we get the same members back, in the same order.

I generally try to avoid these types of test because its not testing functionality, or such little functionality that it doesn't really warrant testing, and whether or not these simple things work can be inferred by tests that test more complex functionality.

So for example the test I alluded to earlier where we could check that GetProposer chooses proposers fairly, indirectly would assert that the CommitteeSet did not "lose" any of the members provided at construction time.

I would propose getting rid of this test and any test that is essentially checking that a container works, IE "I gave you a variable, now give it back to me"

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,374 @@
...
+		size := size
+		t.Run(fmt.Sprintf("get committee of %v member/s", size), func(t *testing.T) {
+			assertSetCommittee(t, size)
+		})
+	}
+}
+
+func TestSet_GetByIndex(t *testing.T) {
```
piersy, line 132: GetByIndex is only called by tests, and I guess that is because only tests have the context to know which validator is at which index.

Instead of adding methods to the public api of CommitteeSet we could instead make the validators available from the test code by just returning the list of created validators in each case.

If we did that then we could get rid of the GetByIndex and this test.

Looking through the code I see that we actually create test validators in a number of different places.

```
	vset, keys := newTestValidatorSet(5) // crypto
	vset, keys := newTestValidatorSet(5) // backend
	committeeSet, keysMap := newTestCommitteeSetWithKeys(4) // core 
	committeeSet := newTestCommitteeSet(2) // core
	committeeMembers := createTestCommitteeMembers(t, 4, genRandUint64(4, maxRange)) // committee
```

Four of these functions are executing almost identical code, so we could replace them with a package to be used in testing that could be imported by other test packages (like httptest in the standard library) and make it return the list of created validators, the fifth (`createTestCommitteeMembers`) seems to be doing something a bit more specific but is contained in the committee package so could just be modified in place to return the validators.

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+	t.Run("error on accessing member index not in committee", func(t *testing.T) {
+		_, err := set.GetByIndex(6)
+		assertError(t, consensus.ErrCommitteeMemberNotFound, err)
+	})
+}
+
+func TestSet_GetByAddress(t *testing.T) {
+	committeeMembers := createTestCommitteeMembers(t, 4, genRandUint64(4, maxRange))
+	sort.Sort(committeeMembers)
```
piersy, line 156:
```suggestion
	sort.Sort(committeeMembers)
```

I don't think we need to sort here

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+	})
+
+	t.Run("error on accessing member index not in committee", func(t *testing.T) {
+		_, err := set.GetByIndex(6)
+		assertError(t, consensus.ErrCommitteeMemberNotFound, err)
+	})
+}
+
+func TestSet_GetByAddress(t *testing.T) {
```
piersy, line 154: With the same reasoning as to the GetByIndex function we can change the signature of the GetByAddress function from:
```go
GetByAddress(addr common.Address) (int, types.CommitteeMember, error)
```
to:
```go
GetByAddress(addr common.Address) (types.CommitteeMember, error)
```

Since the only place we use the index is in this test.

Also we shouldn't be checking the index of a node since it is a detail of the internals of CommitteeSet. For this test we should be interested in whether we can retrieve a node that is in the committee set by its address.

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,374 @@
...
+		testCase := testCase
+		t.Run(fmt.Sprintf("validator set size %v and round %v", testCase.size, testCase.round), func(t *testing.T) {
+			committeeMembers := createTestCommitteeMembers(t, testCase.size, genRandUint64(int(testCase.size), maxRange))
+			sort.Sort(committeeMembers)
+			lastBlockProposer := committeeMembers[rand.Intn(int(testCase.size))].Address
+			roundRobinOffset := getMemberIndex(committeeMembers, lastBlockProposer)
+			if len(committeeMembers) > 1 {
+				roundRobinOffset++
+			}
+			expectedProposerIndex := (roundRobinOffset + testCase.round) % testCase.size
```
piersy, line 200: This test is duplicating the logic found in the implementation.

consensus/tendermint/backend/engine_test.go
aneequesafdar: I think the test is appropriately named, we are saying that when the engine is running and we stop it there are no errors

piersy: This test is called "engine is running, no errors", but we are not running any engine, in fact the engine is a mock.

I think what this test is really checking is that "stopping running engine returns no errors", the current title makes no reference to stop or close.

aneequesafdar: On line 474 we use the field `coreStarted` to determine whether the engine is running or not since in the construction of the test we say that `coreStarted = true` ("engine is running") and when we call Close we don't want any errors ("no errors"). If you look at "engine is not running, error returned" there we use `&Backend{}` and by default `coreStarted = false`, so we don't need to explicitly say anything.

Yes, the engine is a mock if it weren't then it would be an integration test which this is not. The same convention is used for every other test for Start and Stop, so I think you should also say that the rest are also badly named. If you feel they are not expressive enough please feel free to add suggestions as before.

piersy: Sorry I hadn't realised that these are all subtests of `TestClose` this makes more sense now

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -189,83 +189,45 @@ func TestSet_GetByAddress(t *testing.T) {
...
-			gotProposer := set.GetProposer(testCase.round)
-			
-			if expectedProposer != gotProposer {
-				t.Fatalf("expected proposer: %v and got: %v", expectedProposer, gotProposer)
+			require.NoError(t, err)
+
+			firstCommitteeMemberAddr := committeeMembers[0].Address
+			for i := 1; i < numOfPasses; i++ {
+				var committeeFromCallingGetProposer types.Committee
+				for j := 0; j < size; j++ {
```
piersy, line 212: I think you want your outer loop to go from 0 to numOfPasses*size, otherwise you are just testing the same round numbers (0..size-1) on each pass.

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -189,83 +189,45 @@ func TestSet_GetByAddress(t *testing.T) {
...
-		{size: 10, round: 0}, {size: 10, round: 1}, {size: 10, round: 2}, {size: 10, round: 8}, {size: 10, round: 7},
-		{size: 10, round: 10},
-	}
-
-	for _, testCase := range testCases {
-		testCase := testCase
-		t.Run(fmt.Sprintf("validator set size %v and round %v", testCase.size, testCase.round), func(t *testing.T) {
-			committeeMembers := createTestCommitteeMembers(t, testCase.size, genRandUint64(int(testCase.size), maxRange))
+	numOfPasses := 10
+	setSizes := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23,
```
piersy, line 194: Whta about looping from 1..100 instead?

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -189,83 +189,45 @@ func TestSet_GetByAddress(t *testing.T) {
...
-		for start, _ := range b{
-			if b[start] == comitteeStart
-			break
-		}
-
-		// Check that the selected proposers were selected in sorted committee set order.
-		assert.Equal(t, committeeMembers, append(firstPass[start:], firstPass[:start]))
-}
-
 func TestSet_GetProposer(t *testing.T) {
```
piersy, line 192: I think this test could do with a comment explaining how it is verifying the properties of round robin, its probably not obvious unless you wrote the code yourself.

consensus/tendermint/committee/committee_set_test.go
aneequesafdar: Yeah I understand what you mean, however, I cannot do  `0 to numOfPasses*size` because for numOfPasses > 1 the committee size will double, triple and so on with duplicates. Please have a look at the latest commit.

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,393 @@
...
+			set, err := NewSet(copyMembers(committeeMembers), lastBlockProposer)
+			require.NoError(t, err)
+
+			firstCommitteeMemberAddr := committeeMembers[0].Address
+			startRound, endRound := 0, size
+			for i := 1; i <= numOfPasses; i++ {
+				if i > 1 {
+					startRound = endRound
+					endRound = size * i
+				}
```
piersy, line 213: I think we could simplify this slightly by doing the following.
```suggestion
			var startRound, endRound int
			for i := 1; i <= numOfPasses; i++ {
				startRound = endRound
				endRound = size * i
```

consensus/tendermint/committee/committee_set_test.go
```diff
@@ -0,0 +1,393 @@
...
+	})
+
+	t.Run("error on accessing member address not in committee", func(t *testing.T) {
+		_, _, err := set.GetByAddress(common.HexToAddress("testaddress"))
+		assertError(t, consensus.ErrCommitteeMemberNotFound, err)
+	})
+}
+
+// TestSet_GetProposer tests the round robin selection of proposers by continuously calling GetProposer and checking
+// the resulting Committee is the same as ordered Committee passed to the set
```
piersy, line 193: I was thinking we could make reference to the properties that are being checked, something like:
```suggestion
// TestSet_GetProposer tests the round robin selection of proposers. It validates that as GetProposer is
// called with consecutive rounds, consecutive proposers are chosen in the sort order defined by
// types.Committee. The consequence of this is that proposers are selected fairly, with N-1 other proposers
// being selected between any two instances of the same proposer being selected twice. It also validates that
// the selection process starts from the committee member that follows lastBlockProposer in a sorted instance
// of types.Committee.
```

The last sentence of this comment I think is something that is currently missing from this test, but should be added.

aneequesafdar: The last sentence is not necessary if you think it is please checkout the branch and make a commit

piersy: Just to clarify my thinking, if we had the following members `ABCDEF` this test could pass in its current form if every round (of round robin) the sequence `ABCDEF` was returned by consecutive calls to `GetProposer` irrespective of what `lastBlockProposer` was provided at construction time.

If this is correct then I think we need that extra check.

PR #515, opened 2020-04-16
Add genesis tool and doc

piersy:
This PR adds a tool to generate the genesis file. It's purpose is to provide an easy way to generate a genesis file, allowing the user to configure the pertinent parameters while automatically setting everything that should not be modified by users. E.G difficulty, or nonce.

I have made an attempt to document all possible, fields of a genesis file, I'm not sure that my descriptions are correct, so please read them carefully :)

The documentation site can be run by running `make serve` from the docs folder, its probably nicer to read the docs in this fashion.


cmd/gengen/new_genesis.go
```diff
@@ -0,0 +1,327 @@
...
+		return nil, fmt.Errorf("failed to construct initial user state: %v", err)
+	}
+
+	parentHash := common.Hash{}
+	_, err = rand.Read(parentHash[:])
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate random parent hash: %v", err)
+	}
+
+	chainID, err := rand.Int(rand.Reader, math.MaxBig256)
```
JekaMas, line 158: The same question?

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	github.com/huin/goupnp v0.0.0-20161224104101-679507af18f3
 	github.com/influxdata/influxdb v1.2.3-0.20180221223340-01288bdb0883
 	github.com/jackpal/go-nat-pmp v1.0.2-0.20160603034137-1fa385a6f458
-	github.com/julienschmidt/httprouter v1.1.1-0.20170430222011-975b5c4c7c21
+	github.com/julienschmidt/httprouter v1.2.0
```
JekaMas, line 40: Is it relevant to the issue?

cmd/gengen/new_genesis.go
```diff
@@ -0,0 +1,327 @@
...
+		users[i] = user
+	}
+
+	operatorAddress, genesisUsers, genesisAlloc, err := generateUserState(users, pubKeys)
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct initial user state: %v", err)
+	}
+
+	parentHash := common.Hash{}
+	_, err = rand.Read(parentHash[:])
```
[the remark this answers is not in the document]
piersy: That is a mistake, we shouldn't be setting it.

cmd/gengen/new_genesis.go
piersy: The reasoning for this is explained here - https://github.com/clearmatics/autonity/pull/515/files#diff-038df58f5433c67046957b02f81f22a6R72

go.mod
piersy: I don't know why exactly, but that was pulled in by dependencies of gengen.

cmd/gengen/new_genesis.go
JekaMas: A bit strange, usually a user defines ID of the network.

piersy: I'm just pasting the doc here so we can discuss it more easily.

```
  // chainId acts as a chain discriminator that is incorporated into
  // transaction signatures to prevent a single transaction being used across
  // multiple chains. It was added after the creation of ethereum classic, so
  // that users could control which chain their transactions should be
  // executed on. The gengen tool sets it to a random 256 bit value to negate
  // the chance that transactions from one private network could be reused on
  // another.
```

piersy: So this is my thinking.

If we let users define this, then I guess it could easily happen that two users pick the same, chainId.

If that situation occurred and there was a user that was a member of both of these chains, with the same account address in each chain when they send a transaction on one chain, any  member of that chain could take that transaction and send it to the other chain and assuming the account nonce was the same in both situations it could be processed.

A malicious user that was also part of both of those chains could simply collect transactions from each chain and wait until the nonce was correct on the other chain to send them on the other chain. In this way they can very easily maliciously manipulate the network.

By using a large random value for chainId we prevent this from happening.

Also I don't see that there is any benefit to being able to choose the chainId, it's just an extra unnecessary burden to the user.

docs/source/genesis_file.rst
```diff
@@ -0,0 +1,305 @@
...
+         "policy": 0
+       },
+
+       // autonityContract defines the configuration for the Autonity contract
+       // that manages who can participate in the network and controls the
+       // minGasPrice. It must be set. The gengen tool will set it based on the
+       // parameters it was provided.
+
+       "autonityContract": {
+         "deployer": "0x0000000000000000000000000000000000000000",
```
piersy, line 143: @clearmatics/autonity Could somebody explain how we use the deployer address. I can see that we call the evm with it to deploy the contract. And then it has control of some core functions.

I'm guessing we wan't the deployer address to be an address that nobody owns the key for, but then in our wiki is see that we say:
> deployer: address deploying the Autonity Contract, use a random one if no special needs

And I'm wondering what special needs?

docs/source/genesis_file.rst
```diff
@@ -0,0 +1,305 @@
...
+         // users specifies the initial set of members of the system. It is left
+         // to the network creator to define the users. The gengen tool will
+         // generate users from the parameters it is provided.
+
+         "users": [
+           {
+
+             // address is the account address for this user, if enode is
+             // specified, then the account address does not need to be specified.
+             // The gengen tool will not set address.
```
piersy, line 195: @clearmatics/autonity  I'm actually not sure this is correct.

In the wiki we say:

>user.enode or user.address must be defined.

But if I don't define address, when I unmarshal a genesis file that does not specify user addresses, users in the unmarshalled struct have a zero valued address in the struct. I guess this would cause problems, could someone confirm?

cmd/gengen/new_genesis.go
```diff
@@ -0,0 +1,320 @@
...
+	// Holds generated or loaded keys for users
+	pubKeys := make([]*ecdsa.PublicKey, len(userStrings))
+
+	_, err := os.Stat(userKeys)
+	if err != nil {
+		// Stores private keys so they can be written to file
+		privKeys := make([]*ecdsa.PrivateKey, len(userStrings))
+		// No file at path so generate keys
+		for i := range pubKeys {
+			k, err := crypto.GenerateKey()
```
piersy, line 55: @clearmatics I don't think this linting rule is helpful.

> shadow: declaration of "err" shadows declaration at line 49 (govet)

The approach to handling  errors that I'm using here is idiomatic go, I'm dealing with each error immediately, and so do not need to worry about shadowing. When I ran this check in the standard go 'net' package, the linter came back with 130 cases of shadowing errors.

Fixing this check by renaming the error variables to some other arbitrary name would just make the code harder to read,  since go developers are used to scanning for "err" when they read code. So I don't want to do this.

So unless anyone objects I'm going to exclude shadowed errors from the linting.

JekaMas: It saved me so many times... Shadowing could be extremely dangerous and hard to debug.

piersy: >It saved me so many times... Shadowing could be extremely dangerous and hard to debug.
Are you referring just to shadowed variables or shadowed errors?

Just to clarify I'm not proposing we disable the shadowing check, but that we just exclude shadowed errors.

The general approach to handling errors in go is to deal with them immediately, when you follow that pattern shadowing is not a problem.

The problem I see with shadowing is when you later want to make use of the original variable but are in fact mistakenly referencing the  shadow declaration.

For this problem to present itself with errors you would need to be deferring handling of one error until after some other error returning function has been called. I've seen this approach used very rarely.

Could you link to some areas of the codebase that are using this approach?

JekaMas: Actually, both. Any shadowing is dangerous. And the fact that standard go vet warns about it mean something.

cmd/gengen/new_genesis.go
JekaMas: It's not a problem. Even with same IDs but different by hash genesises two nodes won't establish a connection.

piersy: Could you explain what is preventing the scenario I outlined above from occurring?

cmd/gengen/new_genesis.go
piersy: >Any shadowing is dangerous. And the fact that standard go vet warns about it mean something.

The fact that the `net` std go library has 130 occurrences of error shadowing tells me that the language creators don't consider it a problem. If such shadowing was seriously dangerous they wouldn't allow it in a package that is imported by around 200,000 other packages in the ecosystem.

In fact the standard go vet tool doesn't warn about shadowing, our go vet tool in the linter is something created by golangci.

I'd like to understand the situations in which this lint check has saved you when it comes to shadowed errors. Could you point me to some places in the code where we are deferring error handling and you were saved by this linter?

JekaMas: It was written 10+ years ago. Nobody knew how to write on golang, all best practices wasn't developed.

My point the same as yours: allowing shadowing is a step to have not so good code with problems, it's a small step but it's a step.

Shadowing is listed in 50 shades of go and in standard linters.

cmd/gengen/new_genesis.go
JekaMas: The transaction structure doesn't have chainID field, so it's not relevant to the issue.

cmd/gengen/new_genesis.go
JekaMas: But if you insist: eth/downloader/statesync.go `func (s *stateSync) loop() (err error)` uses named return `(err error` and also defer with error check. Both things could cause unpredictable errors if we allow shadowing. I'm sure that you can find more examples.

piersy: It sounds like you are in favour of removing shadowing in all situations, but remember shadowing is a language feature, it is explicitly allowed by the language. And I would guess it was explicitly allowed in the language so that people didn't have to come up with arbitrarily different names for variables that really meant the same thing in different scopes, I.E I think they allowed shadowing so that people could write exactly the kind of code I have written here without having to come up with a different name for all of their error variables.

I dug a bit deeper and realised that these lint shadow results are in fact false positives (because the outer err is not referenced after shadow declarations), which explains why I felt this linting wasn't helpful. Here is an excerpt from the shadow [file](https://github.com/golangci/golangci-lint/blob/6771beaee6819347f0f926aee564a6244e4ab45e/vendor/golang.org/x/tools/go/analysis/passes/shadow/shadow.go#L28) in golangci-lint.

```This analyzer check for shadowed variables.
A shadowed variable is a variable declared in an inner scope
with the same name and type as a variable in an outer scope,
and where the outer variable is mentioned after the inner one
is declared.
(This definition can be refined; the module generates too many
false positives and is not yet enabled by default.)
For example:
	func BadRead(f *os.File, buf []byte) error {
		var err error
		for {
			n, err := f.Read(buf) // shadows the function variable 'err'
			if err != nil {
				break // causes return of wrong value
			}
			foo(buf)
		}
		return err
	}
```

We can see the authors of the tool are aware that it is not functioning fully correctly and they don't consider it good enough to be enabled by default.

I don't really see what your point is with the example in eth/downloader/statesync.go, the code is complicated, it uses a named return and a defer statement, this is a case where I can see the benefit of being notified of shadowing (as per golangci-lint's definition). Unfortunately golangci-lint does not pick up shadowing here, when In fact there is shadowing (line 333) according to their definition.

So considering the information I have right now, it looks to me like the shadow check is broken, it flags things it shouldn't flag and it doesn't flag things that it should.

I'd still be interested to understand its benefits though. Can you provide links to some github actions runs that failed because of shadow linting checks on errors that really did contain bugs?

JekaMas: Check works fine if you think it's not, please create an issue in golang repo. I've provided examples where the check 8s useful 8n our code.

piersy: Hmm, this pr is evidence that the tool is generating false positives, so Its broken on that level. Are you seeing the linter pick up the shadowing at line 333 of `eth/downloader/statesync.go` ?

docs/source/genesis_file.rst
piersy: Ok so, I discussed this with @yazzaoui, here is a summary.

The deployer address is a mechanism that we use to restrict certain function calls in the Autonity contract to be called only by the autonity client. The Autonity contract only accepts calls to the restricted functions if they come from the deployer address. Internally the client executes these functions by calling the EVM directly, bypassing any code that processes transactions and verifies transaction signatures, when it makes these calls it uses the deployer address as the sender address.

We want to ensure that no users know the private key for the deployer address, otherwise they would be able to send transactions that could execute the restricted system functions of the autonity contract.

At present users of a network have no way to know that the creator of the genesis does not hold the key to the deployer address, they would need to trust that the creator is not badly behaved.

I can see 2 ways to solve this issue.

 * Update the clients to drop any transactions coming from the deployer address, in this way even if the key was discovered, the key holder would not be able to manipulate the system functions. We can also hardcode the address at this point so we could go for the zero address.

* Generate the address at runtime based on the content of the genesis file (including the users), this ensures that no one party has control over the deployer address, and would be safe as long as no-one discovers the key.

cmd/gengen/new_genesis.go
piersy: @JekaMas bump

docs/source/genesis_file.rst
piersy: @yazzaoui Confirmed this is a problem, I also tried setting the address to the empty string but unmarshalling fails with the following error.

```
hex string has length 0, want 40 for common.Address
```

So I'll need to update the address to be a pointer variable and add the  `omitempty ` marshalling tag.

cmd/gengen/new_genesis.go
piersy: @JekaMas 

>Check works fine if you think it's not, please create an issue in golang repo.

I don't know how you are assessing that the check works fine.

The tool authors state
> the module generates too many false positives and is not yet enabled by default


The only example you've given me in the code, is a file you have never worked on, which highlights another issue with the vet tool which is it doesn't detect shadowing in deferred functions.

@JekaMas The example you provided highlights that the check is not working, that is not a surprise since the creators of the check say

piersy: So I'm going to exclude shadowed errors from the linting.

We've had quite a discussion, but throughout it I have only seen reasons for disabling the shadow check. They are.

* The authors don't consider it ready for general use.
* It produces false positives.
* It also misses instances of shadowing.

I have no counter examples showing when this check was useful in the case of errors.

cmd/gengen/new_genesis.go
```diff
@@ -0,0 +1,304 @@
...
+
+		Timestamp: uint64(time.Now().Unix()),
+		Mixhash:   types.BFTDigest,
+
+		// We need to set this to an empty slice to maintain consistency
+		// between the genesis we return here and the same genesis encoded and
+		// decoded from JSON. This is beause go-ethereum marshals []byte using
+		// hexutil.Bytes which marshals nil to '0x' and '0x' subsequently
+		// unmarshals to an empty slice.
+		ExtraData: []byte{},
```
aneequesafdar, line 147: Could you point me to where is this code

cmd/gengen/main.go
```diff
@@ -0,0 +1,217 @@
...
+// writeGenesis writes a json encoded representation of genesis to path.
+func writeGenesis(path string, genesis *core.Genesis) error {
+	f, err := os.Create(path)
+	if err != nil {
+		return fmt.Errorf("failed to create genesis file: %v", err)
+	}
+	defer f.Close()
+	e := json.NewEncoder(f)
+	e.SetIndent("", "  ")
+	err = e.Encode(genesis)
```
aneequesafdar, line 212: Does this implicitly call json.Marshal?

cmd/gengen/gengen_test.go
```diff
@@ -0,0 +1,127 @@
...
+	_, _, err = newGenesis(10, []string{user}, nil)
+	assert.Error(t, err, "invalid port")
+
+	user = "1e12,v,1,lll:6789"
+	_, _, err = newGenesis(10, []string{user}, nil)
+	assert.Error(t, err, "invalid ip")
+
+	user = "1e12,p,1,:6789"
+	_, _, err = newGenesis(10, []string{user}, nil)
+	assert.Error(t, err, "invalid user type and stake combination")
```
aneequesafdar, line 92: Why don't we care about the type of error, can you please explain why is it sufficient to check that there is an error?

cmd/gengen/gengen_test.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func TestKeysProcessing(t *testing.T) {
+	_, generatedKeys, err := newGenesis(10, validUsers, nil)
+	require.NoError(t, err)
+
+	// Check keys were generated for users.
+	require.Equal(t, len(validUsers), len(generatedKeys))
```
aneequesafdar, line 100: Should we not be checking that the keys generated are ecdsa key or doing some type checking?

.golangci/step2.yml
```diff
@@ -13,4 +13,12 @@ linters:
...
\ No newline at end of file
+      check-shadowing: true
+
+issues:
+  # List of regexps of issue texts to exclude, empty list by default.
+  # But independently from this option we use default exclude patterns,
+  # it can be disabled by `exclude-use-default: false`. To list all
+  # excluded by default patterns execute `golangci-lint run --help`
+  exclude:
+    - 'shadow: declaration of "err" shadows declaration at line'
```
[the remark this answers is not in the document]
piersy: How about  a comment for that line saying `# exclude shadow linter results for errors` ?

cmd/gengen/main.go
```diff
@@ -0,0 +1,217 @@
...
+
+	userDescription = `
+Specifies the parameters for a user.Can be specified multiple times, once for
+each user. The first user to be defined on the command line will become the
+operator (a user with special privileges and the ability to administer the
+system). A user is defined by 4 parameters separated by a comma with no spaces.
+
+Initial eth - The starting eth in wei (1 ETH is 10^18 wei) for this user, a non
+negative 256 bit integer value. This can be specified with scientific notation,
+the max value specified in scientific notation would be 1.1579209e77.
```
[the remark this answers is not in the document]
piersy: I typed `2^256` into a google search and looked at the result.

cmd/gengen/main.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	genesis, userKeys, err := newGenesis(minGasPrice, users, userKeys)
+	if err != nil {
+		return fmt.Errorf("failed to generate genesis: %v", err)
+	}
+	err = writeKeys(userKeysFile, userKeys)
```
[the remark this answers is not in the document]
piersy: "truncate" means remove all content from the file.

If non nil `userKeys` are passed to `newGenesis` then it returns them unchanged, otherwise it generates keys. So in the case that there was a previous keys file, we will be overwriting it but with exactly the same data.


So at this point in the code userKeys will hold the keys, if the variable userKeys was not nil when is was passed

cmd/gengen/main.go
piersy: I don't think so, I think  Marshal and Encode both make use of private json functionality.

cmd/gengen/new_genesis.go
```diff
@@ -0,0 +1,304 @@
...
+	// Check to see the number is positive.
+	if f.Cmp(big.NewFloat(0)) == -1 {
+		return nil, fmt.Errorf("the nunber defined by %q is less than 0", str)
+	}
+
+	// Check to see if the number is an integer.
+	result, accuracy := f.Int(nil)
+	if accuracy != big.Exact {
+		return nil, fmt.Errorf("the nunber defined by %q is not an integer", str)
+	}
```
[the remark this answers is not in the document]
piersy: x is the `big.Float` when you convert it to an int an it returns an argument telling you whether the `big.Float` was really an exact integer or if it was not which direction it was rounded in.

cmd/gengen/gengen_test.go
piersy: The only reason to have an error type, would be if you were going to react to different types in different ways. In this case there is only one course of action taken when an error is encountered, and that is to fail the whole program.

Just checking for an error lets me know that the provided input was invalid.

docs/source/genesis_file.rst
piersy: Opened an issue for this #527

cmd/gengen/new_genesis.go
piersy: You can see here the `genesisSpecMarshaling` struct which somehow configures gencodec, if you look at the marshaling functions defined on `hexutil.Bytes` you can see what I'm talking about in my comment.

https://github.com/clearmatics/autonity/blob/7df8ef052bceeb6b45b51b22a9b7e496f8bafd47/core/genesis.go#L97-L101

docs/source/genesis_file.rst
```diff
@@ -0,0 +1,319 @@
...
+
+ TODO link enode description to documentation - https://github.com/clearmatics/autonity/issues/528
+
+.. code-block:: text
+
+   {
+
+     // config was originally intended by the geth team to hold config for the
+     // ethereum client that did not impact the contents of the block 0 (the
+     // genesis block), for the most part it still does, except that the committee
```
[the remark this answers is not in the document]
piersy: This is meant to say

>for the most part it still does

Meaning that most of the fields of config do not impact the contents of block 0

docs/source/genesis_file.rst
```diff
@@ -0,0 +1,319 @@
...
+     // field in block 0 is derived from the Autonity users specified inside
+     // autonityContract.
```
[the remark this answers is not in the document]
piersy: I think in and of work equally well here, I have however updated the text to make it more clear where autonityContract comes from.

docs/source/genesis_file.rst
```diff
@@ -0,0 +1,319 @@
...
+       "ethash": {},
+
+       // tendermint defines configuration for the tendermint consensus protocol
+       // used by Autonity, and must be set. The gengen tool will set this.
+
+       "tendermint": {
+
+         // block-period defines the minimum time in seconds between blocks.
+         // Without this property limiting the speed at which blocks can be
+         // processed node APIs can become unresponsive while the node consumes
```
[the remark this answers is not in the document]
piersy: It should actually be `processed, node APIs`

piersy:
@aneequesafdar I think I've got back to you on all your questions. With regards to the English suggestions. I've applied them manually to preserve the formatting in the rst file. The ones I applied verbatim I have resolved, there were some where I modified the suggestion that was applied, I have left those unresolved without a comment, in case you want to re-review those parts, and there were some of your suggestions that I think were not correct, those are also unresolved and I left a comment.

cmd/gengen/gengen_test.go
aneequesafdar: @piersy ?

piersy: Hey @aneequesafdar, I must have missed this one, sorry!

This test is more about checking that keys are generated when not provided but used when provided. Let me know if that is not clear and I can add a comment to clarify, or maybe you could suggest a change.

------

We could validate the generated keys here but then I think we start slipping towards duplicating the code in the genesis creation function in the test.

If we did this for the keys, to maintain consistency shouldn't we validate every other field of the genesis? If we went with that approach it would mean a lot more tests and I think we would end up with either lots of duplication of the logic inside the genesis generation function or end up calling the actual code used in the genesis generation function to perform the test (in which case we are not really testing anything).

For all this effort, none of these tests would actually answer the more pertinent question of whether this genesis is valid? To answer that we would need to actually use the genesis in setting up a network. That is something I plan to do by using this code to generate the genesis in our e2e tests but I think is out of scope for this PR. Once we can use this code to generate the genesis for our e2e tests we will be getting for "free" with each test a validation that this code is generating correct genesis files.

PR #518, opened 2020-04-17
[Discussion] Possible proposer elections attack tests

b00ris:


piersy:
Closing this as these attacks assume a malicious operator which is not a situation we are considering. The weakness here stems from the fact that our round robin implementation is sorting the set which it selects from and could be fixed by not ordering the set.

PR #521, opened 2020-04-20
Contract tests for incorrect enodes

### LINKED ISSUE #520: Add test for incorrect enodes in contract tests
By the time we only test Autonity contract calls with correct enodes. We need to add different test cases for incorrect enodes

JekaMas:
Adds contract level tests for incorrect enodes.
Closes #520

### DIFF
--- a/contracts/autonity/contract/test/test.js
+++ b/contracts/autonity/contract/test/test.js
@@ -582,4 +582,81 @@ contract('Autonity', function (accounts) {
             await token.removeUser(accounts[5], {from: operator});
         });
     });
+
+    describe('Enode validationof incorrect cases', function() {
+        let incorrectEnodes = [
+            {
+                case: "missing prefix",
+                enode: ""
+            },
+            {
+                case: "invalid signature on node record",
+                enode: "enr:"
+            },
+            {
+                case: "invalid signature on node record",
+                enode: "enr:x"
+            },
+            {
+                case: "invalid signature on node record",
+                enode: "enr:-EmGZm9vYmFyY4JpZIJ2NIJpcIR_AAABiXNlY3AyNTZrMaEDOlFBdkZvqBXtSB_60JEQotNE9sm3jB0Ur8NRw6Ub4z2DdWRwgnZf"
+            },
+            {
+                case: "missing prefix",
+                enode: "1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439"
+            },
+            {
+                case: "invalid domain or IP address",
+                enode: "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@hostname:3"
+            },
+            {
+                case: "invalid port",
+                enode: "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo"
+            },
+            {
+                case: "invalid public key (wrong length, want 128 hex chars)",
+                enode: "01010101"
+            },
+            {
+                case: "invalid public key (wrong length, want 128 hex chars)",
+                enode: "enode://01010101"
+            },
+            {
+                case: "invalid public key (wrong length, want 128 hex chars)",
+                enode: "enode://01010101@123.124.125.126:3"
+            },
+            {
+                case: "invalid public key",
+                enode: "enode://ffffd65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo"
+            },
+            {
+                case: "invalid URL scheme",
+                enode: "http://foobar"
+            },
+            {
+                case: "missing protocol scheme",
+                enode: "://foo"
+            }
+        ];
+
+        beforeEach(async function() {
+            token = await utils.deployContract(validatorsList, whiteList,
+                userTypes, stakes, commisionRate, operator, minGasPrice, bondPeriod, committeeSize, version, {from: accounts[8]} );
+        });
+
+        let getCallback = function(incorrectEnode) {
+            return async function () {
+                try {
+                    await token.addValidator(accounts[7], 100, incorrectEnode, {from: operator});
+                    assert.fail('Expected throw not received');
+                } catch (e) {
+                    // successful path with an expected exception
+                }
+            }
+        }
+
+        incorrectEnodes.forEach(function(testCase, i, arr) {
+            it('enode should fail ' + i + " '" + testCase.case + "'", getCallback(testCase.enode));
+        });
+    });
 });
--- a/contracts/autonity/contract/test/test.js
+++ b/contracts/autonity/contract/test/test.js
@@ -618,16 +618,20 @@ contract('Autonity', function (accounts) {
                 enode: "01010101"
             },
             {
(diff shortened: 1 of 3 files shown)


contracts/autonity/contract/test/test.js
```diff
@@ -582,4 +582,81 @@ contract('Autonity', function (accounts) {
             await token.removeUser(accounts[5], {from: operator});
         });
     });
+
+    describe('Enode validationof incorrect cases', function() {
+        let incorrectEnodes = [
+            {
+                case: "missing prefix",
```
piersy, line 589: I think these case names are actually mapping to the error string that are returned from the validation function.

I think it would be clearer if the case names were describing what was actually happening in the case, for example in this case the enode is empty. So I would call the case "empty enode". That means that if the implementation of the validation function changes and starts returning different errors these cases will continue to make sense.

Or alternatively, if we really wan't to check that a specific error string is returned then lets check that.

JekaMas: We don't return different errors to the user, error message would be the same in all cases.
So the cases describe what is happening in the test.

piersy: By in the test, I guess you are referring to what happens inside the validation code, in this case it fails because the validation code determines that the there is a missing prefix, but that is a detail based on the internals of the implementation of the validation function.

The internals of that function can change and if they do this description of what is happening in the test can become wrong. If however instead of making reference to what happens in the test we were to name test cases based on the input enode they are more likely to remain meaningful across implementation changes.

JekaMas: If validation function will change than the whole Ethereum enode format will change, if enode format will change then discovery, p2p and consensus will change. I'm not sure should we worry about it...

Also, you can check further tests and how we should name them?

piersy: Not necessarily. Imagine that an empty enode check were added to the beginning of the validation function.

E.G
```
	if enode == "" {
		return errors.New("empty enode")
	}
```

This doesn't affect the format, of an enode but it does change the implementation.

If that change were added  the description "missing prefix" wouldn't make much sense any more.

I'm happy to suggest names for the remaining tests, but it doesn't seem worth it unless we can come to an agreement on this point.

JekaMas: It also will break a few another tests in the codebase. So what's the problem and do you have something more friendly than "correct public key with correct prefix, correct IP and incorrect port", ""correct public key with incorrect prefix, correct IP and port".

Also "empty enode" doesn't reflect the actual problem with the enode.

piersy: > Also "empty enode" doesn't reflect the actual problem with the enode

I would say it does,  an empty enode will always be wrong, and the test case for empty enode would still be a valid testcase even if the entire enode format changed.

But if we don't agree on this point, I think we can just agree to disagree.

contracts/autonity/contract/test/test.js
```diff
@@ -582,4 +582,85 @@ contract('Autonity', function (accounts) {
...
+            {
+                case: "invalid domain or IP address",
+                enode: "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@hostname:3"
+            },
+            {
+                case: "invalid port",
+                enode: "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo"
+            },
+            {
+                case: "invalid public key (wrong length, want 128 hex chars)",
```
piersy, line 617: Why isn't this case called missing prefix?

contracts/autonity/contract/test/test.js
```diff
@@ -582,4 +582,85 @@ contract('Autonity', function (accounts) {
...
+                enode: "enode://01010101"
+            },
+            {
+                case: "invalid public key (wrong length, want 128 hex chars) with correct prefix and IP:PORT",
+                enode: "enode://01010101@123.124.125.126:3"
+            },
+            {
+                case: "invalid port",
+                enode: "enode://ffffd65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:foo"
+            },
```
piersy, line 631:
```suggestion
            },
            {
                case: "invalid port",
                enode: "enode://ffffd65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:0"
            },
```

contracts/autonity/contract/test/test.js
```diff
@@ -582,4 +582,85 @@ contract('Autonity', function (accounts) {
...
+            {
+                case: "invalid public key",
+                enode: "enode://ffffd65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:30303"
+            },
+            {
+                case: "invalid URL scheme",
+                enode: "http://foobar"
+            },
+            {
+                case: "missing protocol scheme",
```
piersy, line 641: What is the difference between missing protocol scheme and missing prefix?

contracts/autonity/contract/test/test.js
```diff
@@ -582,4 +582,85 @@ contract('Autonity', function (accounts) {
...
+                enode: "://foo"
+            }
```
piersy, line 643:
```suggestion
            },
            {
                case: "invalid query parameter",
                enode: "enode://ffffd65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:4040?discpor=30301"
            },
            {
                case: "invalid discport port",
                enode: "enode://ffffd65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:4040?discport=0"
            },
```

core/vm/contract_check_enode.go
```diff
@@ -0,0 +1,28 @@
...
+
+// checkEnode implemented as a native contract.
+type checkEnode struct{}
+
+func (c checkEnode) RequiredGas(_ []byte) uint64 {
+	return params.EnodeCheckGas
+}
+func (c checkEnode) Run(input []byte) ([]byte, error) {
+	if len(input) == 0 {
+		panic(fmt.Errorf("invalid enode - empty"))
```
piersy, line 19: What is the effect of panicking on contract execution? How does panicking compare with returning an error?

contracts/autonity/contract/test/test.js
JekaMas: It's a different test case.

piersy: Yes, the suggestion is to add another test case that checks that ports of 0 are not valid.

JekaMas: Technically there's nothing wrong with port 0

piersy: Binding on port 0 causes the operating system to randomly assign a port. So an enode with port 0 would never be accessible, since they cant actually be listening on port 0.

JekaMas: It's true only for Linux

piersy: That may be so, it is however a reserved port and should not be used according to RFC6335. Also the Autonity client is currently only required to support linux/x64.

contracts/autonity/contract/test/test.js
piersy: @JekaMas I can see that you merged this PR but that this suggestion was not included. I'm wondering why, are these not valid test cases?

PR #545, opened 2020-05-05
Simplify dns resolution

piersy:
This PR replaces the resolveSet implementation with a modified
enode.Node.

The resolveSet approach to dns resolution was problematic for a number
of reasons.

* It made use of global state, which is likely to make testing
difficult.

* It wasted compute resources by perpetually re-resolving its known set
of hosts irrespective of whether we actually needed to resolve the
hosts.

* Because it's resolution timeline was ignorant of when connections were
being made it couldn't ensure that the resolution results were up to
date when they were used to make a connection.

* Although it allowed us to resolve enodes with hosts it didn't change the
definition of enode, meaning that when parsing an enode the caller would
need to know if they have an enode with a host or not in order to call
the correct parsing function.

* By not changing the definition of enode.Node to be host aware the
previous implementation did not provide any mechanism for the p2p
package to resolve hostnames, meaning that once a Node was handed over
to the p2p package it was stuck with that IP.
To solve this shortcoming it looks like code was written that, for every
block loads the whitelist from the contract, tries to resolve all the
hostnames, converts the entries into enode.Node instances and adds them
to the p2p package. This has the effect of updating the ips associated
with the enodes once a block. This approach suffers from the problem
that for each block, we actively disconnect from nodes that we weren't
able to resolve, even if we are currently connected to them and they are
in our whitelist. The approach is very verbose and a waste
of computing resources.
This code still exists and should be removed. In order to remove it we
need to be able to react to participant change events such that we only
add or remove enodes from the p2p package when the whitelist actually
changes.


p2p/enode/node.go
```diff
@@ -138,6 +140,52 @@ func (n *Node) Pubkey() *ecdsa.PublicKey {
...
+		ip = ipv4
+	}
+	n.r.Set(enr.IP(ip))
+	return nil
+}
+
+// // Ensure the IP is 4 bytes long for IPv4 addresses.
+// if ipv4 := ip.To4(); ipv4 != nil {
+// 	ip = ipv4
+// }
```
aneequesafdar, line 187: Repeated commented code from above

p2p/enr/entries.go
```diff
@@ -94,6 +94,12 @@ func (v IP) ENRKey() string {
 	return "ip"
 }
 
+type HOST string
+
+func (v HOST) ENRKey() string {
+	return "host"
+}
+
```
aneequesafdar, line 102: Do you need to define Encode and Decode RlP functions on this record?

p2p/enode/node.go
piersy, line 187: Removing this:
```suggestion
```

p2p/enr/entries.go
piersy: No I don't because the underlying type of HOST is string which doesn't require any special encoding or decoding for RLP

Issue #542, opened 2020-05-01
Add quorum precommit tests for tendermint


aneequesafdar:
# Problem

We don't have tests for what a node should do when it receives quorum precommits.

This is a sub-task for #480.

# Acceptance Criteria

Use lines 49-54 of page 6 of [The latest gossip on BFT consensus](https://arxiv.org/pdf/1807.04938.pdf) to describe the correct behaviours in test function for the current implementation of Tendermint.

Issue #539, opened 2020-05-01
Add quorum prevote tests for tendermint


aneequesafdar:
# Problem

We don't have tests for what a node should do when it receives quorum prevote.

This is a sub-task for #480.

# Acceptance Criteria

Use lines 36-43 of page 6 of [The latest gossip on BFT consensus](https://arxiv.org/pdf/1807.04938.pdf) to describe the correct behaviours in test function for the current implementation of Tendermint.

Issue #538, opened 2020-05-01
Add schedule prevote timeout tests for tendermint


aneequesafdar:
# Problem

We don't have tests for when a node should schedule prevote timeout and what should be done when it expires.

This is a sub-task for #480.

# Acceptance Criteria

Use lines 34-35 & 61-64 of page 6 of [The latest gossip on BFT consensus](https://arxiv.org/pdf/1807.04938.pdf) to describe the correct behaviours in test function for the current implementation of Tendermint.

Issue #537, opened 2020-05-01
Add old proposal tests for tendermint


aneequesafdar:
# Problem

We don't have tests for what a node should do when it receives an old proposal from the proposer.

This is a sub-task for #480.

# Acceptance Criteria

Use lines 28-33 of page 6 of [The latest gossip on BFT consensus](https://arxiv.org/pdf/1807.04938.pdf) to describe the correct behaviours in test function for the current implementation of Tendermint.

PR #547, opened 2020-05-06
Consensus behavior test - Start Round, Proposal, Prevote msg handling.

Jason-Zhangxin-Chen:
Start to test the consensus logic from a higher level rather than the simple pieces of function 
test implemented by us before. 

The PR contains tendermint behavior test: Tendermint StartRound() logics explain on its pseudo-code in the paper:  https://arxiv.org/pdf/1807.04938.pdf, at page-6, line-11.


consensus/tendermint/core/tendermint_start_round_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if c.sentProposal != true {
+		t.Error("Proposer did not send proposal on consensus start round")
+	}
+
+	if c.step != propose {
+		t.Error("Proposer did not step to propose step")
+	}
+
+	if c.Round() != round {
```
piersy, line 83: Because `c.round` is initialised to 0 in a new `core` this check is not checking that round was correctly set, in order to check that we need to use a value for round other than 0.

consensus/tendermint/core/tendermint_start_round_test.go
```diff
@@ -0,0 +1,124 @@
...
+	c.sentProposal = false
+	c.height = currentHeight
+	round := int64(0)
+	c.startRound(context.Background(), round)
+
+	if c.sentProposal != true {
+		t.Error("Proposer did not send proposal on consensus start round")
+	}
+
+	if c.step != propose {
```
piersy, line 79: Because `c.step` is initialised to propose in a new `core`, this check is not checking that step was correctly set by `c.startRound`, in order to check that we should set `c.step` to something other than `propose` before calling `c.startRound`

consensus/tendermint/core/tendermint_start_round_test.go
```diff
@@ -0,0 +1,124 @@
...
+	// create consensus core.
+	c := New(backendMock)
+	c.committeeSet = committeeSet
+	round := int64(1)
+	c.startRound(context.Background(), round)
+	if c.proposeTimeout.started != true {
+		t.Error("Follower did not start proposal timer on consensus start round")
+	}
+
+	if c.step != propose {
```
piersy, line 117: Same comment as on the check for `propose` on the other test.

consensus/tendermint/core/tendermint_start_round_test.go
```diff
@@ -0,0 +1,124 @@
...
+	c.pendingUnminedBlocks[proposalHeight.Uint64()] = proposalBlock
+	c.committeeSet = committeeSet
+	c.sentProposal = false
+	c.height = currentHeight
+	round := int64(0)
+	c.startRound(context.Background(), round)
+
+	if c.sentProposal != true {
+		t.Error("Proposer did not send proposal on consensus start round")
+	}
```
piersy, line 77:
```suggestion
assert.True(t, c.sentProposal, "Proposer did not send proposal on consensus start round")
```

We could use the testify assert package here just to make our tests a bit more streamlined, it contains a lot of matchers beyond just `assert.True`.

Jason-Zhangxin-Chen:
@piersy , thank you so much for the findings that I did not noticed.

aneequesafdar:
Could you please create separate pull requests for each of the issues described in issues #535, #536, #537, #538, #539, #540, #541 and #542, I think I would make reviewing much easier and we won't have to spend a lot of time reviewing one pull request for a long time. The issues describe the lines which they are meant to be testing. So all you have to do is take your testcases for proposal, prevote and precommit and create separate pull request reference each of the issues mentioned above.

We can use different files for each of the issues so that we don't run into conflict and once everything is approved we can refactor it into a single file.

Jason-Zhangxin-Chen:
> Could you please create separate pull requests for each of the issues described in issues #535, #536, #537, #538, #539, #540, #541 and #542, I think it would make reviewing much easier and we won't have to spend a lot of time reviewing one pull request for a long time. The issues describe the lines which they are meant to be testing. So all you have to do is take your testcases for proposal, prevote and precommit and create separate pull request referencing each of the issues mentioned above.
> 
> We can use different files for each of the issues so that we don't run into conflict and once everything is approved we can refactor it into a single file.
> 
> Also, it may be helpful to use the branch name convention as follows, `tdm/test/{title of issue and number}`, just a suggestion.

Thank you, Aneeque, I will break them into seperate ones, then finally we refine them into single file.

aneequesafdar:
closing because of https://github.com/clearmatics/autonity/pull/547#issuecomment-628371662

PR #549, opened 2020-05-07
1.9.15 upstream upgrade

yazzaoui:
- Made miner tests compatible with Tendermint ( see https://github.com/clearmatics/autonity/pull/549/commits/fe5a3023640ca01e88861b642b18d51e5d51e435 ) with the introduction of the Autonity contract and fullChainReader interfaces.
- Removed consensus dependency from Autonity package.
- As usual, there are some linter issues with the upstream branch so I'll have to force merge this branch.


consensus/tendermint/backend/engine.go
```diff
@@ -364,7 +363,7 @@ func (sb *Backend) AutonityContractFinalize(header *types.Header, chain consensu
 	defer sb.contractsMu.Unlock()
 
 	if header.Number.Int64() == 1 {
-		contractAddress, err := sb.blockchain.GetAutonityContract().DeployAutonityContract(chain, header, state)
+		contractAddress, err := sb.blockchain.GetAutonityContract().DeployAutonityContract(sb.blockchain.Config(), header, state)
```
aneequesafdar: Why is this not `chain.Config()`? if it makes no difference can you please remove the unused `chain` variable from this function signature

contracts/autonity/autonity.go
```diff
@@ -29,32 +29,42 @@ type EVMProvider interface {
...
-		operator:           operator,
-		initialMinGasPrice: minGasPrice,
-		bc:                 bc,
-		evmProvider:        evmProvider,
-	}
-	err := contract.upgradeAbiCache(ABI)
-	return &contract, err
+type ChainContext interface {
+	// GetHeader returns the hash corresponding to their hash.
+	GetHeader(common.Hash, uint64) *types.Header
```
aneequesafdar, line 34: `ChainContext` is only used in one place: it is embedded inside `Blockchainer` interface. While the `GetHeader` function is not used anywhere in autonity package, so I think `ChainContext` is not needed and should be removed. Otherwise please point me to where the `GetHeader` function is being called.

aneequesafdar, line 34:
```suggestion
```

contracts/autonity/autonity.go
```diff
@@ -29,32 +29,42 @@ type EVMProvider interface {
...
 }
```
aneequesafdar, line 35:
```suggestion

```

contracts/autonity/autonity.go
```diff
@@ -29,32 +29,42 @@ type EVMProvider interface {
...
 
 type Blockchainer interface {
+	ChainContext
```
aneequesafdar, line 38:
```suggestion
```

contracts/autonity/autonity.go
aneequesafdar, line 35:
```suggestion
```

core/evm.go
```diff
@@ -25,14 +25,18 @@ import (
 	"github.com/clearmatics/autonity/core/vm"
 )
 
+// This interface was introduced to avoid cyclic imports with the autonity package.
```
aneequesafdar, line 28: can you please explain what was causing cyclic import? As far as I can see autonity package was not in the import list before this change

core/evm.go
```diff
@@ -25,14 +25,18 @@ import (
...
+type HeaderGetter interface {
+	// GetHeader returns the hash corresponding to their hash.
+	GetHeader(common.Hash, uint64) *types.Header
+}
+
```
aneequesafdar, line 33: I reverted this change and replaced all usages of HeaderGetter with ChainContext (I also added back the GetHeader to ChainContext), all of them passed locally, therefore I believe this is no longer required.

```suggestion
```

core/evm.go
```diff
@@ -25,14 +25,18 @@ import (
...
 // ChainContext supports retrieving headers and consensus parameters from the
 // current blockchain to be used during transaction processing.
 type ChainContext interface {
+	HeaderGetter
```
aneequesafdar, line 37:
```suggestion
```

core/evm.go
```diff
@@ -25,14 +25,18 @@ import (
...
 	// Engine retrieves the chain's consensus engine.
 	Engine() consensus.Engine
-
-	// GetHeader returns the hash corresponding to their hash.
-	GetHeader(common.Hash, uint64) *types.Header
```
aneequesafdar, line 35: Please add back

core/evm.go
```diff
@@ -59,25 +63,33 @@ func NewEVMContext(msg Message, header *types.Header, chain ChainContext, author
 }
 
 // GetHashFn returns a GetHashFunc which retrieves header hashes by number
-func GetHashFn(ref *types.Header, chain ChainContext) func(n uint64) common.Hash {
-	var cache map[uint64]common.Hash
+func GetHashFn(ref *types.Header, chain HeaderGetter) func(n uint64) common.Hash {
```
aneequesafdar, line 66:
```suggestion
func GetHashFn(ref *types.Header, chain ChainContext) func(n uint64) common.Hash {
```

miner/worker_test.go
```diff
@@ -47,73 +47,51 @@ const (
...
 	}
-	t.newTxs = append(t.newTxs, tx2)
+)
 
+func init() {
+	testTxPoolConfig = core.DefaultTxPoolConfig
+	testTxPoolConfig.Journal = ""
+	ethashChainConfig = params.TestChainConfig
+	tendermintChainConfig = &params.ChainConfig{}
+	tendermintChainConfig = params.AutonityTestChainConfig
```
aneequesafdar, line 80: Can you please explain why is `tendermintChainConfig` set twice? I think the following is sufficient

```suggestion
	tendermintChainConfig = params.AutonityTestChainConfig
```

I ran the worker_test.go file with the above suggestion, it passes.

miner/worker_test.go
```diff
@@ -222,98 +169,101 @@ func (b *testWorkerBackend) newRandomUncle() *types.Block {
...
-		}
-		if err := b.txPool.AddLocal(b.newRandomTx(testCase, false)); err != nil {
-			t.Fatal(err)
+		b.txPool.AddLocal(b.newRandomTx(true))
+		b.txPool.AddLocal(b.newRandomTx(false))
+		if !isTendermint {
+			// Don't create fake uncles as it is in theory an impossible scenario.
+			// We're only testing here the import functionality.
+			w.postSideBlock(core.ChainSideEvent{Block: b.newRandomUncle()})
+			w.postSideBlock(core.ChainSideEvent{Block: b.newRandomUncle()})
```
aneequesafdar, line 240: Just curious to know what would happen without this if condition, does it break the tests? if so, do you know why?

consensus/tendermint/backend/engine.go
piersy: I'm confused as to why this comment appears as outdated.

contracts/autonity/autonity.go
piersy: This is addressed in #596

contracts/autonity/autonity.go
piersy: This is addressed in #596

contracts/autonity/autonity.go
piersy: This is addressed in #596

core/evm.go
piersy: This is addressed in #596

core/evm.go
piersy: This is addressed in #596

core/evm.go
piersy: This is addressed in #596

core/evm.go
piersy: This is addressed in #596

core/evm.go
piersy: This is addressed in #596

consensus/tendermint/backend/engine.go
aneequesafdar: Because I made the comment on a previous commit, instead of the latest one, since I was specifically looking at @yazzaoui commit.

miner/worker_test.go
yazzaoui: great thanks !

Issue #527, opened 2020-04-29
Make deployer address safe


piersy:
# Problem

The deployer address is a mechanism that we use to restrict certain function calls in the Autonity contract to be called only by the autonity client. The Autonity contract only accepts calls to the restricted functions if they come from the deployer address. Internally the client executes these functions by calling the EVM directly, bypassing any code that processes transactions and verifies transaction signatures, when it makes these calls it uses the deployer address as the sender address.

We want to ensure that no users know the private key for the deployer address, otherwise they would be able to send transactions that could execute the restricted system functions of the autonity contract.

At present users of a network have no way to know that the creator of the genesis does not hold the key to the deployer address, they would need to trust that the creator is not badly behaved.

# Implementation

I can see 2 ways to solve this issue.

 * Update the clients to drop any transactions coming from the deployer address, in this way even if the key was discovered, the key holder would not be able to manipulate the system functions. We can also hardcode the address at this point so we could go for the zero address.

* Generate the address at runtime based on the content of the genesis file (including the users), this ensures that no one party has control over the deployer address, and would be safe as long as no-one discovers the key.

# Acceptance Criteria

Participants do not have to trust that the creator of the genesis file does not hold the key to the genesis address.

# Notes

I think I prefer the first approach outlined above (drop transactions from deployer address), it seems conceptually simpler and doesn't run the risk that anyone can discover the key.

aneequesafdar:
I am not sure we can simply drop the transaction from the deployer address since it is meant to be controlled by the operator and the operator is responsible for adding and removing participants from the Autonity contract, thus deployer's address is required to make such transactions.

Unless we somehow provide another way for the operator to change the participants, which I am not sure is a good idea since it would not be recorded on the blockchain.

piersy:
@aneequesafdar I'm not following your explanation.

It seems like you are saying that the operator is able to change the deployer address.
> I am not sure we can simply drop the transaction from the deployer address since it is meant to be controlled by the operator

And also saying that the operator needs to use the deployer address to add or remove users.
> the operator is responsible for adding and removing participants from the Autonity contract, thus deployer's address is required to make such transactions

[...]

PR #552, opened 2020-05-12
Hardcode deployer address

aneequesafdar:
This pull request fixes the problem described in issue #527

### DIFF
--- a/contracts/autonity/autonity.go
+++ b/contracts/autonity/autonity.go
@@ -2,6 +2,7 @@ package autonity
 
 import (
 	"errors"
+	"github.com/clearmatics/autonity/common/acdefault"
 	"math/big"
 	"reflect"
 	"sort"
@@ -76,7 +77,7 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
 	}
 
 	// prepare abi and evm context
-	deployer := ac.bc.Config().AutonityContractConfig.Deployer
+	deployer := acdefault.Deployer()
 	sender := vm.AccountRef(deployer)
 	gas := uint64(0xFFFFFFFF)
 	evm := ac.getEVM(header, deployer, stateDB)
--- a/contracts/autonity/calls.go
+++ b/contracts/autonity/calls.go
@@ -2,6 +2,7 @@ package autonity
 
 import (
 	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/common/acdefault"
 	"github.com/clearmatics/autonity/consensus"
 	"github.com/clearmatics/autonity/core/state"
 	"github.com/clearmatics/autonity/core/types"
@@ -56,8 +57,8 @@ func (ac *Contract) getEVM(header *types.Header, origin common.Address, statedb 
 func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *types.Header, statedb *state.StateDB) (common.Address, error) {
 	// Convert the contract bytecode from hex into bytes
 	contractBytecode := common.Hex2Bytes(chain.Config().AutonityContractConfig.Bytecode)
-	evm := ac.getEVM(header, chain.Config().AutonityContractConfig.Deployer, statedb)
-	sender := vm.AccountRef(chain.Config().AutonityContractConfig.Deployer)
+	evm := ac.getEVM(header, acdefault.Deployer(), statedb)
+	sender := vm.AccountRef(acdefault.Deployer())
 
 	contractABI, err := ac.abi()
 
@@ -124,7 +125,7 @@ func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *
 }
 
 func (ac *Contract) UpdateAutonityContract(header *types.Header, statedb *state.StateDB, bytecode string, abi string, state []byte) error {
-	caller := ac.bc.Config().AutonityContractConfig.Deployer
+	caller := acdefault.Deployer()
 	evm := ac.getEVM(header, caller, statedb)
 	contractBytecode := common.Hex2Bytes(bytecode)
 	data := append(contractBytecode, state...)
@@ -139,7 +140,7 @@ func (ac *Contract) UpdateAutonityContract(header *types.Header, statedb *state.
 }
 
 func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.Header, function string, result interface{}, args ...interface{}) error {
-	caller := ac.bc.Config().AutonityContractConfig.Deployer
+	caller := acdefault.Deployer()
 	contractABI, err := ac.abi()
 	if err != nil {
 		return err
@@ -229,7 +230,7 @@ func (ac *Contract) callRetrieveContract(state *state.StateDB, header *types.Hea
 
 func (ac *Contract) callSetMinimumGasPrice(state *state.StateDB, header *types.Header, price *big.Int) error {
 	// Needs to be refactored somehow
-	deployer := ac.bc.Config().AutonityContractConfig.Deployer
+	deployer := acdefault.Deployer()
 	sender := vm.AccountRef(deployer)
 	gas := uint64(0xFFFFFFFF)
 	evm := ac.getEVM(header, deployer, state)
--- a/params/autonity_contract.go
+++ b/params/autonity_contract.go
@@ -42,15 +42,15 @@ func (ut UserType) GetID() int {
 // Autonity contract config. It'is used for deployment.
 type AutonityContractGenesis struct {
 	// Address of the validator who deploys contract stored in bytecode
-	Deployer common.Address `json:"deployer" toml:",omitempty"`
+	Deployer common.Address `json:"deployer, omitempty"`
 	// Bytecode of validators contract
 	// would like this type to be []byte but the unmarshalling is not working
-	Bytecode string `json:"bytecode" toml:",omitempty"`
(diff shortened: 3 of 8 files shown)


params/autonity_contract.go
```diff
@@ -42,15 +42,15 @@ func (ut UserType) GetID() int {
...
+	Bytecode string `json:"bytecode, omitempty"`
 	// Json ABI of the contract
-	ABI         string         `json:"abi" toml:",omitempty"`
-	MinGasPrice uint64         `json:"minGasPrice" toml:",omitempty"`
-	Operator    common.Address `json:"operator" toml:",omitempty"`
-	Users       []User         `json:"users" toml:",omitempty"`
+	ABI         string         `json:"abi, omitempty"`
+	MinGasPrice uint64         `json:"minGasPrice, omitempty"`
+	Operator    common.Address `json:"operator, omitempty"`
+	Users       []User         `json:"users, omitempty"`
```
piersy: How do you know that these toml tags are unused?

If they are unused then we should remove the `--config value    TOML configuration file` command-line parameter from the autonity client.

Also is this related to hardcoding the deployer?

aneequesafdar: If you look at `ChainConfig` in config.go, which embeds `AutonityContractConfig`, you can see that `ChainConfig` doesn't use toml tags nor does `Genesis` which embeds `ChainConfig`, if I am misunderstanding please let me know and I will revert. 

And this is not related to hardcoding the deployer, however, it seemed weird to have `omitempty` in toml tag instead of json.

piersy: >And this is not related to hardcoding the deployer,

I would suggest not including this change in this PR then.

>however, it seemed weird to have omitempty in toml tag instead of json.

That does seem odd, I had a look at the use of TOML in go-ethereum and couldn't really see a consistent approach.

I can see that they allow the genesis (which contains ChainConfig) to be represented as TOML.

https://github.com/ethereum/go-ethereum/blob/46698d79310ab1c12089f4bc2c13ac2f136128da/eth/config.go#L92-L96

So I think it might be better to try to come up with a cohesive and consistent approach to how we hanlde toml, and I think it probably deserves its own ticket.

contracts/autonity/autonity.go
```diff
@@ -303,18 +274,5 @@ func (ac *Contract) upgradeAbiCache(newAbi string) error {
 }
 
 func (ac *Contract) GetContractABI() string {
-	ac.Lock()
-	defer ac.Unlock()
-
-	var JSONString = ac.bc.Config().AutonityContractConfig.ABI
-	bytes, err := ac.bc.GetKeyValue([]byte(ABISPEC))
```
Jason-Zhangxin-Chen, line 310: Load contract ABI from level db is removed here. When we upgrade the contract, the ABI will be updated too both in the cach and levelDB for persistent, then for the contract ABI buffer, if the ABI is presented in levelDB, then we need to load it from LevelDB to cach otherwise error might happens. I did capture this logics in your pull request yet.

contracts/autonity/autonity.go
```diff
@@ -303,18 +274,5 @@ func (ac *Contract) upgradeAbiCache(newAbi string) error {
 }
 
 func (ac *Contract) GetContractABI() string {
-	ac.Lock()
```
Jason-Zhangxin-Chen, line 306: Loading contract abi from level db happens on chain data initialization. Nice!

contracts/autonity/autonity.go
```diff
@@ -258,35 +257,7 @@ func (ac *Contract) performContractUpgrade(statedb *state.StateDB, header *types
...
-		if err != nil {
-			log.Error("Cant get contract address", "err", err)
-		}
-		return addr
-	}
-	return ac.address
-}
-
 func (ac *Contract) abi() (*abi.ABI, error) {
-	ac.Lock()
```
Jason-Zhangxin-Chen, line 273: Move this abi loading from levelDB to chain init is good for me. Make sure that the draft versino of contract's ABI are also store at levelDB incase of node restart it could load from level DB too.

contracts/autonity/calls.go
```diff
@@ -102,35 +82,31 @@ func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *
...
-	ac.address = contractAddress
-	ac.Unlock()
-	log.Info("Deployed Autonity Contract", "Address", contractAddress.String())
+	log.Info("Deployed Autonity Contract", "Address", ContractAddress.String())
 
-	return contractAddress, nil
+	return nil
 }
 
 func (ac *Contract) UpdateAutonityContract(header *types.Header, statedb *state.StateDB, bytecode string, abi string, state []byte) error {
```
piersy, line 103: It looks like the  `abi` parameter is unused, but I would think we wan't to update the contract's abi at this point.

We could extract the code from the constructor into an unexported method called `parseABI` and use that here and in the constructor to get an instance of `*abi.ABI`

contracts/autonity/calls.go
```diff
@@ -229,10 +204,8 @@ func (ac *Contract) callRetrieveContract(state *state.StateDB, header *types.Hea
 
 func (ac *Contract) callSetMinimumGasPrice(state *state.StateDB, header *types.Header, price *big.Int) error {
 	// Needs to be refactored somehow
-	deployer := ac.bc.Config().AutonityContractConfig.Deployer
-	sender := vm.AccountRef(deployer)
 	gas := uint64(0xFFFFFFFF)
-	evm := ac.getEVM(header, deployer, state)
+	evm := ac.evmProvider.EVM(header, deployer, state)
 
 	ABI, err := ac.abi()
```
piersy, line 210: There is no longer any need for the `abi` method, lets replace uses of it with the field `ac.contractABI`

contracts/autonity/calls.go
piersy: Actually looking at this again, I think this function shouldn't accept abi as a parameter and we should perform the update of the abi in `performContractUpgrade` after this `UpdateAutonityContract` returns successfully.

Also `UpdateAutonityContract` can be made unexported, it is only used in the autonity package.

piersy: Actually now I see that is what `upgradeAbiCache` is doing and that is called at the appropriate place, we should use `upgradeAbiCache` in the constructor instead of duplicating it's logic there.

contracts/autonity/autonity.go
```diff
@@ -79,11 +79,7 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
 	// prepare abi and evm context
 	gas := uint64(0xFFFFFFFF)
 	evm := ac.evmProvider.EVM(header, deployer, stateDB)
-
-	ABI, err := ac.abi()
-	if err != nil {
-		return
-	}
+	ABI := ac.contractABI
```
piersy, line 81: I think it would be clearer to remove this variable and just use `ac.contractABI` in place of it.

contracts/autonity/calls.go
```diff
@@ -38,13 +38,7 @@ func (ac *Contract) DeployAutonityContract(chain consensus.ChainReader, header *
...
 	contractBytecode := common.Hex2Bytes(chain.Config().AutonityContractConfig.Bytecode)
 	evm := ac.evmProvider.EVM(header, deployer, statedb)
-
-	contractABI, err := ac.abi()
-
-	if err != nil {
-		log.Error("abi.JSON returns err", "err", err)
-		return err
-	}
+	contractABI := ac.contractABI
```
piersy, line 41: This variable is only used once so we should just remove it and use `ac.contractABI` in its place.

contracts/autonity/calls.go
```diff
@@ -115,11 +109,7 @@ func (ac *Contract) UpdateAutonityContract(header *types.Header, statedb *state.
 }
 
 func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.Header, function string, result interface{}, args ...interface{}) error {
-	contractABI, err := ac.abi()
-	if err != nil {
-		return err
-	}
-
+	contractABI := ac.contractABI
```
piersy, line 112: I think it would be clearer to remove this variable and just use `ac.contractABI` in place of it.

contracts/autonity/calls.go
```diff
@@ -206,11 +196,7 @@ func (ac *Contract) callSetMinimumGasPrice(state *state.StateDB, header *types.H
 	// Needs to be refactored somehow
 	gas := uint64(0xFFFFFFFF)
 	evm := ac.evmProvider.EVM(header, deployer, state)
-
-	ABI, err := ac.abi()
-	if err != nil {
-		return err
-	}
+	ABI := ac.contractABI
```
piersy, line 199: I think it would be clearer to remove this variable and just use `ac.contractABI` in place of it.

contracts/autonity/autonity.go
```diff
@@ -36,19 +36,18 @@ func NewAutonityContract(
...
 		initialMinGasPrice: minGasPrice,
 		bc:                 bc,
 		evmProvider:        evmProvider,
-	}, nil
+	}
+	err := contract.upgradeAbiCache(ABI)
+	if err != nil {
+		return nil, err
+	}
+	return &contract, nil
```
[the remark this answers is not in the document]
aneequesafdar: is it ok to return non nil contract when there is an error?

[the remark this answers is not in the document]
piersy: Sure, if the error is non nil, then the returned contract will never be accessed.

Issue #506, opened 2020-04-07
Integrate proposer selection


yazzaoui:
## Problem

The proposer selection mechanism `getProposer(height,round)` in the Autonity Contract need to be integrated with the consensus layer1 implementation.

## Acceptance criteria

- Committee.Set `GetProposer`  method should call Autonity Contract `getProposer`

PR #559, opened 2020-05-28
Implement proposer election weighted by voting power

piersy:
This PR now contains changes from 2 PRs, the intital changes were from @Jason-Zhangxin-Chen to add proposer election weighted by voting power, following that there are changes from me which encapsulate the new proposer election logic in a separate implementation of committee.Set.

Relates to #506

## First PR description
Implement the proposer election weighted by voting power and provided by the autonity contract.
1. Autonity contract side, getProposer(height, round) function was added.
2. Add contract unit testcases to cover getProposer() function.
3. Layer1 tendermint engine side, adapt the getProposer(height, round) of autonity contract.
python e2e testing on this integration will be run too.

## Second PR description
This PR Moves logic for proposer selection via weighted round robin into a single place, which is a separate implementation of the core.committee interface. I also moved the committee interface and implementations into the core package.

There was a lot of preparatory work required to achieve this.

It could be summarised as follows.

1. Separate committee creation code from larger structs (backend.Backend).
1. Simplify committee creation code by removing un-needed parameters.
1. Centralise committee creation code (combine multiple creation methods into one).
1. Reduce use of committee and hence use of committe creation code, by substituting simpler objects for committee where possible.
1. Once use of committee had been isolated to the single location where we decide which implementation is required, move the committee creation code and implementations there.

For a better understanding of the process I would suggest looking at each commit individuality, the commit messages are very descriptive.

There are some further simplifications that could be applied but are out of scope of this PR.
They are:
* Remove the Committee method from committee (it does not make sense to use the committee as a container for types.Committee)
* Remove the GetByAddress and GetByIndex methods from committee, these methods are just a different manifestation of using committee as a container for types.Committee.
* Separate Quorum and F into their own interface, Quorum and F are not related to proposer selection.


piersy:
@Jason-Zhangxin-Chen Haha, thanks :) I don't think its anything too complicated, if you look at the commit messages you'll see its just a sequence of fairly straightforward changes.

consensus/tendermint/backend/backend.go
```diff
@@ -132,26 +130,19 @@ type Backend struct {
 	vmConfig                *vm.Config
 }
 
+func (sb *Backend) BlockChain() *core.BlockChain {
+	return sb.blockchain
+}
+
 // Address implements tendermint.Backend.Address
 func (sb *Backend) Address() common.Address {
-	sb.privateKeyMu.RLock()
```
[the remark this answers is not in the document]
piersy: Yes, it is set on construction only, there is no chance that it can change after that.

consensus/tendermint/core/committee.go
```diff
@@ -0,0 +1,219 @@
...
+	// PoS (voting power) weighted round robin.
+	statedb, err := state.New(w.previousBlockStateRoot, w.bc.StateCache())
+	if err != nil {
+		log.Error("cannot load state from block chain.")
+		return types.CommitteeMember{}
+	}
+	proposer := w.autonityContract.GetProposerFromAC(w.previousHeader, statedb, w.previousHeader.Number.Uint64(), round)
+	member := w.previousHeader.CommitteeMember(proposer)
+	return *member
+	// TODO make this return an error
```
piersy, line 199: This todo is covered by #592

contracts/autonity/autonity.go
```diff
@@ -167,6 +167,10 @@ func (ac *Contract) GetMinimumGasPrice(block *types.Block, db *state.StateDB) (u
 	return ac.callGetMinimumGasPrice(db, block.Header())
 }
 
+func (ac *Contract) GetProposerFromAC(header *types.Header, db *state.StateDB, height uint64, round int64) common.Address {
+	return ac.callGetProposerFromAC(db, header, height, round)
```
piersy, line 171: @Jason-Zhangxin-Chen Why not inline the contents of `ac.callGetProposerFromAC` here ?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
 
     }
 
+    /*
+    * getProposer
```
piersy, line 437: Why is the function name duplicated in the first line of the comment?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+    * Implement a stateless weighted-round-robin proposer election from the committee which is in a consistent view
+    * across the nodes, the view is calculated by function computeCommittee() on each block finalization.
```
piersy, line 439: >the view is calculated by function computeCommittee() on each block finalization.

We should avoid writing comments that describe where a function is called from or in this case that the function is called at each block finalization. This is because where or how this function is called is out of scope for this function and it is very easy for the comment to become stale when the way that a function is called or where it is called from changes. Function comments should be concerned with what this function does, what parameters it takes and what result it returns.

piersy, line 439:
```suggestion
    * getProposer returns the address of the proposer for the given height and round. The proposer is 
    * selected from the committee via round robin with selection probability determined by the voting
    * power of each committee member. The selection mechanism is deterministic and will always select 
    * the same address, given the same height, round and contract state.
```

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+    */
+    function getProposer(uint256 height, uint256 round) public view returns(address) {
+        // if no committee set, return operator address.
+        if (committee.length == 0) {
+            return operatorAccount;
+        }
```
piersy, line 445: I don't think we need this here, I believe committees of zero length are not permitted, and we therefore cannot start up an autonity instance with a committee of zero length.

See the `Prepare` function of `AutonityContractGenesis`
https://github.com/clearmatics/autonity/blob/be89a8c72394696b61c59ffaaf17530ccabf5e4b/params/autonity_contract.go#L89-L91

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+        // calculate total voting power from current committee.
+        uint256 total_voting_power = 0;
+        for (uint256 i = 0; i < committee.length; i++) {
+            total_voting_power += committee[i].votingPower;
```
piersy, line 449: We should be using safe math here.
```suggestion
            total_voting_power.add(committee[i].votingPower);
```

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+        }
+
+        // fallback to round robin if total voting power is 0.
+        if (total_voting_power == 0) {
+            uint256 index = (height + round) % committee.length;
+            return committee[index].addr;
+        }
```
piersy, line 456: It feels to me like we shouldn't need this because we shouldn't allow validators without stake, is that allowed?

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+            // skip those members with 0 stake.
+            if (committee[i].votingPower == 0) {
+                continue;
+            }
+            counter += committee[i].votingPower;
+            if (index <= counter - 1) {
+                return committee[i].addr;
+            }
+        }
+    }
```
piersy, line 475: This is not weighted round robin, its weighted random sampling.

contracts/autonity/contract/test/test.js
```diff
@@ -582,4 +582,102 @@ contract('Autonity', function (accounts) {
...
+                let round;
+                for (round = 0; round < 2; round ++){
+                    let proposer = await token.getProposer(height, round);
+                    if (proposer === accounts[2] || proposer === accounts[4] || proposer == accounts[5]) {
+                        assert.fail("proposer is not expected.")
+                    }
+                }
+            }
+        });
+    });
```
piersy, line 682: There doesn't seem to be a test that checks that proposers are selected proportionally to their voting power.

piersy:
@Jason-Zhangxin-Chen I tried to request changes, but github won't allow me to request changes to my own pr :D

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+            uint256 index = (height + round) % committee.length;
+            return committee[index].addr;
+        }
+
+        // distribute seed into a 256bits key-space.
+        uint256 key = height + round;
+        uint256 value = uint256(keccak256(abi.encodePacked(key)));
+        uint256 index = value % total_voting_power;
+
+        //find the index hit which committee member which line up in a sorted committee list.
```
piersy, line 463: The committee list is not necessarily sorted, and it doesn't need to be sorted for this to work.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +433,47 @@ contract Autonity {
...
+        uint256 counter = 0;
+        for (uint256 i = 0; i < committee.length; i++) {
+            // skip those members with 0 stake.
+            if (committee[i].votingPower == 0) {
+                continue;
+            }
```
piersy, line 469: I think it would be neater to drop this condition, I would consider it premature optimisation.

Jason-Zhangxin-Chen: > I think it would be neater to drop this condition, I would consider it premature optimisation.

It filter out those validator with 0 stake, they shouldn't be scheduled as the proposer as there was a requirement.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > The committee list is not necessarily sorted, and it doesn't need to be sorted for this to work.

The deterministic scheduling need the members to have a same committee view on each height, I think the view includes the order of the memebers who were selected as the committee. If we don't promise the order or some bugs in the contract does not promise this ordering, it would break the following voting phases of tendermint.

contracts/autonity/contract/test/test.js
Jason-Zhangxin-Chen: > There doesn't seem to be a test that checks that proposers are selected proportionally to their voting power.

There was one, but it cause a long time period the collect the final proportions since it need time line to be extended as long as the better (infinitly long) to let the scheduling turn to infinitly close to the theorical value.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > It feels to me like we shouldn't need this because we shouldn't allow validators without stake, is that allowed?

Yes, it is not allow, but don't know if some expection or mistake happens can impact this schdueling, if it happens we cannot on-hold the network but to consider the fallback with round robin.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: @piersy , thanks. I was replace the safe add now. Please check from my last commit on this branch.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > I don't think we need this here, I believe committees of zero length are not permitted, and we therefore cannot start up an autonity instance with a committee of zero length.
> 
> See the `Prepare` function of `AutonityContractGenesis`
> https://github.com/clearmatics/autonity/blob/be89a8c72394696b61c59ffaaf17530ccabf5e4b/params/autonity_contract.go#L89-L91

Now this checking is removed.

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen I tried to request changes, but github won't allow me to request changes to my own pr :D

I already commit my fix.

contracts/autonity/contract/test/test.js
piersy: I would suggest checking that the proportions are adhered to within some tolerance.

E.G: If the network is of size 100 check that the proportions are adhered to within the bounds +/-10.

This test is a must have, without it we have no way of knowing if this works or not.

contracts/autonity/contract/contracts/Autonity.sol
piersy: I'm not suggesting that the order of members is not deterministic, I'm just saying that the list cannot be guaranteed to be *sorted*

https://github.com/clearmatics/autonity/blob/be89a8c72394696b61c59ffaaf17530ccabf5e4b/contracts/autonity/contract/contracts/Autonity.sol#L413-L425

contracts/autonity/contract/contracts/Autonity.sol
piersy: My point was that we don't need to filter them out because the the next expression will never return a validator with 0 stake.

Since adding 0 to the counter does not change the `if` condition. *If* the `if` condition actually did evaluate to true then it must have evaluated to true for the previous validator and therefore the previous validator would have been returned.

If this validator with 0 stake happens to be the first validator then it can't ever be returned because the index can never be less than 0 and in this case the counter will be -1.

https://github.com/clearmatics/autonity/blob/10dc9bae05790c44b11c8b542702c16c236fde9e/contracts/autonity/contract/contracts/Autonity.sol#L470-L473

contracts/autonity/contract/contracts/Autonity.sol
piersy: We shouldn't allow invalid states to exist in the system, if it is not allowed and it in fact happens this implies a programming error. Our response to programming errors should be to crash the system immediately since the only solution to a programming error is to change the code.

Having conditions like this delays us finding out about the programming errors, and in the meantime runs the risk that the system is allowed to continue functioning incorrectly, with the ultimate risk in this case that considerable real world value is mishandled or lost.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > I'm not suggesting that the order of members is not deterministic, I'm just saying that the list cannot be guaranteed to be _sorted_
> 
> https://github.com/clearmatics/autonity/blob/be89a8c72394696b61c59ffaaf17530ccabf5e4b/contracts/autonity/contract/contracts/Autonity.sol#L413-L425

Yes, the sorting is only for pick top N stakeowners of the validators who should form the PoS committee. In the getProposer function, I removed the comments.

contracts/autonity/contract/test/test.js
Jason-Zhangxin-Chen: > I would suggest checking that the proportions are adhered to within some tolerance.
> 
> E.G: If the network is of size 100 check that the proportions are adhered to within the bounds +/-10.
> 
> This test is a must have, without it we have no way of knowing if this works or not.

I pull back those 2 testcases that I implemented before, and now they are presented in the latest commit.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > My point was that we don't need to filter them out because the the next expression will never return a validator with 0 stake.
> 
> Since adding 0 to the counter does not change the `if` condition. _If_ the `if` condition actually did evaluate to true then it must have evaluated to true for the previous validator and therefore the previous validator would have been returned.
> 
> If this validator with 0 stake happens to be the first validator then it can't ever be returned because the index can never be less than 0 and in this case the counter will be -1.
> 
> https://github.com/clearmatics/autonity/blob/10dc9bae05790c44b11c8b542702c16c236fde9e/contracts/autonity/contract/contracts/Autonity.sol#L470-L473

0 stake would possible happens when redeem or transfer tokens happens. So skipping it would be valid.
When total stake of committee is 0, it would fallback to round robin and returned a round robin address.
At this if conditoin, it will always find one address at least since there is a none 0 stake validator in the list even if the first one is 0 stake validator. I update the comments of this part for a better understanding of it.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > We shouldn't allow invalid states to exist in the system, if it is not allowed and it in fact happens this implies a programming error. Our response to programming errors should be to crash the system immediately since the only solution to a programming error is to change the code.
> 
> Having conditions like this delays us finding out about the programming errors, and in the meantime runs the risk that the system is allowed to continue functioning incorrectly, with the ultimate risk in this case that considerable real world value is mishandled or lost.

Well, I agree we shouldn't allow invalid states to exist in the system. However we cannot make 100% sure that there isn't an invalid state happens during the run time. So to consider it in pessimism style, why not to check it with those extra condition for a better robustness.

contracts/autonity/contract/contracts/Autonity.sol
piersy: I understand that 0 stake is possible and that skipping committee members with 0 stake is a valid approach.

My  point is that these two code blocks are equivalent, and will produce exactly the same results for any set of validators with any stake values.

```go
        for (uint256 i = 0; i < committee.length; i++) {
            // skip those members with 0 stake.
            if (committee[i].votingPower == 0) {
                continue;
            }
            counter += committee[i].votingPower;
            if (index <= counter - 1) {
                return committee[i].addr;
            }
        }
    }
```

```go
        for (uint256 i = 0; i < committee.length; i++) {
            counter += committee[i].votingPower;
            if (index <= counter - 1) {
                return committee[i].addr;
            }
        }
    }
```

So given that they produce the same results I would go for the shorter one, since I think it is easier to understand.

contracts/autonity/contract/contracts/Autonity.sol
piersy: >why not to check it with those extra condition for a better robustness.

If we see that we have an invalid state, that means that we have a programming error. The only way to fix this is to stop the system fix the code and fix the state.

Allowing the system to continue running after we know that it contains programming errors is *not* robust. Because once you have reached an invalid state you cannot rely on any of the code in the system to work correctly any more, since when we write code, we assume some valid state for the code to act upon.

What I'm advocating here is failing fast [this](https://www.martinfowler.com/ieeeSoftware/failFast.pdf) article gives a more in depth view on why it is better to fail fast.

Jason-Zhangxin-Chen:
@piersy @yazzaoui , please review from the latest commit, now we downgrade the user type once redeem / send stake cause a 0 stake validator. Proposer selection side, we assume there are no 0 stake validator can be presented.

BTW, if validator send all stake to stakeholder / participants community, then it would possible lead to no validator in the network, and we will have no chance to produce block anymore. Everything will be on-hold and cannot be recovered.

I think our current user model system is not robust enough to be ready for production usecase.

contracts/autonity/autonity.go
piersy: @Jason-Zhangxin-Chen Still wondering about this.

contracts/autonity/contract/contracts/Autonity.sol
piersy: @Jason-Zhangxin-Chen  And also wondering about this

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +437,36 @@ contract Autonity {
...
+        // we assume there is no 0 stake/power validators.
+        uint256 counter = 0;
+        for (uint256 i = 0; i < committee.length; i++) {
+            counter = counter.add(committee[i].votingPower);
+            if (index <= counter - 1) {
+                return committee[i].addr;
+            }
+        }
+        // in case of any exceptions, to return the first validator for liveness.
+        return committee[0].addr;
```
piersy, line 467: We should not do this. This is the same problem I was trying to explain in https://github.com/clearmatics/autonity/pull/559#discussion_r445435551. If this happens then it means something has gone badly wrong, at this point we need to stop the system, so that we can fix the problem before allowing the system to continue. So at this point we need to crash.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -540,6 +574,20 @@ contract Autonity {
         emit Transfer(sender, recipient, amount);
     }
 
+    function _checkDowngradeUserType(address _address) internal {
+        require(_address != address(0), "address must be defined");
+        require(users[_address].addr != address(0), "user must exist");
```
piersy, line 579: I think it would be nicer to enforce this at a higher level, so make these checks at the beginning of the publicly available functions.

contracts/autonity/calls.go
```diff
@@ -191,6 +191,18 @@ func (ac *Contract) callGetMinimumGasPrice(state *state.StateDB, header *types.H
 	return minGasPrice.Uint64(), nil
 }
 
+func (ac *Contract) callGetProposerFromAC(state *state.StateDB, header *types.Header, height uint64, round int64) common.Address {
```
yazzaoui, line 194: Suggest renaming this to `callGetProposer` to keep it in line with the other functions.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -540,6 +574,20 @@ contract Autonity {
...
+
+        User memory u = users[_address];
+        if (u.stake != 0 || u.userType != UserType.Validator) {
+            return;
+        }
+
+        _removeFromArray(u.addr, stakeholders);
+        _removeFromArray(u.addr, validators);
+        usersList.push(u.addr);
```
piersy, line 588: I think the user is already in the userList so this push will create a duplicate.

contracts/autonity/contract/test/test.js
```diff
@@ -498,6 +498,40 @@ contract('Autonity', function (accounts) {
                 userTypes, stakes, commisionRate, operator, minGasPrice, bondPeriod, committeeSize, version,  { from:accounts[8]} );
         });
 
+        it('test user type downgrade on redeem stake', async function f() {
+            let initStake = 100;
+            await token.addValidator(accounts[8], initStake, freeEnodes[0], {from: operator});
+            await token.redeemStake(accounts[8], initStake, {from: operator});
+
+            let userType = await token.myUserType({from: accounts[8]});
+            assert (userType == 0, "wrong user type");
```
piersy, line 507: We should probably assert that the user type was something different before the call to redeemStake.

contracts/autonity/contract/test/test.js
```diff
@@ -498,6 +498,40 @@ contract('Autonity', function (accounts) {
...
+        });
+
+        it('test user type not downgrade on redeem stake', async function f() {
+            let initStake = 100;
+            await token.addValidator(accounts[8], initStake, freeEnodes[0], {from: operator});
+            await token.redeemStake(accounts[8], 50, {from: operator});
+
+            let userType = await token.myUserType({from: accounts[8]});
+            assert (userType == 2, "wrong user type");
```
piersy, line 516: I think it would be nice to make some variables for the userTypes rather than using ints.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -540,6 +574,20 @@ contract Autonity {
...
 
+    function _checkDowngradeUserType(address _address) internal {
+        require(_address != address(0), "address must be defined");
+        require(users[_address].addr != address(0), "user must exist");
+
+        User memory u = users[_address];
+        if (u.stake != 0 || u.userType != UserType.Validator) {
+            return;
+        }
+
```
piersy, line 585: We need to check here to make sure that this isn't the last user with stake, in the case that this is the last user with stake we need to revert, so that we don't end up in a situation where we have no stake assigned in the system.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -540,6 +574,20 @@ contract Autonity {
...
+        _removeFromArray(u.addr, stakeholders);
+        _removeFromArray(u.addr, validators);
+        usersList.push(u.addr);
+        users[_address].userType = UserType.Participant;
```
piersy, line 589: I think we should use changeUserType here.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -240,6 +242,8 @@ contract Autonity {
     */
     function send(address _recipient, uint256 _amount) external returns (bool) {
         _transfer(msg.sender, _recipient, _amount);
+        // todo check 0 balance, and downgrade user role.
+        _checkDowngradeUserType(msg.sender);
```
piersy, line 246: I think we should move this call into _transfer, we don't wan't to allow transferring stake and not updating the user type accordingly.

contracts/autonity/contract/contracts/Autonity.sol
piersy: I've updated the comment to not reference computeCommittee

Jason-Zhangxin-Chen: > I've updated the comment to not reference computeCommittee

Thank you, Piers.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > I think we should use changeUserType here.

changeUserType is only reserved for system operator, in the case of validator send stake to the others, it does not need a operactor to do so.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -433,6 +435,38 @@ contract Autonity {
...
+
+        // find the index hit which committee member which line up in the committee list.
+        // we assume there is no 0 stake/power validators.
+        uint256 counter = 0;
+        for (uint256 i = 0; i < committee.length; i++) {
+            counter = counter.add(committee[i].votingPower);
+            if (index <= counter - 1) {
+                return committee[i].addr;
+            }
+        }
```
piersy, line 467: What gets returned if the code reaches here?

contracts/autonity/contract/contracts/Autonity.sol
piersy: Good point, I still think we should use the logic of changeUserType here because it saves us duplicating the logic, and it will also mean that user type change events are emitted consistently. So I would suggest extracting the logic of changeUserType into a private function that does not restrict usage to the operator, and then calling that here and in the public changeUserType method.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -537,9 +571,23 @@ contract Autonity {
     function _transfer(address sender, address recipient, uint256 amount) internal canUseStake(sender) canUseStake(recipient) {
         users[sender].stake = users[sender].stake.sub(amount, "Transfer amount exceeds balance");
         users[recipient].stake = users[recipient].stake.add(amount);
+        // todo check 0 balance, and downgrade user role.
+        _checkDowngradeUserType(sender);
         emit Transfer(sender, recipient, amount);
     }
 
+    function _checkDowngradeUserType(address _address) internal {
+        require(validators.length > 1, "Downgrade user failed due to keep at least 1 validator in the network");
```
piersy, line 580: We should be checking here that the user identified by `_address` is a validator. They could just be a stakeholder.

contracts/autonity/contract/contracts/Autonity.sol
```diff
@@ -227,6 +227,8 @@ contract Autonity {
     function redeemStake(address _account, uint256 _amount) public onlyOperator(msg.sender) canUseStake(_account) {
```
piersy, line 227: I'm just wondering what is being redeemed for this stake here? To me it looks like this function is just burning stake.

Jason-Zhangxin-Chen: > I'm just wondering what is being redeemed for this stake here? To me it looks like this function is just burning stake.

Yes, redeem is burning stake from network for the time being.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > What gets returned if the code reaches here?

If there is no 0 stake validator, the code will not reach here:

The index is unsigned integer with a range [ 0, total_voting_power - 1],
As we don't allow to have 0 voting power validators, so the voting power of per validator 
would range at integers of [ 1,  total_voting_power], which it lets counter to have a range of unsinged integer [1, total_voting_power].
the if condition: (index <= counter -1) will always find a slot, otherwise it have to loop to next item.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > Good point, I still think we should use the logic of changeUserType here because it saves us duplicating the logic, and it will also mean that user type change events are emitted consistently. So I would suggest extracting the logic of changeUserType into a private function that does not restrict usage to the operator, and then calling that here and in the public changeUserType method.

changeUserType is too heavy, it was reserved for public usage by system operator, it remove the user from storage (user map, validator list, participant list, user list, enode list...) and recreate a brand new one into the storage again. It also did some redundant checking as we did in the redeem and send token method. checkDowngrade should be lite than this changeUserType.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > We should be checking here that the user identified by `_address` is a validator. They could just be a stakeholder.

Stakeholder doesn't need to hold stake, so I think that 0 stake stakeholder is valid for the time being. So the checking we apply is only stake is 0 for validator node can apply for the downgrade of user type, and make sure there is at least 1 validator in the network.

piersy: So this is the scenario I am imagining.

We have one validator, and one stakeholder, each with 5 stake.
The stakeholder calls send sending 1 stake to the validator.
_checkDowngradeUserType reverts the transaction because there is only 1 validator.

contracts/autonity/contract/contracts/Autonity.sol
piersy: Do you know why we named it redeem instead of burn?

contracts/autonity/contract/contracts/Autonity.sol
piersy: Imagine there was a bug in the code that manages the validators such that this function is called when there are no validators, what would this return in that case?

contracts/autonity/contract/contracts/Autonity.sol
piersy: Generally I would prioritise safety and simplicity over performance, until there is a real need for a performance gain. In this case I don't see a reason to optimise the performance now at the expense of the maintainability of this code.

Why do we need this to be so performant?

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > Do you know why we named it redeem instead of burn?

No Idea, it's logic is more likely with burning, but right now about the network economic, a lots of things is not clear and consistently declared yet.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > So this is the scenario I am imagining.
> 
> We have one validator, and one stakeholder, each with 5 stake.
> The stakeholder calls send sending 1 stake to the validator.
> _checkDowngradeUserType reverts the transaction because there is only 1 validator.

It make sense, I correct it in my last commit just to re-order the statement. You can check it if it fix this case.

contracts/autonity/contract/contracts/Autonity.sol
Jason-Zhangxin-Chen: > Generally I would prioritise safety and simplicity over performance, until there is a real need for a performance gain. In this case I don't see a reason to optimise the performance now at the expense of the maintainability of this code.
> 
> Why do we need this to be so performant?

In my understanding, just keep this since it is lite and also clear enough for maintainance.

consensus/tendermint/backend/backend_test.go
```diff
@@ -289,51 +286,6 @@ func TestCheckSignature(t *testing.T) {
...
-	addr, err := tendermintCrypto.CheckValidatorSignature(vset, data, sig)
-	if err != tendermintCrypto.ErrUnauthorizedAddress {
-		t.Errorf("error mismatch: have %v, want %v", err, ErrUnauthorizedAddress)
-	}
-	emptyAddr := common.Address{}
-	if addr != emptyAddr {
-		t.Errorf("address mismatch: have %v, want %v", addr, emptyAddr)
-	}
-}
-
```
aneequesafdar, line 336: As discussed in the meeting, this test should be moved to `tendermintCrypto` package.

consensus/tendermint/core/committee.go
```diff
@@ -0,0 +1,206 @@
...
+	return -1, *m, nil
+}
+
+// Get the round proposer
+func (w *weightedRoundRobinCommittee) GetProposer(round int64) types.CommitteeMember {
+
+	// If previous header was the genesis block then we will not yet have
+	// deployed the autonity contract so will take the proposer as the first
+	// defined validator of the genesis block.
+	if w.previousHeader.IsGenesis() {
```
piersy, line 186: Need to check to see if we are in block 1, because in block one the autonity contract has not been deployed yet.

consensus/tendermint/core/committee.go
```diff
@@ -0,0 +1,206 @@
...
+		sort.Sort(w.previousHeader.Committee)
+		return w.previousHeader.Committee[0]
```
piersy, line 188:
```suggestion
		return w.previousHeader.Committee[round%len(w.previousHeader.Committee)]
```

consensus/tendermint/core/committee_test.go
```diff
@@ -246,86 +223,69 @@ func TestSet_IsProposer(t *testing.T) {
...
-	assertNilError(t, err)
-
-	copiedSet := set.Copy()
-	if !reflect.DeepEqual(set, copiedSet) {
-		t.Fatalf("failed to correctly copy set, expected: %v and got: %v", set, copiedSet)
-	}
-}
-
-func TestSet_QandF(t *testing.T) {
+func TestSet_Q(t *testing.T) {
```
piersy, line 255: F needs to be added back in here

consensus/tendermint/core/core.go
```diff
@@ -293,7 +266,7 @@ func (c *core) startRound(ctx context.Context, round int64) {
 
 	// If the node is the proposer for this round then it would propose validValue or a new block, otherwise,
 	// proposeTimeout is started, where the node waits for a proposal from the proposer of the current round.
-	if c.isProposer() {
+	if c.committeeSet().GetProposer(c.Round()).Address == c.address {
```
piersy, line 272: lets add back the isProposer() method of core.

consensus/tendermint/core/precommit_test.go
```diff
@@ -489,16 +487,16 @@ func TestHandleCommit(t *testing.T) {
...
 	// block = block.WithSeal(h)
 
-	chainReaderMock := consensus.NewMockChainReader(ctrl)
+	// chainReaderMock := consensus.NewMockChainReader(ctrl)
 	backendMock := NewMockBackend(ctrl)
 	backendMock.EXPECT().LastCommittedProposal().MinTimes(1).Return(block, addr)
-	backendMock.EXPECT().BlockChain().Return(chainReaderMock)
-	chainReaderMock.EXPECT().GetHeaderByHash(gomock.Any()).Return(h)
+	// backendMock.EXPECT().BlockChain().Return(chainReaderMock)
+	// chainReaderMock.EXPECT().GetHeaderByHash(gomock.Any()).Return(h)
```
piersy, line 499: Should be removed

piersy:
This needs some tests that actually check that the weighted random sampling is working

consensus/tendermint/backend/backend_test.go
piersy: This test already exists in the crypto package, not sure how it ended up getting copied here.

https://github.com/clearmatics/autonity/blob/84a94e10c0a453c6eec9b0ca4c7219bd9ec02e2b/consensus/tendermint/crypto/check_signature_test.go#L15-L59

contracts/autonity/autonity.go
piersy: we put all calls in the autonity_calls file by convention

PR #563, opened 2020-06-01
Fix node baned at topology test

Jason-Zhangxin-Chen:
To fix issue: https://github.com/clearmatics/autonity/issues/431
The problem: Node cannot connect to network again during the topology changing test in local e2e test. Root cause is that in p2p network layer, there is a hardcoded time window (35 s) which limited the reconnect request if the last request was captured in that time window. While in our local e2e testing context, very testcase runs within tens of seconds. The topology checking was finished before the peer have chance to reconnect.
The solution: Make the time window configuarable for the local e2e testing framework, waiting for 35s is too long for local e2e testing. And this change does not impact the p2p network protocol from end to end point of view since the config is not exposed from e2e point of view.


.github/workflows/go.yml
```diff
@@ -375,3 +375,31 @@ jobs:
...
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Embed Autonity contract
+        if:  steps.cache-autonity-code.outputs.cache-hit != 'true'
+        run: make embed-autonity-contract
+
+      - name: TestTopology
+        run: go test ./consensus/test/... -v -run 'TestTendermintStarSuccess|TestTendermintBusSuccess|TestTendermintChangeTopologyFromStarToBusSuccess|TestTendermintAddValidatorsToTopologySuccess|TestTendermintAddParticipantsToTopologySuccess|TestTendermintAddStakeholdersToTopologySuccess' -timeout 40m
```
piersy, line 405: Missing newline here

piersy, line 405: These tests are already being run by the job tendermint_e2e_tests, we don't need another job.

p2p/dial.go
```diff
@@ -73,12 +73,13 @@ type dialstate struct {
...
-	lookupBuf     []*enode.Node // current discovery lookup results
-	static        map[enode.ID]*dialTask
-	hist          expHeap
+	start                 time.Time // time when the dialer was first used
+	lookupRunning         bool
+	dialing               map[enode.ID]connFlag
+	lookupBuf             []*enode.Node // current discovery lookup results
+	static                map[enode.ID]*dialTask
+	hist                  expHeap
+	dialHistoryExpiration time.Duration
```
piersy, line 82: It would be nice to have a comment here explaining what dialHistoryExpansion is

consensus/test/testnode_test.go
```diff
@@ -186,12 +188,15 @@ func (validator *testNode) startService() error {
 
 	validator.subscription = validator.service.BlockChain().SubscribeChainEvent(validator.eventChan)
 
-	if err := ethereum.StartMining(1); err != nil {
-		return fmt.Errorf("cant start mining %s", err)
-	}
+	// stakeholder or participant shouldn't start mining since errUnauthorized would rise.
```
piersy, line 191: I think this comment needs more elaboration. Why would anyone be mining in a BFT system?

consensus/test/testcase_test.go
```diff
@@ -527,7 +527,9 @@ func (t *Topology) ConnectNodesForIndex(index string, nodes map[string]*testNode
 	if err != nil {
 		return err
 	}
-	if _, ok := ch[blockNum]; !ok {
+	_, dropOk := ch[blockNum]
+	_, connectOk := ch[blockNum-1]
+	if !(dropOk || connectOk) {
```
piersy, line 532: This could do with some explanation, looking at this code I have no idea what this logic does or why

consensus/test/testcase_test.go
```diff
@@ -542,22 +544,26 @@ func (t *Topology) ConnectNodesForIndex(index string, nodes map[string]*testNode
...
-		if _, ok := graphConnections[k]; ok {
-			continue
+	if dropOk {
+		for k := range currentConnections {
+			if _, ok := graphConnections[k]; ok {
+				continue
+			}
+			fmt.Println("node", index, "removes to", k)
+			nodes[index].node.Server().RemovePeer(nodes[k].node.Server().Self())
+			nodes[index].node.Server().RemoveTrustedPeer(nodes[k].node.Server().Self())
```
piersy, line 554: This logic is also unclear to me.

consensus/test/testcase_test.go
```diff
@@ -542,22 +544,26 @@ func (t *Topology) ConnectNodesForIndex(index string, nodes map[string]*testNode
...
-		if _, ok := currentConnections[k]; ok {
-			continue
+	if connectOk {
+		for k := range graphConnections {
+			if _, ok := currentConnections[k]; ok {
+				continue
+			}
+			fmt.Println("node", index, "connects to", k)
+			nodes[index].node.Server().AddPeer(nodes[k].node.Server().Self())
+			nodes[index].node.Server().AddTrustedPeer(nodes[k].node.Server().Self())
```
piersy, line 565: And this too. If I knew what dropOk or connectOk meant then I might have a chance.

.github/workflows/go.yml
Jason-Zhangxin-Chen: Yes, I fix all the review findings, please check them from my last commint.

consensus/test/testcase_test.go
Jason-Zhangxin-Chen: Yes, this is chance to move forward to the peer connect or disconnect. 
I remove it since it do make some understanding difficulties. Then in any chance, it will try to add/remove peer connections by according to the pre-define topolgy graph. Now it is much better.

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen Actually, it looks like I was too hasty in giving my approval, the tests are failling!

I just notice that those jobs were canceled for some reason. I am merging the develop into the branch, let's see how about this time.

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen Actually, it looks like I was too hasty in giving my approval, the tests are failing!

Those testcase were cancell by some reason. By the latest merge and comment fix, it seems to be running good.

consensus/test/testcase_test.go
```diff
@@ -63,6 +63,13 @@ type testCase struct {
 	noQuorumTimeout      time.Duration
 	topology             *Topology
 	skipNoLeakCheck      bool
+
+	// To apply loose TX error check in those disaster test cases. For example, pending TX might happens when
+	// the recover node not yet complete the round state synchronization from peers, while the other nodes already come
+	// to the end chain height defined by the test case. Following with block sync, the recover node hook function keep
+	// issuing TXs into the TX memory pool. In such case, we check the engine is not on-hold and move to the final chain
+	// height defined by the test case, pending TX shouldn't be a error to cause test failure.
+	looseTXCheck bool
```
piersy, line 72: Hi @Jason-Zhangxin-Chen, could you try and explain this again, I can't understand what this is doing.

Specifics that I don't understand:

>  those disaster test cases

Which test cases are these?

> pending TX

What is this?

> the recover node

What is the recover node? This is a global option so it doesn't really make sense to talk about specific nodes from specific test cases here.

consensus/test/testnode_test.go
```diff
@@ -186,12 +188,16 @@ func (validator *testNode) startService() error {
 
 	validator.subscription = validator.service.BlockChain().SubscribeChainEvent(validator.eventChan)
 
-	if err := ethereum.StartMining(1); err != nil {
-		return fmt.Errorf("cant start mining %s", err)
-	}
+	// None validator node shouldn't mine and submit candidate blocks to consensus engine since they are not
+	// the member of committee set, they are un-authorized. It fix the un-authorized error in the mining thread.
```
piersy, line 192:
```suggestion
	// Only validator nodes should mine bocks.
```

consensus/test/testnode_test.go
```diff
@@ -186,12 +188,16 @@ func (validator *testNode) startService() error {
...
+	if strings.HasPrefix(validator.Name, ValidatorPrefix) {
+		if err := ethereum.StartMining(1); err != nil {
+			return fmt.Errorf("cant start mining %s", err)
+		}
 
-	for !ethereum.IsMining() {
-		time.Sleep(50 * time.Millisecond)
+		for !ethereum.IsMining() {
```
piersy, line 198:
```suggestion
		// Wait for eth node to start mining
		for !ethereum.IsMining() {
```

consensus/test/testcase_test.go
Jason-Zhangxin-Chen: > Hi @Jason-Zhangxin-Chen, could you try and explain this again, I can't understand what this is doing.
> 
> Specifics that I don't understand:
> 
> > those disaster test cases
> 
> Which test cases are these?
>  For example those two https://github.com/clearmatics/autonity/runs/585426726
https://github.com/clearmatics/autonity/runs/572227361, failed due to pending TX happens.
> > pending TX
> 
> What is this?
>  pending TX means there are un-mined TX in the local TX memory pool.
> > the recover node
> 
> What is the recover node? This is a global option so it doesn't really make sense to talk about specific nodes from specific test cases here.

It should be the recovering node,  I update the code comments in the latest commit.
For example, in a network with: A, B, C, D, E as validators. They are producing blocks for a while, and we shut down E for a while to get E have a chain gap with the alive network, and then we shut down D to let the chain on hold.
And then we start E, E would first sync block from alive nodes,  it would also sync consensus state (round state) from alive validator peers to meet the quorum num of nodes again to produce block, finally they will get to the end height defined in the testcase. During that period we start D, D would do the same sync as E, beside the sync, E also keep issuing TXs as we implemented in the test framework while the other nodes already come to the end of test, letting the TX un-mined in E's TX memory pool.

piersy: Hi @Jason-Zhangxin-Chen 

I'm still unclear as to what are the disaster test cases? But in general comments should not make reference to things outside of the scope of the code they are describing, so if by "those disaster test cases" you mean some specific e2e tests then this commenting is liable to become stale (when "those disaster test cases" are changed/removed ... etc)

> those disaster test cases

I think I understand what `looseTXCheck` does now. I could summarise my understanding as "it disables checking to see if transactions are processed". I if that is the case, why not just stop sending transactions, rather than sending transactions and not checking them?

However the rest of your description still does not make sense to me.

>	// In those case for example stop and start nodes, pending TX might happens when the recovering node not yet
	// complete the round state synchronization from peers to make the engine have quorum members to mine block again,
	// during the time, the other nodes already come to the end chain height defined by the test case and the recovering
	// node is keep issuing TXs into the TX memory pool which cause pending transactions un-mined. In such case, we
	// check the engine is not on-hold and move to the final chain height defined by the test case, pending TX shouldn't
	// cause the test case failure.

TXs should be broadcast to other nodes irrespective of the round state synchronisation. And if those TXs are broadcast to the rest of the network then they should be processed.

Jason-Zhangxin-Chen:
Close this pull request with new one opened: https://github.com/clearmatics/autonity/pull/661 base on latest upstream's refine.

PR #564, opened 2020-06-01
Enforce init before node run

### LINKED ISSUE #473: Autonity client should exit if started without config file and database
# Problem

It's possible to run the autonity client without a config file and without a db, this should be forbidden.

# Acceptance Criteria

When run without a config file and db the autonity client should exit with a non 0 exit code and display a message explaining that a config file argument must be provided in order for it to function correctly.

piersy:
This PR makes sure that you cannot run an autonity node without first running the autonity `init` sub-
command.

This is achieved by checking for the existence of files in the chaindata dir when running autonity and exiting if it doesn't exist. Running the init sub-command will populate the chaindata dir.

Fixes #473


cmd/autonity/run_test.go
```diff
@@ -49,6 +49,13 @@ type testautonity struct {
 func init() {
 	// Run the app if we've been exec'd as "autonity-test" in runAutonity.
 	reexec.Register("autonity-test", func() {
+
+		// if len(os.Args) > 3 {
```
Jason-Zhangxin-Chen, line 53: Remove these comments from this init() if we don't need it now, since it might confuse people why we put such logics and commented it...

piersy: Oh thanks, Jason. This is currently a draft since the tests are failing and I have not fixed that yet. So I will need to re-request your review when it stops being a draft.

piersy:
Closing this as I now have this work on another branch.

Issue #504, opened 2020-04-06
github actions cache usage allows invalid cache keys to be used


piersy:
# Problem

It looks like our github actions caching is configured to allow use to restore code from a cache that does not relate to the branch being processed. Obviously we wan't to only be testing the code of the  branch which was pushed, not other branches.

From the [docs](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action)
>You can optionally provide a list of restore-keys to use when the key doesn't match an existing cache. A list of restore-keys is useful when you are restoring a cache from another branch because restore-keys can partially match cache keys. For more information about matching restore-keys, see "Matching a cache key."

In all our uses of caching we provide the following restore key:
https://github.com/clearmatics/autonity/blob/62e991c62d4e5c5747159478b6734141d3d68ad7/.github/workflows/go.yml#L15

Which looks like it allows restoring a cache from any branch, if you consider that our full cache key definition is:
https://github.com/clearmatics/autonity/blob/62e991c62d4e5c5747159478b6734141d3d68ad7/.github/workflows/go.yml#L14 

# Implementation

Remove all use of restore keys.

# Acceptance Criteria

Github actions for a push  should only be run with the code of the commit that was pushed.

JekaMas:
It will break linters.

JekaMas:
Please check any CI results, eg https://github.com/clearmatics/autonity/pull/472/checks?check_run_id=564564727

You can check that it restores the code that is needed.

piersy:
I don't understand your point @JekaMas?

The run you referenced seems to exhibit the behaviour that I'm concerned about. That being different code caches being restored.

The bootstrap job restores this cache key:
```
Linux-aut-cache-autonity-code-3d867b8e765c09ed01725668f130eb18e039946a
```
And the contract tests job restores this cache key:
```
Linux-aut-cache-autonity-code-907b9c5a16f87fcb6a6816109bbafd25260fc908
```

There is no guarantee that a cache is going to be available, which is why the restore keys are provided as a fallback.

Looking a bit deeper though I see that restore keys are only used when a cache miss is encountered, and since we always check out the code in the case of a cache miss, we will end up overwriting code retrieved from an invalid cache anyway. This will work out most of the time, but could cause a failure in the case that we restore an old cache that contains code that has since been deleted and now may not compile.

piersy:
this has been fixed by #550

Issue #496, opened 2020-03-31
Fix flaky tests


piersy:
# Problem

The following tests fail randomly. I provide links to the failures all the links are related to commit c37a10a. It can be seen that all of these tests have passed in some of the runs.

```
TestContractUpgradeSeveralUpgrades
https://github.com/clearmatics/autonity/runs/588448483
https://github.com/clearmatics/autonity/runs/574380765
https://github.com/clearmatics/autonity/runs/574671480

TestTendermintStartStopFPlusTwoNodes
https://github.com/clearmatics/autonity/runs/585426726
https://github.com/clearmatics/autonity/runs/572227361

TestContractUpgradeSeveralUpgradesOnStarTopology
https://github.com/clearmatics/autonity/runs/573461126
https://github.com/clearmatics/autonity/runs/574047422

TestTendermintStarOverParticipantSuccess
https://github.com/clearmatics/autonity/runs/573651085
https://github.com/clearmatics/autonity/runs/574047654

TestTendermintStarOverParticipantSuccessWithExternalUser
https://github.com/clearmatics/autonity/runs/585096427
```

# Acceptance Criteria

Time successful runs of each test, then run 100 iterations (`-count 100`) of each test with a timeout of 2.5*<successful_run_time>*100. We should not see any failures. 

Provide the commands to run the tests in the PR so that reviewers can easily check for themselves.

 We set the timeout 2.5 times higher than the time for a successful run since that seems like a reasonable amount of leeway to allow for test time fluctuations.

JekaMas:
2500s ?

piersy:
2500s is calculated by allowing 25s per test run and having a count of 100. My understanding is that the timeout flag provides a global timeout, not a per run timeout.

PR #565, opened 2020-06-02
Docker e2e test framework

Jason-Zhangxin-Chen:
Integrate the docker e2e test for CI base on latest develop branch.
Please review it, there is no changes in the client side, the testing take the client as a blackbox.


Jason-Zhangxin-Chen:
@piersy , when you review it, please also help to double check if those autogen files would be impacted. Thank you so much.

Makefile
```diff
@@ -102,6 +102,10 @@ test-contracts:
 	done
 	@cd contracts/autonity/contract/ && $(NPMBIN)/truffle test && cd -
 
+docker-e2e-test: embed-autonity-contract
+	build/env.sh go run build/ci.go install
```
piersy, line 106: If your intent here is to build the autonity binary you can just depend on `autonity` like this, autonity depends on `embed-autonity-contract` so that will also be run.

```suggestion
docker-e2e-test: autonity
```

.github/workflows/go.yml
```diff
@@ -369,3 +369,65 @@ jobs:
...
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-autonity-bin
+        with:
+          path: /home/runner/work/autonity/autonity/build/bin
+          key: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
+
+      - name: docker_e2e_test
+        run: make docker-e2e-test
```
piersy, line 433: You are missing a newline on this file, missing newlines interfere with commandline tools like grep which works on lines.

You should be able to configure your editor to always ensure a newline at the end of a file. There is a setting in goland `Ensure line feed at file end on Save` that looks like it does this.

.github/workflows/go.yml
```diff
@@ -369,3 +369,65 @@ jobs:
...
+        if: steps.cache-autonity-code.outputs.cache-hit != 'true'
+        with:
+          clean: 'false'
+
+      - name: Embed Autonity contract
+        if:  steps.cache-autonity-code.outputs.cache-hit != 'true'
+        run: make embed-autonity-contract
+
+      - name: Build
+        run: make all
```
piersy, line 409: What else do you need to make apart from the autonity binary?

docker_e2e_test/README.md
```diff
@@ -0,0 +1,35 @@
+# ATAutonity
```
piersy, line 1: Is this a typo? If not ATAutonity doesn't mean anything to me.

docker_e2e_test/README.md
```diff
@@ -0,0 +1,35 @@
...
+If your linux is on ubuntu-18.04, the script will auto install it for you.
+`sudo apt-get install --yes docker.io`
+
+# How to use it
+## block chain genesis configuration
+In the docker_e2e_test/planner/networkplanner.py + line: 122, there are default genesis configuration, update it on your demand.
+
+## run it
+You can run it via make:
+`make docker-e2e-test`
```
piersy, line 27: The readme seems to be missing a section that explains exactly what this does. So clearly it runs e2e tests, but what tests? What does this offer that is not provided by the tests under `consensus/test` ?

.github/workflows/go.yml
```diff
@@ -369,3 +369,65 @@ jobs:
...
+        run: sudo apt-get update
+
+      - name: Install Python3
+        run: sudo apt-get install --yes python3
+
+      - name: Install PIP3
+        run: sudo apt-get install --yes python3-pip
+
+      - name: Install 3rd party lib
+        run: sudo pip3 install -r /home/runner/work/autonity/autonity/docker_e2e_test/requirements_docker_test.txt
```
piersy, line 421: Why are you installing the 3rd party packages with sudo?

Makefile
```diff
@@ -102,6 +102,10 @@ test-contracts:
...
+	cd docker_e2e_test && sudo python3 test_via_docker.py ../build/bin/autonity
```
piersy, line 107: Why are you executing the tests with sudo?

Jason-Zhangxin-Chen: > Why are you executing the tests with sudo?

In my testbed, I have to use sudo, otherwise the SW installation might be not allow. I think this is depends on the OS setting. We can try to remove SUDO for CI workers.

.github/workflows/go.yml
Jason-Zhangxin-Chen: > What else do you need to make apart from the autonity binary?

the bootnode binary is also need for the chain setup, so here just make all of those stuffs.

docker_e2e_test/README.md
Jason-Zhangxin-Chen: Hi @piersy , yes when port this e2e framework, I forget to sync all of those README from ATAutonity's better README, now it should be much better than before.
Please check the latest commit.

piersy: Thanks for adding the readme, it does clarify some of my questions, however I'm still not convinced that adding this test framework would actually be useful to us right now.

Looking through the testcases provided by this project It looks like a lot of them are duplicating cases covered by the tests that exist under 'consensus/test'

So the following sets of tests look roughly equivalent.

---
```
Happy case, send 10 token per TX last for 1 minutes.
```
```
TestTendermintSuccess
TestTendermintSlowConnections
TestTendermintMemoryLeak
TestTendermintLongRun
TestTendermintTC7
```
---
```
1 node crashed, and recover it again, not on-holding.
With 2 node crashed, N = 6, Q = ceil(2xN/3) = 4, not on-holding.
With 2 node crashed, 1 node back, after the back, consensus should not on-holding.
With 2 node crashed, 2 node back, after the back, consensus should not on-holding.
With 3 node crashed, 3 of 6 crashed, system on-hold.
With 3 node crashed, 1 node recover, consensus engine should not on-hold.
With 3 node crashed, 2 node recover, consensus engine should not on-hold.
With 3 node crashed, 3 node recover again, consensus engine should not on-hold.
```
```
TestTendermintStopUpToFNodes
TestTendermintStartStopSingleNode
TestTendermintStartStopFNodes
TestTendermintStartStopFPlusOneNodes
TestTendermintStartStopFPlusTwoNodes
TestTendermintStartStopAllNodes
```
---
```
6 nodes in Ring network topology, should not on-hold.
6 nodes in bus line network topology, should not on-hold.
6 nodes in star network topology, should not on-hold.
```
```
TestTendermintStarSuccess
TestTendermintBusSuccess
```
---
These tests look like a combination of the stop start tests we have and the topology tests we have under `consensus/test`
```
6 nodes in star network topology, restart the central node, should not on-hold.
6 nodes in star network topology, restart the central node with 1s traffic control should not on-hold.
```

That leaves the following tests that are not covered by tests under `consensus/test`.

```
network traffic control with package delay, loss, duplication, reordering, corruption.
N = 6, Q = 4, let only 4 validators running with one of them be applied with heavy traffic control.

Network partition. 6 break into 2 slices with 3 node each slice, engine should on-hold.
Network partition. 6 break into 3 and 3, then the two slices recovers, engine recovers too.
Network partition. 6 break into 4 and 2, main party(4 validator) should not on-hold.
Network partition. 6 break into 3 and 3, then the two slices connected by bridge, engine recovers too.
Network partition. 6 break to 3 full mesh plus 1 node, and 2 nodes of bus line, recover with 3 nodes of full mesh plus a 3 nodes of bus line.

Bootstrap from scratch, 2 by 2 on 5 seconds interval
Bootstrap from scratch, with all node start in the same time.
Bootstrap from scratch with 1 node crashed and recover later during the bootstrap.
Bootstrap from scratch with 2 nodes crashed and 2 nodes recover later during the bootstrap.
Bootstrap from scratch with 3 node crashed and 1 node recover later during the bootstrap. Consensus should not no-hold.
Bootstrap from scratch, start n1 to n4 for mining, shutdown all then to start n3,n4,n5,n6. Overlapped n3, n4 like: <<n1, n2, {{n3, n4>> , n5, n6}}
```

I'd like to understand what is the meaning of bootstrap here? Possibly those tests are covered by tests under `consensus/test` but I can't know without a better understanding of bootstrap.

So that leaves for sure  a set of tests that create network partitions and apply traffic control that are not covered by the tests under `consensus/test`.

I think we do obviously want to test situations involving network partitions and with traffic control, but I think we can do this in go much more easily than we can with this framework.

If you think about low level methods that are used in autonity for connecting to other nodes and sending/reciving data they are the following methods from the net package.

```
func Dial(network, address string) (Conn, error)

type Conn interface {
    // Read reads data from the connection.
    // Read can be made to time out and return an Error with Timeout() == true
    // after a fixed time limit; see SetDeadline and SetReadDeadline.
    Read(b []byte) (n int, err error)

    // Write writes data to the connection.
    // Write can be made to time out and return an Error with Timeout() == true
    // after a fixed time limit; see SetDeadline and SetWriteDeadline.
    Write(b []byte) (n int, err error)

    // Close closes the connection.
    // Any blocked Read or Write operations will be unblocked and return errors.
    Close() error
```

Any network failure, whatever it is can only manifest itself as an error returned from one of these methods. That means we can simulate all network errors just by providing our own implementation or wrapping of these methods and returning an error.

When it comes to slow networks, delays, packet loss ... etc we can simulate these problems by delaying message sending, dropping some messages instead of sending them or sending garbage data.

We can do this in go without the need for a whole other test framework.

Having to maintain two frameworks means twice as much work and unless this framework was really offering something that was not achievable in our current framework I don't think we can justify the extra development cost. Not only is it a whole new framework to maintain but it also falls short of our current framework in several ways.

* It takes hours to run - you can't run single tests.
* Its in a different language, and requires a whole new tool-set to work with it.
* Its not debuggable.
* It uses docker which adds a further layer of complexity to understanding what actually happened in the tests.

So unless there is something I'm missing, I really suggest that we do not add this to the project. It seems to me we want to make some quite big changes to the structure of the project going forward and I think adding this will slow that process down a lot and only provide benefit that we could more easily achieve by modifying our existing test framework.

Jason-Zhangxin-Chen: Hi @piersy, actually this test framework was built before those ones implemented in golang. The difference is that the docker e2e test framework is taking the client as a blackbox, at a higher level testing which is closer to the end to end operation. In the history, it do address bugs which was not addressed in the local e2e test framewok built in golang. For example the round state recovery on the node start up phase. It looks like a higher level testing which stand on the view of network operator's. I would not suggest to remove those testcases at a different testing level. 
About the bootstrap from scratch testing, it is about the re-deploy the network from brand new chain state, which is so important for a distributed systems for state/view exchange on the start up phase, to make sure if the network start up in any ordering of the node IN/OUT sequence is successfully done, and the mining is correctly proceeded.
I would say that we can keep those testcases to make sure these test scenarios work from real end to end point of view base on the network operation point.

piersy: Hi @Jason-Zhangxin-Chen I'm not sure I fully understand your response. I'll try to clarify bit by bit.

> The difference is that the docker e2e test framework is taking the client as a blackbox, at a higher level testing which is closer to the end to end operation.

This is clear but we can do black box testing in golang as well, I don't see any need to create a separate framework in a separate language in order to achieve this and having our black box testing in go means that the tests don't suffer from the disadvantages I laid out above.

> I would not suggest to remove those testcases at a different testing level.

If you are saying that we should have black box tests I agree, but I think they should be done in golang. I don't believe that this framework gives us much value. How do I measure value in a test framework?

What is the purpose of testing? Ultimately the purpose of testing is to allow you to build software free of defects. In order to help you do that you need tests that do two things.

* Fail when something is broken.
* Let you know what is broken and how to fix it.

A test that only fails but doesn't let you know what is broken and how to fix it is not useful. For example the flaky tests (#496) that you are working on right now. You can't reproduce the failures on your local system and all you have is the logs of the failures, its difficult to understand what is wrong and the consequence of that is you will need to spend a lot of time trying to figure out what needs fixing (these tests are not particularly useful). Contrast that with a test that when it fails you immediately know what needs to change in the code to fix it, that is the best type of test.

In order to have tests that let you know what is broken and how to fix it, it is important to have a short feedback loop(the time between something breaking and the time between finding out about it). There are some development environments that have optimised testing to the point that they run all the tests for each character typed. This means that the engineer has instant feedback when they break anything, this is extremely useful. The shorter the feedback loop the less specific the tests need to be about what went wrong, because the engineer will have changed less since the last testing feedback, this allows the engineer to narrow the scope of their search for bugs, which allows them to find and fix the bug quicker. Long feedback loops slow development not only by making it hard to find out where bugs originated but also because while an engineer is waiting for results from a test they are effectively blocked from making more changes on top of their latest change while they wait to see if their latest change fixed the problem. They don't want to change too much between runs because it will make it harder to pinpoint what the fix was, in the case that their changes did result in a fix.

The feedback loop for this test framework is extremely long. It takes 1hr 15m to run on the ci(and something like 2 hours on my local machine), this feedback loop will slow development and make it very hard to pinpoint the source of an error. Not only that but the framework offers very little help in identifying errors, when it fails it provides logs in a compressed file, but we can't access that file from ci.

I'm imagining the sequence of steps that I would need to take to fix the code when a test fails in this framework?

It could look something like this.

----

So I've been working away and pushing my branch regularly, maybe 5 times in the last hour(its 10am). 

* I think the work is done, my latest pushes are queued in CI, so I cancel previous runs that haven't finished running the e2e tests, because I wan't feedback on my latest state, quickly.
* I come back in an hour and a half and find that a test failed in the e2e tests. Now I'm not sure if this is something I introduced in my earlier commits or the latest commit, because I cancelled the runs for the earlier commits. So I kick of the first of the five runs again.
* I  come back in an hour and a half. Its now (1 pm). That run passed, so I go for a binary search and pick the middle of the 5 commits and kick of the CI for that.
* I come back in an hour and a half, this time the run failed, but I can't know if it was that commit or the previous commit that introduced the bug, so I run CI for the commit before.
* I come back in an hour and a half (its now 4pm) and see that it has failed. I've identified the commit with the bug (in 6 hours) but I have no logs or any idea what the problem is.
* So now I move to my local machine to run the tests again so I can retrieve the logs.
* The tests take 2.5 hours on my local machine, I now have logs covering the runs where something went wrong, its the end of the day (6:30pm)
* The next day at 9am I investigate the logs, after a couple of hours of investigation I have an idea for a potential fix, and make the change.
* I wait a further 2.5 hours for the test results, the test is still failing, its 1:30pm. I tweak some more things and set off the tests again.
* After 2.5 more hours I still have failing tests so I tweak some more things and set the tests off to run again.
* This run succeeds, its 6:30pm, I've fixed the bug. It took 18 hours or 2 days.

In the time it took me to fix the bug found in this framework, I could probably have written an equivalent test in go.

The go test will run faster, can be run individually, can be debugged, produces visible logs whether run on CI or locally and doesn't require a whole new set of tools to work with. It will also save the next person to encounter this problem roughly 18hours or 2 days of work.

This scenario I think shows why adding this test framework to our project will not be helpful.

piersy: Hi @Jason-Zhangxin-Chen in regard of your second point.

>About the bootstrap from scratch testing, it is about the re-deploy the network from brand new chain state, which is so important for a distributed systems for state/view exchange on the start up phase, to make sure if the network start up in any ordering of the node IN/OUT sequence is successfully done, and the mining is correctly proceeded.
I would say that we can keep those testcases to make sure these test scenarios work from real end to end point of view base on the network operation point.

Are you saying that we run tests were we begin running nodes from a genesis file (brand new chain state) and we try starting nodes in different orders?

If so I still don't see any reason preventing us from having those tests in our existing framework in golang.

piersy:
@Jason-Zhangxin-Chen So I tried running the framework locally and a testcase failed, I can't see any indication of what went wrong.

This is the output in my terminal showing the failure.

[docker_e2e_tests_failure.log](https://github.com/clearmatics/autonity/files/4771531/docker_e2e_tests_failure.log)

These are the logs that were provided in a tar file.

[logs.zip](https://github.com/clearmatics/autonity/files/4771555/logs.zip)

I an imagine spending days or weeks trying to figure out what went wrong here, granted I have not looked too hard here, because I really don't want to spend days figuring out what went wrong.

Maybe you can have a look and see what you think the problem is.

docker_e2e_test/README.md
piersy: @Jason-Zhangxin-Chen So I tried running the framework locally and a testcase failed, I can't see any indication of what went wrong.

This is the output in my terminal showing the failure.

[docker_e2e_tests_failure.log](https://github.com/clearmatics/autonity/files/4771531/docker_e2e_tests_failure.log)

These are the logs that were provided in a tar file.

[logs.zip](https://github.com/clearmatics/autonity/files/4771555/logs.zip)

I an imagine spending days or weeks trying to figure out what went wrong here, granted I have not looked too hard here, because I really don't want to spend days figuring out what went wrong.

Maybe you can have a look and see what you think the problem is.

Jason-Zhangxin-Chen:
> I left a few comments/questions, I also have a PR onto this one but lets wait till it builds successfully before we consider it.

Hi Piers, thank you so much for your comments. 
But I cannot totally agree with your idea. As the python test framework is a higher level testing than the one implemented in golang. When you explore blockchain project like BTC, DASH and PIVX and EOS, you will find that there is python based framework co-exist with those C++ repo to perform the end to end testing taking the client as a blackbox without using any internal implementation details include the P2P layer's feature like dials etc... 

From end to end testing point of view, I believe golang can do so, but python is more cost-efficient than golang.

The so called local e2e testing framework implemented with golang, it is powerfull too, but it does not take the entire autonity as a blackbox, it is not kind of end to end testing at least from my point of view. It looks like a component integration testing framework of autonity client. I cannot 100% sure that if the golang e2e test can address all the bugs before we roll out the release. And this is why we built ATAutoniy (the python base e2e testing framework) at the very beginning, trying to experience the features from real e2e point of view.

About the debuging of failed testcase in python e2e testframework, I agree that this might be harder than those one built via golang since we have to switch the context from different node's logs. This is a part that we debuging and maintain the produciton network from end to end point of view, the framework collects test context log include time/height of the test start and end, steps of the event scheduled in the test also. It could help to faster the debuging.

If customer support teammates say we don't need such test scenarios, then I will remove those testcases.

docker_e2e_test/README.md
Jason-Zhangxin-Chen: > Hi @Jason-Zhangxin-Chen I'm not sure I fully understand your response. I'll try to clarify bit by bit.
> 
> > The difference is that the docker e2e test framework is taking the client as a blackbox, at a higher level testing which is closer to the end to end operation.
> 
> This is clear but we can do black box testing in golang as well, I don't see any need to create a separate framework in a separate language in order to achieve this and having our black box testing in go means that the tests don't suffer from the disadvantages I laid out above.
> 
> > I would not suggest to remove those testcases at a different testing level.
> 
> If you are saying that we should have black box tests I agree, but I think they should be done in golang. I don't believe that this framework gives us much value. How do I measure value in a test framework?
> 
> What is the purpose of testing? Ultimately the purpose of testing is to allow you to build software free of defects. In order to help you do that you need tests that do two things.
> 
> * Fail when something is broken.
> * Let you know what is broken and how to fix it.
> 
> A test that only fails but doesn't let you know what is broken and how to fix it is not useful. For example the flaky tests (#496) that you are working on right now. You can't reproduce the failures on your local system and all you have is the logs of the failures, its difficult to understand what is wrong and the consequence of that is you will need to spend a lot of time trying to figure out what needs fixing (these tests are not particularly useful). Contrast that with a test that when it fails you immediately know what needs to change in the code to fix it, that is the best type of test.
> 
> In order to have tests that let you know what is broken and how to fix it, it is important to have a short feedback loop(the time between something breaking and the time between finding out about it). There are some development environments that have optimised testing to the point that they run all the tests for each character typed. This means that the engineer has instant feedback when they break anything, this is extremely useful. The shorter the feedback loop the less specific the tests need to be about what went wrong, because the engineer will have changed less since the last testing feedback, this allows the engineer to narrow the scope of their search for bugs, which allows them to find and fix the bug quicker. Long feedback loops slow development not only by making it hard to find out where bugs originated but also because while an engineer is waiting for results from a test they are effectively blocked from making more changes on top of their latest change while they wait to see if their latest change fixed the problem. They don't want to change too much between runs because it will make it harder to pinpoint what the fix was, in the case that their changes did result in a fix.
> 
> The feedback loop for this test framework is extremely long. It takes 1hr 15m to run on the ci(and something like 2 hours on my local machine), this feedback loop will slow development and make it very hard to pinpoint the source of an error. Not only that but the framework offers very little help in identifying errors, when it fails it provides logs in a compressed file, but we can't access that file from ci.
> 
> I'm imagining the sequence of steps that I would need to take to fix the code when a test fails in this framework?
> 
> It could look something like this.
> 
> So I've been working away and pushing my branch regularly, maybe 5 times in the last hour(its 10am).
> 
> * I think the work is done, my latest pushes are queued in CI, so I cancel previous runs that haven't finished running the e2e tests, because I wan't feedback on my latest state, quickly.
> * I come back in an hour and a half and find that a test failed in the e2e tests. Now I'm not sure if this is something I introduced in my earlier commits or the latest commit, because I cancelled the runs for the earlier commits. So I kick of the first of the five runs again.
> * I  come back in an hour and a half. Its now (1 pm). That run passed, so I go for a binary search and pick the middle of the 5 commits and kick of the CI for that.
> * I come back in an hour and a half, this time the run failed, but I can't know if it was that commit or the previous commit that introduced the bug, so I run CI for the commit before.
> * I come back in an hour and a half (its now 4pm) and see that it has failed. I've identified the commit with the bug (in 6 hours) but I have no logs or any idea what the problem is.
> * So now I move to my local machine to run the tests again so I can retrieve the logs.
> * The tests take 2.5 hours on my local machine, I now have logs covering the runs where something went wrong, its the end of the day (6:30pm)
> * The next day at 9am I investigate the logs, after a couple of hours of investigation I have an idea for a potential fix, and make the change.
> * I wait a further 2.5 hours for the test results, the test is still failing, its 1:30pm. I tweak some more things and set off the tests again.
> * After 2.5 more hours I still have failing tests so I tweak some more things and set the tests off to run again.
> * This run succeeds, its 6:30pm, I've fixed the bug. It took 18 hours or 2 days.
> 
> In the time it took me to fix the bug found in this framework, I could probably have written an equivalent test in go.
> 
> The go test will run faster, can be run individually, can be debugged, produces visible logs whether run on CI or locally and doesn't require a whole new set of tools to work with. It will also save the next person to encounter this problem roughly 18hours or 2 days of work.
> 
> This scenario I think shows why adding this test framework to our project will not be helpful.



> Hi @Jason-Zhangxin-Chen I'm not sure I fully understand your response. I'll try to clarify bit by bit.
> 
> > The difference is that the docker e2e test framework is taking the client as a blackbox, at a higher level testing which is closer to the end to end operation.
> 
> This is clear but we can do black box testing in golang as well, I don't see any need to create a separate framework in a separate language in order to achieve this and having our black box testing in go means that the tests don't suffer from the disadvantages I laid out above.
> 
> > I would not suggest to remove those testcases at a different testing level.
> 
> If you are saying that we should have black box tests I agree, but I think they should be done in golang. I don't believe that this framework gives us much value. How do I measure value in a test framework?
> 
> What is the purpose of testing? Ultimately the purpose of testing is to allow you to build software free of defects. In order to help you do that you need tests that do two things.
> 
> * Fail when something is broken.
> * Let you know what is broken and how to fix it.
> 
> A test that only fails but doesn't let you know what is broken and how to fix it is not useful. For example the flaky tests (#496) that you are working on right now. You can't reproduce the failures on your local system and all you have is the logs of the failures, its difficult to understand what is wrong and the consequence of that is you will need to spend a lot of time trying to figure out what needs fixing (these tests are not particularly useful). Contrast that with a test that when it fails you immediately know what needs to change in the code to fix it, that is the best type of test.
> 
> In order to have tests that let you know what is broken and how to fix it, it is important to have a short feedback loop(the time between something breaking and the time between finding out about it). There are some development environments that have optimised testing to the point that they run all the tests for each character typed. This means that the engineer has instant feedback when they break anything, this is extremely useful. The shorter the feedback loop the less specific the tests need to be about what went wrong, because the engineer will have changed less since the last testing feedback, this allows the engineer to narrow the scope of their search for bugs, which allows them to find and fix the bug quicker. Long feedback loops slow development not only by making it hard to find out where bugs originated but also because while an engineer is waiting for results from a test they are effectively blocked from making more changes on top of their latest change while they wait to see if their latest change fixed the problem. They don't want to change too much between runs because it will make it harder to pinpoint what the fix was, in the case that their changes did result in a fix.
> 
> The feedback loop for this test framework is extremely long. It takes 1hr 15m to run on the ci(and something like 2 hours on my local machine), this feedback loop will slow development and make it very hard to pinpoint the source of an error. Not only that but the framework offers very little help in identifying errors, when it fails it provides logs in a compressed file, but we can't access that file from ci.
> 
> I'm imagining the sequence of steps that I would need to take to fix the code when a test fails in this framework?
> 
> It could look something like this.
> 
> So I've been working away and pushing my branch regularly, maybe 5 times in the last hour(its 10am).
> 
> * I think the work is done, my latest pushes are queued in CI, so I cancel previous runs that haven't finished running the e2e tests, because I wan't feedback on my latest state, quickly.
> * I come back in an hour and a half and find that a test failed in the e2e tests. Now I'm not sure if this is something I introduced in my earlier commits or the latest commit, because I cancelled the runs for the earlier commits. So I kick of the first of the five runs again.
> * I  come back in an hour and a half. Its now (1 pm). That run passed, so I go for a binary search and pick the middle of the 5 commits and kick of the CI for that.
> * I come back in an hour and a half, this time the run failed, but I can't know if it was that commit or the previous commit that introduced the bug, so I run CI for the commit before.
> * I come back in an hour and a half (its now 4pm) and see that it has failed. I've identified the commit with the bug (in 6 hours) but I have no logs or any idea what the problem is.
> * So now I move to my local machine to run the tests again so I can retrieve the logs.
> * The tests take 2.5 hours on my local machine, I now have logs covering the runs where something went wrong, its the end of the day (6:30pm)
> * The next day at 9am I investigate the logs, after a couple of hours of investigation I have an idea for a potential fix, and make the change.
> * I wait a further 2.5 hours for the test results, the test is still failing, its 1:30pm. I tweak some more things and set off the tests again.
> * After 2.5 more hours I still have failing tests so I tweak some more things and set the tests off to run again.
> * This run succeeds, its 6:30pm, I've fixed the bug. It took 18 hours or 2 days.
> 
> In the time it took me to fix the bug found in this framework, I could probably have written an equivalent test in go.
> 
> The go test will run faster, can be run individually, can be debugged, produces visible logs whether run on CI or locally and doesn't require a whole new set of tools to work with. It will also save the next person to encounter this problem roughly 18hours or 2 days of work.
> 
> This scenario I think shows why adding this test framework to our project will not be helpful.


> Hi @Jason-Zhangxin-Chen I'm not sure I fully understand your response. I'll try to clarify bit by bit.
> 
> > The difference is that the docker e2e test framework is taking the client as a blackbox, at a higher level testing which is closer to the end to end operation.
> 
> This is clear but we can do black box testing in golang as well, I don't see any need to create a separate framework in a separate language in order to achieve this and having our black box testing in go means that the tests don't suffer from the disadvantages I laid out above.
> 
> > I would not suggest to remove those testcases at a different testing level.
> 
> If you are saying that we should have black box tests I agree, but I think they should be done in golang. I don't believe that this framework gives us much value. How do I measure value in a test framework?
> 
> What is the purpose of testing? Ultimately the purpose of testing is to allow you to build software free of defects. In order to help you do that you need tests that do two things.
> 
> * Fail when something is broken.
> * Let you know what is broken and how to fix it.
> 
> A test that only fails but doesn't let you know what is broken and how to fix it is not useful. For example the flaky tests (#496) that you are working on right now. You can't reproduce the failures on your local system and all you have is the logs of the failures, its difficult to understand what is wrong and the consequence of that is you will need to spend a lot of time trying to figure out what needs fixing (these tests are not particularly useful). Contrast that with a test that when it fails you immediately know what needs to change in the code to fix it, that is the best type of test.
> 
> In order to have tests that let you know what is broken and how to fix it, it is important to have a short feedback loop(the time between something breaking and the time between finding out about it). There are some development environments that have optimised testing to the point that they run all the tests for each character typed. This means that the engineer has instant feedback when they break anything, this is extremely useful. The shorter the feedback loop the less specific the tests need to be about what went wrong, because the engineer will have changed less since the last testing feedback, this allows the engineer to narrow the scope of their search for bugs, which allows them to find and fix the bug quicker. Long feedback loops slow development not only by making it hard to find out where bugs originated but also because while an engineer is waiting for results from a test they are effectively blocked from making more changes on top of their latest change while they wait to see if their latest change fixed the problem. They don't want to change too much between runs because it will make it harder to pinpoint what the fix was, in the case that their changes did result in a fix.
> 
> The feedback loop for this test framework is extremely long. It takes 1hr 15m to run on the ci(and something like 2 hours on my local machine), this feedback loop will slow development and make it very hard to pinpoint the source of an error. Not only that but the framework offers very little help in identifying errors, when it fails it provides logs in a compressed file, but we can't access that file from ci.
> 
> I'm imagining the sequence of steps that I would need to take to fix the code when a test fails in this framework?
> 
> It could look something like this.
> 
> So I've been working away and pushing my branch regularly, maybe 5 times in the last hour(its 10am).
> 
> * I think the work is done, my latest pushes are queued in CI, so I cancel previous runs that haven't finished running the e2e tests, because I wan't feedback on my latest state, quickly.
> * I come back in an hour and a half and find that a test failed in the e2e tests. Now I'm not sure if this is something I introduced in my earlier commits or the latest commit, because I cancelled the runs for the earlier commits. So I kick of the first of the five runs again.
> * I  come back in an hour and a half. Its now (1 pm). That run passed, so I go for a binary search and pick the middle of the 5 commits and kick of the CI for that.
> * I come back in an hour and a half, this time the run failed, but I can't know if it was that commit or the previous commit that introduced the bug, so I run CI for the commit before.
> * I come back in an hour and a half (its now 4pm) and see that it has failed. I've identified the commit with the bug (in 6 hours) but I have no logs or any idea what the problem is.
> * So now I move to my local machine to run the tests again so I can retrieve the logs.
> * The tests take 2.5 hours on my local machine, I now have logs covering the runs where something went wrong, its the end of the day (6:30pm)
> * The next day at 9am I investigate the logs, after a couple of hours of investigation I have an idea for a potential fix, and make the change.
> * I wait a further 2.5 hours for the test results, the test is still failing, its 1:30pm. I tweak some more things and set off the tests again.
> * After 2.5 more hours I still have failing tests so I tweak some more things and set the tests off to run again.
> * This run succeeds, its 6:30pm, I've fixed the bug. It took 18 hours or 2 days.
> 
> In the time it took me to fix the bug found in this framework, I could probably have written an equivalent test in go.
> 
> The go test will run faster, can be run individually, can be debugged, produces visible logs whether run on CI or locally and doesn't require a whole new set of tools to work with. It will also save the next person to encounter this problem roughly 18hours or 2 days of work.
> 
> This scenario I think shows why adding this test framework to our project will not be helpful.

Hi @piersy , at this integration phase, I think it might be not stable enough, but we can try to fix those un-stable problems as we did for those golang ones. Finally we will make it stable.

Jason-Zhangxin-Chen: > vered by tests under `consensus/test` but I can't know without a better understanding of bootstrap.

Bootstrap from scratch means the network will be re-init and start from a brand new genisis block. Those ledgers in the last context will be discarded.

.github/workflows/go.yml
```diff
@@ -382,9 +382,60 @@ jobs:
...
+        shell: bash
+
+      - name: Cache code
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-autonity-code
+        with:
+          path: /home/runner/work/autonity/autonity/
+          key: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
```
piersy, line 410: We shouldn't use restore keys  #504 , and this restore key makes no sense since it is the same as the main key.
```suggestion
          restore-keys: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
```

.github/workflows/go.yml
```diff
@@ -382,9 +382,60 @@ jobs:
...
+
+      - uses: actions/checkout@v2
+        if: steps.cache-autonity-code.outputs.cache-hit != 'true'
+        with:
+          clean: 'false'
+
+      - name: Embed Autonity contract
+        if:  steps.cache-autonity-code.outputs.cache-hit != 'true'
+        run: make embed-autonity-contract
```
piersy, line 419: You don't need this, it's generated by the bootstrap job
```suggestion
```

.github/workflows/go.yml
```diff
@@ -382,9 +382,60 @@ jobs:
...
+        run: make all
+
+      - name: Cache binary
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-autonity-bin
+        with:
+          path: /home/runner/work/autonity/autonity/build/bin
+          key: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
```
piersy, line 431:
```suggestion
```

piersy:
Hi @Jason-Zhangxin-Chen I just thought I would respond here for the record, even if we have decided to merge this without a full review.

>But I cannot totally agree with your idea. As the python test framework is a higher level testing than the one implemented in golang. When you explore blockchain project like BTC, DASH and PIVX and EOS, you will find that there is python based framework co-exist with those C++ repo to perform the end to end testing taking the client as a blackbox without using any internal implementation details include the P2P layer's feature like dials etc...

Just because other projects use a python test framework doesn't mean it is a good idea. I would want to know why they use python?

>From end to end testing point of view, I believe golang can do so, but python is more cost-efficient than golang.

I don't see why you would think that testing in python is more cost efficient than golang? I would think that golang would be more cost efficient, for these reasons.

* We are primarily go developers and are hiring primarily go developers to work on this project, so they should be more efficient at writing golang than python and I wouldn't consider these languages to differ greatly in their verbosity and difficulty to write, in contrast to say c++ which is extremely verbose, complex and has the overhead of having to manage memory.

* We can leverage existing golang test code rather than having to write it from scratch in python.

* Having to handle a different language will increase the amount of work required for our build/delploy/ci pipelines.

>The so called local e2e testing framework implemented with golang, it is powerfull too, but it does not take the entire autonity as a blackbox, it is not kind of end to end testing at least from my point of view. It looks like a component integration testing framework of autonity client. I cannot 100% sure that if the golang e2e test can address all the bugs before we roll out the release. And this is why we built ATAutoniy (the python base e2e testing framework) at the very beginning, trying to experience the features from real e2e point of view.

I agree the existing e2e framework is not using a black box approach, I'm not suggesting that it would be our black box testing framework. We can make a black box testing framework by simply only relying on the external APIs of the nodes that we run. Its not necessary to use another language to do that.

Jason-Zhangxin-Chen:
> Hi @Jason-Zhangxin-Chen I just thought I would respond here for the record, even if we have decided to merge this without a full review.
> 
> > But I cannot totally agree with your idea. As the python test framework is a higher level testing than the one implemented in golang. When you explore blockchain project like BTC, DASH and PIVX and EOS, you will find that there is python based framework co-exist with those C++ repo to perform the end to end testing taking the client as a blackbox without using any internal implementation details include the P2P layer's feature like dials etc...
> 
> Just because other projects use a python test framework doesn't mean it is a good idea. I would want to know why they use python?
> 
> > From end to end testing point of view, I believe golang can do so, but python is more cost-efficient than golang.
> 
> I don't see why you would think that testing in python is more cost efficient than golang? I would think that golang would be more cost efficient, for these reasons.
> 
> * We are primarily go developers and are hiring primarily go developers to work on this project, so they should be more efficient at writing golang than python and I wouldn't consider these languages to differ greatly in their verbosity and difficulty to write, in contrast to say c++ which is extremely verbose, complex and has the overhead of having to manage memory.
> * We can leverage existing golang test code rather than having to write it from scratch in python.
> * Having to handle a different language will increase the amount of work required for our build/delploy/ci pipelines.
> 
> > The so called local e2e testing framework implemented with golang, it is powerfull too, but it does not take the entire autonity as a blackbox, it is not kind of end to end testing at least from my point of view. It looks like a component integration testing framework of autonity client. I cannot 100% sure that if the golang e2e test can address all the bugs before we roll out the release. And this is why we built ATAutoniy (the python base e2e testing framework) at the very beginning, trying to experience the features from real e2e point of view.
> 
> I agree the existing e2e framework is not using a black box approach, I'm not suggesting that it would be our black box testing framework. We can make a black box testing framework by simply only relying on the external APIs of the nodes that we run. Its not necessary to use another language to do that.

I unstand you point, Python, C++, Golang they are tools, for end to end testing, python is the best tool at least for me.

.github/workflows/go.yml
```diff
@@ -382,9 +382,48 @@ jobs:
       - name: test contract
         run: make test-contracts
 
-  # Tests the helloworld example, this indirectly ensures that the autonity
-  # docker image can be built since the helloworld example uses it. We don't
-  # want to use any cache here since we want a clean checkout before any code
```
piersy, line 387: This shouldn't be removed, this should replace the comments on lines 424-426

piersy (approved):
The CI config is looking good :+1:

yazzaoui (approved):
LGTM

aneequesafdar (approved):

PR #572, opened 2020-06-02
Change tendermint apis backend #479

### LINKED ISSUE #479: Change Tendermint API's TednermintCore.Backend to *Backend
# Problem

In `consensus/tendermint/backend/api.go` the API struct is defined as follow:

```go
type API struct {
	chain      consensus.ChainReader
	tendermint tendermintCore.Backend
```

`tendermintCore.Backend` interface is imported from the `core` package, however, there is no need for this because API is defined in `backend` package, this also means that in tests a `tendermintCore.MockBackend` is used to test the functionality, which is not useful because in practice the `tendermintCore` doesn't call the function required by API thus rendering the test useless and not achieving as much test coverage as possible. Therefore the API should be changed to:

```go
type API struct {
	chain      consensus.ChainReader
	tendermint *Backend
```
This also means the test needs to be fixed so that they don't use a `tendermintCore.MockBackend`.

# Acceptance Criteria

- Ensure test no longer use `tendermintCore.MockBackend` and all the tests pass.
- Remove relevant TODO from the code

# Notes

This problem was noticed during pull request #472

aneequesafdar:
closes #479


go.mod
```diff
@@ -13,7 +13,8 @@ require (
 	github.com/alecthomas/participle v0.4.1
 	github.com/aristanetworks/goarista v0.0.0-20170210015632-ea17b1a17847
 	github.com/aws/aws-sdk-go v1.25.48
-	github.com/btcsuite/btcd v0.0.0-20171128150713-2e60448ffcc6
+	github.com/btcsuite/btcd v0.20.1-beta
+	github.com/btcsuite/btcutil v1.0.2 // indirect
```
piersy, line 17: I think this must be a mistake

aneequesafdar: I don't really know the go.mod was updated it was done automatically via make, I can exclude them for the time being

piersy: You haven't added any new external dependencies, so nothing should have changed. I would suggest reverting the changes to go.mod and go.sum and rebuilding everything and running the tests and see if they come back. They shouldn't as per my understanding.

PR #575, opened 2020-06-10
Reward distribution test

Jason-Zhangxin-Chen:
- The problem: 

https://github.com/clearmatics/autonity/issues/412, we don't have testcase to cover an accurate 
reward distribution checking on the network.

- How it is implemented: 

Reuse the local e2e testing framework to spin up the test network, and plug a reward distribution checker into the test framework, each node in the network would check it's balance includes those rewards base on each new block mined.


Jason-Zhangxin-Chen:
To be close via PR: https://github.com/clearmatics/autonity/pull/594 since this branch is out of update.

Issue #543, opened 2020-05-01
Add future round change tests for tendermint


aneequesafdar:
# Problem

We don't have tests for what a node should do when it 2f+1 messages from distinct validator for future round.

This is a sub-task for #480.

# Acceptance Criteria

Use lines 55-56 of page 6 of [The latest gossip on BFT consensus](https://arxiv.org/pdf/1807.04938.pdf) to describe the correct behaviours in test function for the current implementation of Tendermint.

PR #584, opened 2020-06-13
Tendermint tests for upon conditions

### LINKED ISSUE #535: Add consensus start round tests
# Problem

We don't have tests for core start rounds which determines how the proposer will send the Proposal and what non-proposers should do.

This is a sub-task for #480.

# Acceptance Criteria

Use lines 1-21 of page 6 of [The latest gossip on BFT consensus](https://arxiv.org/pdf/1807.04938.pdf) to describe the correct behaviours in test function for the current implementation of Tendermint.

aneequesafdar:
closes #535 #536 #537 #538 #539 #540 #541 #542 #543


consensus/tendermint/core/propose.go
```diff
@@ -141,11 +141,7 @@ func (c *core) handleProposal(ctx context.Context, msg *Message) error {
 
 		// Line 22 in Algorithm 1 of The latest gossip on BFT consensus
 		if vr == -1 {
-			var voteForProposal = false
-			if c.lockedValue != nil {
-				voteForProposal = c.lockedRound == -1 || h == c.lockedValue.Hash()
-			}
-			c.sendPrevote(ctx, voteForProposal)
+			c.sendPrevote(ctx, !(c.lockedRound == -1 || h == c.lockedValue.Hash()))
```
yazzaoui, line 144: Are we sure that `lockedValue` will never be nil in that branch ?

aneequesafdar: If you look at this test case, it addresses your concern: https://github.com/clearmatics/autonity/blob/12cccf7384b27004d35fe99661ac97f13cb4f9cb/consensus/tendermint/core/upon_condition_test.go#L315-L345

yazzaoui: I see, even though it works I feel it's a smelly pattern to call methods of nil objects, if ever the upstream team decides to change what fields the `Hash` method is accessing, this would panic. Any view on that @piersy ?

aneequesafdar: The code is completely in our control, we know that if `lockedRound > -1` then `lockedValue != nil`, so there is no reason to check for nil on `lockedValue` when we already have a check on `lockedRound`. Also, the following code show how `||` conditions are evaluated in go lang:

```go
package main

import (
	"fmt"
)

type testStruct struct {
	number int
}

func main() {
	t1 := &testStruct{2}
	t2 := &testStruct{1}
	t2 = nil

	fmt.Println(true || t1 == t2)
}
```
Therefore we know the first part of the `||` statement is evaluated first and if it is true then the rest of the statement is ignored, hence, it will not panic. 

If upstream were change the implementation of the `Hash()` function as long as we create the `lockedValue` block properly it will be of no consequence and we are not creating any blocks here. So the bigger problem is what would happen if `lockedValue == nil` and since our implementation mandates `lockedValue == nil` iff `lockedRound = -1`, thus a check on `lockedRound` serves as a check on nil for `lockedValue`

piersy: @yazzaoui 
>I see, even though it works I feel it's a smelly pattern to call methods of nil objects, if ever the upstream team decides to change what fields the Hash method is accessing, this would panic. Any view on that @piersy ?

This is not a case of calling a method on a nil object, the method is not being called, because as @aneequesafdar said the first condition of the or expression is evaluating to true.

I don't consider defining methods on nil object receivers to be smelly. It's a language feature and whether it's use is smelly depends on how it is used.

With respect to the upstream changing and causing a panic, the same argument could be made for any method or field we access on any code from upstream. The upstream could change their code such that a field we were previously relying on being non nil in a certain code path, is now nil in that code path. The only way to protect against this is to test all your code paths.

I do think this code warrants being commented though.

consensus/tendermint/core/propose.go
```diff
@@ -155,11 +151,7 @@ func (c *core) handleProposal(ctx context.Context, msg *Message) error {
 		// Line 28 in Algorithm 1 of The latest gossip on BFT consensus
 		// vr >= 0 here
 		if vr < c.Round() && rs.PrevotesPower(h) >= c.CommitteeSet().Quorum() {
-			var voteForProposal = false
-			if c.lockedValue != nil {
-				voteForProposal = c.lockedRound <= vr || h == c.lockedValue.Hash()
-			}
-			c.sendPrevote(ctx, voteForProposal)
+			c.sendPrevote(ctx, !(c.lockedRound <= vr || h == c.lockedValue.Hash()))
```
piersy, line 154: Are we sure that lockedValue will never be nil in this branch ?

aneequesafdar: `lockedValue` can be nil, however `c.lockedRound <= vr` will be true if `lockedValue = nil`. please look at the following testcase (where I mandate `lockedValue = nil` regardless of what `lockedRound` is which should never be possible): https://github.com/clearmatics/autonity/blob/12cccf7384b27004d35fe99661ac97f13cb4f9cb/consensus/tendermint/core/upon_condition_test.go#L428-L467

PR #586, opened 2020-06-17
Fix flaky tests

Jason-Zhangxin-Chen:
To fix the unstable test case listed by https://github.com/clearmatics/autonity/issues/496, this pull request fix the pending transaction issue happens on node re-join into the network. For the other unstable testcase, they are fixed by this pull request: https://github.com/clearmatics/autonity/pull/563

To apply loose TX error check in those disaster recovery test cases. 
For example, pending TX might happens when the recover node not yet complete the round state synchronization from peers, while the other nodes already come to the end chain height defined by the test case. Following with block sync, the recover node hooks keep issuing TXs into the TX memory pool. 
In such case, we check the engine is not on-hold and move forward to the final chain height defined by the test case, pending TX shouldn't be an error to cause test failure.


consensus/test/testcase_test.go
```diff
@@ -63,6 +63,13 @@ type testCase struct {
 	noQuorumTimeout      time.Duration
 	topology             *Topology
 	skipNoLeakCheck      bool
+
+	// To apply loose TX error check in those disaster test cases. For example, pending TX might happens when
+	// the recover node not yet complete the round state synchronization from peers, while the other nodes already come
+	// to the end chain height defined by the test case. Following with block sync, the recover node hook function keep
+	// issuing TXs into the TX memory pool. In such case, we check the engine is not on-hold and move to the final chain
+	// height defined by the test case, pending TX shouldn't be a error to cause test failure.
```
piersy, line 71: I think this could be a more concise description of this flag.
```suggestion
	// When set to true, this disables the check for un-mined transactions at the end of the test,
	// which if found would normally cause the test to fail.
```

consensus/test/testcase_test.go
```diff
@@ -63,6 +63,10 @@ type testCase struct {
 	noQuorumTimeout      time.Duration
 	topology             *Topology
 	skipNoLeakCheck      bool
+
+	// When set to true, this disables the check for un-mined transactions at the end of the test,
+	// which if found would normally cause the test to fail.
+	looseTXCheck bool
```
piersy, line 69: Now that I understand what this flag is doing, I'm thinking instead of sending transactions that we don't check, why not just stop sending transactions?

Jason-Zhangxin-Chen: Issuing TX would be better to verify if the basic features works from this local e2e test.

piersy: Could you explain what basic features are being verified by submitting txs that we do not check?

aneequesafdar (approved):

consensus/test/testcase_test.go
aneequesafdar: >  why not just stop sending transactions?

The major changes the take place in `wgLoop` of `consensus/test/test_helpers_test.go:runNode` where the changes are determined by the block which arrives from the `peer.eventChan`. Each test case defined for how many blocks a node should be mining as well as the number of transaction that needs to be sent by each node per block. Once the test number of block has reached the nodes will stop sending further transactions, however, the test case will keep running until `blocksToWait` blocks have been mined in order to allow for pending/queued transactions to be mined.

In the current test cases, the total amount of blocks waited on is not enough to allow for the all transactions to be mined, so the simplest solution is to either reduce the number of transaction issued or increase the `blocksToWait` parameter. Increasing the `blocksToWait` parameter will cause the total duration of the test case to increase.

This pull request simply ignores the transactions. Another way to ignore the pending and queued transaction would be to manually remove them from each node's transaction pool as done by the `testnode_test.go:stopNode()` when the before hook is called from start and stop tests. However, this approach is not straightforward and it will need to ensure that the transactions are removed from the transactions pool of all the nodes and the test cases also remove these transactions from its transaction map as it is shared by all the peers.

piersy: Hi @aneequesafdar, what I'm suggesting is just setting this to 0 in the relevant test cases.

https://github.com/clearmatics/autonity/blob/06aea15de7950dd0f7edbd6c5e484d36d9b6a84b/consensus/test/testcase_test.go#L45

That is the field that controls how many transactions will be sent. So instead of adding extra code to ignore the transactions that we send, we can just not send the transactions in the first place.

aneequesafdar: Yeah sounds like a good idea too, I wasn't sure what you were suggesting, I thought you wanted to stop sending the transactions after enough blocks were mined. Which was already the case, so I was confused.

piersy: >TX issuing and TX mining is a basic feature, I think this should be involved in the test.

Hi @Jason-Zhangxin-Chen, when `looseTxCheck == true`  we don't verify that transactions are being mined, and tx issuing is performed by our test code shown below.

So the only additional feature I can see being verified by sending transactions is that there is a transaction pool that can have transactions added to it without returning an error (line 47 below). Given that all other end to end tests are also verifying that there is a transaction pool that can have transactions added to it without returning an error and additionally are verifying that tx mining is functioning, I don't see a need to send transactions in these tests, especially when it involves further complicating the test framework by the addition of yet another parameter.

https://github.com/clearmatics/autonity/blob/876929411d71985495e23ed546f7ecec69956f61/consensus/test/test_helpers_test.go#L35-L58

Jason-Zhangxin-Chen: @piersy, so you prefer to set num of tx to be 0 for those testcases? I just prefer to let the TX issuing traffic happens during the test, just take those pending TX as normal scenario since in real production, TX pending happens all the time.

piersy: @Jason-Zhangxin-Chen I understand that having pending transactions is a common occurrence in real networks, it shouldn't be the case here though. Because we created this framework, so we should be able to construct a test in such a way that the transactions we submit get mined and provide sufficient buffer such that it works reliably on a variety of hardware.

Regarding not sending the transactions, my point is if we are not checking that txs are being processed then the test cases do not let us know that transaction mining is working, so we are just wasting resources sending transactions.

piersy:
@Jason-Zhangxin-Chen 

I think we agreed that we were not going to follow the approach of adding a `looseTxCheck` flag and instead would like to try to extend the number of blocks to see if we can solve the problems seen in these tests. So I am waiting for you to make these changes before I continue my review.

yazzaoui:
> @Jason-Zhangxin-Chen
> 
> I think we agreed that we were not going to follow the approach of adding a `looseTxCheck` flag and instead would like to try to extend the number of blocks to see if we can solve the problems seen in these tests. So I am waiting for you to make these changes before I continue my review.

Yes, let's see first the effect of just increasing `blocksToWait` . `looseTxCheck` might be required though for some cases where nodes are offline ..

piersy:
> > @Jason-Zhangxin-Chen
> > I think we agreed that we were not going to follow the approach of adding a `looseTxCheck` flag and instead would like to try to extend the number of blocks to see if we can solve the problems seen in these tests. So I am waiting for you to make these changes before I continue my review.
> 
> Yes, let's see first the effect of just increasing `blocksToWait` . `looseTxCheck` might be required though for some cases where nodes are offline ..

@Jason-Zhangxin-Chen @yazzaoui As far as I can see none of the tests end up with nodes being offline, so there is no need to use looseTxCheck. If we did have nodes that end up offline, then the solution would be to not check the transactions for that node, rather than not check any transactions at all.

Jason-Zhangxin-Chen:
Close this PR since we aggreed to solve it by extending the block hight of the testcase, it is under testing.

Jason-Zhangxin-Chen:
This pull request is closed, but the fix_flaky_tests branch has unmerged commits.

PR #591, opened 2020-06-18
Reward test with stake changes

Jason-Zhangxin-Chen:
https://github.com/clearmatics/autonity/issues/590
During the local e2e reward distribution test, manage the staking on the run time to verify if reward distribution works correctly by according to the stake changes via mint stake, redeem stake and transfer take functions.


Jason-Zhangxin-Chen:
To be close via PR: https://github.com/clearmatics/autonity/pull/594 since this branch is out of update.

PR #595, opened 2020-06-23
Remove the HeaderGetter interface

piersy:
Changes to the autonity contract in "Hardcode deployer address (#552)"
meant that this interface was no longer required.


piersy:
I think I may have found a way to not have FullChainReader, this change will be included in that PR

PR #603, opened 2020-07-01
Tdm/test/multi conditions test

Jason-Zhangxin-Chen:
Add 3 new test scenarios which is failed currently to cover tendermint behavior test. They are all about to simulate inbound msg handling with different order on those upon conditions.
@yazzaoui @aneequesafdar @piersy , please double check if those test scenario is valid. From my sense they should be valid to trigger the action describe in the pseudo code: line 28, 36 and 49.


Jason-Zhangxin-Chen:
Close this pull request since in current implementation those wrong ordered msg will be backloged and handled on step change.

PR #611, opened 2020-07-02
Deploy autonity contract on init

piersy:
This pr moves autonity contract deployment to the init phase.

It turned out to be quite simple!


cmd/autonity/genesis_test.go
```diff
@@ -28,34 +28,6 @@ var customGenesisTests = []struct {
...
-						}
-					]
-				},
-				"tendermint" : {}
-			}
-		}`,
-		query:  "eth.getBlock(0).nonce",
-		result: "0x0000000000001338",
-	},
 	// Genesis file with specific chain configurations
```
aneequesafdar, line 31: Can you please explain why you are removing this testcase?

eth/handler_test.go
```diff
@@ -477,7 +477,7 @@ func testCheckpointChallenge(t *testing.T, syncmode downloader.SyncMode, checkpo
 	// Initialize a chain and generate a fake CHT if checkpointing is enabled
 	var (
 		db     = rawdb.NewMemoryDatabase()
-		config = new(params.ChainConfig)
+		config = params.TestChainConfig
```
aneequesafdar, line 480: I am not sure why this was required, as far as I can see the difference between `new(params.ChainConfig)` and `params.TestChainConfig` is that some values are initialised, such as `big.Int`. I cannot see what changed to break the tests, can you please explain?

cmd/autonity/genesis_test.go
piersy: This test was checking that we could run autonity with an invalid configuration. The empty chain configuration in this case would have meant that the autonity contract deployment would have failed, but since this test never got beyond block 0 this problem never manifested itself.

eth/handler_test.go
piersy: The `TestChainConfig` is initialising the hard fork blocks that we rely on being set for the autonity contract to work. In order for this test to now work we need to set the hard fork blocks appropriately so that we can deploy the autonity contract on init.

yazzaoui (approved):
Happy with this PR, let's get it rebased and merged :+1:

aneequesafdar (approved):

PR #614, opened 2020-07-09
Blockperiod fixes

yazzaoui:
- Improve support for block periods greater than one second by making the mining logic block-period aware and making sure that the initial propose timeout value is set accordingly.
- Potentially fix for https://github.com/clearmatics/autonity/issues/604
- Bump Autonity version to 0.5

### DIFF
--- a/consensus/tendermint/core/timeout.go
+++ b/consensus/tendermint/core/timeout.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	initialProposeTimeout   = 3000 * time.Millisecond
+	initialProposeTimeout   = 2000 * time.Millisecond
 	proposeTimeoutDelta     = 500 * time.Millisecond
 	initialPrevoteTimeout   = 1000 * time.Millisecond
 	prevoteTimeoutDelta     = 500 * time.Millisecond
@@ -108,15 +108,15 @@ func (t *timeout) reset(s Step) {
 func (c *core) measureMetricsOnTimeOut(step uint64, r int64) {
 	switch step {
 	case msgProposal:
-		duration := timeoutPropose(r)
+		duration := c.timeoutPropose(r)
 		tendermintProposeTimer.Update(duration)
 		return
 	case msgPrevote:
-		duration := timeoutPrevote(r)
+		duration := c.timeoutPrevote(r)
 		tendermintPrevoteTimer.Update(duration)
 		return
 	case msgPrecommit:
-		duration := timeoutPrecommit(r)
+		duration := c.timeoutPrecommit(r)
 		tendermintPrecommitTimer.Update(duration)
 		return
 	}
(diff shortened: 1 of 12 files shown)


consensus/tendermint/core/timeout.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	initialProposeTimeout   = 3000 * time.Millisecond
+	initialProposeTimeout   = 2000 * time.Millisecond
```
Jason-Zhangxin-Chen, line 12: Why we change it to 2 seconds?

miner/worker.go
```diff
@@ -887,6 +887,16 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
 	if parent.Time() >= uint64(timestamp) {
 		timestamp = int64(parent.Time() + 1)
 	}
+
```
Jason-Zhangxin-Chen, line 890: I used to know the block period is controlled somehow by difficulty as BTC network did. Never know in our case, it is controlled in such way before.

consensus/tendermint/core/timeout.go
```diff
@@ -183,15 +183,15 @@ func (c *core) handleTimeoutPrecommit(ctx context.Context, msg TimeoutEvent) {
 
 /////////////// Calculate Timeout Duration Functions ///////////////
 // The timeout may need to be changed depending on the Step
-func timeoutPropose(round int64) time.Duration {
-	return initialProposeTimeout + time.Duration(round)*proposeTimeoutDelta
+func (c *core) timeoutPropose(round int64) time.Duration {
+	return initialProposeTimeout + time.Duration(c.blockPeriod)*time.Second + time.Duration(round)*proposeTimeoutDelta
```
aneequesafdar, line 187: Are we sure we include the block period part of the linear function? I would recommend something like the following:
`c.blockPeriod + time.Duration(round)*proposeTimeoutDelta`. So we should delete `initialProposeTimeout`.

consensus/tendermint/core/timeout.go
```diff
@@ -183,15 +183,15 @@ func (c *core) handleTimeoutPrecommit(ctx context.Context, msg TimeoutEvent) {
 
 /////////////// Calculate Timeout Duration Functions ///////////////
 // The timeout may need to be changed depending on the Step
-func timeoutPropose(round int64) time.Duration {
-	return initialProposeTimeout + time.Duration(round)*proposeTimeoutDelta
+func (c *core) timeoutPropose(round int64) time.Duration {
```
aneequesafdar, line 186: I would recommend passing the block period as a parameter. Otherwise, we are passing core to functions which don't use them at all

yazzaoui: Noted, feel free to do it when you'll be working on untangling.

consensus/tendermint/core/timeout.go
yazzaoui: `initialProposeTimeout` can't be equal to `c.blockPeriod` as we have to consider for network delays and such.

miner/worker.go
yazzaoui: It's a very convoluted logic unfortunately

PR #615, opened 2020-07-10
Tendermint behavior improvement

Jason-Zhangxin-Chen:
@yazzaoui @aneequesafdar @piersy 
This PR is about an enhancement of our tendermint implementation. It apply TDM line 28 condition checking for prevote handler when an old round prevote is apply to round state. The scenario was simulated by a new testcase added under tendermint behavior testset, also there is detail comments on the testcase context. The correction is about to add line 28 conditon checking on the prevote handler once the old prevote is appected as an old round msg. Please review it.

### DIFF
--- a/consensus/tendermint/core/prevote.go
+++ b/consensus/tendermint/core/prevote.go
@@ -70,6 +70,21 @@ func (c *core) handlePrevote(ctx context.Context, msg *Message) error {
 			// We only process old rounds while future rounds messages are pushed on to the backlog
 			oldRoundMessages := c.messages.getOrCreate(preVote.Round)
 			c.acceptVote(oldRoundMessages, prevote, preVote.ProposedBlockHash, *msg)
+
+			// Line 28 in Algorithm 1 of The latest gossip on BFT consensus.
+			if c.step == propose {
+				if c.curRoundMessages != nil && c.curRoundMessages.proposal != nil {
+					vr := c.curRoundMessages.proposal.ValidRound
+					h := c.curRoundMessages.proposal.ProposalBlock.Hash()
+					rs := c.messages.getOrCreate(vr)
+
+					if vr >= 0 && vr < c.Round() && rs.PrevotesPower(h) >= c.committeeSet().Quorum() {
+						c.sendPrevote(ctx, !(c.lockedRound <= vr || h == c.lockedValue.Hash()))
+						c.setStep(prevote)
+						return nil
+					}
+				}
+			}
 		}
 		return err
 	}
--- a/consensus/tendermint/core/upon_condition_test.go
+++ b/consensus/tendermint/core/upon_condition_test.go
@@ -557,6 +557,64 @@ func TestOldProposal(t *testing.T) {
 		assert.Equal(t, clientLockedValue, c.lockedValue)
 		assert.Equal(t, clientLockedValue, c.validValue)
 	})
+
+	// line 28 check upon condition on prevote handler.
+	t.Run("handle proposal before full quorum prevote on valid round is satisfied, exe action by applying old round prevote into round state", func(t *testing.T) {
+		currentHeight := big.NewInt(int64(rand.Intn(maxSize) + 1))
+		currentRound := int64(rand.Intn(committeeSizeAndMaxRound))
+		// vr >= 0 && vr < round_p
+		proposalValidRound := int64(rand.Intn(int(currentRound)))
+		// -1 <= c.lockedRound <= vr
+		clientLockedRound := int64(rand.Intn(int(proposalValidRound+2) - 1))
+		// the new round proposal
+		proposalMsg, proposal := generateBlockProposal(t, currentRound, currentHeight, proposalValidRound, members[currentRound].Address, false)
+
+		// the old round prevote msg to be handled to get the full quorum prevote on old round vr with value v.
+		prevoteMsg, _, _ := prepareVote(t, msgPrevote, proposalValidRound, currentHeight, proposal.ProposalBlock.Hash(), clientAddr, privateKeys[clientAddr])
+
+		// the expected prevote msg to be broadcast for the new round with <currentHeight, currentRound, proposal.ProposalBlock.Hash()>
+		prevoteMsgToBroadcast, prevoteMsgRLPNoSig, prevoteMsgRLPWithSig := prepareVote(t, msgPrevote, currentRound, currentHeight, proposal.ProposalBlock.Hash(), clientAddr, privateKeys[clientAddr])
+
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		backendMock := NewMockBackend(ctrl)
+		backendMock.EXPECT().Address().Return(clientAddr)
+
+		c := New(backendMock, config.RoundRobin)
+		c.setCommitteeSet(committeeSet)
+		// construct round state with: old round's quorum-1 prevote for v on valid round.
+		c.messages.getOrCreate(proposalValidRound).AddPrevote(proposal.ProposalBlock.Hash(), Message{Code: msgPrevote, power: c.committeeSet().Quorum() - 1})
+
+		// client on new round's step propose.
+		c.setHeight(currentHeight)
+		c.setRound(currentRound)
+		c.setStep(propose)
+		c.lockedRound = clientLockedRound
+		c.validRound = clientLockedRound
+		c.lockedValue = nil
+		c.validValue = nil
+
+		backendMock.EXPECT().VerifyProposal(*proposal.ProposalBlock).Return(time.Duration(1), nil)
+		backendMock.EXPECT().Sign(prevoteMsgRLPNoSig).Return(prevoteMsgToBroadcast.Signature, nil)
+		backendMock.EXPECT().Broadcast(context.Background(), committeeSet.Committee(), prevoteMsgRLPWithSig).Return(nil)
+
+		// now we handle new round's proposal with round_p > vr on value v.
+		err := c.handleCheckedMsg(context.Background(), proposalMsg, members[currentRound])
+		assert.Nil(t, err)
+
+		// now we receive the last old round's prevote MSG to get quorum prevote on vr for value v.
+		// the old round's prevote is accepted into the round state which now have the line 28 condition satisfied.
+		// now to take the action of line 28 which was not align with pseudo code before.
+		sender := 0
+		err = c.handleCheckedMsg(context.Background(), prevoteMsg, members[sender])
+		assert.Nil(t, err)
+		assert.Equal(t, prevote, c.step)
+		assert.Nil(t, c.lockedValue)
+		assert.Equal(t, clientLockedRound, c.lockedRound)
+		assert.Nil(t, c.validValue)
+		assert.Equal(t, clientLockedRound, c.validRound)
+	})
 }
 
 // The following tests aim to test lines 34 - 35 & 61 - 64 of Tendermint Algorithm described on page 6 of
--- a/consensus/test/autonity_contract_test.go
+++ b/consensus/test/autonity_contract_test.go
@@ -227,6 +227,8 @@ func TestCheckBlockWithSmallFee(t *testing.T) {
 }
 
 func TestRemoveFromValidatorsList(t *testing.T) {
+	// to be tracked by https://github.com/clearmatics/autonity/issues/604
+	t.Skip("skipping test since the upstream update cause local e2e test framework go routine leak.")
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}
--- a/consensus/test/start_stop_test.go
+++ b/consensus/test/start_stop_test.go
@@ -149,6 +149,8 @@ func TestTendermintStartStopSingleNode(t *testing.T) {
 }
 
 func TestTendermintStartStopFNodes(t *testing.T) {
+	// to be tracked by https://github.com/clearmatics/autonity/issues/604
+	t.Skip("skipping test since the upstream update cause local e2e test framework go routine leak.")
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}


consensus/tendermint/core/prevote.go
```diff
@@ -70,6 +70,21 @@ func (c *core) handlePrevote(ctx context.Context, msg *Message) error {
 			// We only process old rounds while future rounds messages are pushed on to the backlog
 			oldRoundMessages := c.messages.getOrCreate(preVote.Round)
 			c.acceptVote(oldRoundMessages, prevote, preVote.ProposedBlockHash, *msg)
+
+			// Line 28 in Algorithm 1 of The latest gossip on BFT consensus.
+			if c.step == propose {
```
yazzaoui, line 75: Can't happen, if we receive a prevote message while the core is at propose step, it will be put in the backlog.
After timeoutPropose, we will process the prevote with `c.step == prevote` .

consensus/tendermint/core/prevote.go
```diff
@@ -70,6 +70,21 @@ func (c *core) handlePrevote(ctx context.Context, msg *Message) error {
...
+				if c.curRoundMessages != nil && c.curRoundMessages.proposal != nil {
```
yazzaoui, line 76: `c.curRoundMessage` can't never be `nil`, it is initalized in `setInitialState`

consensus/tendermint/core/prevote.go
piersy: This is happening, I took this branch and just added a panic immediately inside this block.

See the failing test here - https://github.com/clearmatics/autonity/runs/890599873?check_suite_focus=true

I would suggest either keeping this block if it makes sense (Id have to give it some more scrutiny to know if it did) or keeping just a panic here to ensure that we know when our assumptions are violated and to help us track down the problematic code.

consensus/tendermint/core/prevote.go
piersy: I think it is a good idea to omit these nil checks since if either of these fields happen to be nil, we will be alerted to the problem by a panic.

consensus/tendermint/core/prevote.go
```diff
@@ -70,6 +70,21 @@ func (c *core) handlePrevote(ctx context.Context, msg *Message) error {
...
+					vr := c.curRoundMessages.proposal.ValidRound
+					h := c.curRoundMessages.proposal.ProposalBlock.Hash()
+					rs := c.messages.getOrCreate(vr)
+
+					if vr >= 0 && vr < c.Round() && rs.PrevotesPower(h) >= c.committeeSet().Quorum() {
+						c.sendPrevote(ctx, !(c.lockedRound <= vr || h == c.lockedValue.Hash()))
```
piersy, line 82: I'm not convinced this is correct, line 30 says that we should be broadcasting either v (the value in the prevote message)   but we are sending `c.curRoundMessages.GetProposalHash()` how can we be sure that `c.curRoundMessages.GetProposalHash()` matches v? In any case it would be much clearer if we just broadcast a prevote for the v of the received message.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -557,6 +557,64 @@ func TestOldProposal(t *testing.T) {
...
+		c.messages.getOrCreate(proposalValidRound).AddPrevote(proposal.ProposalBlock.Hash(), Message{Code: msgPrevote, power: c.committeeSet().Quorum() - 1})
+
+		// client on new round's step propose.
+		c.setHeight(currentHeight)
+		c.setRound(currentRound)
+		c.setStep(propose)
+		c.lockedRound = clientLockedRound
+		c.validRound = clientLockedRound
+		c.lockedValue = nil
+		c.validValue = nil
```
aneequesafdar, line 596: these cannot be nil while `lockedRound` and `validRound` are set, they should be equal to some proposal that was sent in `clientLockedRound`, unless `clientLockedRound = -1`.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -557,6 +557,64 @@ func TestOldProposal(t *testing.T) {
...
 	})
+
+	// line 28 check upon condition on prevote handler.
+	t.Run("handle proposal before full quorum prevote on valid round is satisfied, exe action by applying old round prevote into round state", func(t *testing.T) {
+		currentHeight := big.NewInt(int64(rand.Intn(maxSize) + 1))
+		currentRound := int64(rand.Intn(committeeSizeAndMaxRound))
+		// vr >= 0 && vr < round_p
+		proposalValidRound := int64(rand.Intn(int(currentRound)))
+		// -1 <= c.lockedRound <= vr
+		clientLockedRound := int64(rand.Intn(int(proposalValidRound+2) - 1))
```
aneequesafdar, line 568: I think the scene is depicted here is a follows:

Round 0:
- Quorum clients sent precommit nil hence there was a round change

Round 1:
- The proposer of round 1 had received quorum prevotes for the proposal of round 0, thus, the proposer of round 1 reproposes the proposal with vr = 0
- More than quorum of network is yet to receive the prevotes from the previous round, thus the are their lockedValue and lockedRound is -1.
- The proposer of round 1 has a really good connection with the rest of the network, thus it is able to send the proposal to its peers before they receive enough prevotes from round 0 to form a quorum.
- Now the proposal will only be able to satisfy line 28 of the tendermint sudo code, however, the prevotes from the previous round are yet to arrive.
- A quorum prevotes for round 0 finally arrive, however they would be put into the backlog, so now the client has enough messages to send a prevote for round 1, but will not be able to do so because of the backlog.
- The client will have to wait till the propose timeout expiry to move to step prevote while sending a prevote nil for round 1. Once it is in step prevote, it will process the prevotes from round 0 and nothing would happen even though the client had enough messages to send prevote for the proposal.
- Thus this round is wasted because more than a quorum on the network sent prevote nil even though they had enough messages to send prevote for the proposal.

This can also happen where the client locked round < vr, it cannot happen for lockedRound = vr because that mean the client had received enough prevote in a timely manner and there are no old prevote to arrive.

Therefore the comment and the clientLockedRound should be changed as follows:

```suggestion
		// -1 <= c.lockedRound < vr, client is either unlocked or is locked on a round less than the proposal's valid round, the later case is less likely than the client being unlocked
		clientLockedRound := int64(rand.Intn(int(proposalValidRound)) - 1)
```

consensus/tendermint/core/prevote.go
aneequesafdar: line 72 and `rs.PrevotesPower(h)` ensure that `c.curRoundMessages.GetProposalHash() = prevote.ProposedBlockHash`, otherwise, if the hashes were different there would not be a quorum, please have a look at the rest of the function, similar logic is used.

aneequesafdar: Also, there is an implicit assumption that there can only be only proposal per round, thus any value different than the proposal received by the client will be from a malicious guy, this is also why we can assume that if we have received a proposal and prevote which come from a correct client will be for same proposal

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -557,6 +557,64 @@ func TestOldProposal(t *testing.T) {
...
+		// now to take the action of line 28 which was not align with pseudo code before.
+		sender := 0
+		err = c.handleCheckedMsg(context.Background(), prevoteMsg, members[sender])
+		assert.Nil(t, err)
+		assert.Equal(t, prevote, c.step)
+		assert.Nil(t, c.lockedValue)
+		assert.Equal(t, clientLockedRound, c.lockedRound)
+		assert.Nil(t, c.validValue)
+		assert.Equal(t, clientLockedRound, c.validRound)
+	})
```
aneequesafdar, line 617:
```suggestion
	t.Run("handle proposal before full quorum prevote on valid round is satisfied, exe action by applying old round prevote into round state", func(t *testing.T) {
		currentHeight := big.NewInt(int64(rand.Intn(maxSize) + 1))
		currentRound := int64(rand.Intn(committeeSizeAndMaxRound))

		// vr >= 0 && vr < round_p
		proposalValidRound := int64(rand.Intn(int(currentRound)))

		// -1 <= c.lockedRound < vr, if the client lockedValue = vr then the client had received the prevotes in a
		// timely manner thus there are no old prevote yet to arrive
		clientLockedRound := int64(rand.Intn(int(proposalValidRound)) - 1)

		// the new round proposal
		proposalMsg, proposal := generateBlockProposal(t, currentRound, currentHeight, proposalValidRound, members[currentRound].Address, false)

		// old proposal some random block
		clientLockedValue := generateBlock(currentHeight)

		// the old round prevote msg to be handled to get the full quorum prevote on old round vr with value v.
		prevoteMsg, _, _ := prepareVote(t, msgPrevote, proposalValidRound, currentHeight, proposal.ProposalBlock.Hash(), clientAddr, privateKeys[clientAddr])

		// the expected prevote msg to be broadcast for the new round with <currentHeight, currentRound, proposal.ProposalBlock.Hash()>
		prevoteMsgToBroadcast, prevoteMsgRLPNoSig, prevoteMsgRLPWithSig := prepareVote(t, msgPrevote, currentRound, currentHeight, proposal.ProposalBlock.Hash(), clientAddr, privateKeys[clientAddr])

		ctrl := gomock.NewController(t)
		defer ctrl.Finish()

		backendMock := NewMockBackend(ctrl)
		backendMock.EXPECT().Address().Return(clientAddr)

		c := New(backendMock, config.DefaultConfig())
		c.setCommitteeSet(committeeSet)
		// construct round state with: old round's quorum-1 prevote for v on valid round.
		c.messages.getOrCreate(proposalValidRound).AddPrevote(proposal.ProposalBlock.Hash(), Message{Code: msgPrevote, power: c.committeeSet().Quorum() - 1})

		// client on new round's step propose.
		c.setHeight(currentHeight)
		c.setRound(currentRound)
		c.setStep(propose)
		c.lockedRound = clientLockedRound
		c.validRound = clientLockedRound
		c.lockedValue = clientLockedValue
		c.validValue = clientLockedValue

		backendMock.EXPECT().VerifyProposal(*proposal.ProposalBlock).Return(time.Duration(1), nil)
		backendMock.EXPECT().Sign(prevoteMsgRLPNoSig).Return(prevoteMsgToBroadcast.Signature, nil)
		backendMock.EXPECT().Broadcast(context.Background(), committeeSet.Committee(), prevoteMsgRLPWithSig).Return(nil)

		// now we handle new round's proposal with round_p > vr on value v.
		err := c.handleCheckedMsg(context.Background(), proposalMsg, members[currentRound])
		assert.Nil(t, err)

		// now we receive the last old round's prevote MSG to get quorum prevote on vr for value v.
		// the old round's prevote is accepted into the round state which now have the line 28 condition satisfied.
		// now to take the action of line 28 which was not align with pseudo code before.
		sender := 0
		err = c.handleCheckedMsg(context.Background(), prevoteMsg, members[sender])
		assert.Nil(t, err)
		assert.Equal(t, prevote, c.step)
		assert.Equal(t, clientLockedValue, c.lockedValue)
		assert.Equal(t, clientLockedRound, c.lockedRound)
		assert.Equal(t, clientLockedValue, c.validValue)
		assert.Equal(t, clientLockedRound, c.validRound)
	})
```

consensus/tendermint/core/prevote.go
aneequesafdar: @yazzaoui old round messages are not added to the backlog:

```go 
// We want to store only future messages in backlog
		if err == errFutureHeightMessage {
			logger.Debug("Storing future height message in backlog")
			c.storeBacklog(msg, sender)
		} else if err == errFutureRoundMessage {
			logger.Debug("Storing future round message in backlog")
			c.storeBacklog(msg, sender)
			// decoding must have been successful to return
			c.handleFutureRoundMsg(ctx, msg, sender)
		} else if err == errFutureStepMessage {
			logger.Debug("Storing future step message in backlog")
			c.storeBacklog(msg, sender)
		}

		return err
```

In `checkMessage()` old round message take precedence over future step message, thus none of the old round messages are added to the backlog, they are directly added to message set as mentioned on line 71-72.

Jason-Zhangxin-Chen:
@yazzaoui @piersy @aneequesafdar , thanks for the valuable comments. Yes, this scenario would happens since those valid old round prevote msg might be coming when node is on propose step even if we are using a reliable broadcast than the native p2p gossip, and those old round msg was directly apply to the round state (msg set) which should trigger the actions defined in the pseudo code.

consensus/tendermint/core/prevote.go
piersy: > line 72 and `rs.PrevotesPower(h)` ensure that `c.curRoundMessages.GetProposalHash() = prevote.ProposedBlockHash`, otherwise, if the hashes were different there would not be a quorum, please have a look at the rest of the function, similar logic is used.

I don't think they do. Line 72 just accepts this prevote,  `rs.PrevotesPower(h)` checks for quorum threshold on `h` but there is no check to see that `h` equals `preVote.ProposedBlockHash`.

I don't actually think that is a problem though. Since actually line 30 is saying we should be broadcasting the value of prevotes that have reached quorum threshold, which is what this does.

consensus/tendermint/core/prevote.go
Jason-Zhangxin-Chen: @yazzaoui , this will happens since we didn't store the old round msg in backlog yet.

consensus/tendermint/core/prevote.go
Jason-Zhangxin-Chen: @yazzaoui , @piersy , yes, the curRoundMessage shouldn't be nil ever, so I'd remove this checking in my new commit.

consensus/tendermint/core/prevote.go
Jason-Zhangxin-Chen: > > line 72 and `rs.PrevotesPower(h)` ensure that `c.curRoundMessages.GetProposalHash() = prevote.ProposedBlockHash`, otherwise, if the hashes were different there would not be a quorum, please have a look at the rest of the function, similar logic is used.
> 
> I don't think they do. Line 72 just accepts this prevote, `rs.PrevotesPower(h)` checks for quorum threshold on `h` but there is no check to see that `h` equals `preVote.ProposedBlockHash`.
> 
> I don't actually think that is a problem though. Since actually line 30 is saying we should be broadcasting the value of prevotes that have reached quorum threshold, which is what this does.



> > line 72 and `rs.PrevotesPower(h)` ensure that `c.curRoundMessages.GetProposalHash() = prevote.ProposedBlockHash`, otherwise, if the hashes were different there would not be a quorum, please have a look at the rest of the function, similar logic is used.
> 
> I don't think they do. Line 72 just accepts this prevote, `rs.PrevotesPower(h)` checks for quorum threshold on `h` but there is no check to see that `h` equals `preVote.ProposedBlockHash`.
> 
> I don't actually think that is a problem though. Since actually line 30 is saying we should be broadcasting the value of prevotes that have reached quorum threshold, which is what this does.

Yes, for the enhancement, this piece of code just check the potential senario to move forward into next voting step if the Line 28 condition is matched.

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: @aneequesafdar , thank you for the details explainations, it is a good catch. By writing the test, we are trying to align with the pseudo code as much as possible, this enhancement do help to close the gaps with it.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,86 @@ func TestOldProposal(t *testing.T) {
 		assert.Equal(t, clientLockedValue, c.lockedValue)
 		assert.Equal(t, clientLockedValue, c.validValue)
 	})
+
+	// line 28 check upon condition on prevote handler.
+	t.Run("handle proposal before full quorum prevote on valid round is satisfied, exe action by applying old round prevote into round state", func(t *testing.T) {
+		clientIndex := len(members) - 1
+		clientAddr = members[clientIndex].Address
+
+		currentHeight := big.NewInt(int64(rand.Intn(maxSize) + 1))
```
piersy, line 579: Why are you using `rand.Intn`?

Jason-Zhangxin-Chen: > Why are you using `rand.Intn`?

Aneeque and I decide to use rand value to make it flexible for potential values without hard code it.

piersy: I actually think this use of `rand.Intn` is not helpful because when the test fails you don't know what parameters caused it to fail.

And so you can't recreate the failure.

Although in fact in this case the randomly chosen values will always be the same because you are not calling `rand.Seed`.

What I would suggest is making sure you call `rand.Seed` and then ensuring that if the test fails it prints out the seed, that way you can easily recreate failed cases by hardcoding the seed.

Also given that you are randomising the values I would run each test a few times rather than just once.

piersy (changes requested):
I left my suggestions in my last comment.

consensus/tendermint/core/upon_condition_test.go
aneequesafdar: So the idea was to use random round numbers and ensure that random numbers were always in a certain range, this allows for testing different round at each run while the test assumes that as long as the random number is in a certain range the behaviour should remain the same.

Your suggestion is good to print the seed in order to determine the round number when test case fail. However, this issue is not unique to this test case, the whole `upon_condition_test.go` has this issue, so I think we should create a separate issue to address this for all the test. Otherwise, please advise on how to proceed.

Jason-Zhangxin-Chen: This is a very low level unit test rather than the local e2e test framework, so if it failed with random values, it would be really simple to reproduce the problem with a debugger attach to the testcase.

piersy: @Jason it would not be so easy!

The following blocks show the randomly selected variables.
https://github.com/clearmatics/autonity/blob/6b7fa2f565241b35121b0c5aba6fab9f8925e39e/consensus/tendermint/core/upon_condition_test.go#L435-L436
https://github.com/clearmatics/autonity/blob/6b7fa2f565241b35121b0c5aba6fab9f8925e39e/consensus/tendermint/core/upon_condition_test.go#L563-L568

They are:
committeeSet (4-99) range of 96 possible values
currentHeight  (1-100) range of 100 possible values
currentRound (4-99) range of 96 possible values
proposalValidRound (4-99) range of 96 possible values
clientLockedRound ((-1)-100) range of 101 possible values

So there are potentially `96*100*96*96*101 = 8935833600` (roughly 9 billion) different random states that could be generated here. If you don't know what the random variables were when the test failed, it could take you a very long time to reproduce that particular state.

piersy: @aneequesafdar I will open a separate issue.

Jason-Zhangxin-Chen: Okay, let's create a seperate issue to remove those random values.

piersy: I created #659

aneequesafdar (approved):

consensus/tendermint/core/upon_condition_test.go
piersy: @aneequesafdar Could you add your scenario description as a comment for this testcase

piersy (changes requested):
Requested an additional explanatory comment

aneequesafdar:
@piersy the test case is commented well, so I think adding the scenario would be redundant

piersy:
> @piersy the test case is commented well, so I think adding the scenario would be redundant

@aneequesafdar I disagree, if the comments in the code were sufficient you wouldn't have felt the need for elaborating here in the PR.  Having the scenario explained really helps to understand what is happening in the test.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,128 @@ func TestOldProposal(t *testing.T) {
 		assert.Equal(t, clientLockedValue, c.lockedValue)
 		assert.Equal(t, clientLockedValue, c.validValue)
 	})
+
+	// line 28 check upon condition on prevote handler.
+	/*
+	  Please refer to the discussion history: https://github.com/clearmatics/autonity/pull/615, this test case address
+	the un-handled valid scenario which is not processed by our Tendermint implementation. The scene is depicted here
+	is a follows:
```
piersy, line 578: This is a bit misleading, since the scenario is now handled.

```suggestion
	Please refer to the discussion history: https://github.com/clearmatics/autonity/pull/615, this test case address
	the scenario laid out below.
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,128 @@ func TestOldProposal(t *testing.T) {
...
+
+	autonity/consensus/tendermint/core/propose.go, Lines 131 to 133 in 78f199d
+
+	 if err := c.proposeTimeout.stopTimer(); err != nil {
+	 	return err
+	 }
+
+	A quorum prevotes for round 0 finally arrive, however, these will be added to message set but nothing will happen
+	even though enough messages are present in the message set to send a prevote for the old proposal.
+	autonity/consensus/tendermint/core/prevote.go, Lines 69 to 74 in 78f199d
```
piersy:
```suggestion
	A quorum prevotes for round 0 finally arrive, however, these will be added to message set and without 
	the existence of the line 28 upon condition nothing would happen, even though enough  messages 
	are present in the message set to send a prevote for the old proposal.
	
	This was previously the case in:
	autonity/consensus/tendermint/core/prevote.go, Lines 69 to 74 at 78f199d
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,128 @@ func TestOldProposal(t *testing.T) {
...
+	Round 0:
+		Quorum clients sent precommit nil hence there was a round change
+	Round 1:
+		The proposer of round 1 had received quorum prevotes for the proposal of round 0, thus, the proposer of round 1
+	re-proposes the proposal with vr = 0, More than quorum of the network is yet to receive the prevotes from the
+	previous round, thus the are their lockedValue = nil and lockedRound = -1. The proposer of round 1 has a really good
+	connection with the rest of the network, thus it is able to send the proposal to its peers before they receive enough
+	prevotes from round 0 to form a quorum.	Now the proposal will only be able to satisfy line 28 of the Tendermint
+	pseudo code, however, the prevotes from the previous round are yet to arrive. Since the proposal of the current
+	round has been received the timer would be stopped.
```
piersy:
```suggestion
	prevotes from round 0 to form a quorum. Now the proposal received by the peers would only be able to satisfy line
	28 of the Tendermint pseudo code, however, the prevotes from the previous round are yet to arrive. Since the
	proposal of the current round has been received the timer would be stopped.
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,128 @@ func TestOldProposal(t *testing.T) {
...
+	 return err
+
+	Now the client is stuck since the timer has been stopped thus a prevote nil cannot be sent and the timer cannot be
+	restarted until startRound() is called for a new round. The resending of the message set will also not help because
+	it would only send messages to peers which they haven't seen and since there were no new messages the peers will not
+	be able to make progress. This can also happen where the client's lockedRound < vr, it cannot happen for
+	lockedRound = vr because that means the client had received enough prevote in a timely manner and there are no old
+	prevote to arrive.
+
+	Therefore we have a liveness bug in the current implementation of Tendermint.
```
piersy:
```suggestion
	Without the line 28 upon conditoin the client is stuck since the timer has been stopped, thus a prevote nil cannot be sent and the timer cannot be
	restarted until startRound() is called for a new round. The resending of the message set will also not help because
	it would only send messages to peers which they haven't seen and since there were no new messages the peers will not
	be able to make progress. This can also happen where the client's lockedRound < vr, it cannot happen for
	lockedRound = vr because that means the client had received enough prevote in a timely manner and there are no old
	prevote to arrive.

	Therefore we had a liveness bug in implementations of Tendermint in commits prior to this one.
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,127 @@ func TestOldProposal(t *testing.T) {
 		assert.Equal(t, clientLockedValue, c.lockedValue)
 		assert.Equal(t, clientLockedValue, c.validValue)
 	})
+
+	// line 28 check upon condition on prevote handler.
+	/*
+		  Please refer to the discussion history: https://github.com/clearmatics/autonity/pull/615
+		This test case address the scenario laid out below.
```
piersy, line 577: This is a bit misleading, since the scenario is now handled.

```suggestion
	Please refer to the discussion history: https://github.com/clearmatics/autonity/pull/615
	This test case validates the need for the upon condition defined in line 28 of the tendermint
	whitepaper, which addresses the scenario laid out below. 
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,127 @@ func TestOldProposal(t *testing.T) {
...
+
+		autonity/consensus/tendermint/core/propose.go, Lines 131 to 133 in 78f199d
+
+		 if err := c.proposeTimeout.stopTimer(); err != nil {
+		 	return err
+		 }
+
+		A quorum prevotes for round 0 finally arrive, however, these will be added to message set but nothing will happen
+		even though enough messages are present in the message set to send a prevote for the old proposal.
+		autonity/consensus/tendermint/core/prevote.go, Lines 69 to 74 in 78f199d
```
piersy, line 598:
```suggestion
		A quorum prevotes for round 0 finally arrive, however, these will be added to message set and without 
		the existence of the line 28 upon condition nothing would happen, even though enough  messages 
		are present in the message set to send a prevote for the old proposal.
	
		This was previously the case in:
		autonity/consensus/tendermint/core/prevote.go, Lines 69 to 74 at 78f199d
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+
+		autonity/consensus/tendermint/core/propose.go, Lines 131 to 133 in 78f199d
+
+		 if err := c.proposeTimeout.stopTimer(); err != nil {
+		 	return err
+		 }
+
+		A quorum prevotes for round 0 finally arrive, however, these will be added to message set and without 
+		the existence of the line 28 upon condition nothing would happen, even though enough  messages 
+		are present in the message set to send a prevote for the old proposal.
```
piersy, line 599:
```suggestion
	A quorum of prevotes for round 0 finally arrive, however, these will be added to message set and 
	without the existence of the line 28 upon condition nothing would happen, even though enough  
	messages are present in the message set to send a prevote for the old proposal.
	
	This was previously the case in:
	autonity/consensus/tendermint/core/prevote.go, Lines 69 to 74 at 78f199d
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		Round 0:
+			Quorum clients sent precommit nil hence there was a round change
+		Round 1:
+			The proposer of round 1 had received quorum prevotes for the proposal of round 0, thus, the proposer of round 1
+		re-proposes the proposal with vr = 0, More than quorum of the network is yet to receive the prevotes from the
+		previous round, thus the are their lockedValue = nil and lockedRound = -1. The proposer of round 1 has a really good
+		connection with the rest of the network, thus it is able to send the proposal to its peers before they receive enough
+		prevotes from round 0 to form a quorum.	Now the proposal will only be able to satisfy line 28 of the Tendermint
+		pseudo code, however, the prevotes from the previous round are yet to arrive. Since the proposal of the current
+		round has been received the timer would be stopped.
```
piersy, line 589:
```suggestion
		prevotes from round 0 to form a quorum. Now the proposal received by the peers would only be able to satisfy line
		28 of the Tendermint pseudo code, however, the prevotes from the previous round are yet to arrive. Since the
		proposal of the current round has been received the timer would be stopped.
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		 return err
+
+		Now the client is stuck since the timer has been stopped thus a prevote nil cannot be sent and the timer cannot be
+		restarted until startRound() is called for a new round. The resending of the message set will also not help because
+		it would only send messages to peers which they haven't seen and since there were no new messages the peers will not
+		be able to make progress. This can also happen where the client's lockedRound < vr, it cannot happen for
+		lockedRound = vr because that means the client had received enough prevote in a timely manner and there are no old
+		prevote to arrive.
+
+		Therefore we have a liveness bug in the current implementation of Tendermint.
```
piersy, line 618:
```suggestion
		Without the line 28 upon conditoin the client is stuck since the timer has been stopped, thus a prevote nil
		cannot be sent and the timer cannot be restarted until startRound() is called for a new round. The
		resending of the message set will also not help because it would only send messages to peers which they
		haven't seen and since there were no new messages the peers will not be able to make progress. This can
		also happen where the client's lockedRound < vr, it cannot happen for lockedRound = vr because that means
		the client had received enough prevote in a timely manner and there are no old prevote to arrive.

		Therefore we had a liveness bug in implementations of Tendermint in commits prior to this one.
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		Round 1:
+			The proposer of round 1 had received quorum prevotes for the proposal of round 0, thus, the proposer of round 1
+		re-proposes the proposal with vr = 0, More than quorum of the network is yet to receive the prevotes from the
+		previous round, thus the are their lockedValue = nil and lockedRound = -1. The proposer of round 1 has a really good
+		connection with the rest of the network, thus it is able to send the proposal to its peers before they receive enough
+		prevotes from round 0 to form a quorum.	Now the proposal will only be able to satisfy line 28 of the Tendermint
+		pseudo code, however, the prevotes from the previous round are yet to arrive. Since the proposal of the current
+		round has been received the timer would be stopped.
+
+		autonity/consensus/tendermint/core/propose.go, Lines 131 to 133 in 78f199d
```
piersy, line 591:
```suggestion
		autonity/consensus/tendermint/core/propose.go, Lines 131 to 133 at 78f199d
```

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		c.lockedValue = clientLockedValue
+		c.validValue = clientLockedValue
+
+		//schedule the proposer timeout since the client is not the proposer for this round
+		c.proposeTimeout.scheduleTimeout(1*time.Second, c.Round(), c.Height(), c.onTimeoutPropose)
+
+		backendMock.EXPECT().VerifyProposal(*proposal.ProposalBlock).Return(time.Duration(1), nil)
+		backendMock.EXPECT().Sign(prevoteMsgRLPNoSig).Return(prevoteMsgToBroadcast.Signature, nil)
+		backendMock.EXPECT().Broadcast(context.Background(), committeeSet.Committee(), prevoteMsgRLPWithSig).Return(nil)
+
```
piersy, line 677: This setup is quite involved, I feel like this test would be easier to follow and more robust if instead of manually constructing a specific state in core and then submitting messages, we actually created the state by submitting messages. I.E we totally recreate the scenario.

This manual setup is liable to break every time we change the core implementation, and its also very possible that theer could be a mistake in this setup which would render the test invalid. If we instead achieved the desired state by passing messages to core then we solve these issues.

aneequesafdar: The idea behind these tests was to write unit tests for all possible cases of upon conditions from the perspective of one node, where we want to ensure that if conditions for upon conditions are fulfilled the state is changed correctly. This is by no mean a comprehensive list test case rather only a bare minimum of cases derived from how the upon conditions can be satisfied. This also took into account the current implementation of how Tendermint is implemented and I agree it may not be the best way to test things.

However, I don't think your suggestion will lead to any less involved set up, I think it would be even more involved since you will have to manage different node separately and create and control numerous messages and timeouts. This would also require massive changes to the current implementation of Tendermint. 

Although I agree with your following assessment

> This manual setup is liable to break every time we change the core implementation, and its also very possible that there could be a mistake in this setup which would render the test invalid. If we instead achieved the desired state by passing messages to core then we solve these issues.

Therefore, I see this comment as a separate issue otherwise the changes suggest would be too large and not be inline with what this pull request is trying to achieve.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		c.setHeight(currentHeight)
+		c.setRound(currentRound)
+		c.setStep(propose)
+		c.lockedRound = clientLockedRound
+		c.validRound = clientLockedRound
+		c.lockedValue = clientLockedValue
+		c.validValue = clientLockedValue
+
+		//schedule the proposer timeout since the client is not the proposer for this round
+		c.proposeTimeout.scheduleTimeout(1*time.Second, c.Round(), c.Height(), c.onTimeoutPropose)
```
piersy, line 672: Why not use `startRound` here instead of manually starting the timer?

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		assert.False(t, c.proposeTimeout.timerStarted())
+
+		// now we receive the last old round's prevote MSG to get quorum prevote on vr for value v.
+		// the old round's prevote is accepted into the round state which now have the line 28 condition satisfied.
+		// now to take the action of line 28 which was not align with pseudo code before.
+		sender := 0
+		err = c.handleCheckedMsg(context.Background(), prevoteMsg, members[sender])
+		assert.Nil(t, err)
+
+		assert.Equal(t, prevote, c.step)
```
piersy, line 692: Should we not check height and round here as well?

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+
+		// check timer was stopped after receiving the proposal
+		assert.False(t, c.proposeTimeout.timerStarted())
+
+		// now we receive the last old round's prevote MSG to get quorum prevote on vr for value v.
+		// the old round's prevote is accepted into the round state which now have the line 28 condition satisfied.
+		// now to take the action of line 28 which was not align with pseudo code before.
+		sender := 0
+		err = c.handleCheckedMsg(context.Background(), prevoteMsg, members[sender])
+		assert.Nil(t, err)
```
piersy, line 690:
```suggestion
		assert.NoError(t, err)
```
You could make this change in a number of other places.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -570,6 +570,131 @@ func TestOldProposal(t *testing.T) {
...
+		backendMock.EXPECT().Sign(prevoteMsgRLPNoSig).Return(prevoteMsgToBroadcast.Signature, nil)
+		backendMock.EXPECT().Broadcast(context.Background(), committeeSet.Committee(), prevoteMsgRLPWithSig).Return(nil)
+
+		// now we handle new round's proposal with round_p > vr on value v.
+		err := c.handleCheckedMsg(context.Background(), proposalMsg, members[currentRound])
+		assert.Nil(t, err)
+
+		// check timer was stopped after receiving the proposal
+		assert.False(t, c.proposeTimeout.timerStarted())
+
```
piersy, line 684: I think we should check height,round,step and the locked values here as well.

piersy (changes requested):
Some more suggestions.

consensus/tendermint/core/upon_condition_test.go
aneequesafdar: Initially, I did use start round however, it calls backend functions when then need to be mocked, so that's why I decided to start the timer manually. Then I also need to take into consideration that the sender is not the proposer otherwise it would start sending proposal.

consensus/tendermint/core/upon_condition_test.go
aneequesafdar: Why is checking for no error better than checking for nil?

consensus/tendermint/core/upon_condition_test.go
aneequesafdar: I thought we should only check for things we expect to change and if they haven't then that is a problem, checking for things that we don't expect to change may lead to checking for negative which may never lead meaningful testcase, since you would be checking everything in every testcase. I may be wrong, I don't know.

consensus/tendermint/core/upon_condition_test.go
aneequesafdar: I don't think so because the behaviour being captured doesn't expect the round or height to change.

consensus/tendermint/core/upon_condition_test.go
piersy: When the test fails, the error is displayed differently to just a generic object, which makes the failure easier to read.

consensus/tendermint/core/upon_condition_test.go
piersy: That's my point, maybe the height or the round changed at this point. We don't know if it did or not though because we didn't check.

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: @piersy I replaced them with the one you suggested to get an easier read once case failed.

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: In the pseudo code, line 28 and its actions, it does not specified a round or height changes is part of the action, we don't have to check it here.

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: > I think we should check height,round,step and the locked values here as well.

Let's just check the action is align with pseudo code, otherwise it would make it hard to understand.

piersy: I think the example we discussed previously was in the case of a function that returned a value and an error.
E.G:
```golang
func foo() (val, error)
```

And there was a test that was testing that when called, `foo` returned no error.
E.G:
```golang
_, err := foo()
assert.NoError(t, foo)
```

The point I made about that test is that it wasn't really testing that anything was happening because the following implementation of `foo` would pass the test.

```golang
func foo() (val, error) {
	return nil, nil
}
```

So in that case, I said we need to check the value returned from `foo` to know that `foo` actually did something. In that case we were expecting that `foo` would take some action, so we needed to assert that action had been taken.

In this case we want to assert that no action has been taken here otherwise we cannot be sure that the correct action was taken when the prevote message was received.

In general when testing you want to check the things that change, because as you point out there are almost infinite things that don't change and most of the time we are not interested in those things, and alse we can reason that they won't be changed because we are not running code that would change them.

But in this case of testing tendermint, we have (height, round, step, lockedValue, lockedRound, validValue, validRound), every message processed by core has the potential to change those values and it is really important to us that those values change in the correct way in response to a message, even if that change is no change at all.

So I would suggest that we should always check all of these values after each message is processed to verify that the tendermint state is being manipulated as expected at each step.

consensus/tendermint/core/upon_condition_test.go
piersy: This is the kind of problematic setup that I was mentioning [here](https://github.com/clearmatics/autonity/pull/615/files#r479351858), we are not running full core, we are just providing enough setup to get core to behave in the way we want it to in this test.

But it's difficult to know if core would behave in the same way in a real situation, what other actions are taken by `startRound` that we have not taken here that could potentially affect the way that core behaves to the messages handled on lines 679 and 689?

It is for this reason that I would favour tests that perform their setup by fully exercising a full instance of core. But currently this is not really feasible because of the dependency on Backend, which requires pretty much a full node to be able to be correctly instantiated.

consensus/tendermint/core/upon_condition_test.go
piersy: Your assumption is that round and height won't change in respect to the message sent at line 689, but you need to verify that assumption, as it stands this test doesn't let us know that height and round did not change.

It could be that height and round are changing  in respect to the message sent at line 689 and if that is the case it indicates a problem with the implementation.

consensus/tendermint/core/upon_condition_test.go
aneequesafdar: I agree with your concerns. This the best we can do for the time being and it did allow us to find some problems,

piersy (approved):
This is looking good now.

Issue #619, opened 2020-07-20
Fix TestTendermintMajorityExternalUsersFullyConnected


piersy:
# Problem

These tests are intermittently failing due to a goroutine leak.

They seem to pass on my machine but fail on CI.

 The below log extract was taken from the github [actions run for the `0.5.0` release](https://github.com/clearmatics/autonity/actions/runs/164686968) showing the failure of TestTendermintMajorityExternalUsersFullyConnected
```
2020-07-10T14:53:33.6621675Z [31mERROR[0m[07-10|14:53:33.661] Error stopping Consensus Engine          [31merror[0m="stopped engine"
2020-07-10T14:53:35.4459616Z --- FAIL: TestTendermintMajorityExternalUsersFullyConnected (154.45s)
2020-07-10T14:53:35.4460520Z     --- FAIL: TestTendermintMajorityExternalUsersFullyConnected/test_case_no_malicious (154.45s)
2020-07-10T14:53:35.4687517Z ##[error]        leaks.go:78: found unexpected goroutines:
2020-07-10T14:53:35.4718455Z             [Goroutine 6772 in state select, with github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleNewUnminedBlockEvent on top of the stack:
2020-07-10T14:53:35.4718685Z             goroutine 6772 [select]:
2020-07-10T14:53:35.4718984Z             github.com/clearmatics/autonity/consensus/tendermint/core.(*core).handleNewUnminedBlockEvent(0xc000233e00, 0x13a3fa0, 0xc0f4da4300)
2020-07-10T14:53:35.4719270Z             	/home/runner/work/autonity/autonity/consensus/tendermint/core/handler.go:101 +0xfd
2020-07-10T14:53:35.4719509Z             created by github.com/clearmatics/autonity/consensus/tendermint/core.(*core).Start
2020-07-10T14:53:35.4719740Z             	/home/runner/work/autonity/autonity/consensus/tendermint/core/handler.go:43 +0x1e5
2020-07-10T14:53:35.4719873Z             
2020-07-10T14:53:35.4720165Z              Goroutine 6773 in state select, with github.com/clearmatics/autonity/consensus/tendermint/core.(*core).mainEventLoop on top of the stack:
2020-07-10T14:53:35.4720338Z             goroutine 6773 [select]:
[...]

PR #621, opened 2020-07-20
Prevent unintended test runs in ci

piersy:
When using a github actions matrix to run tests we were specifying
specific tests to run by name. Using this approach 'go test' will run
all tests with a matching prefix, which  in our case was causing us to
run tests more than once.

For example the following 2 matrix entries would cause
TestTendermintMajorityExternalUsersFullyConnected to be executed on each
run.

  TestTendermintMajorityExternalUsers,
  TestTendermintMajorityExternalUsersFullyConnected,

This commit updates the matching regex to match the word boundaries at
the beginning and end of the test name so that we only run the specified
test.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -299,7 +299,8 @@ jobs:
         run: make embed-autonity-contract
 
       - name: Test
-        run: go test ./consensus/test/... -v -run='${{ matrix.e2e }}' -timeout 40m
+        # \b matches a word boundary
+        run: go test ./consensus/test/... -v -run='\b${{ matrix.e2e }}\b' -timeout 40m
 
   race_tests:
     needs: bootstrap-checkout


aneequesafdar:
How did you find out that the tests were being run multiple times? Also, base on this https://www.regular-expressions.info/wordboundaries.html I am not sure how word boundaries are helping because a you would not be able to search for a specific word in a camel case string.

piersy:
@aneequesafdar When I was investigating the intermittent test failures #619 looking at the logs I saw that actually in the ci failures for TestTendermintMajorityExternalUsers they were in fact failing because TestTendermintMajorityExternalUsersFullyConnected was failing.

Capital letters are not word boundaries.

The definition from the page you linked is.

> * Before the first character in the string, if the first character is a word character.
> * After the last character in the string, if the last character is a word character.
> * Between two characters in the string, where one is a word character and the other is not a word character.

These word boundaries are occurring at the beginning and end of the string. There are no word boundaries inside the test names because test names are made purely of word characters.

aneequesafdar:
After discussing with @piersy I understood the matrix is used as a list to enumerate all the local e2e tests we want to run as separate parallel jobs, however, it is the `go` command which performs the match on regex test cases name present in the package which causes the `go test` command to run the test multiple times if a prefix is shared by multiple tests.

piersy:
I'm closing this because it would be made redundant by the work I have been doing to speed up Ci

PR #622, opened 2020-07-20
Fix docker image build

### LINKED ISSUE #618: Docker image build failing
### Problem

The Autonity 0.5.0 docker image build is failing in dockerhub.

logs (https://hub.docker.com/repository/registry-1.docker.io/clearmatics/autonity/builds/4a403b58-e02c-47b3-8cc0-019991db1142?ref=login) :
```
Cloning into '.'...
Warning: Permanently added the RSA host key for IP address '140.82.114.4' to the list of known hosts.
Submodule 'tests' (https://github.com/ethereum/tests) registered for path 'tests/testdata'
Cloning into 'tests/testdata'...
Submodule path 'tests/testdata': checked out '5841af6da472fb3f19810354cf9a30afd8e72b5f'
Switched to a new branch 'master'
Pulling cache layers for index.docker.io/clearmatics/autonity:latest...
Done!
KernelVersion: 4.4.0-1060-aws
Components: [{u'Version': u'19.03.8', u'Name': u'Engine', u'Details': {u'KernelVersion': u'4.4.0-1060-aws', u'Os': u'linux', u'BuildTime': u'2020-03-11T01:24:30.000000000+00:00', u'ApiVersion': u'1.40', u'MinAPIVersion': u'1.12', u'GitCommit': u'afacb8b7f0', u'Arch': u'amd64', u'Experimental': u'false', u'GoVersion': u'go1.12.17'}}, {u'Version': u'1.2.13', u'Name': u'containerd', u'Details': {u'GitCommit': u'7ad184331fa3e55e52b890ea95e65ba581ae3429'}}, {u'Version': u'1.0.0-rc10', u'Name': u'runc', u'Details': {u'GitCommit': u'dc9208a3303feef5b3839f4323d9beb36df0a9dd'}}, {u'Version': u'0.18.0', u'Name': u'docker-init', u'Details': {u'GitCommit': u'fec3683'}}]
Arch: amd64
BuildTime: 2020-03-11T01:24:30.000000000+00:00
ApiVersion: 1.40
Platform: {u'Name': u'Docker Engine - Community'}
Version: 19.03.8
MinAPIVersion: 1.12
GitCommit: afacb8b7f0
Os: linux
GoVersion: go1.12.17
Starting build of index.docker.io/clearmatics/autonity:latest...
Step 1/9 : FROM golang:1.14-alpine as builder
---> 3289bf11c284
Step 2/9 : RUN apk add --no-cache make gcc musl-dev linux-headers libc-dev git
---> Running in 552ed13beebf
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
(1/21) Upgrading musl (1.1.24-r8 -> 1.1.24-r9)
(2/21) Installing libgcc (9.3.0-r2)
(3/21) Installing libstdc++ (9.3.0-r2)
(4/21) Installing binutils (2.34-r1)
(5/21) Installing gmp (6.2.0-r0)
(6/21) Installing isl (0.18-r0)
(7/21) Installing libgomp (9.3.0-r2)
(8/21) Installing libatomic (9.3.0-r2)
(9/21) Installing libgphobos (9.3.0-r2)
(10/21) Installing mpfr4 (4.0.2-r4)
(11/21) Installing mpc1 (1.1.0-r1)
(12/21) Installing gcc (9.3.0-r2)
(13/21) Installing nghttp2-libs (1.41.0-r0)
(14/21) Installing libcurl (7.69.1-r0)
(15/21) Installing expat (2.2.9-r1)
(16/21) Installing pcre2 (10.35-r0)
(17/21) Installing git (2.26.2-r0)
(18/21) Installing musl-dev (1.1.24-r9)
(19/21) Installing libc-dev (0.7.2-r3)
(20/21) Installing linux-headers (5.4.5-r1)
(21/21) Installing make (4.3-r0)
Executing busybox-1.31.1-r16.trigger
OK: 163 MiB in 35 packages
Removing intermediate container 552ed13beebf
---> 96cbbe4d0931
Step 3/9 : ADD . /autonity
---> 167e2c8c5250
Step 4/9 : RUN cd /autonity && make autonity
---> Running in 0b32f500cf30
mkdir -p ./common/acdefault/generated
[91mmake: sudo: No such file or directory
make: *** [Makefile:47: common/acdefault/generated/bytecode.go] Error 127
[0m
sudo docker run --rm -v /autonity/./contracts/autonity/contract/contracts:/contracts -v /autonity/./common/acdefault/generated:/output ethereum/solc:0.6.4 --overwrite --abi --bin -o /output /contracts/Autonity.sol
Removing intermediate container 0b32f500cf30
The command '/bin/sh -c cd /autonity && make autonity' returned a non-zero code: 2
```

### Acceptance criteria

The image can be built on dockerhub.

### Question

Could we have this build integrated into our own C.I.?

piersy:
Fixes #618

The docker file build was failing since we had updated the Makefile to run docker with or without sudo based on whether the current user was in the docker group. The assumption being that if you are not in the docker group you need to use sudo, which works sometimes but not in the case that the user is root and sudo is not installed.

This PR ensures that if the user is root, we do not try to use sudo.

It also adds a Makefile rule to build the docker image and a ci job to build the docker image.

This PR changed a bit during the course of its existence.

A number of other problems were found and resolved.

* The build would fail if we tried to run docker during a docker build. This would happen when running the embed-autonity-contract rule(it ran a dockerised solc to compile the contract), but it would only happen if make detected that the generated contract code was out of date, so we could avoid running this step inside the docker build by running `make embed-autonity-contract` before the docker build so that the files added to the docker container included up to date generated contract code.

* If the systems 'sh' shell was bash the embed-autonity-contract rule would generate invalid code. This is due to the bash implementation of echo not interpreting `\n`as a newline unless it is provided the `e` flag. This was resolved by removing use of `\n` and instead using gofmt to fmt the generatd files.

* @eastata made us aware that devops require docker builds to succeed on a host with only docker installed, this meant we couldn't rely on running `make embed-autonity-contract` before running the docker build. This was solved by not using a dockerised version of solc. The reason we chose a dockerised version of solc to begin with was it downloaded much more quickly than the binary from github and it meant the process supported both mac and linux machines. As we now no longer have any team members using mac its not so important to support it.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -369,3 +369,30 @@ jobs:
 
       - name: test contract
         run: make test-contracts
+
+  # Builds the docker image. We don't intend to use the built image as part of
+  # a workflow, but we want to know the image can be built.
+  docker-image-build:
+    # Don't try to build the docker image unless autonity itself builds.
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Cache code
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-autonity-code
+        with:
+          path: /home/runner/work/autonity/autonity/
+          key: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
+
+      - uses: actions/checkout@v2
+        if: steps.cache-autonity-code.outputs.cache-hit != 'true'
+        with:
+          clean: 'false'
+
+      - name: Embed Autonity contract
+        if:  steps.cache-autonity-code.outputs.cache-hit != 'true'
+        run: make embed-autonity-contract
+
+      - name: Build
+        run: make build-docker-image
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -4,6 +4,18 @@ jobs:
   bootstrap-checkout:
     runs-on: ubuntu-latest
     steps:
+      - name: Set up Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Set GOPATH
+        run: |
+          echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+          echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+        shell: bash
+
       - name: Cache code
         uses: actions/cache@v1
         id: cache-autonity-code
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -383,28 +383,17 @@ jobs:
         run: make test-contracts
 
   # Builds the docker image. We don't intend to use the built image as part of
-  # a workflow, but we want to know the image can be built.
+  # a workflow, but we want to know the image can be built. We don't want to
+  # use any cache here since we want a clean checkout before any code
+  # generation has taken place.
   docker-image-build:
-    # Don't try to build the docker image unless autonity itself builds.
     runs-on: ubuntu-latest
+    # Don't try to build the docker image unless autonity itself builds.
     needs: build
     steps:
-      - name: Cache code
-        uses: actions/cache@v1
-        env:
-          cache-name: cache-autonity-code
-        with:
-          path: /home/runner/work/autonity/autonity/
-          key: ${{ runner.os }}-aut-${{ env.cache-name }}-${{ github.sha }}
-
       - uses: actions/checkout@v2
-        if: steps.cache-autonity-code.outputs.cache-hit != 'true'
         with:
           clean: 'false'
-
-      - name: Embed Autonity contract
-        if:  steps.cache-autonity-code.outputs.cache-hit != 'true'
-        run: make embed-autonity-contract
 
       - name: Build
         run: make build-docker-image
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Build Autonity in a stock Go builder container
 FROM golang:1.14-alpine as builder
 
-RUN apk add --no-cache make gcc musl-dev linux-headers libc-dev git
+RUN apk add --no-cache make gcc musl-dev linux-headers libc-dev git perl-utils
 
 ADD . /autonity
 RUN cd /autonity && make autonity
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 .PHONY: autonity-windows autonity-windows-386 autonity-windows-amd64
 
 NPMBIN= $(shell npm bin)
-GOBIN = ./build/bin
+BINDIR = ./build/bin
 GO ?= latest
 LATEST_COMMIT ?= $(shell git log -n 1 develop --pretty=format:"%H")
 ifeq ($(LATEST_COMMIT),)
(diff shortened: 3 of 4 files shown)


eastata:
I can propose for this PR 
**acceptance criteria:** Host for docker build must not have any requirements except `docker` installed.

To resolve it we can decompose make function `embed-autonity-contract` for 2 steps:

* `compile-autonity-contract` - should literaly only compile autonity contract using dockerised solc and save output as a file
* `embed-autonity-contract` - shuld only process compiled contract that already exist form file to go-files `abi.go` and `bytecode.go`

How the workflow can looks like:


# Workflow all-in-docker
Should works at any hosts without any additional requirements except docker installed. GitHub Actions/DockerHub/Local Linux/JFrog/other Docker-based CI
Triggered by `docker build -t autonity .`

```
FROM ethereum/solc:0.6.4 as solc-compiler

# Here we are starting with socl environment and compile contract

FROM golang:1.14

# Here we are using golang environment
# Copy compiled contract from previous image to current env

COPY  --from=solc-compiler /compiled-contract /compiled-contract

RUN make embed-autonity-contract
RUN make ...other_build_steps

ENTRYPOINT [...]
```

# Workflow partially-in-docker
Should works for local developer workstation, GitHub Actions with cached go-dependencies
Triggered by `make`

* make compile-autonity-contract `# If need`
* make embed-autonity-contract
* make ...other_build_steps

piersy:
Hi @eastata Thanks for the input.

The acceptance criteria were not something I was previously aware of, is there a process for making people aware of the devops acceptance criteria?

E.G when a new person joins the company, how can they know that any deployable they work on must be docker buildable with only docker installed?

The core problem we are facing here is that we use docker as part of our normal build process and you cannot run docker during a docker build and additionally the build step using docker is invoked with make and so cannot be run outside of the docker build on a machine that only has docker installed.

So to comply with the acceptance criteria you can either duplicate the dockerised build steps in the docker build file (as you are suggesting) or you can avoid using docker as part of your normal build process, so that all the normal build process can be run during docker build.

In this case it turns out that it is not too hard to switch to a non docker approach for the embed-autonity-contract step. So I think this is a good option, since since in this way we avoid duplicating elements of our build process in the Dockerfile.

Out of interest I would like to know when is devops actually building the docker images, given that they are available via docker hub?

eastata:
@piersy  You have made non reviewed by DevOps changes that lead us to broke current build.
To fix it you are offering workaround:

**Option 1:** switch to a non docker approach
*  **(-)** This workaround add for us (as DevOps) and for 3rd party users complexity. With tis approach build require to setup and support specific environment instead just run `docker build` with all dependencies included inside.
* **(+/-)** It reduce duplication for compiling contract, but add it through yet another one building environment. (DockerHub env != GitHub Actions env)

**Option 2:** (described in my previous comment)
*  **(+)** We can keep possibility to build autonity without complex build environment setup, just using only docker
*  **(-)** It contain duplication step that use the same command, binary and solidity docker image. The same duplication as we already have for dockerized Go environment and GitHub Action build environment

**Option 3:** Rewrite GitHub Actions and tests to docker-only
* **(+)** We will have a guaranty that tested image absolutely the same as used in real infrastructure 
* **(+)** We should not support 2 different environments
* **(-)** It require DevOps time to rewrite and test all
* **(-)** Building and testing time could increase if we are loosing GitHub Actions go modules cache

I can offer evaluate profits and minuses of this options and choose one.
My personal vote for **Option 2** now as optimal balance

Issue #72, opened 2019-04-15
Network Permissionning


yazzaoui:
Opening this issue to describe and discuss the current network permissionning approach.

### Rationale 

- The Autonity client is built around and for private network, hence network permissionning is a fundamental feature. 
- The network is symmetric, all the participant in the network have by default the same privileges than the others. Therefore [mutual authentication](https://en.wikipedia.org/wiki/Mutual_authentication) is required.
 
### Description and rules

Currently, network permissionning is using the Glienecke smart-contract interface to retrieve the list of authorized enodes. However, in a blockchain context, where there is no guarantee of synchronized node state, the enforcement of is not straightforward: the NodeID by itself is not sufficient, we need to know as well how far ahead of us a peer is, and this information can be only retrieved post Ethereum sub-protocol handshake via the total difficulty parameter.    
I will give some example later.

Description of the currently implemented protocol:

- The Ethereum service is updating an internal whitelist via the Glienecke contract before block insertion.
- We try dialing every enodes present in the whitelist on a periodic basis. 
- At DevP2P layer, RLPx ensures that a peer possesses the private key associated with his disclosed NodeId. ECDH is being used to establish a secure shared AES session key for encrypting all futures packets.   
- After successful authentication at DevP2P layer, ETH sub protocol will initiate a handshake, the current total difficulty will be exchanged, which is equal to the chain height with xBFT.  Then the following rules are enforced :

1. If the peer total difficulty is greater than ours we accept the incoming connection
2. If the peer is present in the local whitelist we accept the incoming connection. Previously I was only checking the NodeID but @JekaMas introduced a check on the IP address as well.

Otherwise, the connection is rejected.

### Potential issue and vulnerabilities

- The current implementation is unfinished, we need to stop relaying data when synchronizing. Indeed a malicious peer could fake its TD or a connection can be attempted on a previously authorized peer.   

I will update the document to reflect as well the changes that were made with the new enode format.

### Existing attacks
* https://github.com/clearmatics/autonity/pull/71

JekaMas:
Looks like that step `If the peer total difficulty is greater than ours we accept the incoming connection` needs an additional check. Maybe we should ask a new peer to send the highest block header with valid signs and the proven state of the Glienecke contract where the peer is listed in the whitelist. And only if the peer managed to proof its state in the list we should start sync process.

yazzaoui:
[...]

PR #625, opened 2020-07-22
[Draft PR for the review and future discussion] Network permission delta

Jason-Zhangxin-Chen:
To fix https://github.com/clearmatics/autonity/issues/72, @yazzaoui , @aneequesafdar , @piersy , please review this draft PR  for the network permission enhancement which fill the gap between the implemenation and our yellow paper. It is base on my current knowledge of ETH sub-protocol and our yellow paper, please share your comments if I missed something to make sure we are in the same view of this enhancement. The test is still missing since this is a prototype.

1. Manage an untrusted peer list in p2p network layer.
2. Apply permission for sensitive msg handling base on untrusted peer on the Eth sub-protocol layer.
3. Apply permission for consensus msg and consensus state msg handling base on consensus layer (tendemint).

Note:
1. The removal of node already align with the yellow paper since the removed member will not be selected as next consensus instance's committee member.
2. Things to be discussed after today's standup:
 2.1 Wrong statement in pseudo code in Autonity Yellow Paper.
    It confuse somehow when you have a first glance of it.
    Figure 6: Consensus state synchronization module at participan pi, line 10. The variable pi should be replace with pj.
    ` if (pi ∈/ connected peersi ∪ untrusted peersi ∨ pj ∈/ getCommittee(blockchaini)) then return endIf`
 2.2 The triggering of stateSynch() in Figure 5: Block synchronization module at participant pi, line 19.
    In current implementation, it is triggered by conensus layer after blockchain get synced and consensus core is ready once node is startup, the other trigger point is base on a fix timer which will check if chain is onhold to sync the consensus state msg. It have the same functionality, do we need to align with the yellow paper strictly to trigger this follow.
 2.3 Do we need to apply permission handling for untrusted peer of those extra sensitive msg of ETH sub-protocol:
    GetBlock, GetHeader from untrusted peer, ....


piersy:
@Jason-Zhangxin-Chen @yazzaoui  It would be good to have a high level description of what this permissioning is trying to achieve. I have looked at the Yellow paper and #72 and I still don't have a clear idea of what is the purpose of the network permissioning.

Is it that we restrict each participant's view of the network data to be all data between the genesis block and that last block in which they were part of the whitelist? And that entities that are not participants (I.E have never been part of the whitelist) do not get to view any data at all?

Jason-Zhangxin-Chen:
Close this PR since it was used to describe how it was implemented, it is still under testing with couple of testcase planned.

Issue #624, opened 2020-07-21
Intermittent panic in TestTendermintStartStopFPlusTwoNodes


piersy:
# Problem
The test fails intermittently and very rarely in this way, so far we have only seen it fail in the following run.
https://github.com/clearmatics/autonity/pull/623/checks?check_run_id=893465478

The failure log is below
```
ERROR[07-21|09:09:02.092] NewUnminedBlockEvent: old height unminedBlock addr=0x65dEdbae57a5BB4CaddB1A2f4da98196caD4C383 err="old height message"
ERROR[07-21|09:09:02.583] last mined block                         addr=0x7448D58eA926515F28f9E0D4f5831f0acd8f9013 idx=VC peer=VC num=5  hash="ea7fb5…63912e" txCount=6
ERROR[07-21|09:09:06.161] last mined block                         addr=0x8a653AE4DcD64eC5247bE1560C3a20e4C002cdc5 idx=VB peer=VB num=6  hash="ae5dc5…0e219a" txCount=6
ERROR[07-21|09:09:06.162] last mined block                         addr=0x65dEdbae57a5BB4CaddB1A2f4da98196caD4C383 idx=VE peer=VE num=6  hash="ae5dc5…0e219a" txCount=6
ERROR[07-21|09:09:06.173] last mined block                         addr=0xF8a654dAFBcA036944e6E735F5CbC63b43b90BE8 idx=VA peer=VA num=6  hash="ae5dc5…0e219a" txCount=6
ERROR[07-21|09:09:06.212] last mined block                         addr=0xF8a654dAFBcA036944e6E735F5CbC63b43b90BE8 idx=VA peer=VA num=7  hash="464c93…d4e64b" txCount=0
ERROR[07-21|09:09:06.212] last mined block                         addr=0x7B2F0A20FDDa2a185112928da2d03d6ce10eD789 idx=VD peer=VD num=5  hash="ea7fb5…63912e" txCount=6
ERROR[07-21|09:09:06.216] last mined block                         addr=0x8a653AE4DcD64eC5247bE1560C3a20e4C002cdc5 idx=VB peer=VB num=7  hash="464c93…d4e64b" txCount=0
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x592fc8]

goroutine 82508 [running]:
[...]

PR #635, opened 2020-07-28
Add detailed panic to provide better feedback #624

piersy:
The panic mentioned in https://github.com/clearmatics/autonity/issues/624 is occurring when the
second parameter to big.Int.Add is nil which is in turn caused by
blockchain.GetTd returning nil for the parent of the current block. The
panic happens rarely and we need more knowledge of what is the current
state to better understand how it is ocurring.

This commit adds a much more detailed panic that includes a dump of the
child block and the specific arguments in the call to blockchain.GetTd.

### DIFF
--- a/consensus/test/start_stop_test.go
+++ b/consensus/test/start_stop_test.go
@@ -363,7 +363,7 @@ func TestTendermintStartStopFPlusOneNodes(t *testing.T) {
 }
 
 func TestTendermintStartStopFPlusTwoNodes(t *testing.T) {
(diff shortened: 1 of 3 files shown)


yazzaoui:
Yes, I was working on this one yesterday, the fetcher is calling asynchronously `BroadcastBlock` but it could be that we are stopping the node and the leveldb before this method gets actually called: `pm.blockchain.GetTd` will return nil if the db is closed.

Haven't found yet a better way to avoid that without introducing another lock.

piersy:
Closing since I believe I have fixed this issue in my work to speed up CI

PR #639, opened 2020-08-10
fix freezer hash comparison by correctly calculating header hash

### LINKED ISSUE #638: Blocks in the freezer's ancient store cannot be retrieved
# Problem

Once the network produced `ImmuntabilityTreshold` (default value is 90000) number of blocks, nodes would start adding old blocks to the freezer's ancient store to improve performance. However, in Autonity once nodes were added to the freezer they cannot be retrieved, each call for their retrieval would return `nil`. This meant new nodes added to the network were unable to synchronise with the rest of the network.

# Notes on Freezer's retrieval and addition of blocks

`GetBlockByNumber` uses the rawdb to get the canonical hash based on the number by calling `ReadCanonicalHash` and passing the blockchain's db, and the number for which the hash needs to be looked up.
 
The freezer has multiple tables whose names are defined as `const` in `core/rawdb/schema.go:70`. In the case of `ReadCanonicalHash` the `freezerHashTable` is used to find the hash. Each of these tables is of type `freezerTable` defined in `core/rawdb/freezer_table.go`, intended to manage the different aspects of the blockchain (such as the headers, hashes, bodies etc).
​
If the hash is not found in the freezer then the hash is looked up in the levelDB, however, if during the lookup the hash has been moved to the freezer then freezer is checked again for the hash, otherwise, an empty hash is returned, which mean the block doesn't exist in the node's database.
​
Once `ReadCanonicalHash` returns a hash, `GetBlock` is called using the hash and the number, it would check the block cache before asking the blockchain database for the block. Assuming the block is not available from the cache, then `rawdb.ReadBlock` would retrieve the header and body separately and then combining them before returning. `ReadHeader` and `ReadBody` are static functions which take a `ethdb.Reader` to look up the header and body. `ReadHeader` would first check the freezer for the header then levelDB and finally check the freezer (again, because the header may have been moved to the freezer during the lookup in levelDB), raw rlp values are stored in the database which is decoded into a header before returning. If the header is not found or there is an error while decoding nil is returned. In a similar fashion, `ReadBody` retrieves the block body. If for some reason either the header or body is nil then `ReadBlock` would return nil. Therefore, eventually, `GetBlockByNumber` would return nil.
​
`freezerdb` is defined in `core/rawdb/database.go` as:
​
```go
type freezerdb struct {
	ethdb.KeyValueStore
	ethdb.AncientStore
}
```
​
The `NewDatabaseWithFreezer` creates and returns an instance of `freezerdb`. `NewDatabaseWithFreezer` ensure the data in leveldb (abstracted as `KeyValueStore` here) and freezer's ancient store is consistent and is responsible for starting the `freeze(db)` in a separate go-routine which is responsible for periodically moving data from leveldb into flat files. The freeze go-routine is stopped when database's `Close()` is called which eventually will send a quit signal through the `quit` channel.
​
The `freeze(db)` will run an infinite loop where the `f.quit` channel and the `freezerRecheckInterval` decide when the for-loop is returned and when the next iteration is scheduled (respectively). `freezerRecheckInterval` is used to start a timer which is waited on along with the `f.quit` channel in a select statement. However, `backoff` variable is used to determine how often the timer is scheduled based on the data available to be moved to the freezer, this also ensures the files are not constantly written to with small amounts of data.
​
The `freezer(db)` first determines if there are blocks that can be frozen by checking the current block number is available, it's less than `ImmutabilityThreshold`, the block number being considered is more than the total number of frozen blocks, and the block header can be retrieved. If all these conditions are met then there are blocks that can be frozen. At any given point a maximum of `freezerBatchLimit` blocks can be frozen. All the block components' rlp (hash, header, Body, Receipts, Td) are appended to freezer's ancient store one by one until the `limit`.
​
Apart from the genesis block, all frozen blocks are deleted from the leveldb in a `batch` from both the canonical and side chains and if there are no more blocks to be frozen then the for loop schedules a timer for `freezerRecheckInterval`. In practice, the first `ImmutabilityThreshold` blocks would form `ImmutabilityThreshold/freezerBatchLimit` batches after which a batch would consist of only a few blocks since the `freezerRecheckInterval = time.Minute` and in a minute around 60 blocks are mined which is much less than the `freezerBatchLimit`. 

# Debugging

Branches: `reproduce_freezer_issue` (@Jason-Zhangxin-Chen) and ` recreate_freezer_issue` (@piersy) includes a test case which reproduce the above issue in e2e framework.

# Acceptance Criteria

Ensure the blocks can be retrieved from the freezer correctly.

aneequesafdar:
freezer stores raw rlp values and go-ethereum's header hash is
calculated by keccak256(rlp(header)), which is not true for autonity
header hash, where few fields are remove before creating the rlp and
the hash (please see core/types/block.go:155). Therefore, when freezer
reads raw rlp values from disk it needs to ensure that the hash matches
the one being requested (since forks can happen in go-ethereum and for
the same block number there can be different hashes, which is shouldn't
be the case in autonity, thus this can be seen as redundant) we need to
calculate the autonity header hash before comparison. This can only be
done be decoding the header and then creating its hash. The caller of
ReadHeaderRLP will most likely decode the header again which is not
efficient.

fixes #638 

edit
@piersy posted a question about this hashing here - https://github.com/ethereum/go-ethereum/pull/20195#issuecomment-679224715


core/rawdb/accessors_chain.go
```diff
@@ -281,12 +280,28 @@ func ReadHeaderRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValu
...
-	if len(data) > 0 && crypto.Keccak256Hash(data) == hash {
+	if len(data) > 0 && compareHeaderHash(data, hash) {
 		return data
 	}
 	return nil // Can't find the data anywhere.
 }
 
+func compareHeaderHash(data []byte, h common.Hash) bool {
+	// if block header is go-ethereum's original header and the hash matches then return
+	if crypto.Keccak256Hash(data) == h {
```
piersy, line 291: Why do we need this check ?

core/rawdb/accessors_chain.go
```diff
@@ -281,12 +280,28 @@ func ReadHeaderRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValu
...
 	// result in a not found error.
 	data, _ = db.Ancient(freezerHeaderTable, number)
-	if len(data) > 0 && crypto.Keccak256Hash(data) == hash {
+	if len(data) > 0 && compareHeaderHash(data, hash) {
 		return data
 	}
 	return nil // Can't find the data anywhere.
 }
 
+func compareHeaderHash(data []byte, h common.Hash) bool {
```
piersy, line 289: We should add a test or extend the following test to check that this works.

https://github.com/clearmatics/autonity/blob/b4d1f51febdd5cb6279fd8001d088cdbc1c10dc3/core/rawdb/accessors_chain_test.go#L365-L427

piersy (commented):
A couple of points to resolve before I can approve

core/rawdb/accessors_chain.go
aneequesafdar: Just in case someone run `ethash` on autonity, this is similar to how block hashes are created in https://github.com/clearmatics/autonity/blob/b4d1f51febdd5cb6279fd8001d088cdbc1c10dc3/core/types/block.go#L153-L168

Therefore, I am just trying to be consistent.

core/rawdb/accessors_chain.go
aneequesafdar: Please view 1210d036e6a753dbc803f632656491fbe6567b6f

core/rawdb/accessors_chain.go
aneequesafdar: > How could someone run ethash on autonity?

You can start autonity without a genesis file and that would create a network with ethash by default.

> I feel like that should be forbidden and then we don't need this condition

This is true for a lot of the code, so I am not sure what to do. If I were to remove this check then that would mean if the user run ethash on autonity their block will also be decoded twice when there is no need.

Jason-Zhangxin-Chen (approved):
LGTM.

core/rawdb/accessors_chain.go
piersy: I think that is fine, people should not be running ethash (if they want ethash they can run geth), are any of the tests using ethash, if so not having that check could slow them down significantly.

piersy: @aneequesafdar bump!

aneequesafdar: I don't think this applies anymore since we are returning both the raw rlp and decoded value and the places where the caller was decoding the value no longer decode, rather it uses the return value

PR #650, opened 2020-08-14
Fix memory leak in backlog

yazzaoui:
The memory leak issue was due to two reasons:
- The priority queue implementation was not freeing unused allocated memory. I've forked the dependency under the Clearmatics org and fixed the issues. Relevant commits : https://github.com/clearmatics/autonity_cookiejar/commit/8f65f30029fa2d0689c3f06d390c73cf91dd6370 and https://github.com/clearmatics/autonity_cookiejar/commit/6fc22bf072fec697291bd186b34850d329f02286
- The backlog was implemented as a `map[types.CommitteeMember]...` and the access function here https://github.com/clearmatics/autonity/blob/b4d1f51febdd5cb6279fd8001d088cdbc1c10dc3/consensus/tendermint/core/backlog.go#L78 is not internality performing a deep check, hence for the same validator, we could have a new useless allocation.

This PR is only fixing the external dependency, I will create a subsequent ticket for the second one, as there is a more important issue that I've found.
I've also changed along the way the type of the priority from float32 to int64 in order to avoid floating point precision issues, I've done some tests locally and the assert `float32(i+1) > float32(i)` is no longer valid for `i` greater than 1600000 approximately, which can potentially cause some messages in the backlog to not be processed after a certain block height.


aneequesafdar (commented):
I feel as though we are having to manage 2 dependences of essentially the same `prque`. In some parts of the codebase  `karalabe/cookiejar.v2` is used while in consensus package `autonity_cookiejar` is used. This means that we will need to remember to update `autonity_cookiejar` when the `karalabe/cookiejar.v2` is update adding even more overhead. Otherwise, any issues which exist in `autonity_cookiejar` would never be found nor fixed.

[This commit](https://github.com/clearmatics/autonity_cookiejar/commit/8f65f30029fa2d0689c3f06d390c73cf91dd6370) needs a testcase to ensure the priority queue has decreased in size.

[This commit](https://github.com/clearmatics/autonity_cookiejar/commit/6fc22bf072fec697291bd186b34850d329f02286)    goes against the intention of the package as mentioned here
https://github.com/clearmatics/autonity_cookiejar/blob/6fc22bf072fec697291bd186b34850d329f02286/collections/prque/prque.go#L9-L19

piersy:
@yazzaoui My preference between this PR and #652 is with #652.

The reason for this is that we are dropping the cookiejar dependency and we don't have to manage another external dependency (autuonity/cookiejar).

BTW it would have been a lot easier to review your changes on github.com/autonity/cookiejar if you had put them in a PR.

aneequesafdar:
closing in favour of #652

PR #652, opened 2020-08-16
replace *prque data structure with []*Message to manage backlog

aneequesafdar:


consensus/tendermint/core/backlog.go
```diff
@@ -75,59 +75,65 @@ func (c *core) storeBacklog(msg *Message, src types.CommitteeMember) {
...
-				src: src,
-				msg: msg,
-			})
+				backlog = append(backlog[:offset], backlog[offset+1:]...)
+				totalElemRemoved++
+			}
+			// We need to ensure that there is no memory leak by reallocating new memory if the original underlying
+			// array become very large and only a small part of it is being used by the slice. We need len(backlog) > 0
+			// check again since the backlog size can change and cause division by zero errors.
+			if len(backlog) > 0 && cap(backlog)/len(backlog) > capToLenRatio {
```
piersy, line 130:
```suggestion
			if len(backlog) == 0 || cap(backlog)/len(backlog) > capToLenRatio {
```

This covers the case when the backlog is empty

aneequesafdar: Surely `if len(backlog) == 0` condition can cause division by zero errors, or am I missing something? 

Its possible to have `len(s) == 0` and `cap(s) == 0` at the same time, while it is impossible to have `len(s)>0` and `cap(s)==0` (at the same time). I only want to check this condition when backlog is not empty.

piersy: To free the memory held by the backlog. As it stands when `len(backlog) == 0` we don't free anything, regardless of the magnitude of `cap(backlog)`.

In fact I think this would be a better approach.

```suggestion
			if cap(backlog)/capToLenRatio > len(backlog) {
```
It solves the division by 0 issue.

consensus/tendermint/core/backlog.go
```diff
@@ -75,59 +75,65 @@ func (c *core) storeBacklog(msg *Message, src types.CommitteeMember) {
...
-				continue
-			}
-			logger.Debug("Post backlog event", "msg", msg)
 
-			go c.sendEvent(backlogEvent{
-				src: src,
-				msg: msg,
-			})
+				backlog = append(backlog[:offset], backlog[offset+1:]...)
+				totalElemRemoved++
```
piersy, line 125: The continue statement in this loop is somewhat redundant since we could just move the removal from the backlog into the else block  and remove the continue statement.

aneequesafdar: I don't think I follow, I don't want to remove the element if the code path from line 110-113 is true

consensus/tendermint/core/backlog.go
aneequesafdar: Very nice catch! Forgot to think about the `len(s)==0 and cap(backlog) >> len(s)`. could you please also update the comment in your suggestion

consensus/tendermint/core/backlog.go
aneequesafdar: There is 2 different case, if there is an error from `checkMessage` which is not `errFutureHeightMessage || err == errFutureRoundMessage || err == errFutureStepMessage` then I need to remove the element but I am not going to send the message to the handler, however, the error is one of those 3 then I need to make sure that I don't remove it and move to the next iteration

piersy: If you move lines 124-125 to just below line 121 then you only remove the element when you actually send a backlog event, so you don't need a continue statement to skip over element removal, because there is nothing to skip over.

piersy: Actually maybe I'm reading this wrong, why are you removing old round messages from the backlog and not sending them?

aneequesafdar: Although I would say the `c.checkMessage` should never return any errors other than `errFutureHeightMessage || err == errFutureRoundMessage || err == errFutureStepMessage` since we are in complete control of how messages are added to the backlog

aneequesafdar: > Actually maybe I'm reading this wrong, why are you removing old round messages from the backlog and not sending them?

Because the old height/round/step should never be in the backlog in the first place

piersy: > Because the old height/round/step should never be in the backlog in the first place

In that case we should panic since we have encounteredva programming error.

aneequesafdar: A more nuanced view is that as long as `processBacklog` is called at the right time there shouldn't be any old height/round/step messages. Since `checkMessage` is called at different times, it is possible the round has changed causing `checkMessage` to return different errors on the same message.

piersy: Well, if that is the case then shouldn't those messages be sent as a backlog event.


Consider this situation:

Round 0
Nodes are able to lock v for round 0 but send precommits for nil.

Round 1
A node N in the propose step receives a prevote for v. It puts it in the backlog.
2f+1 other nodes send precommits for nil and move to the next round.
Node N receives f+1 messages from the next round and is able to move to the next round propose step. (the prevote for v in the backlog is discarded)

Round 2
Node N now receives 2f prevotes for v from the previous round. If N had processed the prevote for v it received in round 1 it would be able to send a prevote for v in this round, but  that message was dropped from the backlog.

piersy: Or this situation.

Round 0
Most nodes send precommit for nil and move to the next round.
Node N receives a precommit for v from round 1 and puts it in the backlog. N also receives f prevotes from round 1 so can move to round 1.

Round 1

2f+1 nodes send  precommit for v and f-1 nodes send precommit for nil, none of
them see the 2f+1 precommits for v but the all see at least 2f+1 precommits and
so set their timers and move to round 2.

Node N sees f+1 messages from round 2 and so moves to round 2.

Round2

f+1 nodes send prevotes.

N now receives 2f precommits for v from round 1, but has discarded the
precommit for v at round 1 that it had in its backlog and so cannot commit v.

aneequesafdar: Yes I agree that backlog will discard the message where a jump to the future has been made, however, we still have consensus state synchronisation which will ensure that if progress has not been made then peers will send all of their message set to the rest of the network. 

Also, this problem exist in the current implementation of backlog and I just replaced the data structure while keeping the behaviour the same. Seems like a premature optimisation now since there is an assumption that backlog will only be keeping 

You could send all the message to the handler which are not the future messages to solve this issue. However, this behaviour would be different from what exists now.

piersy: I wouldn't consider this an optimisation, this is a bug in the implementation of Tendermint, we are throwing away messages that should be processed.

In a perfect world with loads of time, I would say lets open a separate PR for this. But given our current timeline and the number of bugs we are facing I think it would be pragmatic to just fix this one here, also considering that block has been completely re-written by you, so we are not touching any other areas of the code or adding extra code to this PR.

I'm just thinking we have a lot of tests that timeout intermittently, this could be contributing to that, so lets fix it while our focus is on this part of the code.

consensus/tendermint/core/backlog.go
```diff
@@ -75,59 +75,62 @@ func (c *core) storeBacklog(msg *Message, src types.CommitteeMember) {
...
+				offset := i - totalElemRemoved
+				curMsg := backlog[offset]
+
+				r, _ := curMsg.Round()
+				h, _ := curMsg.Height()
+				if err := c.checkMessage(r, h, Step(curMsg.Code)); err != nil {
+					if err == errFutureHeightMessage || err == errFutureRoundMessage || err == errFutureStepMessage {
+						logger.Debug("Futrue message in backlog", "msg", curMsg, "err", err)
+						continue
+					}
```
piersy, line 113: I think this could be made a bit neater now.
```suggestion
				err := c.checkMessage(r, h, Step(curMsg.Code))
				if err == errFutureHeightMessage || err == errFutureRoundMessage || err == errFutureStepMessage {
					logger.Debug("Futrue message in backlog", "msg", curMsg, "err", err)
					continue
				}
```

piersy (changes requested):
Final couple of suggestions.

Remove the toPriority function in backlog, its not used any more.

run `go mod tidy` to update the modules, cookiejar is removed.

piersy (approved):
Looks good.

yazzaoui (approved):
LGTM

PR #654, opened 2020-08-20
ReadHeaderRLP returns both the encoded and decode header values

aneequesafdar:
This is in order to prevent decoding the value twice, therefore, it is
up to the caller to determine how to use these returned values.make


yazzaoui (approved):
LGTM

Jason-Zhangxin-Chen (approved):
LGTM.

PR #655, opened 2020-08-20
Fix helloworld

dtebbs:
I needed to add some extra dependencies in helloworld/Dockerfile to make the instructions in the README work.
Also added a github action to test it (adds some time to the CI, but worth being sure it always works smoothly as it's the first section of the docs).


piersy (commented):
Hi @dtebbs, thanks for contributing.

Could you elaborate on the problems you faced, and possibly include an excerpt of the errors you encountered.

I am able to use the helloworld example as expected on the develop branch by running.

```
docker-compose up -d
```

from within the `helloworld` directory.

dtebbs:
> Could you elaborate on the problems you faced, and possibly include an excerpt of the errors you encountered.

@piersy  Thanks for the reply.

The first of my commits adds the helloworld commands to the github actions, which reproduces my error, namely:
```
/bin/sh: json_pp: not found
```
(replaced by the equivalent error for git once the package containing json_pp dependency is added to Dockerfile).

See the action for the full console output:
https://github.com/clearmatics/autonity/runs/1009561999?check_suite_focus=true

After the second commit (to add the dependencies), the same command passes:
https://github.com/clearmatics/autonity/runs/1009571977?check_suite_focus=true

Let me know if you need any further info.

Jason-Zhangxin-Chen (approved):
LGTM

piersy:
Hi @dtebbs, thanks for the info.

I realise now that this is the same problem we faced and fixed with the autonity Docker image #622 , which is that we introduced a code generation step as part of the build process which make will run only if necessary. In my case the code generation had been performed on my local machine and so make did not attempt to perform the code generation during docker build. In your case you had not generated the code locally and so the code generation was attempted during docker build which highlighted the lacking dependencies.

I'm going to update this pr so that `helloworld` makes use of the autonity Docker image, which should resolve the issue.

helloworld/README.md
```diff
@@ -51,38 +54,61 @@ When the nodes have all been deployed and connected to each other, the `nodes-co
...
+```
+
+Here is an example of attaching a console to `autonity-node-1` from the host
+machine, note that the host port for a container that corresponds to 8545 can
+be found from the output of `docker-compose ps` in this case the port is 8541.
 
 ```bash
-$ autonity attach http://0.0.0.0:8541
+$
+../build/bin/autonity attach http://0.0.0.0:8541
```
dtebbs, line 95: Doesn't this still assume that autonity has been built on the host? AFAICT, this won't be the case after running just the commands above.

dtebbs: Sorry. Not directly related to this change, but it stood out to me when reading the doc:  `0.0.0.0` here seems like a strange address for a destination.  I guess it will pretty much always route to the local machine, but I wonder if something concrete like `localhost` could be less confusing. (For reference, `ping 0.0.0.0` fails on my machine).

piersy: Hi @dtebbs, yes it does, 

> Doesn't this still assume that autonity has been built on the host? AFAICT, this won't be the case after running just the commands above.

Hi @dtebbs yes it does, I'll update with build instructions for the autonity binary.

piersy: @dtebbs I've added instructions to build the autonity binary and also substituted localhost for 0.0.0.0. 0.0.0.0 is only a valid source address its not a valid destination address. I don't know why it was working for me.

dtebbs: Thanks @piersy.

helloworld has been extremely useful while testing zeth so it's great to have all this updated.

aneequesafdar (approved):

Jason-Zhangxin-Chen (approved):
LGTM

PR #656, opened 2020-08-21
Fix future height message handling

yazzaoui:
closes https://github.com/clearmatics/autonity/issues/651

- introduce the _untrusted backlog buffer_, that will contain future height messages to be correctly verified when the state reach the appropriate height.
- some refactoring to accommodate that.

### DIFF
--- a/consensus/tendermint/core/backlog.go
+++ b/consensus/tendermint/core/backlog.go
@@ -155,11 +155,17 @@ func (c *core) processBacklog() {
 		c.backlogs[src] = backlog
 
 	}
-	for _, msg := range c.backlogUnchecked[c.height.Uint64()] {
-		go c.sendEvent(backlogUncheckedEvent{
-			msg: msg,
-		})
-		c.logger.Debug("Post unchecked backlog event", "msg", msg)
+	for height, _ := range c.backlogUnchecked {
+		if height == c.height.Uint64() {
+			for _, msg := range c.backlogUnchecked[height] {
+				go c.sendEvent(backlogUncheckedEvent{
+					msg: msg,
+				})
+				c.logger.Debug("Post unchecked backlog event", "msg", msg)
+			}
+		}
+		if height <= c.height.Uint64() {
+			delete(c.backlogUnchecked, height)
+		}
 	}
-	delete(c.backlogUnchecked, c.height.Uint64())
 }
--- a/consensus/tendermint/core/backlog.go
+++ b/consensus/tendermint/core/backlog.go
@@ -93,7 +93,7 @@ func (c *core) storeUncheckedBacklog(msg *Message) {
 	// We discard the furthest ahead messages in priority.
 	if c.backlogUncheckedLen == MaxSizeBacklogUnchecked+1 {
 		maxHeight := msgHeight.Uint64()
-		for k, _ := range c.backlogUnchecked {
+		for k := range c.backlogUnchecked {
 			if k > maxHeight && len(c.backlogUnchecked[k]) > 0 {
 				maxHeight = k
 			}
@@ -159,7 +159,7 @@ func (c *core) processBacklog() {
 		c.backlogs[src] = backlog
 
 	}
-	for height, _ := range c.backlogUnchecked {
+	for height := range c.backlogUnchecked {
 		if height == c.height.Uint64() {
 			for _, msg := range c.backlogUnchecked[height] {
 				go c.sendEvent(backlogUncheckedEvent{
--- a/consensus/tendermint/core/backlog.go
+++ b/consensus/tendermint/core/backlog.go
@@ -18,9 +18,10 @@ package core
 
 import (
 	"github.com/clearmatics/autonity/common"
-	"github.com/clearmatics/autonity/core/types"
 	"math/big"
 )
+
+const MaxSizeBacklogUnchecked = 1000
 
 var (
 	// msgPriority is defined for calculating processing priority to speedup consensus
@@ -33,7 +34,9 @@ var (
 )
 
 type backlogEvent struct {
-	src types.CommitteeMember
+	msg *Message
+}
+type backlogUncheckedEvent struct {
 	msg *Message
 }
 
@@ -71,17 +74,44 @@ func (c *core) storeBacklog(msg *Message, src common.Address) {
 	}
 
 	logger.Debug("Store future message")
-
-	c.backlogsMu.Lock()
-	defer c.backlogsMu.Unlock()
-
 	c.backlogs[src] = append(c.backlogs[src], msg)
+}
+
+// storeUncheckedBacklog push to a special backlog future height consensus messages
+// this is done in a way that prevents memory exhaustion in the case of a malicious peer.
+func (c *core) storeUncheckedBacklog(msg *Message) {
+	// future height messages of a gap wider than one block should not occur frequently as block sync should happen
+	// Todo : implement a double ended priority queue (DEPQ)
+
+	msgHeight, errHeight := msg.Height()
+	if errHeight != nil {
+		panic("error parsing height")
+	}
+
+	c.backlogUnchecked[msgHeight.Uint64()] = append(c.backlogUnchecked[msgHeight.Uint64()], msg)
+	c.backlogUncheckedLen += 1
+	// We discard the furthest ahead messages in priority.
+	if c.backlogUncheckedLen == MaxSizeBacklogUnchecked {
+		maxHeight := msgHeight.Uint64()
+		for k, _ := range c.backlogUnchecked {
+			if k > maxHeight {
+				maxHeight = k
+			}
+		}
+
+		// Forget in the local cache that we ever received this message.
+		// It's needed for it to be able to be re-received and processed later, after a consensus sync, if needed.
+		c.backend.RemoveMessageFromLocalCache(c.backlogUnchecked[maxHeight][len(c.backlogUnchecked[maxHeight])-1].Payload())
+
+		// Remove it from the backlog buffer.
+		c.backlogUnchecked[maxHeight] = c.backlogUnchecked[maxHeight][:len(c.backlogUnchecked[maxHeight])-1]
+		c.backlogUncheckedLen -= 1
+	}
+
 }
 
 func (c *core) processBacklog() {
 	var capToLenRatio = 5
-	c.backlogsMu.Lock()
-	defer c.backlogsMu.Unlock()
 
 	for src, backlog := range c.backlogs {
 		logger := c.logger.New("from", src, "step", c.step)
@@ -108,7 +138,6 @@ func (c *core) processBacklog() {
 				logger.Debug("Post backlog event", "msg", curMsg)
 
 				go c.sendEvent(backlogEvent{
-					src: *c.lastHeader.CommitteeMember(src),
 					msg: curMsg,
 				})
 
@@ -124,5 +153,13 @@ func (c *core) processBacklog() {
 			}
 		}
 		c.backlogs[src] = backlog
+
 	}
+	for _, msg := range c.backlogUnchecked[c.height.Uint64()] {
+		go c.sendEvent(backlogUncheckedEvent{
+			msg: msg,
+		})
+		c.logger.Debug("Post unchecked backlog event", "msg", msg)
+	}
+	delete(c.backlogUnchecked, c.height.Uint64())
 }
--- a/consensus/tendermint/core/backlog.go
+++ b/consensus/tendermint/core/backlog.go
@@ -91,10 +91,10 @@ func (c *core) storeUncheckedBacklog(msg *Message) {
 	c.backlogUnchecked[msgHeight.Uint64()] = append(c.backlogUnchecked[msgHeight.Uint64()], msg)
 	c.backlogUncheckedLen += 1
 	// We discard the furthest ahead messages in priority.
-	if c.backlogUncheckedLen == MaxSizeBacklogUnchecked {
+	if c.backlogUncheckedLen == MaxSizeBacklogUnchecked+1 {
 		maxHeight := msgHeight.Uint64()
 		for k, _ := range c.backlogUnchecked {
-			if k > maxHeight {
+			if k > maxHeight && len(c.backlogUnchecked[k]) > 0 {
 				maxHeight = k
 			}
 		}
@@ -106,6 +106,10 @@ func (c *core) storeUncheckedBacklog(msg *Message) {
 		// Remove it from the backlog buffer.
 		c.backlogUnchecked[maxHeight] = c.backlogUnchecked[maxHeight][:len(c.backlogUnchecked[maxHeight])-1]
 		c.backlogUncheckedLen -= 1
+
+		if len(c.backlogUnchecked[maxHeight]) == 0 {
+			delete(c.backlogUnchecked, maxHeight)
+		}
 	}
 
 }
--- a/consensus/tendermint/core/backlog.go
+++ b/consensus/tendermint/core/backlog.go
@@ -89,7 +89,7 @@ func (c *core) storeUncheckedBacklog(msg *Message) {
 	}
 
 	c.backlogUnchecked[msgHeight.Uint64()] = append(c.backlogUnchecked[msgHeight.Uint64()], msg)
-	c.backlogUncheckedLen += 1
+	c.backlogUncheckedLen++
 	// We discard the furthest ahead messages in priority.
 	if c.backlogUncheckedLen == MaxSizeBacklogUnchecked+1 {
 		maxHeight := msgHeight.Uint64()
@@ -105,7 +105,7 @@ func (c *core) storeUncheckedBacklog(msg *Message) {
 
 		// Remove it from the backlog buffer.
 		c.backlogUnchecked[maxHeight] = c.backlogUnchecked[maxHeight][:len(c.backlogUnchecked[maxHeight])-1]
-		c.backlogUncheckedLen -= 1
+		c.backlogUncheckedLen--
 
 		if len(c.backlogUnchecked[maxHeight]) == 0 {
 			delete(c.backlogUnchecked, maxHeight)
--- a/consensus/tendermint/core/backlog_test.go
+++ b/consensus/tendermint/core/backlog_test.go
@@ -1,16 +1,14 @@
 package core
 
 import (
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/core/types"
+	"github.com/clearmatics/autonity/log"
+	"github.com/golang/mock/gomock"
 	"math/big"
 	"reflect"
 	"testing"
 	"time"
-
-	"github.com/golang/mock/gomock"
-
-	"github.com/clearmatics/autonity/common"
-	"github.com/clearmatics/autonity/core/types"
-	"github.com/clearmatics/autonity/log"
 )
 
 func TestCheckMessage(t *testing.T) {
@@ -265,7 +263,6 @@ func TestProcessBacklog(t *testing.T) {
 		val, _ := committeeSet.GetByIndex(0)
 
 		expected := backlogEvent{
-			src: val,
 			msg: msg,
 		}
 
--- a/consensus/tendermint/core/backlog_test.go
+++ b/consensus/tendermint/core/backlog_test.go
@@ -327,7 +327,6 @@ func TestProcessBacklog(t *testing.T) {
 		val, _ := committeeSet.GetByIndex(0)
 
 		expected := backlogEvent{
-			src: val,
 			msg: msg,
 		}
 
@@ -402,7 +401,6 @@ func TestProcessBacklog(t *testing.T) {
 		val, _ := committeeSet.GetByIndex(0)
 
 		expected := backlogEvent{
-			src: val,
 			msg: msg,
 		}
 
--- a/consensus/tendermint/core/backlog_test.go
+++ b/consensus/tendermint/core/backlog_test.go
@@ -4,8 +4,10 @@ import (
 	"github.com/clearmatics/autonity/common"
 	"github.com/clearmatics/autonity/core/types"
 	"github.com/clearmatics/autonity/log"
+	"github.com/clearmatics/autonity/rlp"
 	"github.com/golang/mock/gomock"
 	"github.com/influxdata/influxdb/pkg/deep"
+	"github.com/stretchr/testify/require"
 	"math/big"
 	"reflect"
 	"testing"
@@ -562,7 +564,6 @@ func TestProcessBacklog(t *testing.T) {
 		defer ctrl.Finish()
 
 		backendMock := NewMockBackend(ctrl)
-		backendMock.EXPECT().Post(gomock.Any()).Times(0)
 
 		committeeSet := newTestCommitteeSet(2)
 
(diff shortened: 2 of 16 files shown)


consensus/tendermint/core/handler.go
```diff
@@ -238,21 +245,42 @@ func (c *core) sendEvent(ev interface{}) {
...
+	}
+	msgHeight, err := msg.Height()
 	if err != nil {
-		logger.Error("Failed to decode message from payload", "err", err)
 		return err
 	}
+	if msgHeight.Cmp(c.Height()) > 0 {
+		// Future height message. Skip procesing and put it in the untrusted backlog buffer.
+		c.storeUncheckedBacklog(msg)
+		return errFutureHeightMessage // No gossip
```
piersy, line 264: Why not gossip here? I'm thinking that this message could be important for nodes at a future height.

consensus/tendermint/core/handler.go
```diff
@@ -238,21 +245,42 @@ func (c *core) sendEvent(ev interface{}) {
...
+	}
+	if msgHeight.Cmp(c.Height()) < 0 {
+		// Old height messages. Do nothing.
+		return errOldHeightMessage // No gossip
```
piersy, line 268: Why not gossip here? I'm thinking that this message could be important for nodes at a previous height.

consensus/tendermint/core/handler.go
```diff
@@ -238,21 +245,42 @@ func (c *core) sendEvent(ev interface{}) {
...
 	// Decode message and check its signature
 	msg := new(Message)
 
-	sender, err := msg.FromPayload(payload, c.lastHeader, crypto.CheckValidatorSignature)
+	err := msg.FromPayload(payload)
+	if err != nil {
+		c.logger.Error("Failed to decode message from payload", "err", err)
+		return err
+	}
+	msgHeight, err := msg.Height()
```
piersy, line 257: This error here doesn't make sense, it should have been returned from `FromPayload` and so there is no need for an error here.

consensus/tendermint/core/message.go
```diff
@@ -85,23 +86,40 @@ func (m *Message) DecodeRLP(s *rlp.Stream) error {
...
+	}
+}
 
+func (m *Message) Validate(validateFn func(*types.Header, []byte, []byte) (common.Address, error), previousHeader *types.Header) (*types.CommitteeMember, error) {
 	// Validate message (on a message without Signature)
 	if validateFn == nil {
-		log.Error("validateFn is not set")
-		return nil, nil
+		panic("validateFn is not set")
+	}
```
piersy, line 116: This is a bit redundant since if you remove this check the code would panic at line 132

consensus/tendermint/core/message.go
```diff
@@ -123,7 +141,7 @@ func (m *Message) FromPayload(b []byte, previousHeader *types.Header, validateFn
 
 	v := previousHeader.CommitteeMember(addr)
 	if v == nil {
-		return nil, fmt.Errorf("validator was not a committee member %q", v)
+		return nil, errNotCommittee
```
piersy, line 144: The previous error message was more helpful since it actually told you who was not a committee member. This is why I advocate for not using error instances.

consensus/tendermint/core/backlog.go
```diff
@@ -124,11 +149,18 @@ func (c *core) processBacklog() {
...
 		}
 	}
+
+	for _, msg := range c.backlogUnchecked[c.height.Uint64()] {
+		go c.sendEvent(backlogUncheckedEvent{
+			msg: msg,
+		})
+		c.logger.Debug("Post unchecked backlog event", "msg", msg)
+	}
+	c.backlogUnchecked[c.height.Uint64()] = nil
```
piersy, line 163: This is freeing the memory held by the value, but its not removing the key from the map, to remove the key you need to use delete.

I also can't see that backlogs is ever deleted from.

So it looks like both backlogUnchecked and backlogs are leaking memory.

consensus/tendermint/core/handler.go
yazzaoui: Once future height messages become current height messages they will be gossiped post-validation.

consensus/tendermint/core/handler.go
yazzaoui: It's a waste of time to verify old height message correctness, the messages that made us able to progress to the next height should have already been gossiped by us,

consensus/tendermint/core/message.go
yazzaoui: To be fair I haven't researched  (or forgot) why we didn't return an error in the case of `validateFn == nil` ( could be for testing purposes? ) but you're right it's unnecessary and the code would panic too if `previousHeader == nil` L212.

consensus/tendermint/core/handler.go
piersy: That forces the rest of the network to wait for this node to catch up before they can progress, by not gossiping here I think we could be limiting the performance of the network.

consensus/tendermint/core/handler.go
```diff
@@ -143,19 +142,27 @@ eventLoop:
...
 				p, err := e.msg.Payload()
+				if err != nil { // decoding happened successfully, why handling errors here ? We should consider panic.
+					c.logger.Debug("core.mainEventLoop Get message payload failed", "err", err)
+				}
+
+				c.backend.Gossip(ctx, c.committeeSet().Committee(), p)
+
+			case backlogUncheckedEvent:
+				c.logger.Debug("Started handling backlogUncheckedEvent")
+				err := c.handleCurrentHeightMsg(ctx, e.msg)
```
piersy, line 160: We need to check here to make sure that `e.msg` has the same height as core.

We can't be sure that core has not progressed to the next height while a `backlogUncheckedEvent` was being sent. We could have received a `messageEvent`, `backlogEvent`, `TimeoutEvent` or `CommitEvent` in the meantime.

consensus/tendermint/core/backlog.go
```diff
@@ -88,9 +89,33 @@ func (c *core) storeBacklog(msg *Message, src types.CommitteeMember) {
...
+				maxHeight = k
+			}
+		}
+		c.backlogUnchecked[maxHeight] = c.backlogUnchecked[maxHeight][:len(c.backlogUnchecked[maxHeight])-1]
+		c.backlogUncheckedLen -= 1
+	}
+
+}
+
 func (c *core) processBacklog() {
```
piersy, line 118: It looks like this is called whenever we change step and it is assumed that we change step sequentially with no gaps, but I'm not so sure that is the case.

If we look here:
https://github.com/clearmatics/autonity/blob/ac1b9dd3a68aa8dfc3044582191574d047256d34/core/blockchain.go#L1661-L1680

We can see that only one `ChainHeadEvent` is sent for the whole chain insertion. The `ChainHeadEvent` eventually leads to a `CommitEvent` being recieved by core which causes core to call `setInitialState` where the height is set according to the latest block. This means we could set the height to something greater than the current height plus one.

https://github.com/clearmatics/autonity/blob/ac1b9dd3a68aa8dfc3044582191574d047256d34/consensus/tendermint/core/core.go#L305-L311

After calling `setInitialState` `setStep` is called.

https://github.com/clearmatics/autonity/blob/ac1b9dd3a68aa8dfc3044582191574d047256d34/consensus/tendermint/core/core.go#L268-L274
`setStep` will call `processBacklog` but process backlog will process the backlog for the current height only, if the last time the height was set the value increased by more than one, then the intervening backlog entries will never be processed.

That would be another source of memory leak.

consensus/tendermint/core/handler.go
yazzaoui: We can't gossip unverified messages, they could be anything, so that would be an important DoS attack vector.
It's not trivial as well to put some limitations, as the round value could be anything.

consensus/tendermint/core/handler.go
yazzaoui: Yeah that's correct.

consensus/tendermint/core/handler.go
piersy: I think we can protect against a dos attack but still forward reasonable looking future messages.

So we could say that we don't handle messages that are some delta ahead of our current height.

And then we could limit the round numbers that we handle for messages at a future height. So discard messages for rounds > X ( because higher rounds are less likely to happen)

yazzaoui: I think the performance gains are going to be very minimal and don't worth really the investment with potential issues arising. TBH we are considering with the researchers is to remove this gossip mechanism altogether.

(But again the messages should gossiped once processed and very likely the node is going to do that very quickly)

consensus/tendermint/core/backlog.go
yazzaoui: Good point, should be resolved

consensus/tendermint/core/handler.go
yazzaoui: Reworked the flow, should be resolved

consensus/tendermint/core/backlog.go
yazzaoui: Should be resolved

consensus/tendermint/core/message.go
yazzaoui: It was for testing, but now I've added a test to check if it panics

consensus/tendermint/core/handler.go
yazzaoui: Should be resolved

consensus/tendermint/core/message.go
```diff
@@ -123,15 +139,25 @@ func (m *Message) FromPayload(b []byte, previousHeader *types.Header, validateFn
...
+		payload, err := rlp.EncodeToBytes(m)
+		if err != nil {
+			// We panic if there is an error, reasons:
+			// Either we received the message and we managed to decode it, hence it must be possible to encode it.
+			// If we can't encode the payload for our own generated messages, that's a programming error.
+			panic("could not decode message payload")
+		}
+		m.payload = payload
+	}
+	return m.payload
```
piersy, line 160: This is not safe for concurrent use.

Instead of checking for nil we can use `sync.Once`

```suggestion
	m.once.Do(func() {
		payload, err := rlp.EncodeToBytes(m)
		if err != nil {
			// We panic if there is an error, reasons:
			// Either we received the message and we managed to decode it, hence it must be possible to encode it.
			// If we can't encode the payload for our own generated messages, that's a programming error.
			panic("could not decode message payload")
		}
	})
	m.payload = payload
```

consensus/tendermint/core/handler.go
```diff
@@ -131,31 +130,33 @@ eventLoop:
 			// A real ev arrived, process interesting content
 			switch e := ev.Data.(type) {
 			case events.MessageEvent:
-				if len(e.Payload) == 0 {
-					c.logger.Error("core.mainEventLoop Get message(MessageEvent) empty payload")
+				// ecrire test of len(0)
```
piersy, line 133: I'm not sure what this comment is talking about

consensus/tendermint/core/handler.go
```diff
@@ -131,31 +130,33 @@ eventLoop:
...
+				msg := new(Message)
+				if err := msg.FromPayload(e.Payload); err != nil {
+					c.logger.Error("core.mainEventLoop Get message(MessageEvent) invalid payload")
```
piersy, line 136: I'd advise against hardcoding package and function names in log messages, because they are likely to go out of sync when code is modified.

Instead we could use the following handlers on logging in core, which provides safe file line number and function name information consistently in all log messages.

https://github.com/clearmatics/autonity/blob/0d83a5f4bc78c9aa134c5e019c8ebd4772b9a773/log/handler.go#L96-L112

consensus/tendermint/core/message.go
```diff
@@ -203,3 +229,11 @@ func (m *Message) Height() (*big.Int, error) {
 func Encode(val interface{}) ([]byte, error) {
 	return rlp.EncodeToBytes(val)
 }
+
+func MustEncode(val interface{}) []byte {
```
piersy, line 233: I'd suggest removing this, it is only used in tests, and it would be more consistent to simply write:
```golang
ret, err := rlp.EncodeToBytes(val)
require.NoError(t, err)
```

In the tests.

consensus/tendermint/core/message_test.go
```diff
@@ -53,73 +53,84 @@ func TestMessageString(t *testing.T) {
...
+		payload := msg.Payload()
 		wantErr := errors.New("some error")
 
 		validateFn := func(previousHeader *types.Header, data []byte, sig []byte) (common.Address, error) {
 			return common.Address{}, wantErr
 		}
 
 		decMsg := &Message{}
-		_, err := decMsg.FromPayload(payload, nil, validateFn)
+		if err := decMsg.FromPayload(payload); err != nil {
```
piersy, line 89: I think you can just use `msg` in all of these tests, no need to get the payload and then decode it.

consensus/tendermint/core/message_test.go
```diff
@@ -53,73 +53,84 @@ func TestMessageString(t *testing.T) {
...
+		}}
 		decMsg := &Message{}
-		_, err := decMsg.FromPayload(payload, nil, validateFn)
-		if err != ErrUnauthorizedAddress {
-			t.Fatalf("have %v, want %v", err, ErrUnauthorizedAddress)
+		if err := decMsg.FromPayload(payload); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		val, err := decMsg.Validate(validateFn, lastHeader)
+		if val != nil {
```
piersy, line 118: We don't need to check the `val` here, if `err` is not nil, we can assume that `val` will not be inspected.

consensus/tendermint/core/message_test.go
```diff
@@ -53,73 +53,84 @@ func TestMessageString(t *testing.T) {
...
+			t.Fatal("should return nil validator")
+		}
+		if err.Error() != "message received is not from a committee member: 0000000000000000000000000000001234567890" {
```
piersy, line 121: We are not really concerned with what the error string is, just that there is an error. Checking the string makes the test more brittle.

What we wan't to ascertain is that when the message sender is not a committee member validation fails.

consensus/tendermint/core/message.go
yazzaoui: yes it would appear that the `Payload()` method is not thread safe, but as far as I can see it is never called concurrently. The messages used to synchronize peers in `SyncPeers` have always this payload field being set by the main thread, before being saved in the current height message buffer.
I would like to avoid using unnecessary mutexes and I have run tests with the `-race` flag, couldn't see anything

piersy: Its called in SyncPeer and in mainEventLoop. Those are 2 separate goroutines.

yazzaoui: What I'm trying to say it that when it SyncPeer is called m.payload is always set but let's discuss at standup.

consensus/tendermint/core/handler.go
yazzaoui: My bad, me neither

consensus/tendermint/core/handler.go
yazzaoui: It's a nice idea but instantiating 4 new loggers is a bit overkill, in my opinion, I'll change the error messages tho.

consensus/tendermint/core/message_test.go
```diff
@@ -53,73 +53,84 @@ func TestMessageString(t *testing.T) {
...
-func TestMessageFromPayload(t *testing.T) {
-	t.Run("nil validator function given, nil validator returned", func(t *testing.T) {
-		msg := &Message{
-			Code:    msgProposal,
-			Address: common.HexToAddress("0x1234567890"),
-		}
-
-		payload, _ := msg.Payload()
+func TestMessageValidate(t *testing.T) {
+	t.Run("nil validator function given, panic", func(t *testing.T) {
```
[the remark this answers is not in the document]
yazzaoui: I was just trying to test that if for any reason someone would have called this function with a nil argument we would have seen a panic to help spot the error, but yeah I guess it's not very useful.

consensus/tendermint/core/message_test.go
```diff
@@ -53,73 +53,84 @@ func TestMessageString(t *testing.T) {
...
-			Code:    msgProposal,
-			Address: common.HexToAddress("0x1234567890"),
-		}
-
-		payload, _ := msg.Payload()
+	t.Run("validate function fails, nil returned", func(t *testing.T) {
+		msg := createPrevote(t, common.Hash{}, 1, new(big.Int).SetUint64(26), types.CommitteeMember{VotingPower: big.NewInt(0)})
+		lastHeader := &types.Header{Number: new(big.Int).SetUint64(25)}
+		payload := msg.Payload()
 		wantErr := errors.New("some error")
```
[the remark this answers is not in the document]
yazzaoui: True, we might want to start using error wrapping here.

consensus/tendermint/core/message_test.go
yazzaoui: `FromPayload` is decoding the message which is needed to call `Height()`

consensus/tendermint/core/handler.go
piersy: I don't see that as overkill, its just the way that loggers are configured. The geth team could have set it up such that you provide options eg:
```
FileName: true,
FunctionName: true,
```

But instead they decided you have to construct some handlers.

consensus/tendermint/core/message_test.go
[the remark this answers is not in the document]
piersy: The only time that you wouldn't see a panic in that situation is either if the function checks for nil and provides a default value, or if the value is never used in the entire call tree of the function.

It makes sens to test for the first situation (I.E: check that the nil is replaced with a default value) because that is not standard.

The second situation I would not bother testing for, or if we do test for it, lets be consistent and write this test for every parameter of every function in our codebase.

consensus/tendermint/core/message.go
piersy: So that may be the case right now, but there is nothing to stop someone from changing the code and making it not so. It's hard to verify that this code is not called concurrently and there  is no hint in the code that we need to be careful about that.

The neatest way to handle this in my opinion is to avoid mutable messages. That means providing constructor functions that make sure all fields are set, which would ensure that you cannot have an instance of Message without all its fields being defined.

consensus/tendermint/core/backlog_test.go
```diff
@@ -541,6 +539,59 @@ func TestProcessBacklog(t *testing.T) {
 		<-timeout.C
 	})
 
+	t.Run("untrusted messages are processed when height change", func(t *testing.T) {
+		nilRoundVote := &Vote{
+			Round:  2,
+			Height: big.NewInt(4),
+		}
```
piersy, line 546: The naming of this looks strange, the round is clearly being set.

consensus/tendermint/core/backlog_test.go
```diff
@@ -541,6 +539,59 @@ func TestProcessBacklog(t *testing.T) {
...
+			height:           big.NewInt(3),
+		}
+
+		c.lastHeader = &types.Header{Committee: committeeSet.Committee()}
+
+		c.storeUncheckedBacklog(msg)
+		c.storeUncheckedBacklog(msg2)
+		c.setStep(prevote)
+		c.processBacklog()
+		c.setHeight(big.NewInt(4))
```
piersy, line 587: Define the height as a variable, it makes the test clearer.

consensus/tendermint/core/backlog_test.go
```diff
@@ -541,6 +539,59 @@ func TestProcessBacklog(t *testing.T) {
...
+
+		backendMock.EXPECT().Post(gomock.Any()).Times(2)
```
piersy, line 589: Critical

This doesn't make sense to me, there are 5 lines that could  be responsible for calling `Post`

```
		c.storeUncheckedBacklog(msg)
		c.storeUncheckedBacklog(msg2)
		c.setStep(prevote)
		c.processBacklog()
		c.setHeight(big.NewInt(4))
```

Also you are calling `processBacklog` before changing the height.

consensus/tendermint/core/backlog_test.go
```diff
@@ -591,3 +642,128 @@ func TestProcessBacklog(t *testing.T) {
...
+		c := &core{
+			logger:           log.New("backend", "test", "id", 0),
+			backend:          backendMock,
+			address:          common.HexToAddress("0x1234567890"),
+			backlogs:         make(map[common.Address][]*Message),
+			backlogUnchecked: make(map[uint64][]*Message),
+			step:             prevote,
+			round:            1,
+			height:           big.NewInt(4),
+		}
```
piersy, line 662: We are repeating this a lot, lets make a function to create core, ideally using (core.New) and then setting the required step, round and height.

consensus/tendermint/core/backlog_test.go
```diff
@@ -591,3 +642,128 @@ func TestProcessBacklog(t *testing.T) {
...
+		var messages []*Message
+
+		for i := int64(0); i < MaxSizeBacklogUnchecked; i++ {
```
piersy, line 665: Probably nicer to make `MaxSizeBacklogUnchecked` a parameter to core and then we can set a lower value like 10 for these tests.

consensus/tendermint/core/handler_test.go
```diff
@@ -139,22 +149,104 @@ func TestHandleCheckedMessage(t *testing.T) {
...
+			ProposedBlockHash: common.BytesToHash([]byte{0x1}),
+		}
+		payload, err := rlp.EncodeToBytes(vote)
+		require.NoError(t, err)
+		msg := &Message{
+			Code:       msgPrevote,
+			Msg:        payload,
+			decodedMsg: vote,
+			Address:    common.Address{},
+		}
```
piersy, line 206: This is a lot of setup for a simple thing. The problem with this is that it makes it hard for the reviewer to understand the pertinent values.

We could wrap this code up to look like this.

```golang
height = 2
round = 1
step = prevote
c := newCore(t, height, round, step)
m := newMessage(t , height-1, round, step)
if err := c.handleMsg(context.Background(), msg); err != errOldHeightMessage {
	t.Fatal("errOldHeightMessage not returned")
}

```

This makes verifying if the test makes sense much easier because all the irrelevant information has been hidden from the reader.

consensus/tendermint/core/message.go
```diff
@@ -85,23 +85,38 @@ func (m *Message) DecodeRLP(s *rlp.Stream) error {
...
+		return m.Decode(&proposal)
+	case msgPrevote, msgPrecommit:
+		var vote Vote
+		return m.Decode(&vote)
+	default:
+		return errMsgPayloadNotDecoded
 	}
+}
 
+func (m *Message) Validate(validateFn func(*types.Header, []byte, []byte) (common.Address, error), previousHeader *types.Header) (*types.CommitteeMember, error) {
```
piersy, line 112: This needs documentation to tell you it will panic.

go.mod
```diff
@@ -40,10 +40,11 @@ require (
 	github.com/karalabe/usb v0.0.0-20190919080040-51dc0efba356
 	github.com/kr/pretty v0.2.0 // indirect
 	github.com/kylelemons/godebug v1.1.0 // indirect
-	github.com/mattn/go-colorable v0.1.0
-	github.com/mattn/go-isatty v0.0.5-0.20180830101745-3fb116b82035
+	github.com/mattn/go-colorable v0.1.7
+	github.com/mattn/go-isatty v0.0.12
```
piersy, line 44: Critical

You've updated the module files but this is not a change related to the changes in this PR.

PR #661, opened 2020-08-27
Node not reconnect on topology test

Jason-Zhangxin-Chen:
To fix issue: #431 https://github.com/clearmatics/autonity/issues/431
The problem: Node cannot connect to network again during the topology changing test in local e2e test. Root cause is that in p2p network layer, there is a hardcoded time window (35 s) which limited the reconnect request if the last request was captured in that time window. While in our local e2e testing context, very testcase runs within tens of seconds. The topology checking was finished before the peer have chance to reconnect.
The solution: Make the time window configuarable for the local e2e testing framework, waiting for 35s is too long for local e2e testing. And this change does not impact the p2p network protocol from end to end point of view since the config is not exposed from e2e point of view.


piersy (approved):
All good :+1:

aneequesafdar (approved):

PR #594, opened 2020-06-20
reward, stake, member management test base on local e2e test framework

Jason-Zhangxin-Chen:
It contains e2e test case to cover:
1. reward distribution test with stake changes on the run time.
2. stake management test.
3. member management test.
It would close: 
https://github.com/clearmatics/autonity/issues/412
https://github.com/clearmatics/autonity/issues/590
https://github.com/clearmatics/autonity/issues/587

PR #665, opened 2020-09-04
Introduce interactor for use in stake management tests

piersy:
The interactor makes for a very fluid interaction with autonity and should help to really streamilne the testcases in the pr.

I decided to open a PR because it would have been difficult to view the code in a change suggestion.

This won't compile because I haven't updated all the tests to use it, its just the initial implementation which we can build on.

I make use of the interactor in one test case just to show how it can cut down the amount of code.

### DIFF
--- a/consensus/test/autonity_interactor_test.go
+++ b/consensus/test/autonity_interactor_test.go
@@ -1,0 +1,174 @@
+package test
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


piersy:
Closing this since @Jason-Zhangxin-Chen has manually integrated it into #594

PR #667, opened 2020-09-07
Tendermint behavior improvement rebased

piersy:
This is a rebased version of #615, which was rebased onto master before its history could be cleaned.


aneequesafdar (approved):

yazzaoui (approved):
LGTM

PR #669, opened 2020-09-07
Fix concurrent map write in Header.CommitteeMember

piersy:

It turns out that Header instances are accessed concurrently and so
should be safe for concurrent use.


aneequesafdar (approved):

core/types/block.go
```diff
@@ -92,6 +93,8 @@ type Header struct {
 	Committee Committee `json:"committee"           gencodec:"required"`
 	// used for committee member lookup, lazily initialised.
 	committeeMap map[common.Address]*CommitteeMember
+	// Used to ensure the committeeMap is created only once.
+	once sync.Once
```
Jason-Zhangxin-Chen, line 97: Does this impact the header json object marsalling or the block persistent in storage?

piersy: It does not, because the field is private and therefore cannot be seen by json or rlp serialisation.

yazzaoui (approved):
Looking good.

PR #670, opened 2020-09-07
Fix miner start/stop race conditions

piersy:

miner.go is syncronised with atomics, but the syncronisation was broken
and atomics can't really provide the syncronisation we require.

See below for an in depth description
https://github.com/ethereum/go-ethereum/pull/21381


miner/miner.go
```diff
@@ -122,26 +140,43 @@ func (miner *Miner) update() {
 	}
 }
 
+// Start starts the miner mining, unless it has been paused by the downloader
+// during sync, in which case it will start mining once the sync has completed.
 func (miner *Miner) Start(coinbase common.Address) {
-	atomic.StoreInt32(&miner.shouldStart, 1)
+	miner.startStopMutex.Lock()
+	defer miner.startStopMutex.Unlock()
```
aneequesafdar, line 147: Is it not possible that same go routine calls `Lock()` twice on `miner.startStopMutex` leading to a panic. For example first lock is acquired on line 128 and then on line 131 you enter this function where you will try to acquire the lock again from the same mutex on line 146? Or can that not be possible? if so, can you please explain how?

aneequesafdar: I already see my mistake, there 2 start function, one is private and the other is public, which I presume will not be called from the same go routine

aneequesafdar (approved):

Jason-Zhangxin-Chen (approved):
LGTM.

yazzaoui (approved):
LGTM

PR #671, opened 2020-09-07
Ensure BlockFetcher goroutines ended on Stop

piersy:

This should fix the problem seen in
https://github.com/clearmatics/autonity/issues/624.


yazzaoui (approved):
LGTM, do you think we can have an upstream PR for it ?

aneequesafdar (approved):

piersy:
> think we can have an upstream PR for it ?

Will do when, but first I wan't to get the miner fix merged into upstream.

PR #674, opened 2020-09-10
Extend blockToWait to make pending TX to be mined to fix issue #496

Jason-Zhangxin-Chen:
To fix pending TX un-mined happpens at #496 for local e2e test, we extend the blockToWait from 15s to 60s, to get sufficient time to let TX to be mined by consensus engine.


aneequesafdar (approved):

Jason-Zhangxin-Chen:
@yazzaoui @piersy  This is a small PR to extend the block to wait in local e2e test framework to get the pending TX be mined in some flaky testcases. Please have a review on it. Thanks.

piersy:
@Jason-Zhangxin-Chen I can see that `TestTendermintStartStopAllNodes` has failed, so I't doesn't look like this has solved the issue.

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen I can see that `TestTendermintStartStopAllNodes` has failed, so I't doesn't look like this PR has solved the issue we were facing.

@piersy , This PR only fix the pending TX umined at the end of the test, if you explore the failed case you pointed, it was due to the different problem like data race or go-routine leaks that we try to address during the week.

piersy:
> @piersy , This PR only fix the pending TX umined at the end of the test, if you explore the failed case you pointed, it was due to the different problem like data race or go-routine leaks that we try to address during the week.

:+1: Thanks for clarifying

Jason-Zhangxin-Chen:
> > @piersy , This PR only fix the pending TX umined at the end of the test, if you explore the failed case you pointed, it was due to the different problem like data race or go-routine leaks that we try to address during the week.
> 
>  Thanks for clarifying, just waiting for CI to re-run

If it is fine, please approved it, it would help to stablize the tests.

piersy (approved):

PR #676, opened 2020-09-18
Add node console

piersy:
Added a script that starts up the nodejs console with a web3 instance
connected via the provided websocket rpc address. It also provides
bindings for the Autonity contract.

Try it out on the helloworld example network
```sh
(cd helloworld && docker-compose up -d)
./nodejsconsole/console localhost:8641
```

Note: The 8641 port is the websocket rpc port of the first node in the helloworld network, it's probably best to use that node since it is the only node that is permissioned to be part of the network and hence the only node that will progress beyond block 0.


yazzaoui (approved):
LGTM

PR #688, opened 2020-09-25
Rpc dump tendermint state

### LINKED ISSUE #162: RPC to get consensus state
add a REST or rpc request to make it easy to get consensus state of the node, something like https://tendermint.com/rpc/#consensusstate (this will reduce a long "reading logs" stage)

Jason-Zhangxin-Chen:
This PR dumps tendermint core state via HTTP RPC API: tendermint_getCoreState, to close #162 https://github.com/clearmatics/autonity/issues/162, 
It does not use mutex to solve the race condition between RPC callback routine and the tendermint core handler routine, instead we use an event channel and event handler to fetch the state. The unit test and manual e2e test was done.


consensus/tendermint/core/message_set.go
```diff
@@ -106,3 +106,13 @@ func (ms *messageSet) Values(blockHash common.Hash) []Message {
 	}
 	return messages
 }
+
+func (ms *messageSet) Keys() []common.Hash {
```
piersy, line 110: This is badly named, anything can be a key in a map, so `Keys` doesn't really tell you anything about what this function is returning, I'd suggest `BlockHashes`

consensus/tendermint/core/message_test.go
```diff
@@ -42,14 +42,21 @@ func TestMessageEncodeDecode(t *testing.T) {
...
+			CommittedSeal: []byte{0x22, 0x33},
+			power:         100,
+		}
 
-	want := "{Code: 0, Address: 0x0000000000000000000000000000001234567890}"
-	if got := msg.String(); got != want {
-		t.Errorf("Expected %v, got %v", want, got)
+		// decode error would return ""
+		want := ""
+		if got := msg.String(); got != want {
```
piersy, line 57: cThis test seems to be checking that an empty message string is returned, this is undesirable behaviour. Instead of allowing the system to create malformed messages we should enforce that messages are well formed. In this case we can mandate that `Decode` must be called before `String` and then we can extend the ConsensusMsg interface to Have  the `String` function. Then if `String` is called before decode we will panic when trying to access `decodedMsg`.

consensus/tendermint/core/round_messages.go
```diff
@@ -76,6 +77,60 @@ func (s *messagesMap) GetMessages() []*Message {
 	return result
 }
 
+func (s *messagesMap) CopyMessages() []string {
```
piersy, line 80: Having this function convert messages to string is a violation of the single responsibility principle, the `messagesMap` should be concerned with managing messages not with formatting them for a state dump. I suggest we move this functionality into `state_dumper.go`, it doesn't need its own function, it can just go inline in `handleStateDump`

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (c *core) getValidValue() common.Hash {
+	v := common.Hash{}
+	if c.validValue != nil {
+		v = c.validValue.Hash()
+	}
+	return v
+}
+
+func (c *core) getParentCommittee() types.Committee {
```
piersy, line 94: This is duplicating `committeeSet().Committee()`

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,117 @@
...
+		SentProposal:          c.sentProposal,
+		SentPrevote:           c.sentPrevote,
+		SentPrecommit:         c.sentPrecommit,
+		SetValidRoundAndValue: c.setValidRoundAndValue,
+		// timer state
+		ProposeTimerStarted:   c.proposeTimeout.timerStarted(),
+		PrevoteTimerStarted:   c.prevoteTimeout.timerStarted(),
+		PrecommitTimerStarted: c.precommitTimeout.timerStarted(),
+		// known msgs in case of gossiping.
+		KnownMsgHash: c.backend.KnownMsgHash(),
```
piersy, line 64: I think we can drop these, they don't constitute core state, the known messages is just a cache held by the backend. We already know what messages core has received because of `CurHeightMessges`.

consensus/tendermint/core/round_messages.go
piersy: In fact all the additional methods added here violate the single responsibility principle because they are all exclusively for generating the state dump, so I would suggest moving all this functionality into the `state_dumper.go` since both these objects exist in core they can freely access each others internals.

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,117 @@
...
+	return v
+}
+
+func (c *core) getValidValue() common.Hash {
+	v := common.Hash{}
+	if c.validValue != nil {
+		v = c.validValue.Hash()
+	}
+	return v
+}
```
piersy, line 92: I'd like to understand why you have set these functions up to return default empty values?

The problem I see with this is that when you do a state dump you will see these empty values and it will look like they were actually proposed. Instead I would think that we need to indicate when these values are nil, so that people can understand when they have been set or not.

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"github.com/clearmatics/autonity/common"
+	"github.com/clearmatics/autonity/core/types"
+	"time"
+)
+
+type coreStateRequestEvent struct {
+}
+
+func (c *core) CoreState() types.TendermintState {
```
piersy, line 12: Rather than cluttering `core` with lots of functions to generate the state, I think it would be nicer to define a standalone function that takes core as a parameter and returns a state dump. In this way we separate the formatting concerns from the actual work of core.

core/types/tendermint_state.go
```diff
@@ -0,0 +1,64 @@
...
+	// timer state
+	BlockPeriod           uint64 `json:"blockPeriod"           gencodec:"required"`
+	ProposeTimerStarted   bool   `json:"proposeTimerStarted"   gencodec:"required"`
+	PrevoteTimerStarted   bool   `json:"prevoteTimerStarted"   gencodec:"required"`
+	PrecommitTimerStarted bool   `json:"precommitTimerStared"  gencodec:"required"`
+
+	// current height messages and known message in case of gossip.
+	CurHeightMessages []string      `json:"CurHeightMessages"     gencodec:"required"`
+	KnownMsgHash      []common.Hash `json:"KnownMsgHash"          gencodec:"required"`
+}
```
piersy, line 64: I think these would be better defined in `consensus/tendermint/core/state` since they are only required to generate a state dump, the rest of the codebase does not need to know about them. I would definitely not consider them "core" types, in the general sense of the word "core".

piersy (changes requested):
Hi @Jason-Zhangxin-Chen I left quite a few comments.

consensus/tendermint/core/message_set.go
Jason-Zhangxin-Chen: > This is badly named, anything can be a key in a map, so `Keys` doesn't really tell you anything about what this function is returning, I'd suggest `BlockHashes`

Resolved in my last commit.

consensus/tendermint/core/message_test.go
Jason-Zhangxin-Chen: > This test seems to be checking that an empty message string is returned, this is undesirable behaviour. Instead of allowing the system to create malformed messages we should enforce that messages are well formed. In this case we can mandate that `Decode` must be called before `String` and then we can extend the ConsensusMsg interface to Have the `String` function. Then if `String` is called before decode we will panic when trying to access `decodedMsg`.
> 
> I think we can get rid of this test entirely, in general I don't think testing string representations is a useful thing to do.

Removed the test case in my last commit.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > This is duplicating `committeeSet().Committee()`

getParentCommittee() returns last block's committee, while committeeSet().Committee returns current height's committee.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > I think we can drop these, they don't constitute core state, the known messages is just a cache held by the backend. We already know what messages core has received because of `CurHeightMessges`.

CurHeightMessages are about current height, while the backend one's tracks the entire gossip history buffered at the peer.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > I'd like to understand why you have set these functions up to return default empty values?
> 
> The problem I see with this is that when you do a state dump you will see these empty values and it will look like they were actually proposed (you cannot tell the difference between proposed or default value). Instead I would think that we need to indicate when these values are nil, so that people can understand when they have been set or not.

Resloved in my last commit.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > Rather than cluttering `core` with lots of functions to generate the state, I think it would be nicer to define a standalone function that takes core as a parameter and returns a state dump. In this way we separate the formatting concerns from the actual work of core.

For (c *core) CoreState() and (c *core) handleStateDump(), I keep them. CoreState() inherit the style which is assembled as a part of interface of Tendermint which is refered by backend, where the RPC server will refer to, and handleStateDump is a handler of object core.

For the others helper functions, they are now standalone functions.
See my last commits.

core/types/tendermint_state.go
Jason-Zhangxin-Chen: Resolved in my last commits.

consensus/tendermint/core/round_messages.go
Jason-Zhangxin-Chen: > Having this function convert messages to string is a violation of the single responsibility principle, the `messagesMap` should be concerned with managing messages not with formatting them for a state dump. I suggest we move this functionality into `state_dumper.go`, it doesn't need its own function, it can just go inline in `handleStateDump`

Resolved in my last commits.

Jason-Zhangxin-Chen: > In fact all the additional methods added in this file violate the single responsibility principle because they are all exclusively for generating the state dump, so I would suggest moving all this functionality into the `state_dumper.go` since both these objects exist in core they can freely access each others internals.

Resolved as below:
getRounds() access internal mutex of messageMap, so I keep it in the file.
CopyMessage were removed, and getVoteState was move to state_dumper.go

consensus/tendermint/core/message_test.go
piersy: But you also need to change the implementation of String to just use `decodedMsg.String()` as it stands calling `String()` on a `Message` before calling decode wont panic and will return you a string with an empty representation of the consensus message.

consensus/tendermint/core/state_dumper.go
piersy: They are the same thing, the committee for a block is defined by it's parent block.

See the implementation in `committee.go`
https://github.com/clearmatics/autonity/blob/4389477fc78a8464829bd66826f9caa851bfb989/consensus/tendermint/core/committee.go#L156-L158

consensus/tendermint/core/state_dumper.go
piersy: It's not actually the entire gossip history, its the most recently seen 1024 messages. I don't see that this is helpful or has anything to do with core. How would someone use the known message hash information?

Jason-Zhangxin-Chen:
> Hi @Jason-Zhangxin-Chen I left quite a few comments.

@piersy , please review my latest bunch of refines to reslove your comments.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > It's not actually the entire gossip history, its the most recently seen 1024 messages. I don't see that this is helpful or has anything to do with core. How would someone use the known message hash information?

Yeah, I shouldn't say it is the entire history, people cannot buffer all the history of gossip. Yeah it should be the most recently ones. They not only contains known message hash on current height, it could also contain old height msg hash. 

Adding this was suggested by Youssef. @yazzaoui , please share your comments on this.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > They are the same thing, the committee for a block is defined by it's parent block.
> 
> See the implementation in `committee.go`
> https://github.com/clearmatics/autonity/blob/4389477fc78a8464829bd66826f9caa851bfb989/consensus/tendermint/core/committee.go#L156-L158

Thank you, Piers. I did not notice this before. Removed it.

consensus/tendermint/core/state_dumper.go
piersy: @yazzaoui, @Jason-Zhangxin-Chen Maybe we should take a step back and consider what people would use the state dump as a wole for?

Its not clear to me why this is needed, what are the scenarios where someone would use this?

Jason-Zhangxin-Chen: In my understanding, it about debugging purpose, for example we want to debug consensus issues (message view on different node, liveness issues, etc...) on production network when system log is not available.

Jason-Zhangxin-Chen: > @Jason-Zhangxin-Chen
> 
> > on production network when system log is not available.
> 
> In what situation would we not have a system log? This rpc call gives access to all the consensus messages, including proposed blocks, so by calling it regularly you can learn the entire state of the blockchain. Therefore any node operator who is not willing to share their logs will also have disabled this RPC.

To collect information for system log would be hard and time cost-ful, dumps the core state of the consensus engine by RPC would be more efficient. 

It is node operator's demand to disable it.

piersy: >To collect information for system log would be hard and time cost-ful, dumps the core state of the consensus engine by RPC would be more efficient

I don't really follow this. If we have logs and we assume that node operators will have logs, why would it be hard to collect information from the logs?

Also I'm not sure what you are referring to by
> It is node operator's demand to disable it.

My original question still stands. When would we not have a system log?

piersy: @yazzaoui Could do with your input here. I'm not understanding why we would need this feature at all, and if we do, we need to decide what its intended purpose is. Right now it looks like a duplication of the information we log and I don't see any reason to add this to the system.

yazzaoui: @piersy If you're talking about `KnownMsgHash` then when I was investigating logs for networks having lost liveness, I used to see very quite often messages being supposedly received but not processed. Having quick access to this information would be very helpful, even though we might want to filter it to only output hash messages for the current height.

If you are talking about this feature as a whole, an RPC endpoint to dump the core state, well there are many use-cases, let's discuss that during standup?

yazzaoui (approved):
LGTM

consensus/tendermint/core/state_dumper.go
piersy: So from the discussion we had at startup, I also understood that we can't expect to log all the information that we may dump because it would be just too much information, so we need a mechanism to gather this fine grained information when something goes wrong.

The issue with the use of `KnownMsgHash` here is that we don't have the hashes of `CurHeightMessages` so `KnownMsgHash` cannot be compared with the other messages.

piersy: So I suggest that instead of providing `KnownMsgHash` we instead provide `backlogs` and `backlogs` unchecked so that it is clear what pending unprocessed messages a node has.

consensus/tendermint/core/message_test.go
piersy: So I'm suggesting we make the implemntation of `Message.String()` look like this:
```go
func (m *Message) String() string {
	return fmt.Sprintf("{sender: %v, power: %v, msgCode: %v, msg: %v}", m.Address.String(), m.power, m.Code, m.decodedMsg.String())
}
```

piersy: You will need to add `String() string` to the `ConsensusMessage` interface.

consensus/tendermint/core/state_dumper.go
piersy: These functions are also duplicating logic, we can replace them with a single function.

```go
function getBlockHash(b *types.Block) *common.Hash {
	if b == nil {
		return nil
	}
	h := b.Hash()
	return &h
}
```

consensus/tendermint/core/state_dumper.go
yazzaoui: I think it's important to keep`KnownMsgHash` and add the missing hash information in `CurHeightMessages`, 
I'm fine adding `backlogs` and `backlogsUnchecked` though.

consensus/tendermint/core/round_messages.go
piersy: I don't think you need a mutex in getRounds since it is only called when trying to dump the state from the main goroutine of core and `messagesMap.inernal` is also only updated in calls from the main routine of core.

So I think we can just remove `getRounds` and put its functionality directly into `getRoundState`.

consensus/tendermint/core/round_messages.go
```diff
@@ -104,6 +117,18 @@ func (s *roundMessages) SetProposal(proposal *Proposal, msg *Message, verified b
...
+	defer s.mu.RUnlock()
+	return s.prevotes.BlockHashes()
+}
+
+func (s *roundMessages) GetPrecommitValues() []common.Hash {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.precommits.BlockHashes()
+}
+
```
piersy, line 131: I think these functions can be removed since `prevotes` and `precommits` are only modified from core's main goroutine. So it is safe to access these fields directly in the state dump.

consensus/tendermint/backend/backend.go
```diff
@@ -407,6 +416,10 @@ func (sb *Backend) GetContractABI() string {
 	return sb.blockchain.GetAutonityContract().GetContractABI()
 }
 
+func (sb *Backend) GetCoreState() tendermintCore.TendermintState {
```
piersy, line 419: We shouldn't be naming getters with a 'Get' prefix see [Effective Go](https://golang.org/doc/effective_go.html#Getters)

consensus/tendermint/core/core.go
```diff
@@ -134,13 +135,13 @@ type core struct {
...
-	round      int64
-	committee  committee
-	lastHeader *types.Header
+	stateMu   sync.RWMutex
+	height    *big.Int
+	round     int64
+	committee committee
 	// height, round and committeeSet are the ONLY guarded fields.
 	// everything else MUST be accessed only by the main thread.
-	stateMu               sync.RWMutex
```
piersy, line 143: You've move the `stateMu` away from its associated comment.

consensus/tendermint/core/message_set.go
```diff
@@ -106,3 +106,13 @@ func (ms *messageSet) Values(blockHash common.Hash) []Message {
...
+
+func (ms *messageSet) BlockHashes() []common.Hash {
+	ms.messagesMu.RLock()
+	defer ms.messagesMu.RUnlock()
+	blockHashes := make([]common.Hash, 0, len(ms.votes))
+	for key := range ms.votes {
+		blockHashes = append(blockHashes, key)
+	}
+	return blockHashes
+}
```
piersy, line 118: I don't think we need a mutex here. And seeing as this is a function added only for state dumping I think it would be nicer to represent it as a standalone function with the other state dumping functions.

like this:
```go
func blockHashes(messageMap map[common.Hash]map[common.Address]Message) []common.Hash {
	blockHashes := make([]common.Hash, 0, len(messageMap))
	for key := range messageMap {
		blockHashes = append(blockHashes, key)
	}
	return blockHashes
}
```

internal/web3ext/web3ext.go
```diff
@@ -830,6 +830,11 @@ web3._extend({
 			name: 'getWhitelist',
 			call: 'tendermint_getWhitelist',
 			params: 0
+		}),
+		new web3._extend.Method({
```
piersy, line 834: You'll need to update `nodejsconsole/console` with this change as well

consensus/tendermint/core/state_dumper.go
piersy: @yazzaoui Why do we want to keep `knownMsgHash` if we can see from `backlogs` and `backlogsUnchecked` messages that have been received but not processed?

consensus/tendermint/core/state_dumper_test.go
```diff
@@ -0,0 +1,240 @@
...
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		backendMock := NewMockBackend(ctrl)
+		backendMock.EXPECT().Address().Return(clientAddr)
+
+		core := New(backendMock, config.DefaultConfig())
+		core.setCommitteeSet(committeeSet)
+		core.lastHeader = prevBlock.Header()
+	})
```
piersy, line 140: This tests doesn't appear to be testing anything

consensus/tendermint/core/state_dumper_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	"gotest.tools/assert"
+	"math/big"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+func TestStateDumper_GetProposal(t *testing.T) {
```
piersy, line 17:
```suggestion
func TestStateDumper(t *testing.T) {
```

consensus/tendermint/core/state_dumper_test.go
Jason-Zhangxin-Chen: > This test doesn't appear to be testing anything

Removed.

internal/web3ext/web3ext.go
Jason-Zhangxin-Chen: > You'll need to update `nodejsconsole/console` with this change as well

Updated with the new API spec

consensus/tendermint/core/message_set.go
Jason-Zhangxin-Chen: > I don't think we need a mutex here. And seeing as this is a function added only for state dumping I think it would be nicer to represent it as a standalone function with the other state dumping functions.
> 
> like this:
> 
> ```go
> func blockHashes(messageMap map[common.Hash]map[common.Address]Message) []common.Hash {
> 	blockHashes := make([]common.Hash, 0, len(messageMap))
> 	for key := range messageMap {
> 		blockHashes = append(blockHashes, key)
> 	}
> 	return blockHashes
> }
> ```

@piersy  The Read Mutex should be fine since the data is concurently accessed by main loop routine and message sync routine (core/handler.go:194 syncLoop).

consensus/tendermint/core/core.go
Jason-Zhangxin-Chen: > You've moved the `stateMu` away from its associated comment.

No, I didn't. it was moved to line 138.

consensus/tendermint/backend/backend.go
Jason-Zhangxin-Chen: > We shouldn't be naming getters with a 'Get' prefix see [Effective Go](https://golang.org/doc/effective_go.html#Getters)

Renamed with CoreState()

consensus/tendermint/core/round_messages.go
Jason-Zhangxin-Chen: > I think these functions can be removed since `prevotes` and `precommits` are only modified from core's main goroutine. So it is safe to access these fields directly in the state dump.

Removed.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > I think it's important to keep`KnownMsgHash` and add the missing hash information in `CurHeightMessages`,
> I'm fine adding `backlogs` and `backlogsUnchecked` though.

@yazzaoui , yes it is. I added them in my last commit.

consensus/tendermint/backend/api_test.go
```diff
@@ -138,3 +138,40 @@ func TestAPIGetWhitelist(t *testing.T) {
...
+	assert.True(t, tendermintState.IsProposer)
+	assert.True(t, tendermintState.SetValidRoundAndValue)
+	assert.True(t, tendermintState.SentProposal)
+	assert.True(t, tendermintState.SentPrevote)
+	assert.True(t, tendermintState.SentPrecommit)
+	assert.False(t, tendermintState.ProposeTimerStarted)
+	assert.False(t, tendermintState.PrevoteTimerStarted)
+	assert.False(t, tendermintState.PrecommitTimerStarted)
+	assert.Equal(t, uint64(67), tendermintState.QuorumVotePower)
+}
```
aneequesafdar, line 177: I am not how useful this test case is because it is not checking that the correct information is returned when the state changes only what the state is at the start. If we were to add test cases where the state change is checked then it would be a duplication of tests `state_dumper_test.go` since `API.GetCoreState()` is just a placeholder for the core.CoreState(). So I would rather have this removed or only check that we receive something, what do you think @piersy?

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,252 @@
...
+		msg.Power = m.GetPower()
+		msg.Hash = types.RLPHash(m.payload)
+
+		// in case of haven't decode msg yet, set round and height as -1.
+		msg.Round = -1
+		msg.Height = big.NewInt(-1)
+		round, err := m.Round()
+		if err != nil {
+			continue
+		}
```
aneequesafdar, line 179: is it correct to not include decoded messages? Shouldn't we decode them here and add them to the results?

Seems to me that messages that are added to `backlog` and `backlogUnchecked` are always decoded since `storeBacklog` and `storeUnchekedBacklog` are added after `msg.FromPayload(e.Payload)` is called which would decode the message. So I would even argue that this check is redundant. 

Please let me know your though @Jason-Zhangxin-Chen @piersy

eth/fetcher/block_fetcher.go
```diff
@@ -464,6 +464,7 @@ func (f *BlockFetcher) loop() {
 					defer f.wg.Done()
 					announce.fetchBodies(hashes) //nolint
 				}(hashes)
+
```
aneequesafdar, line 467:
```suggestion
```

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,252 @@
...
+	go c.sendEvent(e)
+	// wait for response with timeout.
+	timeout := time.After(time.Second)
+	select {
+	case s := <-c.coreStateCh:
+		state = s
+	case <-timeout:
+		state.Code = -1
+		c.logger.Debug("Waiting for tendermint core state timed out", "elapsed", time.Second)
+	}
```
aneequesafdar, line 98: I feel as though this is poor synchronisation mechanism, this will only work if the sending, receiving and handling of `coreStateRequestEvent` is completed under 1 second. If the time out was reduced such that sending, receiving and handling of `coreStateRequestEvent` cannot be handled it that time the mainEvent loop will hang because `c.soreStateCh` is unbuffered thus it will block until another go routine reads for the channel. This is a huge vulnerability and will cause to liveness issues if it happens at f+1 nodes simultaneously.

Consider the following scenario:
- rpc call for state dump is received
- `coreStateRequestEvent` is sent
- `timeout` channel is ready before `c.coreStateCh` therefore CoreState() will return `TendermintState{Code:-1}`
- `mainEventLoop` finally calls `handleStateDump()` when it handles `coreStateRequestEvent`
- `c.coreStateCh <- state` on line 141 will block the mainEventLoop since there's no one reading from `c.coreStateCh` and now consensus messages cannot be processed thus we have lost liveness

Note the solution is not making `c.coreStateCh` a buffered channel because even if it happens once that the timeout was ready before `c.coreStateCh` you will have off by one error where the state will not be the latest and can no longer be relied on.

Also this synchronisation doesn't consider shutdown at all. There is no place we are closing this channel.

I am also concerned about non-validators sending 1000000s of rpc calls which would delay the consensus message processing leading to undefined behaviour. And there would be no mechanism to punish non-validators unlike how accountability can be used to punish malicious nodes. So I am not sure whether it is a good idea to have this being processed in the same go routine as the rest of the consensus messages. You don't want this to interfere with consensus.

I may be wrong, what do you think @yazzaoui @Jason-Zhangxin-Chen and @piersy?

consensus/tendermint/core/state_dumper_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	"gotest.tools/assert"
+	"math/big"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+func TestStateDumper(t *testing.T) {
```
piersy, line 17: I've opened a PR on this one to try and simplify these tests.

The approach you have used here makes it very hard to understand what is going on in these tests.

This is for a number of reasons:

* There are ~80 lines of closures that separate the variables used in the tests from the test definitions, this means you cant see the variables that are being referenced in the test and need to hold them in mind or constantly scroll back to see them. This is easily solvable by making the closures named functions and moving them outside  the test.
* The closures have access to the variables defined at the top, so its not clear without very carefully going over all the closure code if they actually read or modify the variables in their scope. This is compounded by the closures depending on each other making it very hard to understand what effect on the locally defined variables the closures may have. This again can be easily remedied by using named functions instead, which ensures that the functions only have access to their parameters, this frees the reader from having to worry about the internals of each function, they can instead just focus on the interfaces.
* The closures mix receiving of parameters and referencing of local state, for exampe `checkState`
https://github.com/clearmatics/autonity/blob/86f975773764e146a198786e7f74dfbe92aba51c/consensus/tendermint/core/state_dumper_test.go#L66-L83
This will confuse people because they will assume that the function operates on its parameters, and will likey not realise that hidden in amongst all of its checks it is also referencing the local variables `comitteeSet`, `clientAddr` and `members`. Those variables could be passed as parameters, so there is no reason not to.
* The tests themselves are very different to each other and do not use all of the elements in scope, this makes what should be simple tests very hard to understand since the reader has to mentally filter out all the unused elements in the context to clearly see what a test is testing. This is solved by separating the tests so that they have their own context.
* The closures themselves throw around a huge amount of input and output parameters of the same type, this makes them hard to read and makes it easy to accidentally swap parameters around. There is no need for such complexity, prepareRoundMessages and genRoundVoteMessages can be merged, which dramatically reduces the number of parameters being pushed around and consequently dramatically simplifies the tests.

I have made these changes in #714. It is a start to simplifying these tests but I think there is a lot more we could do to make them even simpler.

consensus/tendermint/backend/api_test.go
Jason-Zhangxin-Chen: > this removed or only check that we re

Removed.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > is it correct to not include decoded messages? Shouldn't we decode them here and add them to the results?
> 
> Seems to me that messages that are added to `backlog` and `backlogUnchecked` are always decoded since `storeBacklog` and `storeUnchekedBacklog` are added after `msg.FromPayload(e.Payload)` is called which would decode the message. So I would even argue that this check is redundant.
> 
> Please let me know your though @Jason-Zhangxin-Chen @piersy

Good catch, I will fix it.

Jason-Zhangxin-Chen: > > is it correct to not include decoded messages? Shouldn't we decode them here and add them to the results?
> > Seems to me that messages that are added to `backlog` and `backlogUnchecked` are always decoded since `storeBacklog` and `storeUnchekedBacklog` are added after `msg.FromPayload(e.Payload)` is called which would decode the message. So I would even argue that this check is redundant.
> > Please let me know your though @Jason-Zhangxin-Chen @piersy
> 
> Good catch, I fix it in my last commits.

consensus/tendermint/core/round_messages.go
```diff
@@ -104,6 +117,18 @@ func (s *roundMessages) SetProposal(proposal *Proposal, msg *Message, verified b
 	s.proposal = proposal
 }
 
+func (s *roundMessages) GetPrevoteValues() []common.Hash {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.prevotes.BlockHashes()
```
piersy:
```go
return blockHashes(s.prevotes.votes)
```

piersy:
```suggestion
	return s.prevotes.BlockHashes()
```

consensus/tendermint/core/round_messages.go
```diff
@@ -104,6 +117,18 @@ func (s *roundMessages) SetProposal(proposal *Proposal, msg *Message, verified b
...
+}
+
+func (s *roundMessages) GetPrecommitValues() []common.Hash {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.precommits.BlockHashes()
```
piersy:
```suggestion
	return blockHashes(s.prevotes.votes)
```

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > I feel as though this is poor synchronisation mechanism, this will only work if the sending, receiving and handling of `coreStateRequestEvent` is completed under 1 second. If the time out was reduced such that sending, receiving and handling of `coreStateRequestEvent` cannot be handled it that time the mainEvent loop will hang because `c.soreStateCh` is unbuffered thus it will block until another go routine reads for the channel. This is a huge vulnerability and will cause to liveness issues if it happens at f+1 nodes simultaneously.
> 
> Consider the following scenario:
> 
> * rpc call for state dump is received
> * `coreStateRequestEvent` is sent
> * `timeout` channel is ready before `c.coreStateCh` therefore CoreState() will return `TendermintState{Code:-1}`
> * `mainEventLoop` finally calls `handleStateDump()` when it handles `coreStateRequestEvent`
> * `c.coreStateCh <- state` on line 141 will block the mainEventLoop since there's no one reading from `c.coreStateCh` and now consensus messages cannot be processed thus we have lost liveness
> 
> Note the solution is not making `c.coreStateCh` a buffered channel because even if it happens once that the timeout was ready before `c.coreStateCh` you will have off by one error where the state will not be the latest and can no longer be relied on.
> 
> Also this synchronisation doesn't consider shutdown at all. There is no place we are closing this channel.
> 
> I am also concerned about non-validators sending 1000000s of rpc calls which would delay the consensus message processing leading to undefined behaviour. And there would be no mechanism to punish non-validators unlike how accountability can be used to punish malicious nodes. So I am not sure whether it is a good idea to have this being processed in the same go routine as the rest of the consensus messages. You don't want this to interfere with consensus.
> 
> I may be wrong, what do you think @yazzaoui @Jason-Zhangxin-Chen and @piersy?

@aneequesafdar , good catch. I will fix it by using a buffered channel and an unique ID of the request. If the ID is not expected, it indicates that the state is not latest one, then resent the request.

consensus/tendermint/core/round_messages.go
Jason-Zhangxin-Chen: > 

Resloved.

consensus/tendermint/core/round_messages.go
Jason-Zhangxin-Chen: > 

Resolved in my last commit.

consensus/tendermint/core/state_dumper.go
Jason-Zhangxin-Chen: > > I feel as though this is poor synchronisation mechanism, this will only work if the sending, receiving and handling of `coreStateRequestEvent` is completed under 1 second. If the time out was reduced such that sending, receiving and handling of `coreStateRequestEvent` cannot be handled it that time the mainEvent loop will hang because `c.soreStateCh` is unbuffered thus it will block until another go routine reads for the channel. This is a huge vulnerability and will cause to liveness issues if it happens at f+1 nodes simultaneously.
> > Consider the following scenario:
> > 
> > * rpc call for state dump is received
> > * `coreStateRequestEvent` is sent
> > * `timeout` channel is ready before `c.coreStateCh` therefore CoreState() will return `TendermintState{Code:-1}`
> > * `mainEventLoop` finally calls `handleStateDump()` when it handles `coreStateRequestEvent`
> > * `c.coreStateCh <- state` on line 141 will block the mainEventLoop since there's no one reading from `c.coreStateCh` and now consensus messages cannot be processed thus we have lost liveness
> > 
> > Note the solution is not making `c.coreStateCh` a buffered channel because even if it happens once that the timeout was ready before `c.coreStateCh` you will have off by one error where the state will not be the latest and can no longer be relied on.
> > Also this synchronisation doesn't consider shutdown at all. There is no place we are closing this channel.
> > I am also concerned about non-validators sending 1000000s of rpc calls which would delay the consensus message processing leading to undefined behaviour. And there would be no mechanism to punish non-validators unlike how accountability can be used to punish malicious nodes. So I am not sure whether it is a good idea to have this being processed in the same go routine as the rest of the consensus messages. You don't want this to interfere with consensus.
> > I may be wrong, what do you think @yazzaoui @Jason-Zhangxin-Chen and @piersy?
> 
> @aneequesafdar , good catch. I will fix it by using a buffered channel and an unique ID of the request. If the ID is not expected, it indicates that the state is not latest one, then resent the request. With this none blocking send and receive example: https://gobyexample.com/non-blocking-channel-operations

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,266 @@
...
+	timeout := time.After(time.Second)
+	select {
+	case s := <-e.stateChan:
+		state = s
+	case <-timeout:
+		state.Code = -1
+		c.logger.Debug("Waiting for tendermint core state timed out", "elapsed", time.Second)
+	}
+
+	return state
```
aneequesafdar, line 102: time out is no longer required

```suggestion
	var e = coreStateRequestEvent{
		stateChan: make(chan TendermintState),
	}
	go c.sendEvent(e)
	return <-e.stateChan
```

consensus/tendermint/core/state_dumper.go
```diff
@@ -0,0 +1,266 @@
...
+		Code:         0,
+	}
+
+	// for none blocking send state.
+	select {
+	case e.stateChan <- state:
+		c.logger.Debug("core state msg sent.")
+	default:
+		c.logger.Debug("core state msg dropped.")
+	}
```
aneequesafdar, line 150: this problematic because it introduces a race condition where without timeout expiry you would have different result depending on which go routine checks it select statement first (line 94 vs line 145)

```suggestion
      c.logger.Debug("sending core state msg")
	e.stateChan <- state
```

consensus/tendermint/core/state_dumper_test.go
Jason-Zhangxin-Chen: @piersy , thank you. I have merged #714 into my branch, and remove the timeout test since it is not needed anymore. Now we have 4 test cases left in the state_dumper_test.go

aneequesafdar (approved):

PR #689, opened 2020-09-25
To fix out of range panic for F+1 node reseting test.

Jason-Zhangxin-Chen:
To close: https://github.com/clearmatics/autonity/issues/646 By according to tendermint pseudo code:
```
Line55: upon f + 1 <∗, h_p, round, ∗, ∗> with round > round_p do
Line56: StartRound(round)
```
the round change in client would jumps (from N to N+3, etc..) rather than liner increasing by round ++, https://github.com/clearmatics/autonity/blob/develop/consensus/tendermint/core/round_messages.go#L62-L68 would crash at line 65. The variable “i” is round number which as Key of the map s.internal stores round messages, taking “i” as the index of the 2D slice would make the memory write out of range if the “i” (round number) exceed the capacity of 2D slice. For the time being, round jump could hardly reproduced, but it would happens.

### DIFF
--- a/consensus/tendermint/core/round_messages.go
+++ b/consensus/tendermint/core/round_messages.go
@@ -61,9 +61,11 @@ func (s *messagesMap) GetMessages() []*Message {
 
 	msgs := make([][]*Message, len(s.internal))
 	var totalLen int
-	for i, state := range s.internal {
+	i := 0
+	for _, state := range s.internal {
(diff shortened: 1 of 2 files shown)


piersy (approved):
Looking good!

yazzaoui (approved):
LGTM

PR #691, opened 2020-09-28
Autonity rpc

### LINKED ISSUE #562: Create autonity JSON RPC-API namespace
# Problem

It is complicated to interact with the autonity contract. Accessing important data from the contract should be easy as possible, e.g. minimum gas price, current committee. current whitelist... without having to use a library like web3. 
JSON RPC is the way to go but the endpoint would need to be created in a dynamic way as layer 1 is unaware of all of these functions. 

# Implementation

- Parse the autonity contract ABI when the node starts and extract all the view functions. In the JSON ABI this translates to `stateMutability: "view"` and  "type: function" .
- Create a JSON RPC endpoint for every one of those functions with the `autonity` namespace.

e.g. 
```
 curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"autonity_getCommittee","params":[],"id":74}' http://localhost:6000 
```

will internally execute an evm call (via `AutonityContractCall` ideally ) on the head state  to return :

```json
{                                                                                                                                        
  "jsonrpc": "2.0",                                                                                                                      
  "id": 74,                                                                                                                              
  "result": [                                                                                                                       
      {                                                                                                                                  
        "address": "0xe0608a4754b5558dec04fefe0bb898a75fe01c95",                                                                         
        "votingPower": 2                                                                                                                 
      },
      {                                                                                                                                  
        "address": "0xe0608a4754b5558dec04fefe0bb898a75fe01c95",                                                                         
        "votingPower": 2                                                                                                                 
      },
      {                                                                                                                                  
        "address": "0xe0608a4754b5558dec04fefe0bb898a75fe01c95",                                                                         
        "votingPower": 2                                                                                                                 
      }, 
      {                                                                                                                                  
        "address": "0xe0608a4754b5558dec04fefe0bb898a75fe01c95",                                                                         
        "votingPower": 2                                                                                                                 
      }                                                                                                                                    
    ]                                                                                                                                                                                                                                                                                                                                                                                               
}    
```
# Acceptance Criteria

- Implement the `autonity` namespace with RPC apis for **every** view function of the Autonity Contract.  The list of available functions should not be hardcoded but dynamically retrieved.

- The returned results should be in the JSON format (not ABI encoded).

-  Support for parametrized functions is not required (might be a bit tricky to pull it off) but would be a good improvement, even in some limited form (e.g no ABIv2 support).

aneequesafdar:
closes #562 

This pull request adds support for accessing view functions of the autonity contract. The following functions can now be accessed through an rpc call:

```
	aut_operatorAccount
	aut_committeeSize
	aut_deployer
	aut_allowance
	aut_getNewContract
	aut_getState
	aut_getVersion
	aut_getCommittee
	aut_getValidators
	aut_getStakeholders
	aut_getWhitelist
	aut_balanceOf
	aut_totalSupply
	aut_getUser
	aut_getMaxCommitteeSize
	aut_getMinimumGasPrice
	aut_getProposer
	aut_dumpEconomicMetrics

```
For these rpc's to be accessible `aut` must be added to `--rpcapi` flag when autonity is started, for example, `rpcapi` flag would be: `--rpcapi 'personal,db,eth,net,web3,txpool,miner,tendermint,autonity,aut'`. 

Examples of curl command with its output:
```
curl -X POST -H "Content-type: application/json" --data '{"jsonrpc":"2.0","method":"aut_getCommittee","id":2}' http://localhost:6003 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   381  100   316  100    65   308k  65000 --:--:-- --:--:-- --:--:--  372k
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": [
    {
      "addr": "0x298ecef96a2c6708ec2f23750103683f5769f69b",
      "votingPower": 2
    },
    {
      "addr": "0x7a0b8f8db79455d169981802bc75e85d9fab404b",
      "votingPower": 1
    },
    {
      "addr": "0xcbd17d2512c0b94f22ec709c38dce78a84f2e888",
      "votingPower": 1
    },
    {
      "addr": "0xd2ba3d8a432095eb93b6a0e59664d7d7fb37a68f",
      "votingPower": 1
    }
  ]
}

curl -X POST -H "Content-type: application/json" --data '{"jsonrpc":"2.0","method":"aut_checkMember","params":["0x9b4dfa42a868cd3899f59fb22d96a155752fe1cb"],"id":2}' http://localhost:6000  | jq

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   160  100    40  100   120  40000   117k --:--:-- --:--:-- --:--:--  156k
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": false
}
```


Jason-Zhangxin-Chen (approved):
LGTM. Just one comment: to add a local e2e testcase which calls one of the API, to make sure that no one will break this feature accidently in the future.

eth/backend.go
```diff
@@ -318,6 +318,15 @@ func (s *Ethereum) APIs() []rpc.API {
 		apis = append(apis, s.lesServer.APIs()...)
 	}
 
+	if _, ok := s.engine.(consensus.BFT); ok {
+		apis = append(apis, rpc.API{
+			Namespace: "autonityContract",
```
yazzaoui, line 323:
```suggestion
			Namespace: "aut",
```

piersy (changes requested):
This is really your comment @aneequesafdar, we should have some consistency in the naming. so either we prefix all calls with 'get' or we remove the 'get' prefix. Looking at the ethereum rpcs it looks like having the 'get' prefix is the standard approach so I'd suggest we add that to them all.

aneequesafdar:
Which get are you talking about, @piersy ?

accounts/abi/argument.go
```diff
@@ -75,6 +75,15 @@ func (arguments Arguments) isTuple() bool {
 	return len(arguments) > 1
 }
 
+// isTuple returns true for non-atomic constructs, like (uint,uint) or uint[]
```
piersy, line 78: This documentation looks out of place, this function should have a godoc comment.

contracts/autonity/autonity.go
```diff
@@ -263,9 +263,14 @@ func (ac *Contract) upgradeAbiCache(newAbi string) error {
 	}
 
 	ac.contractABI = &newABI
+	ac.stringContractABI = newAbi
 	return nil
 }
 
 func (ac *Contract) GetContractABI() string {
```
piersy, line 270: I'd suggest renaming this as `StringABI `to distinguish it from `ABI`

contracts/autonity/calls.go
```diff
@@ -135,6 +126,36 @@ func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.H
 	return nil
 }
 
+func (ac *Contract) AutonityContractCallUnpackIntoMap(statedb *state.StateDB, header *types.Header, function string, result map[string]interface{}, args ...interface{}) error {
```
piersy, line 129: This needs documentation.

contracts/autonity/autonity.go
```diff
@@ -263,9 +263,14 @@ func (ac *Contract) upgradeAbiCache(newAbi string) error {
...
 	return ac.stringContractABI
 }
+
+func (ac *Contract) ABI() *abi.ABI {
```
piersy, line 274: This needs documentation.

contracts/autonity/calls.go
```diff
@@ -135,6 +126,36 @@ func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.H
...
+	}
+
+	if err := ac.contractABI.UnpackIntoMap(result, function, ret); err != nil {
+		log.Error("Could not unpack returned value into map[string]interface{}", "function", function)
+		return err
+	}
+	return nil
+}
+
+func (ac *Contract) callContractFunc(statedb *state.StateDB, header *types.Header, function string, args []interface{}) ([]byte, error) {
```
piersy, line 142: This needs documentation.

eth/api.go
```diff
@@ -528,3 +529,66 @@ func (api *PrivateDebugAPI) getModifiedAccounts(startBlock, endBlock *types.Bloc
 	}
 	return dirty, nil
 }
+
```
piersy, line 532: This whole file needs documentation.

rpc/service.go
```diff
@@ -116,19 +121,36 @@ func (r *serviceRegistry) subscription(service, name string) *callback {
...
+			}
+			name := formatName(method.Name)
+			callbacks[name] = cb
 		}
-		cb := newCallback(receiver, method.Func)
+	}
+	return callbacks
+}
+
+// registerCallbacks register new callbacks is the receiver implement DynamicCallbacks.
```
piersy, line 144: This needs some better documentation

rpc/service.go
```diff
@@ -58,6 +58,11 @@ type callback struct {
 	isSubscribe bool           // true if this is a subscription callback
 }
 
+// DynamicCallbacks is used for receivers with dynamic API's method to retrieve.
```
piersy, line 61: I think we need to expand this comment to explain what a receiver is and explain more what a dynamic api is.

contracts/autonity/autonity.go
aneequesafdar: Seems to be self-explanatory but I have added a comment.

eth/api.go
aneequesafdar: We didn't write the whole file so I can't document the whole file

accounts/abi/argument.go
aneequesafdar: Please look at the latest changes

contracts/autonity/autonity.go
aneequesafdar: Please look at the latest changes

contracts/autonity/calls.go
aneequesafdar: Please look at the latest changes

contracts/autonity/autonity.go
aneequesafdar: Please look at the latest changes

contracts/autonity/calls.go
aneequesafdar: Please look at the latest changes

eth/api.go
aneequesafdar: Please look at the latest changes for the comments for our work

rpc/service.go
aneequesafdar: Please look at the latest changes

rpc/service.go
aneequesafdar: Please look at the latest changes

accounts/abi/argument.go
```diff
@@ -75,6 +75,15 @@ func (arguments Arguments) isTuple() bool {
 	return len(arguments) > 1
 }
 
+// Types returns the list of underlying []reflect.Type for Arguments
```
piersy, line 78:
```suggestion
// Types returns the list of underlying []reflect.Type for Arguments.
```

The first sentence of godoc is formatted differently so the full stop is important.

contracts/autonity/calls.go
```diff
@@ -135,6 +126,41 @@ func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.H
 	return nil
 }
 
+// AutonityContractCallUnpackIntoMap uses the state and header to create an evm object which will call the autonity
+// contract function along with the arguments, in any, and return the result in a map[string]interface{} object.
+// The caller need to provide the object into which the results need to be returned.
```
piersy, line 131:
```suggestion
// The caller needs to provide the object into which the results need to be returned.
```

contracts/autonity/calls.go
```diff
@@ -135,6 +126,41 @@ func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.H
 	return nil
 }
 
+// AutonityContractCallUnpackIntoMap uses the state and header to create an evm object which will call the autonity
+// contract function along with the arguments, in any, and return the result in a map[string]interface{} object.
```
piersy, line 130:
```suggestion
// contract function along with the arguments, if any, and return the result in a map[string]interface{} object.
```

piersy:
> Which "get" are you talking about, @piersy?
> 
> If you are referring to the names of the rpc calls, those are all dynamically generated from the names of the functions present in the autonity contract. Ideally you would want the names of the functions to remain the same as the names of the functions in the contract so that people know that calling the contract via web3 or transaction or through this rpc will all return the same result.

I agree that we want the methods to match the names of the functions, but I think before we expose these as an API we should get our functions named in a consistent way. If we put this out there as it is and people start building on it we'll either be stuck with the current function names or we could change the function names without changing the rpc names and then we'd be in the situation where the names differ.

piersy (changes requested):
I don't see any tests for this, we should have them.

aneequesafdar:
>I agree that we want the methods to match the names of the functions, but I think before we expose these as an API we should get our functions named in a consistent way. If we put this out there as it is and people start building on it we'll either be stuck with the current function names or we could change the function names without changing the rpc names and then we'd be in the situation where the names differ.

But that can be said for the autonity contract as well and the transactions people are sending will make it difficult to change the function name. So I don't see this as an issue for this pr as they are dynamically generated and the rpc will change when the autonity contract is changed.

piersy:
> This PR is just describing the mechanism for creating the rpc, so once the function names of auotnity contract are finalised it will be automatically reflected. The function names that I mention in the description are based on the current autonity contract, if you add another abi in genesis, they will be different.
> 
> For me rpc name will always be from autonity contract and it should take precedence because that is what the API is giving access to.

If we make it clear to everyone that these rpcs are not stable and that people cannot rely on the API from release to release then I think this is not a problem.

eth/api.go
```diff
@@ -533,3 +534,72 @@ func (api *PrivateDebugAPI) getModifiedAccounts(startBlock, endBlock *types.Bloc
...
+// AutonityContractAPI implements DynamicCallbacks defined in rpc package to dynamically expose view function of the
+// autonity contract through rpc endpoint.
+type AutonityContractAPI struct {
+	eth   *Ethereum
```
piersy, line 541: I don't think this is used. And if that is the case we should be constructing `AutonityContractAPI` with an instance of `Blockchain` since that is the smallest dependency that will satisfy `AutonityContractAPI's` requirements.

contracts/autonity/calls.go
piersy: The documentation says it returns a map but it just returns an error

eth/api.go
```diff
@@ -533,3 +534,71 @@ func (api *PrivateDebugAPI) getModifiedAccounts(startBlock, endBlock *types.Bloc
...
+							return makereturn(v, nil)
+						}
+					}
+					return makereturn(res, nil)
+				})
+		}
+	}
+
+	return contractViewMethods
+}
```
piersy, line 604: I think this should be in the autonity package. It's definitely not part of the eth api

rpc/service.go
```diff
@@ -116,19 +131,36 @@ func (r *serviceRegistry) subscription(service, name string) *callback {
...
+			name := formatName(method.Name)
+			callbacks[name] = cb
 		}
-		cb := newCallback(receiver, method.Func)
+	}
+	return callbacks
+}
+
+// newDynamicCallbacks creates callbacks for types which implement DynamicCallbacks and add them to callbacks' map.
+func newDynamicCallbacks(callbacks map[string]*callback, receiver reflect.Value, methods map[string]reflect.Value) {
```
piersy, line 155: Why not have this return the callbacks, the name is also confusing because you would expect it to make an instance of DynamicCallbacks.

contracts/autonity/calls.go
```diff
@@ -135,6 +126,41 @@ func (ac *Contract) AutonityContractCall(statedb *state.StateDB, header *types.H
...
+	}
+
+	if err := ac.contractABI.UnpackIntoMap(result, function, ret); err != nil {
+		log.Error("Could not unpack returned value into map[string]interface{}", "function", function)
+		return err
+	}
+	return nil
+}
+
+// callContractFunc create creaties an evm object required to call the autonity contract function and returns the result
```
piersy, line 145:
```suggestion
// callContractFunc create creates an evm object required to call the autonity contract function and returns the result
```

rpc/service.go
```diff
@@ -58,6 +58,21 @@ type callback struct {
...
+// Any instance of golang type with methods at the time of calling take a receiver as the first argument. Take the
+// following method declaration as example:
+//
+//`func (a aType) someMethod(){}`
+//
+// The instance of aType at runtime will be the receiver for someMethod() and will also be the first argument passed
+// someMethod().
+type DynamicCallbacks interface {
+	// Call should return a map of function names to function signatures, which will be used to create callbacks.
+	Calls() map[string]reflect.Value
```
piersy, line 73:
```suggestion
	MethodSignatures() map[string]reflect.Value
```

rpc/service.go
```diff
@@ -58,6 +58,21 @@ type callback struct {
...
+//
+// Any instance of golang type with methods at the time of calling take a receiver as the first argument. Take the
+// following method declaration as example:
+//
+//`func (a aType) someMethod(){}`
+//
+// The instance of aType at runtime will be the receiver for someMethod() and will also be the first argument passed
+// someMethod().
+type DynamicCallbacks interface {
+	// Call should return a map of function names to function signatures, which will be used to create callbacks.
```
piersy, line 72:
```suggestion
	// Call should return a map of function names to functions, which will be used to create callbacks.
```

rpc/service.go
```diff
@@ -58,6 +58,21 @@ type callback struct {
...
+// and expose them through rpc endpoint.
+//
+// Any instance of golang type with methods at the time of calling take a receiver as the first argument. Take the
+// following method declaration as example:
+//
+//`func (a aType) someMethod(){}`
+//
+// The instance of aType at runtime will be the receiver for someMethod() and will also be the first argument passed
+// someMethod().
+type DynamicCallbacks interface {
```
piersy, line 71: I'd probably just call this `Methods` there's nothing intrinsically 'dynamic' about this interface it just provides a map of methods.

Actually its not clear if this is returning methods or functions?

piersy:
@yazzaoui @Jason-Zhangxin-Chen @aneequesafdar  I'm not sure what to do about this PR.

It seems that deep inside the rpc mechanism there is an assumption that pointer arguments to a method are optional. This allows rpc calls to progress when they have not been called with the corresponding pointer arguments. It looks like the existing eth API methods are written to avoid the use of pointers fro required arguments. In the case of getProposer its arguments are uint256 which transalte to *big.Int on the go side. This allows the call to continue when neither height or round is provided, resulting in the rpc handler crashing.

I've tried removing the check for optional pointer parameters, but this resulted in lots of other stuff breaking.

https://github.com/clearmatics/autonity/blob/25e42b420098cf4d28a1dcb762fe56c7794699ee/rpc/json.go#L303-L305

yazzaoui:
@piersy Please check my latest commit.

eth/api.go
piersy: Now that I have tried moving this, I can say that it can't go in the autonity package because of the circular dependencies we would encounter.

piersy (approved):

eth/api.go
```diff
@@ -583,29 +582,33 @@ func contractABIMethods(bc *core.BlockChain) map[string]reflect.Value {
...
 					}
-					res := make(map[string]interface{})
-					err = contract.AutonityContractCallUnpackIntoMap(stateDB, bc.CurrentHeader(), functionName, res, iargs...)
+
+					// Pack the arguments call the function and then unpack the result and return it.
+					packedArgs, err := contractABI.Pack(functionName, iargs...)
 					if err != nil {
 						return makereturn(nil, err)
 					}
-					if len(res) == 1 {
```
yazzaoui: Can we have this back? Found it would make the APIs easier to use.

piersy: I've restored previous functionality, but in a simpler way

eth/api.go
```diff
@@ -602,6 +602,12 @@ func NewAutonityContractAPI(bc *core.BlockChain, ac *autonity.Contract) *Autonit
 						return makereturn(nil, err)
 					}
 					result, err := contractABI.Unpack(functionName, packedResult)
+
+					// If the result is a slice with one element only then just return the element.
+					v := reflect.ValueOf(result)
```
yazzaoui, line 607: Wouldn't
```golang
if len(result) == 1 {
  return makereturn(result[0], err)
}
```
be enough ?

yazzaoui (approved):
LGTM

eth/api.go
piersy: Yeah, it would :) I thought I was dealing with a `reflect.Value`

Jason-Zhangxin-Chen (approved):
LGTM

PR #692, opened 2020-09-30
Combine Autonity init-chain and start-up.

Jason-Zhangxin-Chen:
To close issue: https://github.com/clearmatics/autonity/issues/682
This PR is an enhancement of autonity client, it does not need user to execute a seperate init command to init the chain data before node start. By specifying a flag: --genesis with your genesis json file, client will try to init genesis the chain data for user, and then start node. You don't need to run extra command to start node anymore.

### DIFF
--- a/cmd/autonity/chaincmd.go
+++ b/cmd/autonity/chaincmd.go
@@ -199,14 +199,21 @@ Use "ethereum dump 0" to dump the genesis block.`,
 // initGenesis will initialise the given JSON format genesis file and writes it as
 // the zero'd block (i.e. genesis) or will fail hard if it can't succeed.
 func initGenesis(ctx *cli.Context) error {
-	// Make sure we have a valid genesis JSON
+	// If user does not specify a genesis file on genesis flag, start node from current data dir.
 	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) {
-		log.Info("Genesis flag is not set, start node directly.")
-		log.Info("In case of 1st time start, please start client with setting genesis flag, " +
+		log.Info("genesis flag is not set, start node directly from data dir.")
+		log.Info("In case of 1st time start, please start client with setting --genesis flag, " +
 			"the client will auto init genesis block before attach to network.")
 		return nil
 	}
+
+	// If there is no genesis block presented, init a genesis block with genesis file,
+	// otherwise it checks if genesis block is compatible and matched with genesis file
+	// and return error if they are not, the error would terminate autonity client.
+
+	// Make sure we have a valid genesis JSON.
 	genesisPath := ctx.GlobalString(utils.InitGenesisFlag.Name)
+	log.Info("Trying to init genesis block with genesis file", "file", genesisPath)
 	if len(genesisPath) == 0 {
 		utils.Fatalf("Must supply path to genesis JSON file")
 		return fmt.Errorf("must supply path to genesis JSON file")
--- a/cmd/autonity/chaincmd.go
+++ b/cmd/autonity/chaincmd.go
@@ -45,22 +45,6 @@ import (
 )
 
 var (
-	initCommand = cli.Command{
-		Action:    utils.MigrateFlags(initGenesis),
-		Name:      "init",
-		Usage:     "Bootstrap and initialize a new genesis block",
-		ArgsUsage: "<genesisPath>",
-		Flags: []cli.Flag{
-			utils.DataDirFlag,
-		},
-		Category: "BLOCKCHAIN COMMANDS",
-		Description: `
-The init command initializes a new genesis block and definition for the network.
-This is a destructive action and changes the network in which you will be
-participating.
-
-It expects the genesis file as argument.`,
-	}
 	dumpGenesisCommand = cli.Command{
 		Action:    utils.MigrateFlags(dumpGenesis),
 		Name:      "dumpgenesis",
@@ -216,29 +200,41 @@ Use "ethereum dump 0" to dump the genesis block.`,
 // the zero'd block (i.e. genesis) or will fail hard if it can't succeed.
 func initGenesis(ctx *cli.Context) error {
 	// Make sure we have a valid genesis JSON
-	genesisPath := ctx.Args().First()
+	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) {
+		log.Info("Genesis flag is not set, start node directly.")
+		log.Info("In case of 1st time start, please start client with setting genesis flag, " +
+			"the client will auto init genesis block before attach to network.")
+		return nil
+	}
+	genesisPath := ctx.GlobalString(utils.InitGenesisFlag.Name)
 	if len(genesisPath) == 0 {
 		utils.Fatalf("Must supply path to genesis JSON file")
+		return fmt.Errorf("must supply path to genesis JSON file")
 	}
 	file, err := os.Open(genesisPath)
 	if err != nil {
 		utils.Fatalf("Failed to read genesis file: %v", err)
+		return err
 	}
 	defer file.Close()
 
 	genesis := new(core.Genesis)
 	if err := json.NewDecoder(file).Decode(genesis); err != nil {
 		utils.Fatalf("invalid genesis file: %v", err)
+		return err
 	}
 	// Make AutonityContract and Tendermint consensus mandatory for the time being.
 	if genesis.Config == nil {
 		utils.Fatalf("No Autonity Contract and Tendermint configs section in genesis")
+		return fmt.Errorf("no Autonity Contract and Tendermint configs section in genesis")
 	}
 	if genesis.Config.AutonityContractConfig == nil {
 		utils.Fatalf("No Autonity Contract config section in genesis")
+		return fmt.Errorf("no Autonity Contract config section in genesis")
 	}
 	if genesis.Config.Tendermint == nil {
 		utils.Fatalf("No Tendermint config section in genesis")
+		return fmt.Errorf("no Tendermint config section in genesis")
 	}
 
 	if err := genesis.Config.AutonityContractConfig.Prepare(); err != nil {
@@ -256,10 +252,12 @@ func initGenesis(ctx *cli.Context) error {
 		chaindb, err := stack.OpenDatabase(name, 0, 0, "")
 		if err != nil {
 			utils.Fatalf("Failed to open database: %v", err)
+			return err
 		}
 		_, hash, err := core.SetupGenesisBlock(chaindb, genesis)
 		if err != nil {
 			utils.Fatalf("Failed to write genesis block: %v", err)
+			return err
 		}
 		chaindb.Close()
 		log.Info("Successfully wrote genesis state", "database", name, "hash", hash)
--- a/cmd/autonity/main.go
+++ b/cmd/autonity/main.go
@@ -202,7 +202,6 @@ func init() {
 	app.Copyright = "Copyright 2013-2020 The go-ethereum Authors"
 	app.Commands = []cli.Command{
 		// See chaincmd.go:
-		initCommand,
 		importCommand,
 		exportCommand,
 		importPreimagesCommand,
@@ -241,7 +240,12 @@ func init() {
 	app.Flags = append(app.Flags, metricsFlags...)
 
 	app.Before = func(ctx *cli.Context) error {
-		return debug.Setup(ctx)
+		err := debug.Setup(ctx)
+		if err != nil {
+			return err
+		}
+		// check genesis flag and try to init genesis block, error will stop the app execution.
+		return initGenesis(ctx)
 	}
 	app.After = func(ctx *cli.Context) error {
 		debug.Exit()
--- a/cmd/autonity/main.go
+++ b/cmd/autonity/main.go
@@ -64,6 +64,7 @@ var (
 		utils.LegacyBootnodesV4Flag,
 		utils.LegacyBootnodesV5Flag,
 		utils.DataDirFlag,
+		utils.InitGenesisFlag,
 		utils.AncientFlag,
 		utils.KeyStoreDirFlag,
 		utils.ExternalSignerFlag,
--- a/cmd/autonity/main.go
+++ b/cmd/autonity/main.go
@@ -245,7 +245,7 @@ func init() {
 		if err != nil {
 			return err
 		}
-		// check genesis flag and try to init genesis block, error will stop the app execution.
+		// Try to init chain before autonity start up. Any error from initGenesis would terminate node start up.
 		return initGenesis(ctx)
 	}
 	app.After = func(ctx *cli.Context) error {
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -134,6 +134,11 @@ func printHelp(out io.Writer, templ string, data interface{}) {
 
 var (
 	// General settings
+	InitGenesisFlag = cli.StringFlag{
+		Name:  "genesis",
+		Usage: "Path to the genesis json file",
+		Value: "",
+	}
 	DataDirFlag = DirectoryFlag{
 		Name:  "datadir",
 		Usage: "Data directory for the databases and keystore",
--- a/contracts/autonity/contract/test/autonity/autonity-start.sh
+++ b/contracts/autonity/contract/test/autonity/autonity-start.sh
@@ -19,7 +19,7 @@ RPC_API="tendermint,console,eth,web3,admin,debug,miner,personal,txpool,net"
 # start the node with the keystore and nodekey
 echo "Autonity START"
 $AUTONITY \
-  --genesis "./genesis-tendermint.json" \
+  --genesis genesis-tendermint.json \
   --datadir $DATADIR \
   --nodekey $NODEKEY \
   --keystore $KEYSTORE \
--- a/contracts/autonity/contract/test/autonity/autonity-start.sh
+++ b/contracts/autonity/contract/test/autonity/autonity-start.sh
@@ -19,7 +19,7 @@ RPC_API="tendermint,console,eth,web3,admin,debug,miner,personal,txpool,net"
 # start the node with the keystore and nodekey
 echo "Autonity START"
 $AUTONITY \
-  --genesis genesis-tendermint.json \
+  --genesis "./genesis-tendermint.json" \
   --datadir $DATADIR \
   --nodekey $NODEKEY \
   --keystore $KEYSTORE \
--- a/contracts/autonity/contract/test/autonity/autonity-start.sh
+++ b/contracts/autonity/contract/test/autonity/autonity-start.sh
@@ -13,12 +13,13 @@ RPC_ADDR=127.0.0.1
 RPC_API="tendermint,console,eth,web3,admin,debug,miner,personal,txpool,net"
 
 # init the data directory
-echo "Autonity INIT $RPC_ADDR"
-$AUTONITY init --datadir $DATADIR genesis-tendermint.json
+# echo "Autonity INIT $RPC_ADDR"
+# $AUTONITY init --datadir $DATADIR genesis-tendermint.json
 
 # start the node with the keystore and nodekey
 echo "Autonity START"
 $AUTONITY \
+  --genesis genesis-tendermint.json \
   --datadir $DATADIR \
   --nodekey $NODEKEY \
   --keystore $KEYSTORE \
--- a/cmd/autonity/usage.go
+++ b/cmd/autonity/usage.go
@@ -66,6 +66,7 @@ var AppHelpFlagGroups = []flagGroup{
 		Name: "ETHEREUM",
 		Flags: []cli.Flag{
 			configFileFlag,
+			utils.InitGenesisFlag,
 			utils.DataDirFlag,
 			utils.AncientFlag,
 			utils.KeyStoreDirFlag,


yazzaoui:
- What's the behavior if `--genesis` is supplied but the blockchain is already initialized? Would it still retain the capability of dealing with new hard-fork configs if we have a new compatible genesis file?
- Did you keep or did you remove the `autonity init` command?

Jason-Zhangxin-Chen:
> * What's the behavior if `--genesis` is supplied but the blockchain is already initialized? Would it still retain the capability of dealing with new hard-fork configs if we have a new compatible genesis file?

If the chain is initialized before, the initGenesis checks if the genesis block is exactly the same by comparing the new genesis block generated from the genesis file, if it is mis-match, then an error will rise to stop the scenario and also node start will be terminated. After genesis block is matched, then it continue to check chain config includes fork configurations, if the chain configuration is missing from current database, it will apply the new chain configuration provided by the new genesis file, if the chain configuration is presented from current database, then it will check if the chain conf is compatible between the two, if it is not compatible, then error rise to stop the scenario, and also node start will be termintated too, finally if chain conf is compatible, the new chain conf will be apply into database too.

> * Did you keep or did you remove the `autonity init` command?
I removed the init command, so user will not confuse this init command when we have auto init and start up.

yazzaoui:
Ok sounds good, and what if there is no `--genesis` flag supplied and no chain initialized in the datadir? Does the client terminate as it should?

Jason-Zhangxin-Chen:
> Ok sounds good, and what if there is no `--genesis` flag supplied and no chain initialized in the datadir? Does the client terminate as it should?

If there is no --genesis flag set, client will start from the chain state of the data dir, init-chain will be be performed. 
If the data-dir contains proper chain state, client starts as usual as it does, otherwise it will terminated since there is no chain state.

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
 // initGenesis will initialise the given JSON format genesis file and writes it as
 // the zero'd block (i.e. genesis) or will fail hard if it can't succeed.
 func initGenesis(ctx *cli.Context) error {
-	// Make sure we have a valid genesis JSON
-	genesisPath := ctx.Args().First()
+	// If user does not specify a genesis file on genesis flag, start node from current data dir.
```
aneequesafdar, line 202:
```suggestion
	// If the user does not specify a genesis file, start node will use current `data-dir`.
```

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
+	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) {
+		log.Info("genesis flag is not set, start node directly from data dir.")
+		log.Info("In case of 1st time start, please start client with setting --genesis flag, " +
+			"the client will auto init genesis block before attach to network.")
+		return nil
+	}
```
aneequesafdar, line 208: I don't think there is a need for the second log statement as the info below will provide the same information. Also instead of saying `start node directly from data dir` in the first log statement is confusing because the user doesn't know which data dir is being reffered.

```suggestion
	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) {
		log.Info("--genesis flag is not set")
		return nil
	}
```

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
+
+	// If there is no genesis block presented, init a genesis block with genesis file,
+	// otherwise it checks if genesis block is compatible and matched with genesis file
+	// and return error if they are not, the error would terminate autonity client.
+
+	// Make sure we have a valid genesis JSON.
+	genesisPath := ctx.GlobalString(utils.InitGenesisFlag.Name)
+	log.Info("Trying to init genesis block with genesis file", "file", genesisPath)
```
aneequesafdar, line 216:
```suggestion
	log.Info("Trying to initialise genesis block with genesis file", "filepath", genesisPath)
```

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
 	if len(genesisPath) == 0 {
 		utils.Fatalf("Must supply path to genesis JSON file")
+		return fmt.Errorf("must supply path to genesis JSON file")
```
aneequesafdar, line 219: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
 	}
 	file, err := os.Open(genesisPath)
 	if err != nil {
 		utils.Fatalf("Failed to read genesis file: %v", err)
+		return err
```
aneequesafdar, line 224: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
 	}
 	defer file.Close()
 
 	genesis := new(core.Genesis)
 	if err := json.NewDecoder(file).Decode(genesis); err != nil {
 		utils.Fatalf("invalid genesis file: %v", err)
+		return err
```
aneequesafdar, line 231: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
 	}
 	// Make AutonityContract and Tendermint consensus mandatory for the time being.
 	if genesis.Config == nil {
 		utils.Fatalf("No Autonity Contract and Tendermint configs section in genesis")
+		return fmt.Errorf("no Autonity Contract and Tendermint configs section in genesis")
```
aneequesafdar, line 236: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
 	}
 	if genesis.Config.AutonityContractConfig == nil {
 		utils.Fatalf("No Autonity Contract config section in genesis")
+		return fmt.Errorf("no Autonity Contract config section in genesis")
```
aneequesafdar, line 240: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -215,30 +199,49 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
 	}
 	if genesis.Config.Tendermint == nil {
 		utils.Fatalf("No Tendermint config section in genesis")
+		return fmt.Errorf("no Tendermint config section in genesis")
```
aneequesafdar, line 244: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -256,10 +259,12 @@ func initGenesis(ctx *cli.Context) error {
 		chaindb, err := stack.OpenDatabase(name, 0, 0, "")
 		if err != nil {
 			utils.Fatalf("Failed to open database: %v", err)
+			return err
```
aneequesafdar, line 262: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/chaincmd.go
```diff
@@ -256,10 +259,12 @@ func initGenesis(ctx *cli.Context) error {
...
 		}
 		_, hash, err := core.SetupGenesisBlock(chaindb, genesis)
 		if err != nil {
 			utils.Fatalf("Failed to write genesis block: %v", err)
+			return err
```
aneequesafdar, line 267: This return statement is not required because `utils.Fatalf()` cause the program to terminate thus this return statement is never executed.

cmd/autonity/main.go
```diff
@@ -241,7 +241,12 @@ func init() {
 	app.Flags = append(app.Flags, metricsFlags...)
 
 	app.Before = func(ctx *cli.Context) error {
-		return debug.Setup(ctx)
+		err := debug.Setup(ctx)
+		if err != nil {
+			return err
+		}
+		// Try to init chain before autonity start up. Any error from initGenesis would terminate node start up.
+		return initGenesis(ctx)
```
aneequesafdar, line 249: We really should not be trying to call `initGenesis` here because the documentation of `app.Before` says the following:

```go
// An action to execute before any subcommands are run, but after the context is ready
// If a non-nil error is returned, no subcommands are run
Before BeforeFunc
```
This means that for all subcommands `initGenesis` will be executed which we don't want. For example, `autonity account new` returns:

```
./build/bin/autonity account new
INFO [10-07|16:09:32.900] genesis flag is not set, start node directly from data dir. 
INFO [10-07|16:09:32.900] In case of 1st time start, please start client with setting --genesis flag, the client will auto init genesis block before attach to network. 
.
.
.
```
Instead, we should be calling `initGenesis` inside `cmd/autonity/main.go:autointy()`.

We also need to take care so that we are not calling `makeFullNode()` twice, currently, it is called once in `initGenesis()` and the second time in `autointy()`. `makeFullNode()` sets the configuration of the autonity client so we don't want to do the same work twice.

I recommend to move the following lines (`cmd/autonity/chainmd.go:254-271`) :
```go
// Open an initialise both full and light databases
	stack := makeFullNode(ctx)
	defer stack.Close()

	for _, name := range []string{"chaindata", "lightchaindata"} {
		chaindb, err := stack.OpenDatabase(name, 0, 0, "")
		if err != nil {
			utils.Fatalf("Failed to open database: %v", err)
			return err
		}
		_, hash, err := core.SetupGenesisBlock(chaindb, genesis)
		if err != nil {
			utils.Fatalf("Failed to write genesis block: %v", err)
			return err
		}
		chaindb.Close()
		log.Info("Successfully wrote genesis state", "database", name, "hash", hash)
	}
```
after `cmd/autonity/main.go:319` :

Thus, the 2 functions will look something like the following:

```go
// initGenesis will initialise the given JSON format genesis file and writes it as
// the zero'd block (i.e. genesis) or will fail hard if it can't succeed.
func initGenesis(ctx *cli.Context) *core.Genesis {
	// If user does not specify a genesis file on genesis flag, start node from current data dir.
	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) {
		log.Info("--genesis flag is not set")
		return nil
	}

	// Make sure we have a valid genesis JSON.
	genesisPath := ctx.GlobalString(utils.InitGenesisFlag.Name)
	log.Info("Trying to init genesis block with genesis file", "file", genesisPath)
	if len(genesisPath) == 0 {
		utils.Fatalf("Must supply path to genesis JSON file")
	}
	file, err := os.Open(genesisPath)
	if err != nil {
		utils.Fatalf("Failed to read genesis file: %v", err)
	}
	defer file.Close()

	genesis := new(core.Genesis)
	if err := json.NewDecoder(file).Decode(genesis); err != nil {
	}
	// Make AutonityContract and Tendermint consensus mandatory for the time being.
	if genesis.Config == nil {
		utils.Fatalf("No Autonity Contract and Tendermint configs section in genesis")
	}
	if genesis.Config.AutonityContractConfig == nil {
		utils.Fatalf("No Autonity Contract config section in genesis")
	}
	if genesis.Config.Tendermint == nil {
		utils.Fatalf("No Tendermint config section in genesis")
	}

	if err := genesis.Config.AutonityContractConfig.Prepare(); err != nil {
		spew.Dump(genesis.Config.AutonityContractConfig)
		utils.Fatalf("autonity contract section is invalid. error:%v", err.Error())
	}

	setupDefaults(genesis)
	return genesis
}
```
```go
// autonity is the main entry point into the system if no special subcommand is ran.
// It creates a default node based on the command line arguments and runs it in
// blocking mode, waiting for it to be shut down.
func autonity(ctx *cli.Context) error {
	if args := ctx.Args(); len(args) > 0 {
		return fmt.Errorf("invalid command: %q", args[0])
	}
	prepare(ctx)
	node := makeFullNode(ctx)
	defer node.Close()

	genesis := initGenesis(ctx)
	if genesis != nil {
		// Do something when initGenesis returns nil
		for _, name := range []string{"chaindata", "lightchaindata"} {
			chaindb, err := node.OpenDatabase(name, 0, 0, "")
			if err != nil {
				utils.Fatalf("Failed to open database: %v", err)
				return err
			}
			_, hash, err := core.SetupGenesisBlock(chaindb, genesis)
			if err != nil {
				utils.Fatalf("Failed to write genesis block: %v", err)
				return err
			}
			chaindb.Close()
			log.Info("Successfully wrote genesis state", "database", name, "hash", hash)
		}
	}
	
	startNode(ctx, node)
	node.Wait()
	return nil
}
```

cmd/autonity/chaincmd.go
```diff
@@ -243,7 +244,7 @@ func initGenesis(ctx *cli.Context) error {
 
 	if err := genesis.Config.AutonityContractConfig.Prepare(); err != nil {
 		spew.Dump(genesis.Config.AutonityContractConfig)
-		return fmt.Errorf("autonity contract section is invalid. error:%v", err.Error())
+		utils.Fatalf("autonity contract section is invalid. error:%v", err.Error())
```
piersy, line 247: We shouldn't use `utils.Fatalf` because we lose all the context about where an error occurred.

We should just return errors since it:
A)  Allows us to easily reuse that code elsewhere.
B ) Allows us to add the context to the error as it bubbles up the call stack.

The cli `Actionn` function signature returns an error so we can do this easily.

cmd/autonity/main.go
Jason-Zhangxin-Chen: @aneequesafdar , greate findings, thank you!

cmd/autonity/chaincmd.go
aneequesafdar: I just changed it to remain consistent with the rest of the code before this change no one was using the error.

Jason-Zhangxin-Chen: > We shouldn't use `utils.Fatalf` because we lose all the context about where an error occurred.
> 
> We should just return errors since it:
> A) Allows us to easily reuse that code elsewhere.
> B ) Allows us to add the context to the error as it bubbles up the call stack.
> 
> The cli `Actionn` function signature returns an error so we can do this easily.

I removed utils.Fatalf from initGenesis by returning errors.

cmd/autonity/main.go
```diff
@@ -314,7 +314,11 @@ func autonity(ctx *cli.Context) error {
 	node := makeFullNode(ctx)
 	defer node.Close()
 
-	genesis := initGenesis(ctx)
+	genesis, err := initGenesis(ctx)
+	if err != nil {
+		utils.Fatalf("Failed to validate genesis file: %v", err)
```
aneequesafdar, line 319:
```suggestion
		return fmt.Errorf("Failed to validate genesis file: %v", err)
```

Jason-Zhangxin-Chen: I made some refine, it include this change now. Please view my latest commits.

cmd/autonity/genesis_test.go
```diff
@@ -82,14 +165,145 @@ func TestCustomGenesis(t *testing.T) {
...
+		waitForEndpoint(t, ipc, 5*time.Second)
+		autonity.Interrupt()
+
+		// Make sure genesis block is created.
+		autonity = runAutonity(t, "--nousb",
+			"--datadir", datadir, "--maxpeers", "0", "--port", "0",
+			"--nodiscover", "--nat", "none", "--ipcdisable",
+			"--exec", tt.query, "console")
+		autonity.ExpectRegexp(tt.result)
+		autonity.ExpectExit()
```
aneequesafdar, line 222: this block of code is repeated from `TestCustomGenesis`, therefore, it should be refactored

cmd/autonity/genesis_test.go
```diff
@@ -82,14 +165,145 @@ func TestCustomGenesis(t *testing.T) {
...
+			"--port", "0", "--maxpeers", "0", "--nodiscover", "--nat", "none",
+			"--etherbase", coinbase, "--ipcpath", ipc, "--datadir", datadir, "--genesis", json)
+
+		defer func() {
+			autonity.Interrupt()
+			autonity.ExpectExit()
+		}()
+
+		// Check RPC endpoint service is loaded and RPC service is a available now.
+		waitForEndpoint(t, ipc, 5*time.Second)
```
aneequesafdar, line 235: I am not sure whether just checking for endpoint has been started is sufficient. I think you should check the nonce again as on line 221

cmd/autonity/genesis_test.go
```diff
@@ -82,14 +165,145 @@ func TestCustomGenesis(t *testing.T) {
...
 }
+
+// Test that if there is chain data with genesis block presented, start node with same genesis file, the node should
+// perform a normal start up.
+func TestStartNodeWithSameGenesis(t *testing.T) {
+	for i, tt := range customGenesisTests {
+		coinbase := "0x8605cdbbdb6d264aa742e77020dcbc58fcdce182"
+		var ipc string
+		if runtime.GOOS == "windows" {
+			ipc = `\\.\pipe\autonity` + strconv.Itoa(trulyRandInt(100000, 999999))
```
aneequesafdar, line 192: I am not sure why this is required since before your change `TestCustomGenesis` does not mention windows at all

cmd/autonity/genesis_test.go
```diff
@@ -82,14 +165,145 @@ func TestCustomGenesis(t *testing.T) {
...
+			waitForEndpoint(t, ipc, 5*time.Second)
+			autonity.Interrupt()
+
+			// Make sure genesis block is created.
+			autonity = runAutonity(t, "--nousb",
+				"--datadir", datadir, "--maxpeers", "0", "--port", "0",
+				"--nodiscover", "--nat", "none", "--ipcdisable",
+				"--exec", tt.query, "console")
+			autonity.ExpectRegexp(tt.result)
+			autonity.ExpectExit()
```
aneequesafdar, line 276: Again, this code block is repeated. It should be refactored.

cmd/autonity/genesis_test.go
```diff
@@ -82,14 +165,145 @@ func TestCustomGenesis(t *testing.T) {
...
+				"--port", "0", "--maxpeers", "0", "--nodiscover", "--nat", "none",
+				"--etherbase", coinbase, "--ipcpath", ipc, "--datadir", datadir, "--genesis", jsonV2)
+
+			defer func() {
+				autonity.Interrupt()
+				autonity.ExpectExit()
+			}()
+
+			// With a incompatible genesis, client should do nothing with initGenesis by exit the start up.
+			autonity.ExpectExit()
```
aneequesafdar, line 293: The reference to autonity will change at line 296 by the time this defer function is executed and this piece of code would be redundant since on line 301 and 302 will cause the autonity to exit. The autonity reference should be renamed so that the correct autonity is closed.

```suggestion
			autonity1 = runAutonity(t,
				"--port", "0", "--maxpeers", "0", "--nodiscover", "--nat", "none",
				"--etherbase", coinbase, "--ipcpath", ipc, "--datadir", datadir, "--genesis", jsonV2)

			defer func() {
				autonity1.Interrupt()
				autonity1.ExpectExit()
			}()

			// With a incompatible genesis, client should do nothing with initGenesis by exit the start up.
			autonity1.ExpectExit()
```

contracts/autonity/contract/test/autonity/autonity-start.sh
```diff
@@ -13,12 +13,13 @@ RPC_ADDR=127.0.0.1
 RPC_API="tendermint,console,eth,web3,admin,debug,miner,personal,txpool,net"
 
 # init the data directory
-echo "Autonity INIT $RPC_ADDR"
-$AUTONITY init --datadir $DATADIR genesis-tendermint.json
+# echo "Autonity INIT $RPC_ADDR"
+# $AUTONITY init --datadir $DATADIR genesis-tendermint.json
```
aneequesafdar, line 17: These should be removed instead of commented

aneequesafdar:
You can make use of `t.Run` and only have one function to test the scenarios instead of 3 different test functions

cmd/autonity/genesis_test.go
Jason-Zhangxin-Chen: > this block of code is repeated from `TestCustomGenesis`, therefore, it should be refactored

Resolved at my latest commit: Make testcase of initGenesis to be lean.

cmd/autonity/genesis_test.go
Jason-Zhangxin-Chen: > I am not sure whether just checking for endpoint has been started is sufficient. I think you should check the nonce again as on line 221

Resolved at my latest commit: Make testcase of initGenesis to be lean.

cmd/autonity/genesis_test.go
Jason-Zhangxin-Chen: > I am not sure why this is required since before your change `TestCustomGenesis` does not mention windows at all

It is a part of the raw testcase of upstream, anyway I removed it since all of us develop it under linux. Resolved at my latest commit: Make testcase of initGenesis to be lean.

cmd/autonity/genesis_test.go
Jason-Zhangxin-Chen: > Again, this code block is repeated. It should be refactored.

Resolved at my latest commit: Make testcase of initGenesis to be lean.

cmd/autonity/genesis_test.go
Jason-Zhangxin-Chen: > The reference to autonity will change at line 296 by the time this defer function is executed and this piece of code would be redundant since on line 301 and 302 will cause the autonity to exit. The autonity reference should be renamed so that the correct autonity is closed.

Thank you, Anneeque. It's resolved at my latest commit: Make testcase of initGenesis to be lean.

Jason-Zhangxin-Chen:
> You can make use of `t.Run` and only have one function to test the scenarios instead of 3 different test functions

Resolved at my latest commit: Make testcase of initGenesis to be lean.

contracts/autonity/contract/test/autonity/autonity-start.sh
Jason-Zhangxin-Chen: > These should be removed instead of commented

Resolved at commit:  Remove init step in hello-world script and contract test script.

cmd/autonity/chaincmd.go
```diff
@@ -212,59 +196,48 @@ Use "ethereum dump 0" to dump the genesis block.`,
...
-func initGenesis(ctx *cli.Context) error {
-	// Make sure we have a valid genesis JSON
-	genesisPath := ctx.Args().First()
+// validateGenesis will validate and initialise the given JSON format genesis file
+func validateGenesis(ctx *cli.Context) (*core.Genesis, error) {
+	// If the user does not specify a genesis file, start node will use current `data-dir`.
+	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) {
+		log.Info("--genesis flag is not set")
+		return nil, nil
+	}
```
piersy, line 205: This is an example of what I would call a neutralized function, see [here](https://github.com/piersy/programming-principles/blob/main/neutralized-function.md) for an explanation.

Jason-Zhangxin-Chen: > This is an example of what I would call a neutralized function, see [here](https://github.com/piersy/programming-principles/blob/main/neutralized-function.md) for an explanation.

I resovled it in a new way, please find it at my latest commit.

cmd/autonity/main.go
```diff
@@ -302,6 +306,85 @@ func prepare(ctx *cli.Context) {
 	go metrics.CollectProcessMetrics(3 * time.Second)
 }
 
+func setupDefaults(genesis *core.Genesis) {
```
piersy, line 309: This is another instance of [neutralized function](https://github.com/piersy/programming-principles/blob/main/neutralized-function.md)

cmd/autonity/main.go
```diff
@@ -302,6 +306,76 @@ func prepare(ctx *cli.Context) {
...
+	if genesis.Config.Tendermint == nil {
+		return nil, fmt.Errorf("no Tendermint config section in genesis")
+	}
+
+	if err := genesis.Config.AutonityContractConfig.Prepare(); err != nil {
+		spew.Dump(genesis.Config.AutonityContractConfig)
+		return nil, err
+	}
+
+	if genesis.Config.Tendermint.BlockPeriod == 0 {
```
piersy, line 343: Hi @Jason-Zhangxin-Chen, I think this is a case of [defensive programming](https://github.com/piersy/programming-principles/blob/main/defensive-programming.md)

cmd/autonity/main.go
```diff
@@ -302,6 +306,76 @@ func prepare(ctx *cli.Context) {
...
+		defaultPeriod := config.DefaultConfig().BlockPeriod
+		log.Info("Invalid tendermint block period 0, set to default", "validate genesis", "block period", defaultPeriod)
+		genesis.Config.Tendermint.BlockPeriod = defaultPeriod
+	}
```
piersy, line 347: Hi @Jason-Zhangxin-Chen, I think this is a case of [defensive programming](https://github.com/piersy/programming-principles/blob/main/defensive-programming.md)

cmd/autonity/main.go
```diff
@@ -302,6 +306,76 @@ func prepare(ctx *cli.Context) {
 	go metrics.CollectProcessMetrics(3 * time.Second)
 }
 
+// validateGenesis will validate and initialise the given JSON format genesis file
+func validateGenesis(ctx *cli.Context) (*core.Genesis, error) {
```
piersy, line 310: validateGenesis is not quite the right name for this function. Although the function is performing validation, it is also loading the genesis from disk. If the function was purely performing validation I would expect it to have the following signature `func validateGenesis(g *core.Genesis) error`.

I think just renaming this function to `loadGenesis` is fine, I think it is reasonable to consider validation part of the loading process.

Jason-Zhangxin-Chen: Renamed with loadGenesis

cmd/autonity/main.go
```diff
@@ -302,6 +306,76 @@ func prepare(ctx *cli.Context) {
...
+	}
+
+	// check if genesis is matched with block #0, and check to apply compatible chain configuration.
+	if genesis != nil {
+		for _, name := range []string{"chaindata", "lightchaindata"} {
+			chaindb, err := node.OpenDatabase(name, 0, 0, "")
+			if err != nil {
+				return fmt.Errorf("failed to open database: %v", err)
+			}
+			_, hash, err := core.SetupGenesisBlock(chaindb, genesis)
```
piersy, line 362: If we run autonity without a genesis or db, it sets itself up as an ethash node. This feels incorrect to me.

cmd/utils/flags.go
```diff
@@ -134,6 +134,15 @@ func printHelp(out io.Writer, templ string, data interface{}) {
 
 var (
 	// General settings
+	InitGenesisFlag = cli.StringFlag{
+		Name: "genesis",
+		Usage: "Path to the genesis json file. If the flag is not set, node will start by loading chain-data from \"data-dir\". " +
+			"Otherwise node will start by validating the genesis file, checks if genesis file is match with genesis block, and " +
+			"checks if chain configuration of the genesis file is compatible with current chain-data, apply new compatible chain " +
+			"configuration into chain-data and start the node. Otherwise node start will end up with a mis-match genesis error or an " +
+			"incompatible chain configuration error.",
```
piersy, line 143:
```suggestion
		Usage: `Path to the genesis json file, the genesis file contains
	configuration that defines the first (aka genesis) block of the chain
	and also controls how the client behaves. This flag must be provided
	the first time a node is run so that the node can initialise its
	database with the provided config. Subsequent to the first run, if the
	flag is not set, the node will load all configuration from its
	database, otherwise the node will verify that the config for the
	genesis block exactly matches the genesis block it has in its database
	and then see if the configuration that controls client behaviour is
	compatible with the configuration in the datbase. If so the client will
	update the configuration that controls client behaviour. This is the
	mechanism is used for handling forks.`,
```

I've expanded on the comment here since there is quite a lot going on with this flag. I've also used backtics to enclose the string which allows for easier formatting of the text. I also formatted it into a block as opposed to a really long line because it makes it much easier to read when it is displayed to the user.

piersy: Note the indentation here was intentonal, it results in the displayed text being left aligned.

cmd/autonity/main.go
```diff
@@ -302,16 +305,104 @@ func prepare(ctx *cli.Context) {
...
 func autonity(ctx *cli.Context) error {
 	if args := ctx.Args(); len(args) > 0 {
 		return fmt.Errorf("invalid command: %q", args[0])
 	}
+
 	prepare(ctx)
+
+	// If the user does not specify a genesis and a db, stop the start up.
+	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) && !ctx.GlobalIsSet(utils.DataDirFlag.Name) {
+		return fmt.Errorf("--%s and --%s are not set, autonity client is going to stop", utils.InitGenesisFlag.Name, utils.DataDirFlag.Name)
```
piersy, line 388: If just the datadir flag is set you can still start the node in ethash mode

cmd/autonity/main.go
```diff
@@ -302,16 +305,104 @@ func prepare(ctx *cli.Context) {
...
 }
 
+// loadGenesis will validate and initialise the given JSON format genesis file
+func loadGenesis(ctx *cli.Context) (*core.Genesis, error) {
+	// Make sure we have a valid genesis JSON.
+	genesisPath := ctx.GlobalString(utils.InitGenesisFlag.Name)
+	log.Info("Trying to initialise genesis block with genesis file", "filepath", genesisPath)
+	if len(genesisPath) == 0 {
+		return nil, fmt.Errorf("must supply path to genesis JSON file")
+	}
```
piersy, line 315: This is an example of a [neutralized function](https://github.com/piersy/programming-principles/blob/main/neutralized-function.md), it would make a lot more sense if we provided the filename to the function and only call the function if a filename has been provided.

cmd/autonity/main.go
```diff
@@ -302,16 +305,104 @@ func prepare(ctx *cli.Context) {
...
+}
+
+func initGenesis(ctx *cli.Context, node *node.Node) error {
+	genesis, err := loadGenesis(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to validate genesis file: %v", err)
+	}
+
+	// check if genesis is matched with block #0, and check to apply compatible chain configuration.
+	if genesis != nil {
```
piersy, line 356: In cases such as this its neater to [return early](https://medium.com/swlh/return-early-pattern-3d18a41bba8#:~:text=Return%20early%20is%20the%20way,when%20conditions%20are%20not%20met.) and reduce the indentation by one level.

So instead of doing all your logic inside the if statement, turn the statement around and exit early if the genesis is nil.

```go
if genesis == nil {
    return nil
}
```

This means the reader knows immediately what happens when the genesis is nil, as opposed to the current situation where the reader needs to scroll to the bottom of the if block only to see that we return nil.

piersy: This is something I have resovled in #710

cmd/autonity/main.go
```diff
@@ -302,16 +305,104 @@ func prepare(ctx *cli.Context) {
...
 // blocking mode, waiting for it to be shut down.
 func autonity(ctx *cli.Context) error {
 	if args := ctx.Args(); len(args) > 0 {
 		return fmt.Errorf("invalid command: %q", args[0])
 	}
+
 	prepare(ctx)
+
+	// If the user does not specify a genesis and a db, stop the start up.
+	if !ctx.GlobalIsSet(utils.InitGenesisFlag.Name) && !ctx.GlobalIsSet(utils.DataDirFlag.Name) {
```
piersy, line 387: This check doesn't make sense with the way that the `data-dir` flag is defined. Because the `data-dir` flag has a default value it is acceptable to start a node with neither of these flags after first initialisation.

We want to fail when:
* A node is started without an existing database and no genesis flag is provided.
* A node is started without an existing database and the provided genesis file is invalid.
* A node is started with an existing database and the provided genesis file is not compatible with the database contents.

cmd/autonity/main.go
Jason-Zhangxin-Chen: > If just the datadir flag is set you can still start the node in ethash mode

Agree with that, I fix it in my last commit.

cmd/autonity/main.go
Jason-Zhangxin-Chen: > This check doesn't make sense with the way that the `data-dir` flag is defined. Because the `data-dir` flag has a default value it is acceptable to start a node with neither of these flags after first initialisation.
> 
> We want to fail when:
> 
> * A node is started without an existing database and no genesis flag is provided.
> * A node is started without an existing database and the provided genesis file is invalid.
> * A node is started with an existing database and the provided genesis file is not compatible with the database contents.
> 
> Whether or not the `data-dir` flag is provided is not relevant.

Agree with that, I fix it within my last commit: Not start client with ethash mode when there is no genesis block and genesis file find.

cmd/autonity/main.go
Jason-Zhangxin-Chen: > This is something I have resovled in #710

Thank you, Piers.

cmd/autonity/main.go
Jason-Zhangxin-Chen: > If we run autonity without a genesis or db, it sets itself up as an ethash node. This feels incorrect to me.
> 
> I think we should expand this PR to cover making sure that we can't start up autonity in ethash mode. The original problem that precipitated this PR was that nodes were not able to start up properly on the new testnet. This was because the devops team had forgotten to update the version of autonity used for initing the chain. That was one way of failing to run autonity properly. But we still have the case where anyone who runs autonity without a db and a genesis will then not be able to correctly start autonity even if they subsequently provide a genesis. So I think we should just make this PR about making sure that you can't set up autonity in an inconsistent state.

Resolved it in my latest commits.

core/genesis.go
```diff
@@ -166,8 +173,8 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
 	stored := rawdb.ReadCanonicalHash(db, 0)
```
piersy, line 173: There's a check above this point here lines 162-164 I think it should be removed, it seems like it must now be redundant if we cannot start in ethash mode.

```go
	if genesis != nil && genesis.Config == nil {
		return params.AllEthashProtocolChanges, common.Hash{}, errGenesisNoConfig
	}
```

core/genesis_test.go
```diff
@@ -77,8 +77,9 @@ func TestSetupGenesis(t *testing.T) {
 			fn: func(db ethdb.Database) (*params.ChainConfig, common.Hash, error) {
 				return SetupGenesisBlock(db, nil)
 			},
-			wantHash:   params.MainnetGenesisHash,
-			wantConfig: params.MainnetChainConfig,
+			wantErr:    &NoGenesisError{},
```
piersy, line 80: This is an example of [overfitting](https://github.com/piersy/programming-principles/blob/main/overfitting-tests.md). I can see that there isn't really another choice if you want to add to this test. I think we could just extract this test case into a more simple test that checks that an error is returned when the genesis is nil and there is no block in the db.

Then we could remove `NoGenesisError` since it is only being used by tests.

core/genesis.go
```diff
@@ -145,12 +145,19 @@ func (e *GenesisMismatchError) Error() string {
...
+	return fmt.Sprintf("DB has no genesis block and there is no genesis file set by user")
+}
+
 // SetupGenesisBlock writes or updates the genesis block in db.
 // The block that will be used is:
 //
 //                          genesis == nil       genesis != nil
 //                       +------------------------------------------
-//     db has no genesis |  main-net default  |  genesis
+//     db has no genesis |  NoGenesisError    |  genesis
```
piersy, line 160: This comment doesn't make sense any more.

cmd/autonity/main.go
```diff
@@ -302,16 +306,95 @@ func prepare(ctx *cli.Context) {
...
 	defer node.Close()
+
+	// If genesis flag is not set, node will load chain-data from data-dir. If the flat is set, node will load the
+	// genesis file, check if genesis file is match with genesis block, and check if chain configuration of the genesis
+	// file is compatible with current chain-data, apply new compatible chain configuration into chain db.
+	// Otherwise client will end up with a mis-match genesis error or an incompatible chain configuration error.
+
+	genesisPath := ctx.GlobalString(utils.InitGenesisFlag.Name)
+	if genesisPath != "" {
+		log.Info("--genesis flag is set")
```
piersy, line 386: I think I reintroduced this hardcoded flag in my PR

core/genesis_test.go
Jason-Zhangxin-Chen: > This is an example of [overfitting](https://github.com/piersy/programming-principles/blob/main/overfitting-tests.md). I can see that there isn't really another choice if you want to add to this test. But I think we could just extract this test case into a separate more simple test that checks that an error is returned when the genesis is nil and there is no block in the db.
> 
> Then we could remove `NoGenesisError` since it is only being used by this test.

I thought to align with the upstream's code which defines errors (GenesisMismatchError, GenesisIncompatibleError, etc...), and in the test case side, it also have such framework to expect different type of error to be returned.

core/genesis.go
Jason-Zhangxin-Chen: > This comment doesn't make sense any more.

Yeah, I already update the comments too.

cmd/autonity/main.go
Jason-Zhangxin-Chen: > I think I reintroduced this hardcoded flag in my PR

I removed it in one of the commits I made today.

core/genesis.go
Jason-Zhangxin-Chen: > There's a check above this point here lines 162-164 I think it should be removed, it seems like it must now be redundant if we cannot start in ethash mode.
> 
> ```go
> 	if genesis != nil && genesis.Config == nil {
> 		return params.AllEthashProtocolChanges, common.Hash{}, errGenesisNoConfig
> 	}
> ```

Why does this check is redundant?

piersy: Because we already checked this when we loaded the genesis file. It's not possible to load a genesis file and have a nil genesis.Config.

https://github.com/clearmatics/autonity/blob/8caaa7bc4072d4c3bbe3243ab71a29d515c296e1/cmd/autonity/main.go#L321-L324

core/genesis_test.go
piersy: The upstream code in this case is bad. So we shouldn't align with it unless we have to.

In this case the upstream code is causing you to have to define a new error type just so that you can use the test framework from the upstream code.

We don't actually need that error outside of the test and we don't even need that error to write a test to check that when "no block in DB, genesis == nil" `SetupGenesis` fails. We only need that error in order to be able to use the test framework from upstream.

Errors are an important part of the public API of a package and we definitely should not be adding errors to the public API to satisfy badly designed test frameworks.

We often change code in response to tests, sometimes that change is good for the code as a whole and so It is acceptable to do so, sometime however we end up bending our production code around our tests in a way that degrades the quality of the production code. This is one of those cases, because the error is not used by any production code and serves no purpose outside of the test.

core/genesis.go
piersy: I don't think you changed this comment in your updates

core/genesis.go
Jason-Zhangxin-Chen: > Because we already checked this when we loaded the genesis file. It's not possible to load a genesis file and have a nil genesis.Config.
> 
> https://github.com/clearmatics/autonity/blob/8caaa7bc4072d4c3bbe3243ab71a29d515c296e1/cmd/autonity/main.go#L321-L324

Removed this checker.

core/genesis_test.go
Jason-Zhangxin-Chen: > The upstream code in this case is bad. So we shouldn't align with it unless we have to.
> 
> In this case the upstream code is causing you to have to define a new error type just so that you can use the test framework from the upstream code.
> 
> We don't actually need that error outside of the test and we don't even need that error to write a test to check that when "no block in DB, genesis == nil" `SetupGenesis` fails. We only need that error in order to be able to use the test framework from upstream.
> 
> Errors are an important part of the public API of a package and we definitely should not be adding errors to the public API to satisfy badly designed test frameworks.
> 
> We often change code in response to tests, sometimes that change is good for the code as a whole and so It is acceptable to do so, sometimes however we end up bending our production code around our tests in a way that degrades the quality of the production code. This is one of those cases, because the error is not used by any production code and serves no purpose outside of the test.

I removed the NoGenesisError.

core/genesis.go
Jason-Zhangxin-Chen: What do you mean it doesn't make sense any more? I replace the "main-net default" into "return error". It still clarify the result of this SetupGenesisBlock().

cmd/autonity/accountcmd_test.go
```diff
@@ -32,14 +33,20 @@ import (
...
 	source := filepath.Join("..", "..", "accounts", "keystore", "testdata", "keystore")
 	if err := cp.CopyAll(keystore, source); err != nil {
 		t.Fatal(err)
 	}
-	return datadir
+
+	genesisFile := filepath.Join(datadir, "genesis.json")
+	if err := ioutil.WriteFile(genesisFile, []byte(genesis), 0600); err != nil {
+		t.Fatalf("failed to write genesis file: %v", err)
+	}
```
piersy, line 47: This duplicates lines 77-80 of `consolecmd_test.go`

cmd/autonity/consolecmd_test.go
```diff
@@ -31,23 +33,83 @@ import (
...
+		t.Fatalf("failed to write genesis file: %v", err)
+	}
+	return dir, genesisFile
+}
+
+func getCoinBase(t *testing.T, datadir string) string {
+	keyfile := filepath.Join(datadir, "autonity", "nodekey")
+	key, err := crypto.LoadECDSA(keyfile)
+	if err != nil {
+		t.Fatalf("cannot get node key")
```
piersy, line 88: You should include the error in the string passed to `Fatalf` so when the test fails we can have a better idea of why.

Jason-Zhangxin-Chen: @piersy ,  it is resolved in my latest commit.

cmd/autonity/accountcmd_test.go
Jason-Zhangxin-Chen: Also resolved in my latest commit.

PR #698, opened 2020-10-08
Simplification of core

piersy:
This is a draft PR, for the benefit of @aneequesafdar, its a long way from being ready so I wouldn't advise reviewing it at present.


common/keygenerator/generator_test.go
```diff
@@ -10,3 +14,21 @@ func TestGenerator_Next(t *testing.T) {
...
+}
+
+func BenchmarkEcdsaGenerate(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		_, err := crypto.GenerateKey()
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}
```
aneequesafdar, line 34: I feel as though this is experimental code so either it should be removed or moved to a different branch

consensus/tendermint/algorithm/tendermint.go
```diff
@@ -0,0 +1,340 @@
...
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+type ValueID [32]byte
+
+func (v ValueID) String() string {
+	return hex.EncodeToString(v[:3])
+}
```
aneequesafdar, line 12: Why do you want to only print the first three bytes of ValueID?

consensus/tendermint/algorithm/tendermint.go
```diff
@@ -0,0 +1,340 @@
...
+type ConsensusMessage struct {
+	MsgType    Step
+	Height     uint64
+	Round      int64
+	Value      ValueID
+	ValidRound int64
+}
+
+func (cm *ConsensusMessage) String() string {
+	return fmt.Sprintf("s:%-3s h:%-3d r:%-3d v:%-6s", cm.MsgType.ShortString(), cm.Height, cm.Round, cm.Value.String())
```
aneequesafdar, line 81: Just to confirm `%-3` and `%-6" is padding 3 and 6 spaces respectively to the right?

consensus/tendermint/algorithm/tendermint.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+// Oracle is used to answer questions the algorithm may have about its
+// state, such as 'Am I the proposer' or 'Have i reached prevote quorum
+// threshold for value with id v?'
+type Oracle interface {
+	Valid(ValueID) bool
+	MatchingProposal(*ConsensusMessage) *ConsensusMessage
+	PrevoteQThresh(round int64, value *ValueID) bool
+	PrecommitQThresh(round int64, value *ValueID) bool
```
aneequesafdar, line 91: Do we need to have separate functions for Prevote and Precommit Quroum threshold, when we know both thresholds will always be the same?

consensus/tendermint/algorithm/tendermint.go
```diff
@@ -0,0 +1,340 @@
...
+	// FThresh indicates whether we have messages whose voting power exceeds
+	// the failure threshold for the given round.
+	FThresh(round int64) bool
+	Proposer(round int64, nodeID NodeID) bool
```
aneequesafdar, line 95: I feel as though height also needs to be provided to the implementation of oracle for it to correctly determine the proposer with nodeID is the proposer for the current height and round

common/keygenerator/generator_test.go
piersy: Yep

consensus/tendermint/algorithm/tendermint.go
piersy: For readability of logs. 3 bytes gives enough entropy to be very sure that we won't have collisions for the kind of network sizes that we will be testing and debugging.

aneequesafdar:
closing in favour of #759

Issue #84, opened 2019-04-26
Persist Tendermint state


IronGauntlets:
If we don't store tendermint state then nodes can lose important state (such as their locked value) when stopped and started. This can cause the network to get stuck.

JekaMas:
Also we need WAL+reply to be able to reproduce any incorrect state after cluster failures.

piersy:
Just a note, we should probably just store the tendermint internal state, that's all we should need to avoid nodes sending inconsistent messages.

piersy:
Just to clarify this is a scenario we see happening in the `TestTendermintStartStop` tests, in these tests stopping and starting causes the entire tendermint engine to be reconstructed, meaning it loses all state, which sometimes causes the tests to hang.

Consider this scenario, in a nework of size 4, all nodes prevote for a
value, then 2 nodes lock a value and two restart and lose their state
(that they were in the prevote step). When they come back they will
propose new values and the locked nodes will propose the locked value,
niether value can ever be committed because each can garner at most 2
votes.

lorenzo-dev1:
superseded by https://github.com/autonity/autonity/issues/969

PR #700, opened 2020-10-09
Extend the measurement window of TC-19.

### LINKED ISSUE #697: Docker e2e tests for TC-19 are not stable
#Problem

The tests failed in this run linked below

https://github.com/clearmatics/autonity/runs/1215675810?check_suite_focus=true

I'm unable to upload the logs because the file is larger than 10mb.

Jason-Zhangxin-Chen:
To close #697 
In TC-19, bootstrap network with 1 node restart during the bootstrap phase.
the test scenario is to start node one by one, node_0 is the proposer which started as the first node of the network, after it send the proposal for a while, node_0 was shut down on purpose, while not all the nodes in the network are started yet to get the 1st proposal, so the proposal was not distributed to all members in the network, then node_0 started again, since it is the proposer of height_1, round_0, so it send a new proposal to the network,  with different proposal on the same round, members did not reach the consensus until pre-commit time out happens. then voting moved to the round_1. During the time, the measurement time window exceeded, the test engine cannot observe that the consensus engine is generating blocks until the time window expired, so it failed the testcase as the engine is on-holding. So I extend the measurement window for this test case.


aneequesafdar (approved):

piersy:
@Jason-Zhangxin-Chen 

Could you clarify the meaning of

>During the time, the measurement time window exceeded, got engine does not provide blocks yet.

I understand that the measurement time window was exceeded, I don't understand what you are saying about the engine.

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen
> 
> Could you clarify the meaning of
> 
> > During the time, the measurement time window exceeded, got engine does not provide blocks yet.
> 
> I understand that the measurement time window was exceeded, I don't understand what you are saying about the engine.

The statement "got engine does not provide blocks yet." is replaced with "the test engine cannot observe that the consensus engine is generating blocks until the time window expired, so it failed the testcase as the engine is on-holding. "

piersy:
@Jason-Zhangxin-Chen What exactly is this test case waiting to see in 300 seconds?

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen What exactly is this test case waiting to see in 300 seconds?

After the simulations(node crash, delay, restart, etc...) is done, the test engine will check consensus engine's liveness and also the data consistency between nodes, for example tendermint is keep producing blocks, and the block hash on the latest hight is same across the network.

piersy:
@Jason-Zhangxin-Chen 

> > @Jason-Zhangxin-Chen What exactly is this test case waiting to see in 300 seconds?
> 
> After the simulations(node crash, delay, restart, etc...) is done, the test engine will check consensus engine's liveness and also the data consistency between nodes, for example tendermint is keep producing blocks, and the block hash on the latest hight is same across the network.

300 seconds seems excessive, why does it take so long?

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen
> 
> > > @Jason-Zhangxin-Chen What exactly is this test case waiting to see in 300 seconds?
> > 
> > 
> > After the simulations(node crash, delay, restart, etc...) is done, the test engine will check consensus engine's liveness and also the data consistency between nodes, for example tendermint is keep producing blocks, and the block hash on the latest hight is same across the network.
> 
> 300 seconds seems excessive, why does it take so long?

The time costs, one is from the simulation, the others are about the causality of the test context, for example the scenario I explained in below happens, it would cost more time to let the consensus engine to be recovered.
`
the test scenario is to start node one by one, node_0 is the proposer which started as the first node of the network, after it send the proposal for a while, node_0 was shut down on purpose, while not all the nodes in the network are started yet to get the 1st proposal, so the proposal was not distributed to all members in the network, then node_0 started again, since it is the proposer of height_1, round_0, so it send a new proposal to the network, with different proposal on the same round, members did not reach the consensus until pre-commit time out happens. then voting moved to the round_1. During the time, the measurement time window exceeded, the test engine cannot observe that the consensus engine is generating blocks until the time window expired, so it failed the testcase as the engine is on-holding.
`

piersy:
Hi @Jason-Zhangxin-Chen 

> `the test scenario is to start node one by one, node_0 is the proposer which started as the first node of the network, after it send the proposal for a while, node_0 was shut down on purpose, while not all the nodes in the network are started yet to get the 1st proposal, so the proposal was not distributed to all members in the network, then node_0 started again, since it is the proposer of height_1, round_0, so it send a new proposal to the network, with different proposal on the same round, members did not reach the consensus until pre-commit time out happens. then voting moved to the round_1. During the time, the measurement time window exceeded, the test engine cannot observe that the consensus engine is generating blocks until the time window expired, so it failed the testcase as the engine is on-holding.`

What you are describing here is incorrect behaviour, nodes should not send conflicting messages for the same height round and step. Extending the window is not a solution because we can have scenarios where  the network gets stuck forever. See #84.

So I don't think we should make this change because I think we are seeing a valid failure here, and increasing the timeout will mean that we have to wait that much longer for feedback.

Issue #693, opened 2020-09-30
Update Autonity with geth 1.9.22


yazzaoui:
##  Acceptance criteria

Update Autonity with 1.9.22 Geth https://github.com/ethereum/go-ethereum/releases/tag/v1.9.22

PR #703, opened 2020-10-19
Merge go-ethereum 1.9.23

### LINKED ISSUE #694: Unavailable pending state for non-mining nodes
# Problem

We have disabled the worker's initialization when creating the miner object in order to fix some data races. https://github.com/clearmatics/autonity/commit/9005eab249ed8b3fe72fdbe6ce31148721a42d51#diff-7fc2e947b54d13e96c3477dc4e6e6c06R77-R78

This fix made fail RPC calls requesting the internal `pending` state/block if the node is not mining.

Affected APIs:
- personal_sendTransaction with unset gas value.
- eth_SignTransaction with unset gas value.
- eth_FillTransaction with unset gas value.
- eth_Resend with unset gas value.
- every API that can accept a block number or hash parameter if it set as "pending", e.g. `eth_getBlockByNumber("pending")` 

# Implementation

- Update autonity with upstream geth  (#693) 
- Replace our data race fixes with the upstream logic.
- Enable worker initialization when creating the miner object.
 
# Acceptance Criteria

- Pending state/block rpc requests are working.
- Worker free from race conditions.

yazzaoui:
closes #694 #693


cmd/autonity/chaincmd.go
```diff
@@ -284,10 +284,7 @@ func setupDefaults(genesis *core.Genesis) {
 }
 
 func dumpGenesis(ctx *cli.Context) error {
-	genesis := utils.MakeGenesis(ctx)
-	if genesis == nil {
-		genesis = core.DefaultGenesisBlock()
-	}
```
aneequesafdar: I am not sure why do you want to always dump the default genesis. Is there a specific reason?

core/rawdb/database.go
```diff
@@ -345,7 +345,7 @@ func InspectDatabase(db ethdb.Database) error {
 		case bytes.HasPrefix(key, preimagePrefix) && len(key) == (len(preimagePrefix)+common.HashLength):
 			preimages.Add(size)
 		case bytes.HasPrefix(key, bloomBitsPrefix) && len(key) == (len(bloomBitsPrefix)+10+common.HashLength):
-			bloomBitsSize += size
```
aneequesafdar, line 291: Could you give some context for this change?

metrics/registry.go
```diff
@@ -190,20 +189,6 @@ func (r *StandardRegistry) UnregisterAll() {
...
-// stopMeterTicker Stops Meter Ticker.
-// It's needed to avoid go-routine leaks.
-func (r *StandardRegistry) stopMeterTicker() {
-	arbiter.Lock()
-	started := arbiter.started
-	arbiter.Unlock()
-	if started && atomic.LoadUint32(arbiter.stopped) != 1 {
-		// stop the ticker to avoid go-routine leak
-		arbiter.stopTicker()
-	}
```
aneequesafdar: similar to the previous comment, can you please give some context to why we want to remove this function?

contracts/autonity/autonity.go
```diff
@@ -95,7 +95,7 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
 	v := EconomicMetaData{make([]common.Address, 32), make([]uint8, 32), make([]*big.Int, 32),
 		make([]*big.Int, 32), new(big.Int), new(big.Int)}
 
-	if err := ABI.Unpack(&v, "dumpEconomicsMetricData", ret); err != nil {
+	if err := ABI.UnpackIntoInterface(&v, "dumpEconomicsMetricData", ret); err != nil {
```
aneequesafdar, line 98: the type of `v` was not changed so why are you using `UnpackIntoInterface`, based on the comment it is meant to be called when we are not passing a type which conforms to the result, which is not the case here

ethclient/ethclient_test.go
```diff
@@ -222,7 +222,7 @@ func generateTestChain() (*core.Genesis, []*types.Block) {
 		g.OffsetTime(5)
 		g.SetExtra([]byte("test"))
 	}
-	gblock := genesis.ToBlock(db)
+	gblock, _ := genesis.ToBlock(db)
```
aneequesafdar, line 225: we shouldn't ignore errors

piersy (changes requested):
This is not compiling, so that needs to be resolved before we can even consider this PR

p2p/server.go
```diff
@@ -1177,3 +1174,35 @@ func (srv *Server) PeersInfo() []*PeerInfo {
...
+
+func (c *testTransport) doProtoHandshake(our *protoHandshake) (*protoHandshake, error) {
+	pubkey := crypto.FromECDSAPub(c.rpub)[1:]
+	return &protoHandshake{ID: pubkey, Name: "test"}, nil
+}
+
+func (c *testTransport) close(err error) {
+	c.conn.Close()
+	c.closeErr = err
+}
```
piersy, line 1208: Can we move this back?

consensus/consensus.go
```diff
@@ -112,7 +112,7 @@ type Engine interface {
 	//
 	// Note, the method returns immediately and will send the result async. More
 	// than one result may also be returned depending on the consensus algorithm.
-	Seal(chain ChainHeaderReader, block *types.Block, results chan<- *types.Block, stop <-chan struct{}) error
+	Seal(chain ChainReader, block *types.Block, results chan<- *types.Block, stop <-chan struct{}) error
```
piersy: This seems like a dangerous change. If the go-ethereum team start passing objects here that are only ChainHeaderReaders we could get stuck.

cmd/autonity/chaincmd.go
piersy: This definitely looks wrong.

consensus/test/test_helpers_test.go
```diff
@@ -161,11 +160,11 @@ func makeGenesis(nodes map[string]*testNode, stakeholderName string) *core.Genes
 	return genesis
 }
 
-func makePeer(genesis *core.Genesis, nodekey *ecdsa.PrivateKey, listenAddr string, rpcPort int, inRate, outRate int64, cons func(basic consensus.Engine) consensus.Engine) (*node.Node, error) { //здесь эта переменная-функция называется cons
+func makeNodeConfig(genesis *core.Genesis, nodekey *ecdsa.PrivateKey, listenAddr string, rpcPort int, inRate, outRate int64) (*node.Config, *eth.Config) {
 	// Define the basic configurations for the Ethereum node
 	datadir, err := ioutil.TempDir("", "")
 	if err != nil {
-		return nil, err
+		panic(err)
```
piersy, line 167: Instead of panicking use `require.NoError`

consensus/test/testcase_test.go
```diff
@@ -141,7 +141,7 @@ func runTest(t *testing.T, test *testCase) {
 	// see: metrics/meter.go:55
 	defer metrics.DefaultRegistry.UnregisterAll()
 
-	log.Root().SetHandler(log.LvlFilterHandler(log.LvlError, log.StreamHandler(os.Stderr, log.TerminalFormat(true))))
+	log.Root().SetHandler(log.LvlFilterHandler(log.LvlDebug, log.StreamHandler(os.Stderr, log.TerminalFormat(true))))
```
piersy, line 144: Do we really want this change?

miner/worker.go
```diff
@@ -286,13 +287,23 @@ func (w *worker) pendingBlock() *types.Block {
 
 // start sets the running status as 1 and triggers new work submitting.
 func (w *worker) start() {
+	if pos, ok := w.engine.(consensus.BFT); ok {
+		err := pos.Start(context.Background())
```
piersy, line 291: Shouldn't we be using a cancellable context here?

metrics/meter.go
```diff
@@ -281,17 +281,44 @@ type meterArbiter struct {
...
+	defer ma.Unlock()
+
+	ma.ticker.Stop()
+	atomic.StoreUint32(ma.stopped, 1)
+}
+
+func (ma *meterArbiter) stopTicker() {
+	ma.Lock()
+	defer ma.Unlock()
+	ma.doneCh <- struct{}{}
```
piersy: You don't need to lock here, channel communication is synchronised

metrics/registry.go
```diff
@@ -189,6 +190,20 @@ func (r *StandardRegistry) UnregisterAll() {
...
+// stopMeterTicker Stops Meter Ticker.
+// It's needed to avoid go-routine leaks.
+func (r *StandardRegistry) stopMeterTicker() {
+	arbiter.Lock()
+	started := arbiter.started
+	arbiter.Unlock()
+	if started && atomic.LoadUint32(arbiter.stopped) != 1 {
+		// stop the ticker to avoid go-routine leak
+		arbiter.stopTicker()
+	}
```
piersy: This should be encapsulated in `arbiter.stopTicker`

metrics/registry.go
yazzaoui: They made some changes in the metrics package, I needed to test if they've fixed the leak because we did actually introduce this method with https://github.com/clearmatics/autonity/pull/407 .
Ultimately the leak was still present so I reintroduced this fix.

contracts/autonity/autonity.go
yazzaoui: They've changed the signature of `abi.Unpack`, `UnpackIntoInterface` is the old `Unpack` .

ethclient/ethclient_test.go
yazzaoui: It would have panicked but I'll make it more explicit.

p2p/server.go
yazzaoui: It's not that straightforward, would entail changes that I don't think are in the scope of this PR. 
I'm happy to keep it for later, I'll create a ticket once this gets merged.

miner/worker.go
yazzaoui: That was our old code, I didn't change that.

cmd/autonity/chaincmd.go
yazzaoui: I'll remove it then, it was previously calling `utils.Makegensis` which was returning some default hardcoded genesis depending on the CLI flags, see https://github.com/ethereum/go-ethereum/blob/fb2c79df1995b4e8dfe79f9c75464d29d23aaaf4/cmd/utils/flags.go#L1785-L1800

core/rawdb/database.go
yazzaoui: It's a merge conflict that I've missed in my first pass see https://github.com/ethereum/go-ethereum/blob/fb2c79df1995b4e8dfe79f9c75464d29d23aaaf4/core/rawdb/database.go#L348-L349

metrics/meter.go
yazzaoui: See https://github.com/clearmatics/autonity/pull/703#discussion_r510250744

metrics/registry.go
yazzaoui: See https://github.com/clearmatics/autonity/pull/703#discussion_r510250744

metrics/meter.go
piersy: https://github.com/clearmatics/autonity/pull/703#discussion_r510250744 did not help me understand

p2p/server.go
piersy: Why did you move this in the first place?

cmd/autonity/chaincmd.go
```diff
@@ -284,10 +284,7 @@ func setupDefaults(genesis *core.Genesis) {
 }
 
 func dumpGenesis(ctx *cli.Context) error {
```
piersy, line 286: If we are making this command invalid why not just fully remove this command?

yazzaoui: I'll remove it.

p2p/server.go
yazzaoui: I'll move it back in `server_transport.go` if you prefer, the thinking was that code used for test exclusively didn't warrant to be put in an exclusive, non-test, file.

aneequesafdar (approved):

metrics/meter.go
yazzaoui: Yeah

consensus/test/testcase_test.go
yazzaoui: For the CI logs, I'll revert that when we're happy with this branch.

consensus/consensus.go
yazzaoui: I think `ChainHeaderReader` is only needed for light client support, validators calling `Seal` will always have access to the full blockchain object.

Jason-Zhangxin-Chen (approved):

consensus/consensus.go
piersy: What do we do when we wan't to support light clients then? I thought this was fairly close on our roadmap.

p2p/server.go
piersy: > I'll move it back in `server_transport.go` if you prefer, the thinking was that code used for test exclusively didn't warrant to be put in an exclusive, non-test, file.

I'm not following, code used for tests exclusively should be in a test file or a test package. I thought you said we couldn't move this back to the test file easily? If we can then lets do it. Otherwise I'd like to understand why you moved this out of the test file?

yazzaoui: > I'm not following, code used for tests exclusively should be in a test file or a test package.

Totally agree but doing so here is not in the scope of this PR.

>Otherwise I'd like to understand why you moved this out of the test file?

It was like this before, see https://github.com/clearmatics/autonity/blob/develop/p2p/server_transport.go

consensus/consensus.go
yazzaoui: `Seal` is only called from the miner package so not from light clients.

p2p/server.go
piersy: So the comment confused me

>// testTransport was moved from server_test.go as a temporary workaround around import issues.

But I would definitely move it back, being in a separate file that go-ethereum doesn't have means it cant conflict with their changes. As it stands changes to server.go could now conflict with this.

consensus/consensus.go
piersy: Ahh ok.

p2p/server.go
yazzaoui: Fair enough, already done it yesterday.

piersy (approved):
Lets merge it.

PR #707, opened 2020-10-22
Refactor blockAwaiter, extract height from Algorithm and only start new height when commit event is received

aneequesafdar:


piersy:
These changes were incorporated in #727 which was merged.

PR #709, opened 2020-10-23
Remove `ws:` protocol prefix from URL in NodeJS Console

flofrie:
Removes `ws:` protocol prefix from URL, so that users can specify either `ws://<IP>:<port>` or `wss://<IP>:<port>` on the command line, which allows TLS encrypted connections, e.g. `./console wss://rpc4.bakerloo.autonity.network:8546`.

This is also more consistent with overall expectation that a URL is given at the command line with a protocol prefix.

Note: apparently this does not yet work with self-signed certificates on the server side (such as those used by `rpc1.bakerloo.autonity.network:8546` and `rpc3.bakerloo.autonity.network:8546`), but does work with CA-signed certs, e.g on `rpc2.bakerloo.autonity.network:8546` and `rpc4.bakerloo.autonity.network:8546`. -- So it will need further exploration why `ws` currently seems to ignore system-wide configuration of trusted self-signed certs.


piersy (commented):
@flofrie Thanks for this. We will need to update the usage strings to make it clear that a url is now required.

I'm wondering about your statement that
> This is also more consistent with overall expectation that a URL is given at the command line with a protocol prefix.

What is this "overall expectation that a URL is given at the command line" where does this expectation originate from?

flofrie:
> > This is also more consistent with overall expectation that a URL is given at the command line with a protocol prefix.
> 
> What is this "overall expectation that a URL is given at the command line" where does this expectation originate from?

Perhaps just _my_ expectation? -- E.g. when using Autonity / Geth in console mode, it works with a URL that specifies the protocol, `http` or `https`. In general, I get the impression for many command line tools where you provide a url, the format is `<protocol>://<address>:<port>`. -- E.g when using `curl`, `wget`, etc.

flofrie:
On the question of self-signed certs, I'm looking into it a bit. There's for example https://github.com/theturtle32/WebSocket-Node/issues/259 and https://github.com/websockets/ws/issues/1135

But of course we don't want to use `{ rejectUnauthorized: false }`, because we definitely _do_ want to reject unauthorised connections. We just want it to respect system-wide configurations of trusted certificates. (Or, find a way to explicitly tell it about a trusted certificate)

piersy:
> > What is this "overall expectation that a URL is given at the command line" where does this expectation originate from?
> 
> Perhaps just _my_ expectation? -- E.g. when using Autonity / Geth in console mode, it works with a URL that specifies the protocol, `http` or `https`. In general, I get the impression for many command line tools where you provide a url, the format is `<protocol>://<address>:<port>`. -- E.g when using `curl`, `wget`, etc.

I think I originally used a websocket because I was mistakenly believing it allowed you to wait for the result of a transaction, but I just checked and  was able to wait for a transaction result using an http web3 connection.

To check with http i changed this line:
https://github.com/clearmatics/autonity/blob/0dcd1ed5f2d09f1f9fe65086447b0590447c41c7/nodejsconsole/console#L39

To be `let web3 = new Web3('http://localhost:8541');`


Just for reference I checked like this.
```shell
$ cd helloworld
$ docker-compose up -d
Creating network "helloworld_chainnet" with driver "bridge"
Creating autonity-node-3 ... done
Creating autonity-node-2 ... done
Creating autonity-node-1 ... done
Creating autonity-node-4 ... done
$ cat nodekey1
a4b489752489e0f47e410b8e8cbb1ac1b56770d202ffd45b346ca8355c602c91
$ ../nodejsconsole/console localhost:8641
Welcome to the Autonity node console
modules: admin:1.0 debug:1.0 eth:1.0 miner:1.0 net:1.0 personal:1.0 rpc:1.0 tendermint:1.0 txpool:1.0 
Type web3.<Tab> or autonity.<Tab> to get started

> web3.eth.accounts.wallet.add('a4b489752489e0f47e410b8e8cbb1ac1b56770d202ffd45b346ca8355c602c91')
{ address: '0x850C1Eb8D190e05845ad7F84ac95a318C8AaB07f',
  privateKey:
   '0xa4b489752489e0f47e410b8e8cbb1ac1b56770d202ffd45b346ca8355c602c91',
  signTransaction: [Function: signTransaction],
  sign: [Function: sign],
  encrypt: [Function: encrypt],
  index: 0 }
> await web3.eth.sendTransaction({from: web3.eth.accounts.wallet[0].address, to: web3.eth.accounts.wallet[0].address, value: 1, gas: 50000})
{ blockHash:
   '0x7b869cc20c0c08de808e0e25da1dc766cfbcc4a911c280b57bf70bcc8ddda8c0',
  blockNumber: 1035,
  contractAddress: null,
  cumulativeGasUsed: 21000,
  from: '0x850c1eb8d190e05845ad7f84ac95a318c8aab07f',
  gasUsed: 21000,
  logs: [],
  logsBloom:
   '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
  status: true,
  to: '0x850c1eb8d190e05845ad7f84ac95a318c8aab07f',
  transactionHash:
   '0xaa35c4cb805551d232f18e1c48f94567dc77002d0aad61db0418f6ed7c0ae6eb',
  transactionIndex: 0 }
```

I don't think in general that using `<protocol>://<address>:<port>` is more prevalent than `<address>:<port>`. Its just a question of whether the program you are using is specific to a protocol or not.

I don't see any reason to restrict the console to a specific protocol, but it will take a bit more work than just changing the parameter to allow it to work with http and ws and wss.

This is because we need to call the rpc api before invoking the console in order to find out what modules are loaded so that we can initialise the web3 extensions for them before we actually jump into the console.

See here:
https://github.com/clearmatics/autonity/blob/0dcd1ed5f2d09f1f9fe65086447b0590447c41c7/nodejsconsole/console#L861-L882

Currently we are using a websocket `new WebSocket('ws://$1');`  in order to get the modules, if we allow the user to provide a websocket or http url we will need to detect that and use `new XMLHttpRequest();` in the case of http. XMLHttpRequest I think has a totally different api, so we will need some fairly significant code changes.

flofrie:
I'd think we don't need to allow http. It's just too enable easily using
both ws and was. I've tried it, it works.

flofrie:
@piersy , @yazzaoui , what stands in the way of just making the change proposed in the PR, if only so that we can start using the console in the Bakerloo documentation, to connect to the public endpoints?

I agree that there may be other aspects one can improve, but it would seem that it's not too much to prescribe a format of `ws://URL:port` or `wss://URL:port` ... and those nodes are currently set up with the reverse proxy in place, so it simply works right now.

I didn't mean that at this point we should allow a choice of `http` or `ws`; rather, just let a user choose `ws` or `wss`.

piersy:
@flofrie The only thing standing in the way is updating the usage description in the script, if you execute the script without the correct parameters it will output this.

```
Usage: ./console <websocket enabled autonity rpc host and port>
E.G:   ./console localhost:8546
```

I will update this, then I believe we will be ready to merge it. Sorry for not responding sooner, we have been focusing on getting some of our other PRs merged for this release and I wasn't aware that you were keen to have this change for this release.

piersy (approved):

yazzaoui (approved):

flofrie:
@piersy , @yazzaoui , thanks for approving. Who will merge it?

piersy:
@flofrie I merged it, was just waiting for the tests to pass.

PR #722, opened 2020-11-02
Future proposal consensus fix

yazzaoui:
- Fix future proposal handling: A bug in the handling of future proposals was spotted in the DevNet. The engine was incorrectly moving to the prevote round and setting the proposal for the current round as nil which was leading the node to hang and not be able to commit even though it received a quorum of precommits for a valid proposal. 

- Add missing L49 handling: in the scenario in which we receive a proposal P when core is not at the propose step, the proposal is simply ignored but it could be that we already have a quorum of precommits, we must then commit if this occur.

- Renaming of dumpEconomicMetrics method which was forgotten.

Signed-off-by: yazzaoui <ya@clearmatics.com>

### DIFF
--- a/consensus/tendermint/core/propose.go
+++ b/consensus/tendermint/core/propose.go
@@ -114,6 +114,7 @@ func (c *core) handleProposal(ctx context.Context, msg *Message) error {
 					msg: msg,
 				})
 			})
+			return err
 		}
 		c.sendPrevote(ctx, true)
 		// do not to accept another proposal in current round
--- a/consensus/tendermint/core/propose.go
+++ b/consensus/tendermint/core/propose.go
@@ -125,16 +125,22 @@ func (c *core) handleProposal(ctx context.Context, msg *Message) error {
 		return err
 	}
 
-	// Here is about to accept the Proposal
+	// Set the proposal for the current round
+	c.curRoundMessages.SetProposal(&proposal, msg, true)
+
+	c.logProposalMessageEvent("MessageEvent(Proposal): Received", proposal, msg.Address.String(), c.address.String())
+
+	//l49: Check if we have a quorum of precommits for this proposal
+	curProposalHash := c.curRoundMessages.GetProposalHash()
+	if c.curRoundMessages.PrecommitsPower(curProposalHash) >= c.committeeSet().Quorum() {
+		c.commit(proposal.Round, c.curRoundMessages)
+		return nil
+	}
+
 	if c.step == propose {
 		if err := c.proposeTimeout.stopTimer(); err != nil {
 			return err
 		}
-
-		// Set the proposal for the current round
-		c.curRoundMessages.SetProposal(&proposal, msg, true)
-
-		c.logProposalMessageEvent("MessageEvent(Proposal): Received", proposal, msg.Address.String(), c.address.String())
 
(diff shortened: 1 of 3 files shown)


piersy (approved):

piersy (changes requested):
Actually I was too hasty, the tests are failing

Jason-Zhangxin-Chen (approved):
LGTM

consensus/tendermint/core/propose.go
```diff
@@ -114,6 +114,7 @@ func (c *core) handleProposal(ctx context.Context, msg *Message) error {
 					msg: msg,
 				})
 			})
+			return err
```
piersy, line 117: I don't see that we have a test for this error.

consensus/tendermint/core/propose.go
```diff
@@ -124,17 +125,23 @@ func (c *core) handleProposal(ctx context.Context, msg *Message) error {
...
+
+	c.logProposalMessageEvent("MessageEvent(Proposal): Received", proposal, msg.Address.String(), c.address.String())
+
+	//l49: Check if we have a quorum of precommits for this proposal
+	curProposalHash := c.curRoundMessages.GetProposalHash()
+	if c.curRoundMessages.PrecommitsPower(curProposalHash) >= c.committeeSet().Quorum() {
+		c.commit(proposal.Round, c.curRoundMessages)
+		return nil
+	}
+
```
piersy, line 139: I don't see that we have a test for this either.

consensus/tendermint/core/propose.go
yazzaoui: `"future proposal given, backlog event posted"` is checking it.

consensus/tendermint/core/propose.go
yazzaoui: I'll add one.

consensus/tendermint/core/propose_test.go
```diff
@@ -266,9 +264,66 @@ func TestHandleProposal(t *testing.T) {
...
+			logger:           logger,
+			proposeTimeout:   newTimeout(propose, logger),
+			committee:        valSet,
+			round:            2,
+			height:           big.NewInt(1),
 		}
+
+		err = c.handleProposal(context.Background(), msg)
+		assert.Error(t, err)
+		<-time.NewTimer(2 * time.Second).C
```
piersy, line 326: What is going on here ?

yazzaoui: We have a one second timer before we expect the `backlog` event with the proposal to be posted, so waiting for that to happen here.

piersy: Why do we care? This is the last line of the test, nothing happens after it.

yazzaoui: The `backendMock.EXPECT().Post(event).Times(1)` (L311) would fail if we return too quickly.

piersy, line 326: I think you could add some comments to make this much clearer, as it stands its really not clear. You need to understand that returning 1s from VerifyProposal will cause the message to be posted after 1s and then you need to realise that we are waiting for that to happen.

piersy (approved):
Looking good now :+1:

PR #725, opened 2020-11-05
Race conditions fixes

yazzaoui:
- Go-leak when the miner closes. We enforce now that the miner has properly exited before shutting down the consensus engine during the closing flow of the ethereum backend.

Signed-off-by: yazzaoui <ya@clearmatics.com>

### DIFF
--- a/consensus/test/test_helpers_test.go
+++ b/consensus/test/test_helpers_test.go
@@ -193,6 +193,7 @@ func makeNodeConfig(t *testing.T, genesis *core.Genesis, nodekey *ecdsa.PrivateK
 		Genesis:         genesis,
 		NetworkId:       genesis.Config.ChainID.Uint64(),
 		SyncMode:        downloader.FullSync,
+		NoPruning:       true,
 		DatabaseCache:   256,
 		DatabaseHandles: 256,
 		TxPool:          core.DefaultTxPoolConfig,
--- a/consensus/test/test_helpers_test.go
+++ b/consensus/test/test_helpers_test.go
@@ -193,7 +193,6 @@ func makeNodeConfig(t *testing.T, genesis *core.Genesis, nodekey *ecdsa.PrivateK
 		Genesis:         genesis,
 		NetworkId:       genesis.Config.ChainID.Uint64(),
 		SyncMode:        downloader.FullSync,
-		NoPruning:       true,
 		DatabaseCache:   256,
 		DatabaseHandles: 256,
 		TxPool:          core.DefaultTxPoolConfig,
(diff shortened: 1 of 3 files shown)


miner/miner.go
```diff
@@ -152,7 +152,7 @@ func (miner *Miner) Stop() {
 }
 
 func (miner *Miner) Close() {
-	close(miner.exitCh)
+	miner.exitCh <- struct{}{}
```
piersy, line 155: This isn't going to fully solve our problem, because the miner reads from the exitCh and then closes the worker. So we cannot say when the worker will be closed after calling miner.Close()

yazzaoui: no, the problem is not about the worker, the leak was triggered by the miner loop which when receiving a `downloader.FailedEvent` would restart the worker which would then restart the consensus engine.

piersy: That might have been the case in the problems you were seeing. But my point is we still don't have reliable synchronisation here, because we cannot say how long after calling `miner.Close` will `worker.stop` be called and hence `engine.Close` be called.

In my previous fix, I ensured that upon return from `miner.Close` we could be sure that `engine.Close` had been called.

consensus/test/test_helpers_test.go
```diff
@@ -193,6 +193,7 @@ func makeNodeConfig(t *testing.T, genesis *core.Genesis, nodekey *ecdsa.PrivateK
 		Genesis:         genesis,
 		NetworkId:       genesis.Config.ChainID.Uint64(),
 		SyncMode:        downloader.FullSync,
+		NoPruning:       true,
```
piersy, line 196: I'm sceptical that this was actually a problem, I can't see how our tests would have pruned state when they are not even reaching 128 blocks which is the hardcoded state retention limit.

https://github.com/clearmatics/autonity/blob/8b30891254bf971f34679766ccf6030f395b6ca9/core/blockchain.go#L96

If the reward distribution tests are progressing beyond 128 blocks that is another problem and we should fix that.

yazzaoui: Indeed, but we can check that easily, I will dump the current block number when the assert function is called.

miner/miner.go
yazzaoui: So a recent addition from the upstream merge was to call `engine.Close()` when stopping a node.
https://github.com/clearmatics/autonity/blob/96165c72a7146e03cc88e14353396b557351ee8a/eth/backend.go#L587-L601

With my change, I have the guarantee that no actions on the engine can be taken after L595, either starting or stopping it.

piersy: Having engine starting and stopping managed by miner→worker but having engine closing manged completely separately is confusing. This design will trip us up.

If miner start results in engine start and miner stop results in engine stop we should ensure that miner close results in engine close.

PR #731, opened 2020-11-20
Fix the network economic metrics marshaling issue.

Jason-Zhangxin-Chen:
The structure of economic metrics between AC cotract wrapper and AC is miss-matched, this was due to the commissio rate was removed from AC.

Get the commision rate removed also from wrapper side.


piersy:
@Jason-Zhangxin-Chen 

We should probably add a test to check that this works

aneequesafdar (approved):

Jason-Zhangxin-Chen:
> @Jason-Zhangxin-Chen
> 
> We should probably add a test to check that this works

@piersy , I added a e2e testcase to check if economic data are successfully unpacked by contract ABI, and they should presented at the global metric store. Please review it, thanks.

autonity/autonity.go
```diff
@@ -92,8 +92,7 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
 	}
 
 	// marshal the data from bytes arrays into specified structure.
-	v := EconomicMetaData{make([]common.Address, 32), make([]uint8, 32), make([]*big.Int, 32),
-		make([]*big.Int, 32), new(big.Int), new(big.Int)}
+	v := EconomicMetaData{make([]common.Address, 32), make([]uint8, 32), make([]*big.Int, 32), new(big.Int), new(big.Int)}
```
piersy, line 95: Hi @Jason-Zhangxin-Chen thanks for the test, I'm looking at this more closely and thinking that actually this function should just return an error if there is a problem. Currently it just swallows any error encountered and I'm not sure why. But if it returned an error we could be sure that the metrics are always working, maybe we wouldn't even need the test you just added.

Jason-Zhangxin-Chen: > maybe we

Yes, may be you are right.
May be the oringinal thinking was don't let the metric measurement to block the commit of the block, so it just log the warning.

piersy: In this situation, given that we are patching a release to fix this, it seems that the metrics have sufficient importance to block the commit of a block.

So I think we should go ahead with causing the block creation to fail if we encounter an error in the processing of the metrics.

autonity/autonity.go
```diff
@@ -80,30 +81,41 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
 	input, err := ABI.Pack("dumpEconomicMetrics")
 	if err != nil {
 		log.Warn("Cannot pack the method: ", "err", err.Error())
```
piersy, line 83:
```suggestion
```

autonity/autonity.go
```diff
@@ -80,30 +81,41 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
...
-		return
+		return errors.New("cannot pack contract function")
```
piersy, line 84:
```suggestion
		return fmt.Errorf("failed to pack contract function: %v", err)
```

autonity/autonity.go
```diff
@@ -80,30 +81,41 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
...
 	}
 
 	// call evm.
 	value := new(big.Int).SetUint64(0x00)
 	ret, _, vmerr := evm.Call(vm.AccountRef(Deployer), ContractAddress, input, gas, value)
 	if vmerr != nil {
 		log.Warn("Error Autonity Contract dumpNetworkEconomics", err, vmerr)
-		return
+		return errors.New("cannot init EVM")
```
piersy, line 92:
```suggestion
		return fmt.Errorf("EVM call to dumpNetworkEconomics failed: %v", vmerr)
```

autonity/autonity.go
```diff
@@ -80,30 +81,41 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
...
 		log.Warn("Cannot pack the method: ", "err", err.Error())
-		return
+		return errors.New("cannot pack contract function")
 	}
 
 	// call evm.
 	value := new(big.Int).SetUint64(0x00)
 	ret, _, vmerr := evm.Call(vm.AccountRef(Deployer), ContractAddress, input, gas, value)
 	if vmerr != nil {
 		log.Warn("Error Autonity Contract dumpNetworkEconomics", err, vmerr)
```
piersy, line 91:
```suggestion
```

autonity/autonity.go
```diff
@@ -80,30 +81,41 @@ func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateD
...
-		make([]*big.Int, 32), new(big.Int), new(big.Int)}
+	v := EconomicMetaData{make([]common.Address, 32), make([]uint8, 32), make([]*big.Int, 32), new(big.Int), new(big.Int)}
 
 	if err := ABI.UnpackIntoInterface(&v, "dumpEconomicMetrics", ret); err != nil {
-		// can't work with aliased types
-		log.Warn("Could not unpack dumpNetworkEconomicsData returned value",
-			"err", err,
-			"header.num", header.Number.Uint64())
-		return
+		return errors.New("cannot unpack data from return value from contract")
```
piersy, line 99:
```suggestion
		return fmt.Errorf("cannot unpack data from return value from contract: %v", err)
```

autonity/autonity.go
```diff
@@ -66,9 +67,9 @@ func NewAutonityContract(
 }
 
 // measure metrics of user's meta data by regarding of network economic.
-func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateDB *state.StateDB) {
-	if header == nil || stateDB == nil || header.Number.Uint64() < 1 {
-		return
+func (ac *Contract) MeasureMetricsOfNetworkEconomic(header *types.Header, stateDB *state.StateDB) error {
+	if header.Number.Uint64() < 1 {
+		return errors.New("wrong state parameters")
 	}
```
piersy, line 73: Why is it an error if the header number is less than 1?

Also the error message looks incorrect as it mentions state parameters and this is not related to the `stateDB`.

Additionally the message "wrong state parameters" doesn't provide useful feedback, it should at least tell you what the header number was.

Jason-Zhangxin-Chen: > Why is it an error if the header number is less than 1?
> 
> Also the error message looks incorrect as it mentions state parameters and this is not related to the `stateDB`.
> 
> Additionally the message "wrong state parameters" doesn't provide useful feedback, it should at least tell you what the header number was.

I removed this check since now contract was deployed at block 0. This check does not valid anymore.

piersy (approved):
I suggest using the squash and merge approach for this PR, since it doesn't have a clean history.

For the pr comment I suggest something like.

```
Network economics processing code was previously swallowing errors
encountered while generating metrics. This meant that we got no
feedback when the process was broken due to a change in the autonity
contract that rendered the network economic data generated by the contract
incompatible with the representation in the go code.

This PR restores compatibility between the go code and the autonity contract
and also ensures that errors encountered when processing the network
economic data are not swallowed and instead will cause the node to panic.
This means we will be made aware of these issues whenever they arise again.
```

PR #740, opened 2020-12-15
Restore conformance tests at their latest passing version

szemate:
Conformance tests have been failing since commit 7656dcc. That commit
(presumably accidentally) updated the `testdata` submodule to a later
revision.

The first change in conformance tests that breaks Autonity is coming
from ethereum/tests#676. Reset the `testdata` submodule to the latest
passing version until we figure out how to fix or work around the
failures that are caused by recent changes in the test repository.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -103,9 +103,47 @@ jobs:
           yml: codecov.yml
 
   tests-conformance:
+    needs: bootstrap-checkout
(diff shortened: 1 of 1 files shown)


Jason-Zhangxin-Chen (approved):
LGTM

piersy (approved):
Thank you for this @szemate!

PR #743, opened 2020-12-21
Run go workflow on manual trigger and all PRs

piersy:
This change should ensure that we run the workflow on all pull requests including those from forks and also gives us the ability to trigger the go workflow manually.

It removes the push event which means we will no longer automatically run the go workflow every time we push.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,7 +1,7 @@
 name: Go
 on:
   workflow_dispatch:
-  pull_request_target:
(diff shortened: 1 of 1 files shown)


aneequesafdar (approved):

Jason-Zhangxin-Chen (approved):

yazzaoui:
Ah great, we needed this.

yazzaoui (approved):

PR #748, opened 2021-01-20
Speed up and simplify CI

### LINKED ISSUE #679: Complete speed up ci branch
# Problem

our tests are slow, this makes finding and fixing problems slow.

# Acceptance Criteria

* Tests should be faster than they currently are
* Tests must be as stable as they currently are.

piersy:
This PR attempts to speed up and simplify our github actions CI go workflow.

* The workflow file has been reduced in size from 425 lines to 196 (not including comments).
* The build now runs in approximately 10 minutes instead of 17.
* Max timeout is now 10m rather than 40m

In order to achieve this speedup I needed to reduce the block period from 1 to 0 which allowed our e2e tests to run roughly 6 times faster than they were previously. As a result of reducing the block period a number of synchronisation issues were laid bare and these have been fixed as part of this PR. I also skipped 3 tests one of which I realised was never going to be stable because it relied on timing assumptions #751, another is unreliable due to lack of synchronisation #750, and the final one I cannot figure out, see #749. 

Fixes #679


miner/worker.go
```diff
@@ -151,6 +153,9 @@ type worker struct {
 	resubmitIntervalCh chan time.Duration
 	resubmitAdjustCh   chan *intervalAdjust
 
+	// Go routine waiter
+	waiter *gwait.Waiter
+
```
aneequesafdar, line 158: I looked at the dependency and it doesn't seem to be a lot of code, I think it would be better to move that code inside autonity, it would save us a dependency, also I think a waitgroup can also be used here, although it may be a bit more code

core/rawdb/accessors_chain.go
```diff
@@ -298,7 +300,7 @@ func ReadTxIndexTail(db ethdb.KeyValueReader) *uint64 {
 // into database.
 func WriteTxIndexTail(db ethdb.KeyValueWriter, number uint64) {
 	if err := db.Put(txIndexTailKey, encodeBlockNumber(number)); err != nil {
-		log.Crit("Failed to store the transaction index tail", "err", err)
+		panic(fmt.Sprintf("Failed to store the transaction index: %v", err))
```
aneequesafdar, line 303: any particular reason to just change this to panic, there are other log.Crit in the same file. I think we should remain consistent and if this was only for debugging purposes then I think it should be reverted.

core/blockchain.go
```diff
@@ -2294,7 +2304,12 @@ func (bc *BlockChain) maintainTxIndex(ancients uint64) {
 	}
 	// indexBlocks reindexes or unindexes transactions depending on user configuration
 	indexBlocks := func(tail *uint64, head uint64, done chan struct{}) {
-		defer func() { done <- struct{}{} }()
+		defer func() {
+			select {
+			case done <- struct{}{}:
+			case <-bc.quit:
+			}
+		}()
```
aneequesafdar, line 2312: I am not sure how it helps to read from a channel in a defer statement. The defer statement will only be executed after the function returns even then the selection of cases is random so you can't ensure `case done <- struct{}{}` won't execute. Can you please give further context?

.github/workflows/go.yml
```diff
@@ -1,408 +1,221 @@
...
+        # Here we set up a matrix to control our tests, we use go list and
+        # negative grep to get most of the test packages for the main tests
+        # excluding packages of particular interest. The excluded package are
+        # those that are slow to test, our e2e tests and tendermint core.
+        tests: [[main-tests, $(go list ./... | grep -E -v "/les|/p2p|/eth|/consensus/tendermint|/consensus/test|/core$")],
+                [slow-tests, $(go list ./les/... ./p2p/... ./eth/... | grep -E -v "/p2p/simulations")],
+                # These tests do not seem to work when run with other tests.
+                [problematic-tests, -p 1 ./p2p/simulations/... ./core], 
+                [tendermint-tests, ./consensus/tendermint/... -timeout 5m -race],
+                [e2e-tests, ./consensus/test/... -v]]
```
aneequesafdar, line 77: In this commit abcf65c5 you remove the time limit of 2 hours, are you sure `./consensus/test/... -v` will complete under 10 minutes (I think this is default timelimt)?

.github/workflows/go.yml
```diff
@@ -1,408 +1,221 @@
...
+    strategy:
+      matrix:
+        # Here we set up a matrix to control our tests, we use go list and
+        # negative grep to get most of the test packages for the main tests
+        # excluding packages of particular interest. The excluded package are
+        # those that are slow to test, our e2e tests and tendermint core.
+        tests: [[main-tests, $(go list ./... | grep -E -v "/les|/p2p|/eth|/consensus/tendermint|/consensus/test|/core$")],
+                [slow-tests, $(go list ./les/... ./p2p/... ./eth/... | grep -E -v "/p2p/simulations")],
+                # These tests do not seem to work when run with other tests.
+                [problematic-tests, -p 1 ./p2p/simulations/... ./core],
```
aneequesafdar, line 75: any particular reason to move core to problematic-tests?

consensus/test/start_stop_test.go
```diff
@@ -15,7 +15,7 @@ func TestTendermintStopUpToFNodes(t *testing.T) {
 		{
 			name:          "one node stops at block 1",
 			numValidators: 5,
-			numBlocks:     10,
+			numBlocks:     100,
```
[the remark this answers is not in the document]
piersy: While a node goes down, the network continues to process blocks at a significant pace, if you stick to the original block count all other nodes finish before the stopped node can start again.

consensus/test/test_helpers_test.go
```diff
@@ -415,10 +415,6 @@ wgLoop:
 			peer.blocks[ev.Block.NumberU64()] = block{ev.Block.Hash(), len(ev.Block.Transactions())}
 			peer.lastBlock = ev.Block.NumberU64()
 
-			logger.Error("last mined block", "peer", index,
-				"num", peer.lastBlock, "hash", peer.blocks[ev.Block.NumberU64()].hash,
-				"txCount", peer.blocks[ev.Block.NumberU64()].txs)
-
```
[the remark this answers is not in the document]
piersy: As explained [here](https://github.com/clearmatics/autonity/pull/748#discussion_r565695469) we cannot reduce the blocks for the start stop tests. So I think it's better to remove this log statement. Alternatively we could make the network quiescent (I.E only mine blocks when there are transactions) but I think that should be a separate PR.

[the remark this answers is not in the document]
yazzaoui: I feel it was useful output

miner/worker.go
piersy: The point about the waiter is it ensures you use the waitgroup properly. I can move the code inside.

core/rawdb/accessors_chain.go
piersy: This must have been failing for me, so I changed to a panic so I could have a better understanding of what went wrong. I'll update to use panic across the file.

core/blockchain.go
yazzaoui: I think here it's more about avoiding a deadlock condition when nothing is reading the done channel when quitting

.github/workflows/go.yml
piersy: It will now, unless something goes wrong (avg time 7 mins)

.github/workflows/go.yml
piersy: >These tests do not seem to work when run with other tests

yazzaoui (approved):
So I'm happy with this PR, I've run the tests multiple times with no deadlock or go-routine leak spotted and it indeed speed-up considerably testing :+1: 
One suggestion however: I'd think that it would be beneficial to have a "long" testing mode too with the same assumptions taken currently for the block-period (e.g. >= 1), this mode could be triggered when stuff gets merged to develop/master.

consensus/test/start_stop_test.go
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: It's not necessary to wait for the stop one's recovery, since it does not impact the liveness in that case.

miner/worker.go
aneequesafdar: I would prefer that as it makes it easier to see how the synchronisation is working if it was a lot of code using the dependency would be fine

eth/backend.go
```diff
@@ -603,7 +603,6 @@ func (s *Ethereum) Stop() error {
 	s.txPool.Stop()
 	s.miner.Close()
 	s.blockchain.Stop()
-	s.engine.Close()
```
Jason-Zhangxin-Chen, line 606: Then which point will trigger to run this engine.Close()?

miner/miner.go
```diff
@@ -136,23 +136,32 @@ func (miner *Miner) update() {
 		case <-miner.stopCh:
 			shouldStart = false
 			miner.worker.stop()
+			miner.stopCh <- struct{}{}
```
Jason-Zhangxin-Chen, line 139: Why we rise again event into same channel right after we handled the event from the same channel?

eth/backend.go
aneequesafdar: s.miner.Close() -> worker.stop() -> engine.Close()

miner/miner.go
aneequesafdar: It's a clever way (or hack, take your pick) of not using a waitgroup, this means that the go routine which sent on miner.StopCh can block until the other go routine says its finish executing

consensus/test/start_stop_test.go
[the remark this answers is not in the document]
piersy: @Jason-Zhangxin-Chen what about checking that the node that went down has processed the transactions sent to the network?

consensus/test/test_helpers_test.go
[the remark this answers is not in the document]
piersy: I will put it back in, I think you may find the amount of logs unhelpful though.

core/rawdb/accessors_chain.go
piersy: Actually, I've updated `log.Crit` to print a stack trace

miner/miner.go
piersy: Its just a way to signal back that you have stopped. I wouldn't call it a hack, channels can be bidirectional for a reason.

aneequesafdar (approved):
LGTM

piersy:
@yazzaoui 
> So I'm happy with this PR, I've run the tests multiple times with no deadlock or go-routine leak spotted and it indeed speed-up considerably testing 
> One suggestion however: I'd think that it would be beneficial to have a "long" testing mode too with the same assumptions taken currently for the block-period (e.g. >= 1), this mode could be triggered when stuff gets merged to develop/master.


I don't see the benefits, we will just run the tests that much more slowly, the CPUs will sit idle between every block while they wait for the second to elapse. I think it would be better to run tests with block period 0 and let them run for a long time, then you actually cause the blockchain to do lots of work, you will fill up caches fill up databases, trigger the freezer... etc. All these things will push the code through some new edge cases that may allow us to find more problems.

piersy:
@yazzaoui 

So I added a slow-e2e-tests job that should run when we merge to develop.

PR #753, opened 2021-01-26
New test package

### LINKED ISSUE #567: Use gengen in our project
#Problem

The gengen tool is not being used, and as such is likely to go stale when we change the constraints around the genesis file.

In order to easily be notified of required changes to gengen, we can use gengen to build the genesis files for our tests, when gengen goes out of date tests will fail and we will know that gengen needs to be updated.

# Acceptance Criteria

Use gengen to generate the genesis for tests, the helloworld example and our contract tests.

### LINKED ISSUE #576: Update genesis tool and genesis documentation to make istanbulBlock required
# Problem

gengen is not setting the istanbulBlock and the genesis doc states that the istanbulBlock should not be set.

# Acceptance Criteria

Ensure that the istanbul block is set by the gengen tool and that the documentation makes it clear that it should be set.

### LINKED ISSUE #649: Investigate writing integration level tests
# Problem

The e2e test framework is over-complicated, fragile and does not allow us to be specific about what we test. We need a way to be able to easily test specific scenarios.

### LINKED ISSUE #675: Header marshalling is broken
# Problem

There exist 2 methods for marshaling headers in autonity.

https://github.com/clearmatics/autonity/blob/e69cb4cbd472c5d0f88b817180f8f5e0fded375a/internal/ethapi/api.go#L1097

https://github.com/clearmatics/autonity/blob/e69cb4cbd472c5d0f88b817180f8f5e0fded375a/core/types/gen_header_json.go#L18

And a single method to unmarshal in `gen_header_json.go`. Using the marshaling method defined in `api.go` with the unmarshaling defined in `gen_header_json.go` doesn't work, because they are marshaling into different JSON structures.

Additionally both the marshaling methods are also broken:
* `RPCMarshalHeader` - because it doesn't correctly encode the additional fields that we have added to the header.
* `MarshalJSON` - because it is marshaling the fields added by Autonity to the extra data field, and this doesn't make sense because we have added them to the header. The extra data field exists to allow extra data to be passed around without having to add the fields to the header.

# Acceptance Criteria

* There should be just one marshal method for `types.Header`.
* We should only encode data  in extra data that is not a field on `types.Header`

piersy:
This PR adds a package `test` which provides utilities to simplify working with networks of nodes. It also adds some tests showing how the package can be used and some of these tests replace tests that were using our old e2e test framework.

As part of this work the json header marshaling was fixed #675, the gengen package was moved and modified to make its functionality importable, it was also fixed to include the istanbul block and static dials in the p2p package were throttled to prevent excessive failed dials when starting a network.

These new tests are extremely fast, for example the tests that replace `TestCheckFeeRedirectionAndRedistribution` take 0.36s as opposed to 28.87s, so these new tests are roughly 80 times faster.

Fixes #675
Fixes #649
Fixes #567 
Fixes #576


test/node.go
```diff
@@ -122,14 +122,20 @@ func NewNode(u *gengen.User, genesis *core.Genesis) (*Node, error) {
 // starts eth.Ethereum mining.
 func (n *Node) Start() error {
 	var err error
-	n.Node, err = node.New(n.Config)
+	// Provide a copy of the config to node.New, so that we can rely on
+	// Node.Config field not being manipulated by node and hence use our copy
+	// for black box testing.
+	nodeConfigCopy := *n.Config
```
szemate, line 128: To my understanding it makes a shallow copy of `n.Config` and nested structures and arrays will be the same instances in `n.Config` and `nodeConfigCopy`.

piersy: Yes you are correct @szemate, I've been experimenting with serialising/de-serialising the config in order to create a deep copy but there is a problem marshaling ecdsa keys (see - https://github.com/golang/go/issues/33564#issuecomment-627935091). But this should be fixed in go 1.16. So I was thinking that for the time being, we could use a shallow copy and when go 1.16 is released we should be able to deep copy the config by serialising/deserialising via JSON.

piersy: >To my understanding it makes a shallow copy of n.Config and nested structures and arrays will be the same instances in n.Config and nodeConfigCopy.

Actually its not nested structures that would be the same instances, it is pointer fields, since the pointers will be copied directly and therefore will point to the same memory.

szemate: https://stackoverflow.com/questions/46790190/quicker-way-to-deepcopy-objects-in-golang-json-vs-gob
it uses the `gob` package to deep copy a structure

piersy: `gob` Unfortunately didn't work here either, due to `enode.Node` not having any exported fields. In the end I wrote some custom JSON marshaling for the config object and it seems to be working.

Jason-Zhangxin-Chen (approved):

core/types/block.go
```diff
@@ -107,6 +107,25 @@ type CommitteeMember struct {
...
+// UnmarshalText b as a hex string with 0x prefix.
+func (c *CommitteeMember) UnmarshalText(input []byte) error {
+	var b hexutil.Bytes
+	err := b.UnmarshalText(input)
+	if err != nil {
+		return err
+	}
+	return rlp.DecodeBytes(b, &c)
+}
+
```
aneequesafdar, line 128: 1. Can you please explain how these methods would be called?
2. How do these methods help with Json marshalling and unmarshalling?
3. Why use rlp to encode or decode? Is it for convenience? Since you won't have to come up with specific encoding for `CommitteMember` and won't have to worry about how to separate each field while encoding and decoding? The following is one way to convert the `CommitteeMember` to `[]byte`, however, `UnmarshallText()` would need to consider edges cases of how the length of Voting power can vary.

```go
func (c *CommitteeMember) MarshalText() ([]byte, error) {
     return hexutil.Bytes(append(c.Address[:], c.VotingPower.Bytes()...)).MarshalText()
}
```

piersy: Hey @aneequesafdar, In case you didn't see there is a quite detailed [commit comment](https://github.com/clearmatics/autonity/pull/753/commits/1efadc0ea41b8a9aad8c7cbfb687aa3df2f1a2b0) for this change.

As to your questions.

1&2 - These methods implement the `encoding.TextMarshaler` interface which the `json` package uses to to marshal objects to text, such as these, the elements of the committeeMembers slice. `MarshalText` is distinct from `MarshalJSON` since the output is not a valid JSON object.

3 - I needed a way to convert a `CommitteeMember` to `[]byte`, rlp seemed the obvious choice since elsewhere headers are serialised via rlp, so we can rely on rlp working, and the implementation shouldn't need to change if the structure `CommitteeMember` is modified.

p2p/dial.go
```diff
@@ -419,6 +421,16 @@ func (d *dialScheduler) startStaticDials(n int) (started int) {
 	for started = 0; started < n && len(d.staticPool) > 0; started++ {
 		idx := d.rand.Intn(len(d.staticPool))
 		task := d.staticPool[idx]
+
+		// Add this task to the history so we don't go crazy trying to redial it
+		//
+		// The task history is consuled before dialing and if an entry found it
```
aneequesafdar, line 427:
```suggestion
		// The task history is consoled before dialling and if an entry found it
```

aneequesafdar, line 427:
```suggestion
		// The task history is consulted before dialling and if an entry found it
```

p2p/dial.go
```diff
@@ -419,6 +421,16 @@ func (d *dialScheduler) startStaticDials(n int) (started int) {
...
+		// is not dialled. Entries expire and when they do they send an event
+		// on the expiry channel that kicks off another round of dialing.
+		hkey := string(task.dest.ID().Bytes())
+		// We add a bit of randomness to the expiration to try and interleave crossidals.
```
aneequesafdar, line 431:
```suggestion
		// We add a bit of randomness to the expiration to try and interleave cross-dials.
```

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Give this logger context based on the node address so that we can easily
+	// trace single node execution in the logs. We set the logger only on the
+	// copy, since it is not useful for black box testing and it is also not
+	// marshalable since the implementation contains unexported fields.
+	nodeConfigCopy.Logger = log.New("node", n.Address.String()[2:7])
+	// n.Config.P2P.PrivateKey = n.ConfigCopy.P2P.PrivateKey
```
aneequesafdar, line 146: did you mean `n.Config.P2P.PrivateKey = nodeConfigCopy.P2P.PrivateKey`?

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+// Since the node config is not marshalable by default we construct a
+// marshalable struct which we marshal and unmarshal and then unpack into the
+// original struct type.
+func copyNodeConfig(source, dest *node.Config) error {
+	s := &MarshalableNodeConfig{}
+	s.Config = *source
+	p := MarshalableP2PConfig{}
+	p.Config = source.P2P
+
+	crypto.FromECDSA(source.P2P.PrivateKey)
```
aneequesafdar, line 393: Why is the return value being ignored?

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+	nodeConfigCopy := &node.Config{}
+	err := copyNodeConfig(n.Config, nodeConfigCopy)
+	if err != nil {
+		return err
+	}
+
+	// Give this logger context based on the node address so that we can easily
+	// trace single node execution in the logs. We set the logger only on the
+	// copy, since it is not useful for black box testing and it is also not
+	// marshalable since the implementation contains unexported fields.
```
aneequesafdar, line 144: Can you please explain how unexported fields affect marshalling? Won't the unexported fields just not appear in the final JSON?

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+		return nil, err
+	}
+	// Make the tests fast
+	g.Config.Tendermint.BlockPeriod = 0
+	return g, nil
+}
+
+// Since the node config is not marshalable by default we construct a
+// marshalable struct which we marshal and unmarshal and then unpack into the
+// original struct type.
```
aneequesafdar, line 386: 1. Can you please explain why the node config is not marshallable?
2. Can we not define `MasrshallJSON()` and `UnmarshallJSON()` on  `node.Config`?

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return err
+	}
+	*k = MarshalableECDSAPrivateKey(*key)
+	return nil
+}
+
+func (k *MarshalableECDSAPrivateKey) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + hex.EncodeToString(crypto.FromECDSA((*ecdsa.PrivateKey)(k))) + `"`), nil
+}
```
aneequesafdar, line 435: 1. Why do you not these functions directly on `ecdsa.PrivateKey`?
2. Also, how did you determine that only `ecdsa.PrivateKey` needed `MarshallJSON()` and `UnmarshallJSON()` to have a deep copy?
3. What about `*discv5.Node`, `*netutil.Netlist`, `[]Protocol` and others which are exported fields on `p2p.Config` didn't cause an issue with marshalling and unmarshalling?

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+
+// copyObject copies an object so that the copy shares no memory with the
+// original.
+func copyObject(source, dest interface{}) error {
```
aneequesafdar, line 439: This function is called in 3 places and with 2 types: `*node.Config` and `*eth.Config`, neither of which defines `MarshallJSON()` and `UnmarshallJSON()` , so how does this work?

test/node.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+	// Make the tests fast
+	g.Config.Tendermint.BlockPeriod = 0
+	return g, nil
+}
+
+// Since the node config is not marshalable by default we construct a
+// marshalable struct which we marshal and unmarshal and then unpack into the
+// original struct type.
+func copyNodeConfig(source, dest *node.Config) error {
```
aneequesafdar, line 387: Can you please explain why you have both `copyNodeConfig()` and `copyObject()` when both function are called with the same type: `*node.Config`?

p2p/dial.go
piersy: Hey @aneequesafdar, I'm not able to merge these suggestions any more, both suggestions look good so please merge yourself

test/node.go
piersy: No, this line can be removed `copyNodeConfig` makes a full deep copy of the config. This is probably a leftover from, before I was able to marshal/unmarshal the private key.

test/node.go
piersy: This is just redundant, left over from when I was trying to figure out how to marshal/unmarshal the config. So please remove :)

test/node.go
piersy: What I meant is you can't marshal and unmarshal the logger and expect to get something working, because of the unexported fields, since they will be missing.

test/node.go
piersy: 1. The node config is not marshalable because it contains an `ecdsa.PrivateKey` which is not marshalable. I needed a custom type to be able to marshal and unmarshal the key, and because the key is 2 levels deep in the config struct I created 2 levels of marshalable types around the key to get to it. I'm mow thinking that in fact that was unnecessary and I could have just added the `MarshalableECDSAPrivateKey` at the top level in `MarshalableNodeConfig`, so I would see if you could simplify it like that.

2. Yes we could, we could just split the `copyNodeConfig` into a marshal and unmarshal function, but I'm not sure there would be any benefit because we only copy this config in one place, these tests. If you wanted to make the config generally copyable throughout the codebase, then that would be a reason to do it.

test/node.go
piersy: 1. I wouldn't be able to do that without editing the ecdsa package, which is not under our control.
2. Looking at the node config now, I see there are lots of things that can't me marshaled/unmarshaled. the p2p.Config has lots of interfaces, but for the config that I was using in the test package only the ecdsa.PrivateKey needed special attention. This was easy to discover because you would encounter an error when trying to unmarshal the key.
```go
json: cannot unmarshal object into Go struct field PrivateKey.P2P.PrivateKey.Curve of type elliptic.Curve
```
3. As I noted in my answer to 2 they would cause problems, but we are not setting them, you could definitely add some docs to make it clear that these things should not be set, it would be quite a lot of work to make them all marshal/unmarshal - able.

test/node.go
piersy: This is just making use of default json marshaling, the objects that are being copied don't have any tricky fields set so it works. It would probably be good to stick to using `copyNodeConfig` for node config though.

test/node.go
piersy: As noted above `copyObject` works fine if the object only has fields set that can be marshaled/unmarshaled by the default json marshaling approach, `copyNodeConfig` is required if a private key has been set on the node config.

cmd/gengen/gengen/cmd.go
```diff
@@ -0,0 +1,314 @@
...
+		userType = params.UserValidator
+	default:
+		return nil, fmt.Errorf("failed to parse user type %q, not one of u, s or p", fields[1])
+	}
+
+	bigStake, err := ParseUint(fields[2])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse stake: %v", err)
+	}
+	if !bigStake.IsInt64() {
```
aneequesafdar, line 272: Did you mean?

```suggestion
	if !bigStake.IsUint64() {
```

piersy: Yep

aneequesafdar (approved):
LGTM

test/node.go
aneequesafdar: Strangely the removal of this piece of code cause a data race between a map and atomic value, see the log here
https://github.com/clearmatics/autonity/runs/2126282174?check_suite_focus=true

aneequesafdar: From the latest commit it this no longer seem to be the culprit.

piersy: This is a data race due to reading and writing `conn.flags` in an unsynchronised manner. This PR starts running tests with the race detector so we are now detecting lots of data races that previously went unnoticed.

This is a geth problem, if you look into the geth codebase you will see that `p2p.Server.run` and `p2p.dialScheduler.loop` call the methods highlighted in the log you linked and they are each called in their own goroutines.

piersy: Looks good to me.

aneequesafdar:
@piersy I have rebased the commits from this branch on top of develop in #773 to remove the merge commit from `new_test_framework`.

PR #755, opened 2021-01-28
Accountability tx con

IronGauntlets:
Creating this draft pull request for commenting and reviewing


core/vm/contracts.go
```diff
@@ -41,8 +41,9 @@ import (
 // requires a deterministic gas count based on the input size of the Run method of the
 // contract.
 type PrecompiledContract interface {
-	RequiredGas(input []byte) uint64  // RequiredPrice calculates the contract gas use
-	Run(input []byte) ([]byte, error) // Run runs the precompiled contract
+	RequiredGas(input []byte) uint64     // RequiredPrice calculates the contract gas use
+	Run(input []byte) ([]byte, error)    // Run runs the precompiled contract
+	InitChainContext(chain ChainContext) // InitChainContext bring chain context for precompiled contract.
```
piersy, line 46: This is bad interface design, If most of your implementations are not implementing all the interface methods then the interface is wrong.

Look how this decision to add this method here resulted in the need to update 22 separate implementations with methods that they didn't need.

Instead we can implement and construct a precompiled contract outside of this package and add it to the precompiled contract maps at runtime.

afd/accountability_contracts.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/clearmatics/autonity/core/vm"
+	"github.com/clearmatics/autonity/params"
+	"github.com/clearmatics/autonity/rlp"
+)
+
+var (
+	checkAccusationAddress = common.BytesToAddress([]byte{252})
+	checkProofAddress = common.BytesToAddress([]byte{253})
+	checkChallengeAddress = common.BytesToAddress([]byte{254})
+	failure64Byte = make([]byte, 64)
```
aneequesafdar, line 18: Why do you need a failuer64Byte array? Can  you simply not return nil for `Run` function defined on Precompiled contracts?

core/types/afd_types.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// The proof used by accountability precompiled contract to validate the proof of innocent or misbehavior.
+// Since precompiled contract take raw bytes as input, so it should be rlp encoded into bytes before client send the
+// proof into autonity contract.
+type RawProof struct {
+	Rule       Rule       // rule id.
+	Message    []byte      // the raw rlp encoded msg to be considered as suspicious one
+	Evidence   [][]byte    // the raw rlp encoded msgs as proof of innocent or misbehavior.
+}
```
aneequesafdar, line 38: I am really not sure that this is required because `ConsensusMessages` have `Encode()` and `Decode()` defined on them so I think we can rely on the rlp package to call those recursively, therefore, `Proof` defined below, in my opinion, should be enough.

afd/accountability_contracts.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	header := a.chain.GetHeaderByNumber(h.Uint64())
+	if _, err = in.Message.Validate(crypto.CheckValidatorSignature, header); err != nil {
+		return failure64Byte, err
+	}
+
+	msgHash := types.RLPHash(in.Message.Payload()).Bytes()
+	sender := common.LeftPadBytes(in.Message.Address.Bytes(), 32)
+	return append(sender, msgHash...), nil
```
aneequesafdar, line 123: Can you please explain what you are trying to do here? I don't understand the need for padding 32 bytes to the left of the sender's address

consensus/tendermint/core/message.go
```diff
@@ -39,7 +39,7 @@ var (
 	ErrUnauthorizedAddress  = errors.New("unauthorized address")
 )
 
-type Message struct {
+type ConsensusMessage struct {
```
piersy, line 42: Hi @Jason-Zhangxin-Chen I hope you are well, I still get notifications from Autonity :)

Now that you moved the messages back into core I would suggest changing the name back to just `Message`. That name change touched a lot of places in the code which makes the diff quite large making it harder to spot the functional changes.

Jason-Zhangxin-Chen: Hey Piersy, how do you do with the new job? Yes, I renamed it back as it used to be.

piersy: Hi Jason, the new job is going well thanks, still just settling in :)

faultdetector/accountability_precompile_contracts.go
```diff
@@ -184,8 +184,8 @@ func (c *ChallengeVerifier) validateChallenge(p *Proof) ([]byte, error) {
 
 	if c.validEvidence(p) {
 		msgHash := types.RLPHash(p.Message.Payload()).Bytes()
-		senderHash := types.RLPHash(p.Message.Address).Bytes()
-		return append(msgHash, senderHash...), nil
+		sender := common.LeftPadBytes(p.Message.Address.Bytes(), 32)
+		return append(sender, msgHash...)[0:63], nil
```
aneequesafdar, line 194: i think the slice should be `[0:64]` since the final index is not included in the slice. The same applies to the rest of the append statements.

yazzaoui:
Depreciated, closing.

PR #759, opened 2021-02-12
Simplify core

piersy:
Note this PR's base is `new_test-framework` since this PR builds on top of that. This will need to be rebased on top of `develop` when `new_test_framework` is merged.

This PR completely rewrites the core of the system, it removes both the core and backend packages
and replaces them with a single object Bridge. The bridge bridges the gap between the ethereum
system and the now new implementation of the tendermint algorithm in the 'algorithm' package.
    
A lot needed to change to accommodate this but the result is significantly less code
(minus 2300 lines, not including tests) and a much less tangled system that is easier to reason about.
    
Also the new tendermint algorithm implementation has very few dependencies making it much easier to test.

All the old e2e and new e2e tests are passing, so although there is a lot more testing that could be done I would
consider this good enough to merge at this point. You could consider the remaining testing as filling in the gaps.

TODOs:

* `GetWhitelist` in consensus/tendermint/api.go should return errors.
* The api tests in consensus/tendermint/api_test.go need re-implementing.
* The `peerMessageMap` in consensus/tendermint/bridge.go is not yet implemented. This means nodes will re-broadcast messages to nodes that they have already sent messages to.
*  The core state dump functionality has been commented out, given that the implementation of core has completely changed, a lot of the things that the old state dumper was dumping do not exist so what exactly needs to be dumped will need to be decided and then It can be rebuilt

Increased testing:

Basically all components in `consensus/tendermint` could do with some more detailed unit/integration style tests.
* messageStore
* Verifier
* Syncer
* Broadcaster
* Bridge
* Oracle
* blockAwaiter

There could be more, using coverage should help direct your efforts.

Specific things I would write tests for are a new bridge test that checks that the bridge correctly handles multiple round consensus scenarios. In order to have control over the timeouts we will need to implement a "manual" TimeoutScheduler that allows us to trigger a timeout at a certain point in the test.

And similarly I would write a test that checks multiple round consensus scenarios for the tendermint algorithm.

Also I think it could be useful to write a test or a number of tests where a number of `algorithm.Algorithm` instances interact with each other, similar to `TestReachingConsensus` for the bridge but without all the boilerplate, timing and synchronisation concerns. It could be possible to randomly generate many scenarios here and verify that consensus works in all cases. At that level you could probably crunch through a few thousand different scenarios in under a second.

Future work
It would be nice if the bridge had a way to disconnect peers sending bad or conflicting messages. Currently there is no mechanism to communicate this information back to the ethereum system.


piersy:
Hey @yazzaoui I hope you are well. Just letting you know that I'm still happy to provide support for this and other PRs that I raised, but that the details are probably starting to fade from my memory so the sooner you raise any comments on them the easier it will be for me to respond/modify.

yazzaoui:
Hi @piersy , thanks for that! We were busy lately but we're going to spend some time this week to review that.

Jason-Zhangxin-Chen (approved):
LGTM.

PR #773, opened 2021-03-18
New test framework

### LINKED ISSUE #567: Use gengen in our project

### LINKED ISSUE #576: Update genesis tool and genesis documentation to make istanbulBlock required

### LINKED ISSUE #649: Investigate writing integration level tests

### LINKED ISSUE #675: Header marshalling is broken

IronGauntlets:
Original pull request (#753) message:

  This PR adds a package test which provides utilities to simplify working with networks of nodes. It also adds some tests 
  showing how the package can be used and some of these tests replace tests that were using our old e2e test framework.

  As part of this work the json header marshaling was fixed #675, the gengen package was moved and modified to make its 
  functionality importable, it was also fixed to include the istanbul block and static dials in the p2p package were throttled to 
  prevent excessive failed dials when starting a network.

  These new tests are extremely fast, for example the tests that replace TestCheckFeeRedirectionAndRedistribution take 0.36s 
  as opposed to 28.87s, so these new tests are roughly 80 times faster.

  Fixes #675
  Fixes #649
  Fixes #567
  Fixes #576


piersy:
@aneequesafdar It looks good but I would now clean up the history a bit more. Both the commits to do with copying config and the "Prevent panic on incorrect tracker shutdown" commit could be melded into the add test framework commit.

Then the "fix typos and remove redundant" code commit contains fixups to "make gengen importable", "throttle static dials" and "add test framework", so I would split that up and meld it with those commits.

Jason-Zhangxin-Chen (approved):
LGTM.

yazzaoui (approved):
Great !

PR #777, opened 2021-03-30
Simplify core rebase onto develop

IronGauntlets:
This PR rebases `simplify_core` onto develop after `new_test_framework` commits were merged into develop.

The original pull request (#759) message:

Note this PR's base is `new_test_framework` since this PR builds on top of that. This will need to be rebased on top of `develop` when `new_test_framework` is merged.

This PR completely rewrites the core of the system, it removes both the core and backend packages
and replaces them with a single object Bridge. The bridge bridges the gap between the ethereum
system and the now new implementation of the tendermint algorithm in the 'algorithm' package.
    
A lot needed to change to accommodate this but the result is significantly less code
(minus 2300 lines, not including tests) and a much less tangled system that is easier to reason about.
    
Also the new tendermint algorithm implementation has very few dependencies making it much easier to test.

All the old e2e and new e2e tests are passing, so although there is a lot more testing that could be done I would
consider this good enough to merge at this point. You could consider the remaining testing as filling in the gaps. Coverage of the tendermint package has dropped by 23% but when you consider that there are 6825 less lines of tests its actually not that bad. I think a higher than previous coverage could easily be obtained with just a few hundred lines of additional tests. 

TODOs:

* `GetWhitelist` in consensus/tendermint/api.go should return errors.
* The api tests in consensus/tendermint/api_test.go need re-implementing.
* The `peerMessageMap` in consensus/tendermint/bridge.go is not yet implemented. This means nodes will re-broadcast messages to nodes that they have already sent messages to.
*  The core state dump functionality has been commented out, given that the implementation of core has completely changed, a lot of the things that the old state dumper was dumping do not exist so what exactly needs to be dumped will need to be decided and then It can be rebuilt

Increased testing:

Basically all components in `consensus/tendermint` could do with some more detailed unit/integration style tests.
* messageStore
* Verifier
* Syncer
* Broadcaster
* Bridge
* Oracle
* blockAwaiter

There could be more, using coverage should help direct your efforts.

Specific things I would write tests for are a new bridge test that checks that the bridge correctly handles multiple round consensus scenarios. In order to have control over the timeouts we will need to implement a "manual" TimeoutScheduler that allows us to trigger a timeout at a certain point in the test.

And similarly I would write a test that checks multiple round consensus scenarios for the tendermint algorithm.

Also I think it could be useful to write a test or a number of tests where a number of `algorithm.Algorithm` instances interact with each other, similar to `TestReachingConsensus` for the bridge but without all the boilerplate, timing and synchronisation concerns. It could be possible to randomly generate many scenarios here and verify that consensus works in all cases. At that level you could probably crunch through a few thousand different scenarios in under a second.

Future work
It would be nice if the bridge had a way to disconnect peers sending bad or conflicting messages. Currently there is no mechanism to communicate this information back to the ethereum system.


consensus/tendermint/message_store.go
```diff
@@ -0,0 +1,310 @@
...
+	// // println("precommit power", m.votePower(valueHash, round, msgType, header))
+	// println("precommit power --------")
+	vp := m.votePower(valueHash, round, msgType, header) >= header.Committee.Quorum()
+	// println(vp, "----------------")
+	return vp
+}
+
+func (m *messageStore) fail(round int64, header *types.Header) bool {
+	// println("fail power --------")
+	vp := m.votePower(nil, round, nil, header) >= header.Committee.Quorum()
```
piersy, line 102: Hi guys, hope you're all well!

```suggestion
	vp := m.votePower(nil, round, nil, header) >= header.Committee.F()
```

PR #780, opened 2021-05-09
Delete package-lock.json in node js console

yazzaoui:
I'm not a node developer but the web3 dependencies are causing security alerts on the github dashboard, they are very probably not applicable for autonity given how the console is being used but anyway.. let's fix them. 
We don't really have any versioned dependencies for the nodejs console, so I suggest getting rid of package-lock.json.
And to be fair having package-lock.json without package.json looks very weird to me.
If someone wants to figure out how to correctly version node dependencies please have a go at it.


yazzaoui:
NodeJS console is depreciated, closing.

PR #807, opened 2023-07-21
Improved unbonding

yazzaoui:
This implements improved unbonding and fixes some issues with the Bonding/Unbonding requests API.

### Improved unbonding

This implementation is not strictly following the ADR because of the PAS and the slashing rules integration which is making the design much more complicated than anticipated.

Essentially, the protocol needs to track via a "share" mechanism the relative ownership of stake in the `unbonding` and `self-unbonding` pools. This is due to the reason that the unbonding stake might be slashed, reducing, therefore, the released amount of stake at the end of the unbonding period. The validator structure has been updated with the addition of some new elements. 

-  `unbondingStake`: the total amount of NTN in the unbonding staking pool.
-  `unbondingShares`: the total amount of shares issued for the unbonding staking pool.
- `selfUnbondingStake`: the total amount of NTN in the `self-unbonding` staking pool.
- `selfUnbondingShares`: the total amount of shares in the `self-unbonding` staking pool.

The slashing logic has been updated with the following logic:

- Upon a `SlashingEvent`. the total amount of stake available for slashing is: `bondedStake + selfUnbondingStake + UnbondingStake`  with  `bondedStake` being equal to self-bonded stake + delegated stake. 
- self-unbonding stake gets slashed in first priority
- if the total amount slashed previously isn't enough, then self-bonded stake gets slashed
- if the total amount slashed previously isn't enough,  then unbonding stake and delegated stake get slashed in the same proportions.

### API changes

- Added indexing for most events
- Added new events for tracking bonding and unbonding requests.
    - NewBondingRequest event: 
    ```solidity
    /** @notice This event is emitted when a bonding request to a validator node has been registered.
    * This request will only be effective at the end of the current epoch however the stake will be
    * put in custody immediately from the delegator's account.
    * @param validator The validator node account.
    * @param delegator The caller.
    * @param selfBonded True if the validator treasury initiated the request. No LNEW will be issued.
    * @param amount The amount of NEWTON to be delegated.
    */
    event NewBondingRequest(address indexed validator, address indexed delegator, bool selfBonded, uint256 amount);
    ```
    - NewBondingRequest event: 
    ```solidity
    /** @notice This event is emitted when an unbonding request to a validator node has been registered.
    * This request will only be effective after the unbonding period, rounded to the next epoch.
    * Please note that because of potential slashing events during this delay period, the released amount
    * may or may not be correspond to the amount requested.
    * @param validator The validator node account.
    * @param delegator The caller.
    * @param selfBonded True if the validator treasury initiated the request.
    * @param amount If self-bonded this is the requested amount of NEWTON to be unbonded.
    * If not self-bonded, this is the amount of Liquid Newton to be unbonded.
    */
    event NewUnbondingRequest(address indexed validator, address indexed delegator, bool selfBonded, uint256 amount);
    ```
- `getUnbondingReq` and `getBondingReq` have been removed along with the corresponding head and tail getters


autonity/solidity/contracts/Accountability.sol
```diff
@@ -326,23 +325,49 @@ contract Accountability is IAccountability {
...
+                _remaining -= _val.selfBondedStake;
+                _val.bondedStake -= _val.selfBondedStake;
+                _val.selfBondedStake = 0;
+            }
+        }
+        // --------------------------------------------
+        // Remaining stake to be slashed is split equally between the delegated
+        // stake pool and the non-self unbonding stake pool.
+        // As a reminder, the delegated stake pool is bondedStake - selfBondedStake.
+        if (_remaining > 0 && (_val.unbondingStake + (_val.bondedStake - _val.selfBondedStake) > 0)) {
```
vpiyush, line 358: I believe by this point seflBondedStake should always be zero

autonity/solidity/contracts/Accountability.sol
```diff
@@ -326,23 +325,49 @@ contract Accountability is IAccountability {
...
+            uint256 _unbondingSlash = (_remaining * _val.unbondingStake) /
```
vpiyush, line 359: What does this calculation suppose to do? could you explain this

autonity/solidity/contracts/Accountability.sol
yazzaoui: Yes, good point! That simplifies the code here.

autonity/solidity/contracts/Accountability.sol
yazzaoui: I'll add a comment but it's really just a simple proportionality calculation to split fairly the remaining amount of stake to be slashed from the bonded stake pool and the unbonding stake pool.

autonity/solidity/contracts/Accountability.sol
```diff
@@ -326,23 +325,49 @@ contract Accountability is IAccountability {
...
+                                        (_val.unbondingStake + (_val.bondedStake - _val.selfBondedStake));
+            _val.unbondingStake -= _unbondingSlash;
+            _remaining -= _unbondingSlash;
+            _val.bondedStake -= _remaining;
         }
-        _val.bondedStake -= _slashingAmount;
-        _val.totalSlashed += _slashingAmount;
 
+        _val.totalSlashed += _slashingAmount - _remaining;
```
lorenzo-dev1, line 366: Why ` - _remaining`? Shouldn't  this part always be absorbed by the `_val.bondedStake` with `_val.bondedStake -= _remaining;` ?

lorenzo-dev1:
Not modified by this PR but I noticed that in the `redistribute()` function of Liquid.sol we have a check `require(_validatorReward < _reward, "invalid validator reward");`. However we do allow commission rate to be 100%, so it could happen that `_validatorReward == _reward` and it would be correct behaviour.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -979,49 +1051,83 @@ contract Autonity is IERC20, Upgradeable {
...
-            if(_validator.selfBondedStake < _newtonAmount) {
-                _newtonAmount = _validator.selfBondedStake;
+            if (_validator.selfUnbondingStake == 0) {
+                 _unbonding.unbondingShare = _newtonAmount;
+            } else {
+                _unbonding.unbondingShare = (_newtonAmount * _validator.selfUnbondingShares)/_validator.selfUnbondingStake;
             }
-            _validator.selfBondedStake -= _newtonAmount;
+            _validator.selfUnbondingStake += _newtonAmount;
+            _validator.unbondingShares += _unbonding.unbondingShare;
```
lorenzo-dev1, line 1120: should this be: `_validator.selfUnbondingShares` ?

lorenzo-dev1:
Scenario:

Supposing we have a registered validator with no self-bonded stake. Then delegator1 bonds 10 NTN to it.

At the end of the epoch we will have:
```
_validator.liquidSupply = 10
_validator.bondedStake = 10
```

Then delegator1 unbonds. The LNTN gets locked, then at the end of the epoch we will have:
```
_validator.liquidSupply = 0
 _validator.bondedStake = 0
_validator.unbondingStake = 10
_validator.unbondingShares = 10
_unbonding.unbondingShare = 10
```

Then before the unbonding period expires, the validator gets slashed due a past offence with critical severity (100% slash) and all of the unbondingStake gets slashed:
```
_validator.unbondingStake = 0
```

So delegator1 should receive 0 NTN when the unbonding period expires.

But let's assume that, always before the unbonding period expires, delegator2 bonds 10 NTN:
```
_validator.liquidSupply = 10
_validator.bondedStake = 10
```

And unbonds 10 NTN:
```
_validator.liquidSupply = 0
 _validator.bondedStake = 0
_validator.unbondingStake = 10
_validator.unbondingShares = 20
_unbonding.unbondingShare = 10
```

Now when the two unbonding periods expire, what is going to happen is that both delegator1 and delegator2 will take the hit from the slashing, instead of only delegator1.

Agreed upon solution: Enforce minimum of `unbondingPeriod` between bond and unbond operations.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1072,16 +1121,26 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
 
+        tailBondingID = headBondingID;
+        if(tailUnbondingID == headUnbondingID) {
+            // everything else already processed, return early
+            return;
+        }
+        // Process the fresh unbonding requests, unbond NTN and burn LNTN
+        for (uint256 i = headUnbondingID - 1;
+                     i >= tailUnbondingID && !unbondingMap[i].unlocked;
+                      _applyUnbonding(i--)){}
```
lorenzo-dev1, line 1138: This causes arithmetic underflow when `headUnbondingID = 1` --> `i=0` --> `i-- = -1`

yazzaoui: Reviewing this code path I'm no longer sure if it's okay to process those requests in this reverse order...

autonity/solidity/contracts/Autonity.sol
yazzaoui: Yes absolutely!

autonity/solidity/contracts/Autonity.sol
yazzaoui: @lorenzo-dev1 I've updated the logic here, could you please review it

lorenzo-dev1: Looks fine to me. Going to rebase the fast-test branch on top of the new version and see if tests pass.

autonity/solidity/contracts/Accountability.sol
yazzaoui: yes _remaining must always be 0 at this stage. For some reasons I was thinking it could be possible that it wouldn't be the case but I can't think about any of this scenario now.

yazzaoui:
> Not modified by this PR but I noticed that in the `redistribute()` function of Liquid.sol we have a check `require(_validatorReward < _reward, "invalid validator reward");`. However we do allow commission rate to be 100%, so it could happen that `_validatorReward == _reward` and it would be correct behaviour.

Good spot !

PR #809, opened 2023-07-28
Add the Auton Stabilization Mechanism (ASM)

aiman:
This PR adds the first version of the Auton Stabilization Mechanism (ASM), which consists of three smart contracts:

- **ACU Contract**: computes the value of the ACU index 
- **SupplyControl Contract**: controls the supply of ATN on the network
- **Stabilization Contract**: a CDP-based stabilization mechanism for the ATN

The contracts are ready for protocol integration. The ASM smart contracts should be be deployed by the protocol at genesis. The constructor arguments will need to be represented in the genesis configuration.

The PR also adds a set of tests for each of these contracts. The tests make use of the [Ape Framework](https://github.com/ApeWorX/ape), this can co-exist alongside the existing Truffle tests. The installation and setup is fully automated in the Makefile, but you will also need to have `pipx` installed (`apt install pipx`) to install the Ape Framework.

Also, to run the tests you will need to make sure to be using a supported version Node.js, otherwise it breaks a lot of tools.You can check the supported versions of Node.js here: <https://github.com/nodejs/release#release-schedule>.

For example, Ubuntu LTS 22.04 ships with Node.js v12, which reached EOL last year on 2022-04-30 and is not compatible with Hardhat. If you are using Ubuntu LTS then uninstall `nodejs` and `npm` and instead use the version from NodeSource: <https://github.com/nodesource/distributions>.


autonity/solidity/contracts/asm/ACU.sol
```diff
@@ -0,0 +1,203 @@
...
+    ///
+    /// It retrieves the latest prices from the Oracle Contract. If one or
+    /// more prices are unavailable from the Oracle, it will not compute the /
+    // value for that round.
+    ///
+    /// This function is intended to be called by the protocol during block
+    /// finalisation, after the Oracle Contract finalisation has completed.
+    /// @return status Whether the ACU value was updated successfully
+    /// @dev This function is designed to not revert due to arithmetic
+    /// overflow. The implementation could be greatly simplified if we are able
```
yazzaoui, line 103: > The implementation could be greatly simplified if we are able to call it inside of a try/catch block instead.

Agree, we should do that.

autonity/solidity/contracts/asm/ACU.sol
```diff
@@ -0,0 +1,203 @@
...
+            }
+            int256 pqScaled;
+            unchecked {
+                pqScaled = product / oraclePrecision;
+                if (newValue > 0 && pqScaled > type(int256).max - newValue)
+                    return false;
+                if (newValue < 0 && pqScaled < type(int256).max - newValue)
+                    return false;
+            }
+            newValue += pqScaled;
```
yazzaoui, line 138: I already mentioned it in the previous PR but to my understanding, the ACU is a geometric average not an arithmetic one? And if it was an arithmetic one we should probably divide this by the sum of the weights cf @fjribi @rtedwardchen

fjribi: @yazzaoui 

With the new approach, the theoretical optimal basket (Given equal preference objective function) is a geometric one.  However, arithmetic baskets are more suitable for our use case. Indeed, with arithmetic baskets, we can imply an ACU rate and ACU discount factors (under the Absence of Arbitrage AOA assumption) straightforward from the basket components' yield curves. This could be very useful when comparing the Auton borrowing rate or when pricing financial products. Therefore, a fitting step (variance minimization) is applied to get the arithmetic basket that best fits the optimal geometric one (in our case, uniform geometric mean). The resulting arithmetic basket of the fitting step is the retained basket for ACU determination. Sorry if it was confusing, this should be clarified in the documentation.

The retained ACU arithmetic basket is defined by the fixed quantities of each component currency rather than the weights (although it is possible to use weights, this requires a rebase step). Using fixed quantities, the value (exchange rate) of one unit of ACU  in USD terms is simply the sum of the values of each quantity of currency multiplied by its value in USD (the exchange rate in USD):

$$ACUValue_{inUSD}(t)=\sum_{Currencies} CurrencyQuanity \times CurrencyValue_{inUSD}(t)$$

aiman: Thanks @fjribi. Indeed the ACU is a fixed quantity arithmetic basket.

yazzaoui (approved):
LGTM

PR #810, opened 2023-08-02
fix oracle contract voter sort

lorenzo-dev1:
The oracle contract `_votersSort` function sorts the voters in descending address order, but the `_updateVotingInfo` function expects them sorted in ascending order. This PR fixes the sorting algo to sort ascending and adds some tests for it.

Tests were not failing before because the addresses generated by the truffle contract tests were not hitting this edge case.

### DIFF
--- a/autonity/solidity/contracts/Oracle.sol
+++ b/autonity/solidity/contracts/Oracle.sol
@@ -280,14 +280,13 @@ contract Oracle is IOracle {
     */
     function _votersSort(address[] memory _voters, int _low, int _high)
     internal pure {
+        if (_low >= _high) return;
         int _i = _low;
         int _j = _high;
-        if (_i == _j) return;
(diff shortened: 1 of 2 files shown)


autonity/solidity/contracts/Oracle.sol
```diff
@@ -280,14 +280,13 @@ contract Oracle is IOracle {
     */
     function _votersSort(address[] memory _voters, int _low, int _high)
     internal pure {
+        if (_low >= _high) return;
         int _i = _low;
         int _j = _high;
-        if (_i == _j) return;
```
yazzaoui, line 285: Wouldn't that make the sorting unstable?

yazzaoui (approved):
LGTM

vpiyush (approved):
LGTM!

PR #815, opened 2023-08-08
connect to all commmittee members, reagardless of maxPeers configuration

vpiyush:
AGC dials only to the subset of the committee members based on the maxPeers configuration(default: 50), but as per the current design, all committee members should be connected to each other. This PR fixes this.


yazzaoui (approved):
LGTM

Jason-Zhangxin-Chen (commented):
LGTM

PR #821, opened 2023-08-21
Protocol contracts coverage improvements

lorenzo-dev1:
Most of the protocol contracts (accountability, oracle, autonity) are tested using truffle.

The ASM is tested using the APE framework. The APE framework first run is longer because it needs to create some artifacts for each contract. On the CI, these artifacts are re-created at each workflow run. This could be improved by caching APE artifacts (https://github.com/actions/cache) between runs. As of now I did not implement it, since the `contract-test-fast` running time is still less wrt the one of the `integration-tests` and `byzantine-tests`.

fast running time local (APE artifacts cached): 5~7 mins
normal running time local (APE artifacts cached): 1h 20 mins 

fast running time ci (APE artifacts *not* cached): 17 ~ 25 mins
normal running time ci (APE artifacts *not* cached): 1h 30 mins


yazzaoui (approved):
LGTM. Great work!

.github/workflows/go.yml
```diff
@@ -177,6 +184,30 @@ jobs:
...
+      - name: Go cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ${{ env.GO_CACHE }}
+            ${{ env.GO_MODULES_CACHE }}
+          key: ${{ runner.os }}-gocache-${{ env.GO_VERSION }}${{ hashFiles('go.mod', 'go.sum') }}
+
+      - name: Run contract tests
+        run: make test-contracts-fast
```
vpiyush, line 210: test-contracts and test-contracts-fast both run asm-contracts. does this mean when slow and fast tests are to be run,  test-asm-contracts would run twice

lorenzo-dev1: yes, the idea is either you run the fast version or the normal:
- if you want fast feedback ---> `test-contracts-fast`
- if you want maximum correctness --> `test-contracts`

autonity/solidity/test/oracle.js
```diff
@@ -334,6 +337,7 @@ contract("Oracle", accounts => {
         for (let i = 0; i < round.voters.length; i++){
           const voter = round.voters[i];
           const commit = web3.utils.soliditySha3(...voter.pricesWithCommit);
+          //TODO(tariq) low priority. check that vote correctly trigger the update of the `reports` structure
```
vpiyush, line 340: Is this done yet? If yes, remove the TODO.

tbssajal: No, the low priority ones are not done yet. Maybe they can be done in a separate PR?

vpiyush (commented):
It would be a good idea to either tackle TODOs OR create a ticket that would track those if tackling them is low priority right now.

.github/workflows/go.yml
vpiyush: My question is when we push a PR which is not a draft. Does this make `test-contracts-asm` target run twice?

Jason-Zhangxin-Chen (approved):
LGTM, thanks so much for the tests.

lorenzo-dev1:
> It would be a good idea to either tackle TODOs OR create a ticket that would track those if tackling them is low priority right now.

Good idea, I created https://github.com/autonity/autonity/issues/848 so that they do not get lost.

.github/workflows/go.yml
lorenzo-dev1: ah I see good point. I will modify the CI so that the `test-contracts-asm` is run only once for non-draft PRs. I will leave the command `test-contracts` and `test-contracts-fast` intact however, to ensure that they run all tests.

lorenzo-dev1: https://github.com/autonity/autonity/pull/821/commits/78ac16b840a48711258ae8e71f5d69d5b2275188 should solve

vpiyush (approved):

PR #824, opened 2023-08-23
Improve coverage autonity

tbssajal:


autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -13,13 +13,42 @@ import "./Autonity.sol";
 contract AutonityTest is Autonity {
 
     constructor(Validator[] memory _validators,
-                Config memory _config) Autonity(_validators, _config) {
+                Config memory _config,
+                uint256 _unbodingPeriod) Autonity(_validators, _config) {
 
-        config.policy.unbondingPeriod = 0;
+        config.policy.unbondingPeriod = _unbodingPeriod;
```
lorenzo-dev1, line 19: I think this is redundant, we can set the `unbondingPeriod` in the `_config`

autonity/solidity/test/autonity.js
```diff
@@ -863,37 +1030,120 @@ contract('Autonity', function (accounts) {
...
+      assert.equal(validatorInfo.bondedStake, bondedStake - tokenUnBond, "validator bondedStake is not expected");
+      assert.equal((await valLiquid.lockedBalanceOf(newAccount)).toNumber(), 0, "LNTN not unlocked after epoch end");
+      assert.equal((await valLiquid.balanceOf(newAccount)).toNumber(), tokenMint - tokenUnBond, "LNTN not burned after epoch end");
+
+      // after unbonding period, at the next endEpoch the unbonding shares are converted back to NTNs and released.
+      let currentBalance = (await autonity.balanceOf(newAccount)).toNumber();
+      // for this assert to work, the following needs to be true:
+      // UnbondingPeriod > CurrentEpochStartHeight - UnbondingRequestBlock
+      assert.equal(currentBalance, balance, "NTN released before unbonding period");
+      let unbodingHeight = Number(unbondingRequest.requestBlock) + (await autonity.getUnbondingPeriod()).toNumber();
```
lorenzo-dev1, line 1136: I would suggest to be careful with these spelling mistakes, they are harmless in the short term but in the long term they make the code less maintainable

tbssajal: Updated, please check

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -15,11 +15,38 @@ contract AutonityTest is Autonity {
...
+   function getBondingRequest(uint256 _id) public view returns (BondingRequest memory) {
+        return bondingMap[_id];
+   }
+
+   function getUnbondingRequest(uint256 _id) public view returns (UnbondingRequest memory) {
+        return unbondingMap[_id];
+   }
+
+   function getFirstPendingBondingRequest() public view returns (uint256) {
+     require(tailBondingID < headBondingID, "No pending bonding requests");
```
lorenzo-dev1, line 33: Since this is a test contract that gives us improved visibility in the internal structures of the Autonity contract, I think it should not contain any additional logic (e.g. `require` operations). In my opinion it should just consist of getters for the single fields (e.g. `getHeadBondingId`, `getTailBondingId`, etc. etc.) and all the logic should live inside the truffle test cases.

autonity/solidity/test/autonity.js
```diff
@@ -803,40 +936,88 @@ contract('Autonity', function (accounts) {
...
+      let latestBondingReqId = validators.length;      
+      let bondingRequest = await autonity.getBondingRequest(latestBondingReqId);
+      assert.equal(bondingRequest.amount, tokenMint, "stake bonding amount is not expected");
+      assert.equal(bondingRequest.delegator, newAccount, "delegator addr is not expected");
+      assert.equal(bondingRequest.delegatee, validators[0].nodeAddress, "delegatee addr is not expected");
+      
 
-      assert.equal(latestBondingReqId, (await autonity.getLastRequestedBondingRequest()).toNumber())
+      // LNTN is minted to delegator at epoch end
+      let validatorINfo = await autonity.getValidator(validators[0].nodeAddress);
```
lorenzo-dev1, line 974:
```suggestion
      let validatorInfo = await autonity.getValidator(validators[0].nodeAddress);
```
same as spelling errors, they make the code less maintainable over time, especially if other people put hands on it

autonity/solidity/test/autonity.js
```diff
@@ -863,37 +1044,135 @@ contract('Autonity', function (accounts) {
       );
     });
 
-    //TODO(tariq) replicate this test for a non-selfBonded unbond request --> 
-    //                                                                        1. LNTN locked when unbonding request issued
-    //                                                                        2. LNTN burned at the end of the epoch following the unbonding request. Unbonding request becomes unlocked.
-    //                                                                        3. Unbonding shares issued at the end of the epoch
-    //                                                                        4. Unbonding shares converted to NTNs and released at the end of the unbonding period
     it('un-bond from a valid validator', async function () {
```
lorenzo-dev1, line 1047:
```suggestion
    it('un-bond from a valid validator (selfBonded)', async function () {
```

autonity/solidity/test/autonity.js
```diff
@@ -863,37 +1044,135 @@ contract('Autonity', function (accounts) {
...
+      // check effects of unbond (selfBonded):
+      // unbonded NTN enters "unbonding" state at epoch end and unbonding shares are issued. validator voting power (bondedStake) decreases
+      let validatorINfo = await autonity.getValidator(validators[0].nodeAddress);
+      let bondedStake = validatorINfo.bondedStake;
+      await endEpoch(autonity, operator, deployer);
+      unbondingRequest = await autonity.getUnbondingRequest(latestUnbondingReqId);
+      assert.equal(unbondingRequest.unbondingShare, tokenUnBond, "unbonding share is not expected");
+      assert.equal(unbondingRequest.unlocked, true, "unbonding not applied at epoch end");
+      validatorINfo = await autonity.getValidator(validators[0].nodeAddress);
+      assert.equal(validatorINfo.bondedStake, bondedStake - tokenUnBond, "validator bondedStake is not expected");
```
lorenzo-dev1, line 1077: It would be useful to assert also the `selfUnbondingStake` and the `selfUnbondingShares` fields.

autonity/solidity/test/autonity.js
```diff
@@ -863,37 +1044,135 @@ contract('Autonity', function (accounts) {
...
+      let validatorInfo = await autonity.getValidator(validator);
+      const valLiquid = await liquidContract.at(validatorInfo.liquidContract);
+      assert.equal((await valLiquid.lockedBalanceOf(newAccount)).toNumber(), tokenUnBond);
+
+      // LNTN burned at the end of the epoch. Unbonding request becomes unlocked.
+      // Unbonding shares issued at the end of the epoch. validator voting power (bondedStake) decreases
+      let bondedStake = validatorInfo.bondedStake;
+      await endEpoch(autonity, operator, deployer);
+      unbondingRequest = await autonity.getUnbondingRequest(latestUnbondingReqId);
+      assert.equal(unbondingRequest.unbondingShare, tokenUnBond, "unbonding share is not expected");
```
lorenzo-dev1, line 1145: Would be useful to check also the `validator.unbondingShares` and `validator.unbondingStake`

autonity/solidity/test/autonity.js
```diff
@@ -918,77 +1197,72 @@ contract('Autonity', function (accounts) {
...
+      assert.equal(unstaking.delegator, validators[0].treasury, "delegator addr is not expected");
+      assert.equal(unstaking.delegatee, validators[0].nodeAddress, "delegatee addr is not expected");
+      assert.equal(unstaking.unlocked, false, "pending unbonding request unlocked");
     });
+
     it('test unbonding shares logic', async function () {
-      /* TODO(tariq) issue multiple unbonding requests (both selfBonded and not) in different epochs
+      /* issue multiple unbonding requests (both selfBonded and not) in different epochs
        * and check that the unbonding shares related fields change accordingly. Relevant fields to check:
        * struct Validator {
```
lorenzo-dev1, line 1267: These comments about fields to check can be deleted, it was for instruction purposes

autonity/solidity/test/autonity.js
```diff
@@ -36,6 +36,137 @@ function generateMultiSig(nodekey,oraclekey,treasuryAddr) {
...
+  assert(
+    unbondingReleaseHeight > lastEpochBlock,
+    `unbonding period too short for testing, request-block: ${Number(lastRequest.requestBlock)}, unbonding-period: ${currentUnbondingPeriod}, `
+    + `last-epoch-block: ${lastEpochBlock}`
+  );
+  // otherwise NTN is already released at last epoch end
+
+  // check validator state after unbonding applied but before NTN is released
+  // total-unbond amount for each delegatee
+  let totalUnboded = tokenUnbond * treasuryAddrses.length;
```
lorenzo-dev1, line 106: spelling error
```suggestion
  let totalUnbonded = tokenUnbond * treasuryAddrses.length;
```

autonity/solidity/test/autonity.js
```diff
@@ -1006,7 +1280,37 @@ contract('Autonity', function (accounts) {
...
+      const tokenMint = 100;
+      const tokenUnbond = 10;
+      await bulkBondingRequest(autonity, operator, deployer, treasuryAddrses, delegatee, tokenMint);
+      // unbond some amount and check unbonding share
+      await checkUnbondingPhase(autonity, operator, deployer, treasuryAddrses, delegatee, tokenUnbond);
+      // again unbond some amount and check unbonding share
+      await checkUnbondingPhase(autonity, operator, deployer, treasuryAddrses, delegatee, tokenUnbond);
+
+      // repeat
+      await bulkBondingRequest(autonity, operator, deployer, treasuryAddrses, delegatee, tokenMint);
```
lorenzo-dev1, line 1302: What is the reasoning behind repeating the bulk bond-unbond multiple times?

autonity/solidity/test/autonity.js
cmjc: Should `treasuryAddrses` be `treasuryAddresses`?

autonity/solidity/test/autonity.js
```diff
@@ -1006,7 +1280,37 @@ contract('Autonity', function (accounts) {
...
+      for (let i = 0; i < validators.length; i++) {
+        treasuryAddrses.push(validators[i].treasury);
+        delegatee.push(validators[i].nodeAddress);
+      }
+
+      const tokenMint = 100;
+      const tokenUnbond = 10;
+      await bulkBondingRequest(autonity, operator, deployer, treasuryAddrses, delegatee, tokenMint);
+      // unbond some amount and check unbonding share
+      await checkUnbondingPhase(autonity, operator, deployer, treasuryAddrses, delegatee, tokenUnbond);
```
lorenzo-dev1, line 1297: I like the bulk approach, however I think there are two key aspects that are missing in this tests (or at least from my original idea of the test, we can alter it if we think about something better):
1. The delegator balance is not checked for released NTN (only the validator state is checked)
2. The unbonding requests are issued all in the same epoch (thus will all be dealt with in the same epoch)

In my opinion these two aspects are important to verify the correct functioning of the unbonding shares logic. For the shares logic to be correct, it means that a delegator which unbonds receives:
1. the right amount of NTN (which since there are no slashing here is always 1:1 to the amount he unbonded)
2. at the right time (which depends on the epoch in which they issue the unbonding)

I think that if we can structure this test with this key points in mind, the logic can then be re-used for these `protocol.js` test https://github.com/autonity/autonity/blob/de76e67e5b0399ed561e8691a80b13c1fbc56da3/autonity/solidity/test/protocol.js#L174 (unbonding shares logic testing with slashing)

autonity/solidity/test/autonity.js
tbssajal: kept the comments to understand what is done in case test is too long.

autonity/solidity/test/autonity.js
tbssajal: As we are unbonding less amount than what was bonded, I was just curious about the repetition. Should I remove it?

autonity/solidity/test/autonity.js
lorenzo-dev1: Yeah the comment itself it is good to leave, I am talking about the validator structure fields. I think those are quite big and not very useful

autonity/solidity/test/autonity.js
lorenzo-dev1: I think we can keep it unless it makes the test very slow

tbssajal: This test takes 30 seconds in my machine

autonity/solidity/test/autonity.js
tbssajal: updated

autonity/solidity/test/autonity.js
tbssajal: done

autonity/solidity/test/autonity.js
```diff
@@ -863,37 +1044,135 @@ contract('Autonity', function (accounts) {
...
+      // Unbonding shares issued at the end of the epoch. validator voting power (bondedStake) decreases
+      let bondedStake = validatorInfo.bondedStake;
+      await endEpoch(autonity, operator, deployer);
+      unbondingRequest = await autonity.getUnbondingRequest(latestUnbondingReqId);
+      assert.equal(unbondingRequest.unbondingShare, tokenUnBond, "unbonding share is not expected");
+      assert.equal(unbondingRequest.unlocked, true, "unbonding not applied at epoch end");
+      validatorInfo = await autonity.getValidator(validators[0].nodeAddress);
+      assert.equal(validatorInfo.bondedStake, bondedStake - tokenUnBond, "validator bondedStake is not expected");
+      assert.equal((await valLiquid.lockedBalanceOf(newAccount)).toNumber(), 0, "LNTN not unlocked after epoch end");
+      assert.equal((await valLiquid.balanceOf(newAccount)).toNumber(), tokenMint - tokenUnBond, "LNTN not burned after epoch end");
```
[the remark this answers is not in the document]
tbssajal: done

autonity/solidity/test/autonity.js
tbssajal: done

autonity/solidity/test/autonity.js
tbssajal: done

autonity/solidity/contracts/AutonityTest.sol
tbssajal: done

autonity/solidity/test/autonity.js
tbssajal: done

autonity/solidity/test/autonity.js
tbssajal: removed repetition. also capped validator/delegator count to 3

autonity/solidity/test/autonity.js
tbssajal: Added the suggested changes. Please check

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -15,11 +15,34 @@ contract AutonityTest is Autonity {
...
 
+   function getBondingRequest(uint256 _id) public view returns (BondingRequest memory) {
+        return bondingMap[_id];
+   }
+
+   function getUnbondingRequest(uint256 _id) public view returns (UnbondingRequest memory) {
+        return unbondingMap[_id];
+   }
+
+   function getFirstPendingBondingRequest() public view returns (uint256) {
```
lorenzo-dev1, line 32:
```suggestion
   function getTailBondingID() public view returns (uint256) {
```

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -15,11 +15,34 @@ contract AutonityTest is Autonity {
...
+     return tailBondingID;
+   }
+
+   function getHeadBondingID() public view returns (uint256) {
+     return headBondingID;
+   }
+
+   function getFirstPendingUnbondingRequest() public view returns (uint256) {
```
lorenzo-dev1, line 40:
```suggestion
   function getLastUnlockedUnbonding() public view returns (uint256) {
```

autonity/solidity/test/autonity.js
lorenzo-dev1: Ok I think it is good now. The LNTN locked and burned checks are a bit overkill in this test (we care only about shares <--> NTN conversion) but it's not a big deal. We can keep them for now (extra checks don't hurt) but if the test gets too slow they can be removed. Resolving 👍🏻

lorenzo-dev1 (approved):
LGTM

autonity/solidity/test/autonity.js
```diff
@@ -36,6 +36,197 @@ function generateMultiSig(nodekey,oraclekey,treasuryAddr) {
...
+
+  // NTN should be released
+  for (let i = 0; i < treasuryAddresses.length; i++) {
+    assert.equal((await autonity.balanceOf(treasuryAddresses[i])).toNumber(), balanceNTN[i] + totalUnbonded)
+  }
+}
+
+// following functions check validator-info and unbonding-request-info in several phases
+// as no slashing occurs, everything is issued 1:1
+function checkValInfoAfterBonding(valInfo, expectedValInfo, selfBonded, totalBonded) {
```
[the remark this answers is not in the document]
yazzaoui: I totally agree with that @lorenzo-dev1, It's an important issue we currently have with our tendermint testing suite.

PR #825, opened 2023-08-24
Fix autonity unbonding

tbssajal:
Fixed 2 bugs:
- `_validator.selfBondedStake` should decrease in case of self-unbonding at `_applyUnbonding`
- In case of non-self-unbonding-request with `amount = 0` and no prior delegation, the unbonding-request still registers into unbonding-map. As a result, at `_applyUnbonding` in epoch-end, `_newtonAmount` is calculated assuming total `_validator.liquidSupply` is positive and gets error because of division by 0. In case this error is triggered, epoch is never progressed.

### DIFF
--- a/autonity/solidity/contracts/Autonity.sol
+++ b/autonity/solidity/contracts/Autonity.sol
@@ -321,8 +321,6 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     */
     function unbond(address _validator, uint256 _amount) public {
         require(validators[_validator].nodeAddress == _validator, "validator not registered");
-        // in case _amount = 0, it is harmless, yet can trigger contract to fail in end-epoch phase
-        // in some limited case
         require(_amount > 0, "unbonding amount is 0");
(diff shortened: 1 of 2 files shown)


lorenzo-dev1 (approved):
LGTM

autonity/solidity/contracts/Autonity.sol
```diff
@@ -321,6 +321,9 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     */
     function unbond(address _validator, uint256 _amount) public {
         require(validators[_validator].nodeAddress == _validator, "validator not registered");
+        // in case _amount = 0, it is harmless, yet can trigger contract to fail in end-epoch phase
+        // in some limited case
```
lorenzo-dev1, line 325:
```suggestion
```

Jason-Zhangxin-Chen (approved):
LGTM

yazzaoui (approved):
LGTM let's get that merged

PR #827, opened 2023-08-30
Make Solidity licenses consistent with repo

aiman:
Make the license for Solidity code match the same LGPLv3 license of the repository itself. Also, some references to internal docs are removed.


aiman:
FYI it seems I missed a few source files, so I have amended the PR.

lorenzo-dev1 (approved):

PR #829, opened 2023-08-31
Fix self unbonding

tbssajal:
Fixed the following bugs:
- When requesting for self-unbonding, we don't check if the total requested self-unbonding amount exceeds `_validator.selfBondedStake`
- When applying undonding, in case of self-unbonding, we don't check if the `_newtonAmount` exceeds `_validator.selfBondedStake` which can happen due to slashing


yazzaoui:
IMO it's not a big problem if the requested amount is greater than what is available when performing the initial request because it's understood that it can be reduced when slashing happens. It would save us from introducing a new state variable in the validator object.

tbssajal:
> IMO it's not a big problem if the requested amount is greater than what is available when performing the initial request because it's understood that it can be reduced when slashing happens. It would save us from introducing a new state variable in the validator object.

Yeah, it can save some storage cost this way. Then I will remove the check on _unbond function.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1081,7 +1084,11 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
             require(liqBalance >= _amount, "insufficient unlocked Liquid Newton balance");
             _validator.liquidContract.lock(_recipient, _amount);
         } else {
-            require(_validator.selfBondedStake >= _amount, "insufficient self bonded newton balance");
+            require(
+                _validator.selfBondedStake > _validator.selfUnbondingStakeLocked && _validator.selfBondedStake - _validator.selfUnbondingStakeLocked >= _amount,
```
lorenzo-dev1, line 1088: I think that the first condition `_validator.selfBondedStake > _validator.selfUnbondingStakeLocked` is redundant (assuming that PR #825 gets merged). This is because since we are sure that `_amount > 0`, if `_validator.selfBondedStake - _validator.selfUnbondingStakeLocked <= 0` the require will fail.

autonity/solidity/test/autonity.js
```diff
@@ -995,6 +996,92 @@ contract('Autonity', function (accounts) {
...
+      let tx = await autonity.unbond(validator, tokenUnbond, {from: treasury});
+      truffleAssert.eventEmitted(tx, 'NewUnbondingRequest', (ev) => {
+        return ev.validator === validator && ev.delegator === treasury && ev.selfBonded === true && ev.amount.toNumber() === tokenUnbond
+      }, 'should emit newUnbondingRequest event');
+
+      // if the following does not fail, then we will have panic error in epoch end due to arithmetic underflow
+      await truffleAssert.fails(
+        autonity.unbond(validator, tokenUnbond, {from: treasury}),
+        truffleAssert.ErrorType.REVERT,
+        ""
```
lorenzo-dev1, line 1025:
```suggestion
        "insufficient self bonded newton balance"
```

autonity/solidity/contracts/Autonity.sol
tbssajal: I could be wrong but I think if `_validator.selfBondedStake - _validator.selfUnbondingStakeLocked <= 0` then the `_validator.selfBondedStake - _validator.selfUnbondingStakeLocked` would be very big uint instead of negative number, because I don't see SafeMath here. That is why I put it.

lorenzo-dev1:
> IMO it's not a big problem if the requested amount is greater than what is available when performing the initial request because it's understood that it can be reduced when slashing happens. It would save us from introducing a new state variable in the validator object.

I suggest this approach because imo it is cleaner to have only "validated" unbonding requests in the queue. Then the amount of releases NTN can vary due to slashing but at least we know that in the queue we have only valid requests

autonity/solidity/test/autonity.js
tbssajal: This is odd. I don't understand why the test would pass.

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: Yeah you are right, it would cause underflow. we need this check

tbssajal:
> > IMO it's not a big problem if the requested amount is greater than what is available when performing the initial request because it's understood that it can be reduced when slashing happens. It would save us from introducing a new state variable in the validator object.
> 
> I suggest this approach because imo it is cleaner to have only "validated" unbonding requests in the queue. Then the amount of releases NTN can vary due to slashing but at least we know that in the queue we have only valid requests. With no check it could be filled up with bogus unbonding requests

@yazzaoui please comment if it is preferred to keep the check for clarity even if it costs a little storage cost. Then I will keep the check. I also prefer to keep the extra check.

autonity/solidity/test/autonity.js
lorenzo-dev1: I think it gets interpreted as all reverts are fine (the reason is not checked)

tbssajal:
> 

Updated, please check

tbssajal:
> We can keep the extra check. I was more concerned about the additional complexity than the storage cost but the logic looks quite straightforward.
> 
> BTW, in this case, you will have to regenerate the internal bindings/hardcoded bytecode with `make contracts` and fix the validator genesis initialization code ( check `params/protocol_contracts.go`) .

tbssajal:
> We can keep the extra check. I was more concerned about the additional complexity than the storage cost but the logic looks quite straightforward.
> 
> BTW, in this case, you will have to regenerate the internal bindings/hardcoded bytecode with `make contracts` and fix the validator genesis initialization code ( check `params/protocol_contracts.go`) .

Updated, please check

autonity/solidity/contracts/Autonity.sol
yazzaoui: SafeMath is no longer necessary since solidity 0.8.0 https://docs.soliditylang.org/en/v0.8.19/080-breaking-changes.html

tbssajal: Great! That means `require(_validator.selfBondedStake - _validator.selfUnbondingStakeLocked >= _amount)` is enough. `_validator.selfBondedStake > _validator.selfUnbondingStakeLocked` can be removed.

tbssajal: updated

lorenzo-dev1: From the like it seems like underflow and overflow still reverts:
> Arithmetic operations revert on underflow and overflow. You can use unchecked { ... } to use the previous wrapping behaviour.

tbssajal: Sorry I did not get it. If underflow reverts then it is good right? Like if it is false in require, then it reverts.

tbssajal: Do you mean the underflow will give a big uint256 instead of arithmetic error? I think it will give arithmetic error because I think I encountered something similar in Autonity.sol

tbssajal: I think it is fine.

lorenzo-dev1 (approved):

PR #830, opened 2023-09-04
Change the Oracle Contract symbol format

aiman:
Using slash character in the symbol creates problems for users, because the symbol cannot be used directly in URLs or UNIX filenames, which don't permit slashes without additional quoting.

This commit changes the divider to a hyphen symbol (45 in ASCII), which is another conventional divider that is commonly used.

This is better than no divider, because it has the advantage of being unambiguous. If there was no divider, it is possible to have a situation where you cannot identify the two assets in a pair without relying on additional information that is outside the symbol.

This pull request also has minor improvements to Makefile and the ASM.


yazzaoui (approved):
LGTM

autonity/solidity/contracts/asm/Stabilization.sol
```diff
@@ -441,22 +448,11 @@ contract Stabilization is IStabilization {
...
-
-    /// Determine if the CDP is liquidatable at the given timestamp.
-    ///
-    /// The timestamp must be equal or later than the time of the CDP last
-    /// borrow or repayment.
-    /// @return Whether the CDP is liquidatable
-    function isLiquidatable(
-        address account,
-        uint timestamp
-    ) external view goodTime(account, timestamp) returns (bool) {
```
lorenzo-dev1, line 457: is it fine that we lose the `goodTime` check?

lorenzo-dev1: Actually reviewing more the code, is it possible for `goodTime` to ever fail? 

Since it checks if current `block.timestamp` against the timestamp of the last `deposit`/`borrow`/`repay`/`liquidate` operation for that cdp, the >= relationship will always be satisfied (unless we have `block(n+1).timestamp` < `block(n).timestamp`, which should not get approved by consensus).

aiman: I think the way the diff ended up may be creating some confusion. There were two variants of the `isLiquidatable()` function. The main one is logically unchanged, but it relied on the logic of the removed one.

The `goodTime` check isn't relevant for the main `isLiquidatable()` function. It was only relevant for the variant that is removed. In this deprecated variant, the user can specify a timestamp to perform the check, and we wanted to make sure the user-supplied timestamp was not earlier than the last debt valuation timestamp. It is theoretically possible to request a timestamp before the last borrow or repayment, so it could fail. But as described in 48c0e78c6dfb31da40ea4ba9d012d6c2f1c7a8df, this function variant makes no sense unless we are able to fetch collateral prices to match the user-requested timestamp. So it is removed.

The `goodTime` modifier remains in use by the `debtAmount()` function so we don't lose it.

aiman:
I have rebased on develop 82ac1788814197d0a252b045713769ffbb356bab in order to incorporate the fixes in #825.

autonity/solidity/contracts/asm/Stabilization.sol
lorenzo-dev1: Ahh ok, I see. Thanks for clarifying.

lorenzo-dev1 (approved):

PR #831, opened 2023-09-04
Improve coverage protocol

tbssajal:


autonity/solidity/test/protocol.js
```diff
@@ -167,72 +344,290 @@ contract('Protocol', function (accounts) {
...
-          * Then slash of 50 NTNs. Now:
-          *
-          * unbondingShares = 100
-          * unbondingStake = 50
-          * exchange ratio 2 share : 1 NTN
-          *
-          * The value of the share decreased, as you now need 2 shares to redeem 1 NTN
-          */
+      // issue multiple unbonding requests (both selfBonded and not) in different epochs, interleaved with slashing events
+      // and check that the unbonding shares related fields change accordingly. Relevant fields to check:
```
lorenzo-dev1, line 357:
```suggestion
      // and check that the unbonding shares related fields change accordingly.
```

autonity/solidity/test/protocol.js
```diff
@@ -3,10 +3,187 @@ const assert = require('assert');
...
+  truffleAssert.eventEmitted(tx, 'SlashingEvent', (ev) => {
+    txEvent = ev;
+    return ev.amount.toNumber() > 0;
+  });
+  return txEvent;
+}
+
+// only non-self-unbond
+// make sure selfBondedStake = 0 so that slashing can be applied to unbondingStake and delegatedStake
+async function slashAndUnbond(autonity, accountability, delegators, validator, tokenUnbond, operator, deployer) {
```
lorenzo-dev1, line 42:
```suggestion
async function unbondAndSlash(autonity, accountability, delegators, validator, tokenUnbond, operator, deployer) {
```

autonity/solidity/test/protocol.js
```diff
@@ -3,10 +3,187 @@ const assert = require('assert');
...
+    + `last-epoch-block: ${lastEpochBlock}`
+  );
+  // mine blocks until unbonding period is reached
+  while (await web3.eth.getBlockNumber() < unbondingReleaseHeight) {
+    await utils.mineEmptyBlock();
+  }
+  // release NTN
+  await utils.endEpoch(autonity, operator, deployer);
+  for (let i = 0; i < delegators.length; i++) {
+    let balanceIncrease = unbondingStakes * Number(requests[i].unbondingShare) / unbondingShares;
```
lorenzo-dev1, line 105: I think it would be worth to add an assert to verify that `tokenUnbond` - `balanceIncrease` = the amount we expect to be slashed (10% in case of mid severity)

autonity/solidity/test/protocol.js
```diff
@@ -3,10 +3,187 @@ const assert = require('assert');
...
+    balances.push((await autonity.balanceOf(delegators[i])).toNumber());
+  }
+  let valInfo = await autonity.getValidator(validator);
+  let liquidSupply = Number(valInfo.liquidSupply);
+  let delegatedStakes = Number(valInfo.bondedStake) - Number(valInfo.selfBondedStake);
+  let totalCurrentStake = delegatedStakes;
+  let delegatee = [];
+  delegatee.push(validator);
+  let requestID = (await autonity.getHeadUnbondingID()).toNumber();
+  let count = await utils.bulkUnbondingRequest(autonity, delegators, delegatee, tokenUnbond);
```
lorenzo-dev1, line 55: Unless it is too much work to change, I think it would be useful to make every delegator unbond a different amount. That way we can verify that the ones which unbonded more should be impacted more by the slash (in absolute terms of NTNs, percentage should always be 10% for mid severity)

autonity/solidity/test/protocol.js
```diff
@@ -167,72 +344,290 @@ contract('Protocol', function (accounts) {
...
-      beforeEach(async function () {
-        autonity = await utils.deployContracts(validators, autonityConfig, accountabilityConfig,  deployer, operator);
-      });
+    beforeEach(async function () {
+      autonity = await utils.deployAutonityTestContract(validators, autonityConfig, accountabilityConfig,  deployer, operator);
+      accountability = await AccountabilityTest.new(autonity.address, accountabilityConfig, {from: deployer});
+      await autonity.setAccountabilityContract(accountability.address, {from:operator});
+    });
+
     it('unbondingShares:unbondingStake conversion ratio', async function () {
```
lorenzo-dev1, line 355: I think we need to add a similar test also on `selfUnbondingShare:selfUnbondingStake` conversion ratio.

autonity/solidity/test/protocol.js
```diff
@@ -167,72 +344,290 @@ contract('Protocol', function (accounts) {
...
+      await utils.endEpoch(autonity, operator, deployer);
+      // request unbonding and slash
+      await slashAndUnbond(autonity, accountability, delegators, validator, tokenUnbond, operator, deployer);
+      // repeat
+      await slashAndUnbond(autonity, accountability, delegators, validator, tokenUnbond, operator, deployer);
     });
+
     it.skip('unbondingShares:unbondingStake 100% slash edge case', async function () {
       // see https://github.com/autonity/autonity/issues/819
       // low priority for now as we already now that the problem is there, however we need a test for the future fix
```
lorenzo-dev1, line 402:
```suggestion
```

autonity/solidity/test/protocol.js
```diff
@@ -167,72 +344,290 @@ contract('Protocol', function (accounts) {
...
+        delegators.push(validators[i].treasury);
+      }
+      let validator = validators[maxCount].nodeAddress;
+      let treasury = validators[maxCount].treasury;
+      delegatee.push(validator);
+      let valInfo = await autonity.getValidator(validator);
+      // so slashing applies to delegated stake and unbonding stake
+      await autonity.unbond(validator, Number(valInfo.selfBondedStake), {from: treasury});
+
+      let requestID = (await autonity.getHeadUnbondingID()).toNumber() - 1;
```
lorenzo-dev1, line 487: This piece of code to wait for unbonding is repeated multiple times, it might be worth to convert it into a function

autonity/solidity/test/protocol.js
```diff
@@ -3,10 +3,187 @@ const assert = require('assert');
...
+    let balanceIncrease = unbondingStakes * Number(requests[i].unbondingShare) / unbondingShares;
+    assert.equal((await autonity.balanceOf(delegators[i])).toNumber(), balances[i] + balanceIncrease, "unexpected balance");
+  }
+
+}
+
+
+// only non-self-unbond
+// make sure selfBondedStake = 0 so that slashing can be applied to delegatedStake
+async function slashAndBond(autonity, accountability, delegators, validator, tokenBond, tokenUnbond, operator, deployer) {
```
lorenzo-dev1, line 114:
```suggestion
async function bondAndSlash(autonity, accountability, delegators, validator, tokenBond, tokenUnbond, operator, deployer) {
```

autonity/solidity/test/protocol.js
```diff
@@ -3,10 +3,187 @@ const assert = require('assert');
...
+  let unbondingStakes = Number(valInfo.unbondingStake);
+  let unbondingShares = Number(valInfo.unbondingShares);
+  assert(delegatedStakes > 0, "100% slashing");
+  let requestID = (await autonity.getHeadUnbondingID()).toNumber();
+  let count = await utils.bulkUnbondingRequest(autonity, delegators, delegatee, tokenUnbond);
+  // let unbonding apply
+  await utils.endEpoch(autonity, operator, deployer);
+  while (count > 0) {
+    let request = await autonity.getUnbondingRequest(requestID);
+    let newtonAmount = Math.floor(tokenUnbond * delegatedStakes / liquidSupply);
```
lorenzo-dev1, line 166: Also here I think it would be worth to check that the change in conversion rate matches with the amount that has been slashed. similarly to https://github.com/autonity/autonity/pull/831/files#r1316917902

autonity/solidity/test/protocol.js
```diff
@@ -167,72 +344,290 @@ contract('Protocol', function (accounts) {
...
+      await utils.endEpoch(autonity, operator, deployer);
+      const tokenBond = 100000000;
+      const factor = 1000;
+      // request bonding and slash
+      let releaseBlock = await slashAndBond(autonity, accountability, delegators, validator, tokenBond, factor, operator, deployer);
+      while (await web3.eth.getBlockNumber() < releaseBlock) {
+        await utils.mineEmptyBlock();
+      }
+      await autonity.activateValidator(validator, {from: treasury});
+      // repeat
```
lorenzo-dev1, line 505: I would wait that all the previous unbonding requests are dealt with (NTN released) so that the 2nd slashing does not affect the `shares:unbondingstake` conversion ratio. It is better to keep it 1:1 to ensure test robustness. This test should only interact with the `NTN:LNTN` ratio

autonity/solidity/test/protocol.js
```diff
@@ -167,72 +344,290 @@ contract('Protocol', function (accounts) {
...
+      await slashAndBond(autonity, accountability, delegators, validator, tokenBond, factor, operator, deployer);
     });
+
     it.skip('LNTN:NTN 100% slash edge case', async function () {
       // see https://github.com/autonity/autonity/issues/819
       // low priority for now as we already now that the problem is there, however we need a test for the future fix
```
lorenzo-dev1, line 511:
```suggestion
```

tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
```diff
@@ -14,6 +14,18 @@ const ValidatorState = {
   jailed : 2
 }
 
+const midSeveritySlashingRatePercent = 10;
+
+async function modifiedSlashingFeeAccountability(autonity, accountabilityConfig, operator, deployer) {
```
lorenzo-dev1, line 19:
```suggestion
async function customSlashingRate(autonity, accountabilityConfig, operator, deployer) {
```
No need to specify it is for accountability, slashing is part of accountability.

lorenzo-dev1: I just saw it returns accountability, it makes sense then.

autonity/solidity/test/protocol.js
```diff
@@ -84,25 +146,17 @@ async function slashAndUnbond(autonity, accountability, delegators, validator, t
...
-  );
-  // mine blocks until unbonding period is reached
-  while (await web3.eth.getBlockNumber() < unbondingReleaseHeight) {
-    await utils.mineEmptyBlock();
-  }
+  await utils.mineTillUnbondingRelease(autonity, operator, deployer, false);
   // release NTN
   await utils.endEpoch(autonity, operator, deployer);
+  let factor = 1;
+  while (slashCount > 0) {
```
lorenzo-dev1, line 153: I am probably missing something, but I do not understand why we need this loop

tbssajal: slashCount represents the total count of slash till now

lorenzo-dev1: yes but that should not affect the slashing rate right (since the accountability is re-deployed with historyFactor = 0)

lorenzo-dev1: This factor is then multiplied by the token unbond, but the previous unbonding are already finished. So the current unbonds should not be affected by previous slashes?

tbssajal: Actually the previous slashing is affecting both delegated stake and unbonding stake. That's why it affects. Because of the delegated stake

lorenzo-dev1: Ok the delegated stake should be affected twice, but not the unbonding stake imho

autonity/solidity/test/protocol.js
```diff
@@ -3,10 +3,257 @@ const assert = require('assert');
...
 
 // testing protocol contracts interactions.
 
+const ValidatorState = {
+  active : 0,
+  paused : 1,
+  jailed : 2
+}
+
+const midSeveritySlashingRatePercent = 10;
```
lorenzo-dev1, line 17: I think instead of a constant we can use something like the `ruleToRate` function present in `accountability.js` (which could be moved to utils). So that once we implement the actual different rates in the contract, the tests can be easily adapted if needed.

autonity/solidity/test/protocol.js
lorenzo-dev1: solved by the comment

autonity/solidity/test/protocol.js
tbssajal: done

lorenzo-dev1 (approved):

PR #838, opened 2023-09-12
Improve coverage accountability

tbssajal:


autonity/solidity/test/accountability.js
```diff
@@ -705,5 +705,82 @@ contract('Accountability', function (accounts) {
...
+        "block": 10,
+        "epoch": 0,
+        "reportingBlock": 11,
+        "messageHash": 0,
+      }
+      let balance = web3.utils.toWei("10", "ether");
+      await web3.eth.sendTransaction({from: validators[0].treasury, to: reporter, value: balance});
+      
+      await truffleAssert.fails(
+        accountability.handleEvent.call(event, {from: reporter}),
```
lorenzo-dev1, line 763: Can you add a comment explaining why we are using call here for future readers

autonity/solidity/test/accountability.js
```diff
@@ -705,5 +705,82 @@ contract('Accountability', function (accounts) {
...
+      let eventCount = 3;
+      let currentProof = "0x";
+      for (let i = 0; i < eventCount; i++) {
+        let rawProof = [];
+        rawProof.push(i);
+        event.chunkId = i;
+        event.rawProof = rawProof;
+        let request = (await accountability.handleEvent.request(event, {from: reporter}));
+        let receipt = await utils.signAndSendTransaction(reporter, accountability.address, reporterPrivateKey, request);
+        assert.equal(receipt.status, true, "transcation failed");
```
lorenzo-dev1, line 777:
```suggestion
        assert.equal(receipt.status, true, "transaction failed");
```

autonity/solidity/test/accountability.js
```diff
@@ -705,5 +705,82 @@ contract('Accountability', function (accounts) {
...
+        let currentEvent = await accountability.getReporterChunksMap({from: reporter});
+        let hexNumber = (i > 15) ? i.toString(16) : "0" + i.toString(16);
+        currentProof = currentProof + hexNumber;
+        assert.equal(currentEvent.rawProof, currentProof, "proof mismatch");
+      }
```
lorenzo-dev1, line 782: Can you add one more call to _handleEvent? it should revert with the proof verification failure. It is to check that the `readyToProcess` logic works

tbssajal:
> Could you add a test case also for `require(_event.reporter == msg.sender, "event reporter must be caller");`

Added

autonity/solidity/test/accountability.js
```diff
@@ -705,5 +710,265 @@ contract('Accountability', function (accounts) {
...
+        await truffleAssert.fails(
+          accountability.handleEvent.call(event, {from: reporter[i]}),
+          truffleAssert.ErrorType.REVERT,
+          "failed proof verification"
+        );
+      }
+
+    });
+
+    it.skip("has proper value in fields in storage", async function () {
```
lorenzo-dev1, line 942:
```suggestion
    it.skip("can start with chunkId = 1", async function () {
```

This name is more representative, unless I am missing something?

autonity/solidity/test/accountability.js
```diff
@@ -705,5 +710,265 @@ contract('Accountability', function (accounts) {
...
+        "messageHash": 0,
+      };
+
+      let currentProof = "0x";
+      let eventCount = event.chunks - 1;
+      for (let chunkId = 0; chunkId < eventCount; chunkId++) {
+        let rawProof = [];
+        rawProof.push(chunkId);
+        event.chunkId = chunkId;
+        event.rawProof = rawProof;
```
lorenzo-dev1, line 913: would be better to have different rawproofs for the 2 different validators, so we are sure that there are no interferences

autonity/solidity/test/accountability.js
lorenzo-dev1, line 942: what is the purpose of this test? Checking the start chunkid?

tbssajal: Yes, I will update the description. Depending on soln, we may need to delete the test.

lorenzo-dev1 (approved):
LGTM

PR #839, opened 2023-09-12
Ac cache

lorenzo-dev1:


autonity/autonity.go
```diff
@@ -91,30 +129,82 @@ func NewProtocolContracts(config *params.ChainConfig, db ethdb.Database, provide
...
+
+func (c *Cache) Stop() {
+	c.quit <- struct{}{}
+	c.wg.Wait()
+}
+
+func (c *Cache) MinimumBaseFee() *big.Int {
+	defer c.RUnlock()
+	c.RLock()
+	return c.minBaseFee
```
vpiyush, line 198: We can get rid of the lock by using an atomic pointer for minBaseFee.. It will ease  the contention

autonity/autonity.go
```diff
@@ -69,6 +70,7 @@ func NewEVMContract(evmProvider EVMProvider, contractABI *abi.ABI, db ethdb.Data
 //revive:disable:exported - Autonity is one of the contracts, so repetitive naming here is justified
 type AutonityContract struct {
 	EVMContract
+	*AutonityFilterer // allows to watch for Autonity Contract events
```
Jason-Zhangxin-Chen, line 108: ACEventFilter might be more meaningful in my view.

autonity/autonity.go
```diff
@@ -67,6 +68,35 @@ func NewEVMContract(evmProvider EVMProvider, contractABI *abi.ABI, db ethdb.Data
 	}
 }
 
+type Cache struct {
```
Jason-Zhangxin-Chen, line 71: I think Cache is a too general naming, it should be more specific.

core/blockchain.go
```diff
@@ -440,6 +446,17 @@ func NewBlockChain(db ethdb.Database,
 	return bc, nil
 }
 
+// TODO(lorenzo) to remove, debug purposes
+func (bc *BlockChain) Log() {
```
Jason-Zhangxin-Chen: This should be removed or you can add a termination break thus when the blockchain.Stop() function can stop this routine.

lorenzo-dev1: I already removed that in the most recent commit

autonity/autonity.go
lorenzo-dev1: That type is defined in the auto-generated bindings (https://github.com/autonity/autonity/blob/ac-cache/autonity/bindings.go#L2226) so I cannot change its name

autonity/autonity.go
lorenzo-dev1: I named it `Cache` only because it is in the `autonity` package. So the fully qualified name becomes `autonity.Cache`.

Also it is part of the `ProtocolContracts` struct, so adding the `Protocol`- prefix I think it is redundant.

Any suggestions?

autonity/autonity.go
```diff
@@ -91,30 +129,78 @@ func NewProtocolContracts(config *params.ChainConfig, db ethdb.Database, provide
...
+	}
+}
+
+func (c *Cache) Stop() {
+	close(c.quit)
+	<-c.done
+}
+
+func (c *Cache) MinimumBaseFee() *big.Int {
+	return c.minBaseFee.Load()
```
vpiyush, line 194: I don't think we should be returning the pointer to an actual value in the cache. This makes the value mutable from the caller. We should create a copy of the big.Int and return it.

lorenzo-dev1: solved with https://github.com/autonity/autonity/pull/839/commits/02ec5e9c5afd1c527ef9b106b31268d898fb9f77

vpiyush (approved):

yazzaoui (approved):

Issue #835, opened 2023-09-07
Snappy compression performance assessment


yazzaoui:
Snappy compression although designed to be fast and efficient has a significant overhead as shown by the performance profiling report. 
It is unclear if the compression ratio is high enough to be worthy for the kind of data exchange by an autonity node. 

The task would be to assess the bandwidth gains with `snappy.Encode` by monitoring the input and output data size.

https://github.com/autonity/autonity/blob/2f73948bea1c9bb030381fa39127188df5f270c1/p2p/rlpx/rlpx.go#L216-L226

vpiyush:
Snappy Compression results in various lengths of byte arrays.

| Size Before compression   |  Size after Compression  |
|------------------------------------ | ----------------------------------|
| 36 |   38|
122 | 125
197 | 201
1170 | 887
1797 | 1403
2535 | 2006
76807 | 64251

From the above results it’s clear that compression is not effective on smaller buffer sizes, Instead of compressing it adds a few extra bytes. Whereas for the large size buffers (>1k) It does seem to reduce the size **~20%**

We ran a load test on a 40-node setup with 480 TPS.


| |Max CPU | Avg CPU | N/w Ingress(avg) | N/w Egress(avg) |
| ------------- | ------------ | ----------------------- | ---------------------- | ---- |
| With Snappy | 43.87% | 38.35% | 5.38 Mb | 5.65 Mb |
| W/O Snappy | 42.25% | 36.75% | 5.72 Mb | 6.09 Mb |


The same test was then run on an optimized snappy code, There are two major optimizations:
- There is no snappy encoding done on packets smaller than 512 bytes
- The length of the buffer being used for snappy encoding was adjusted before encoding to the actual length needed for encoding.

| | Max CPU | Avg CPU | N/w Ingress(avg) | N/w Egress(avg) |
|-|------------- | ------------ | ----------------------- | ---------------------- | 
| Snappy optimised | 39.25% | 32% | 5.2 Mb | 5.06 Mb |

Here is the CPU profile of all three versions of the code:

Snappy-Enabled
![Snappy-Enabled](https://github.com/autonity/autonity/assets/18163958/6c2f8d31-bf5c-4a7b-b61f-52096955029d)

Snappy-Disabled
![Snappy-Disabled](https://github.com/autonity/autonity/assets/18163958/25229afa-c808-4d8e-a16e-f15f2bcbd3c3)

Snappy-optimized
[...]

PR #843, opened 2023-09-27
Snappy optimized

vpiyush:
Snappy optimizations. For more details refer [here](https://github.com/autonity/autonity/issues/835).


p2p/rlpx/buffer.go
```diff
@@ -117,11 +117,11 @@ func putUint24(v uint32, b []byte) {
 // growslice ensures b has the wanted length by either expanding it to its capacity
 // or allocating a new slice if b has insufficient capacity.
 func growslice(b []byte, wantLength int) []byte {
-	if len(b) >= wantLength {
+	if len(b) == wantLength {
 		return b
 	}
-	if cap(b) >= wantLength {
-		return b[:cap(b)]
+	if len(b) > wantLength || cap(b) >= wantLength {
```
lorenzo-dev1, line 123: I think the first two ifs can be substituted simply with
```
if len(b) >= wantLength {
   return b[:wantLength]
}
```

in any case cap(b) will always be >= wantLength, right?

lorenzo-dev1, line 123: I think we can merge the first if into this one with the condition `len(b) >= wantLength`

p2p/rlpx/rlpx.go
```diff
@@ -144,7 +144,7 @@ func (c *Conn) Read() (code uint64, data []byte, wireSize int, err error) {
 	wireSize = len(data)
 
 	// If snappy is enabled, verify and decompress message.
-	if c.snappyReadBuffer != nil {
+	if c.snappyReadBuffer != nil && wireSize > minSizeToCompress {
```
lorenzo-dev1, line 147: Isn't it possible that a frame whose original size is > 512 bytes, gets compressed to < 512 bytes and therefore it won't be uncompressed on the receiving end?

For example you have on the sending size a packet of size 513 bytes, it gets compressed to 511 bytes, and on the receiving end it does not get uncompressed (wrongfully)

p2p/rlpx/buffer.go
```diff
@@ -117,11 +117,11 @@ func putUint24(v uint32, b []byte) {
...
+		return b[:wantLength]
```
lorenzo-dev1, line 124: I am most probably missing something related to golang here, but how does this help with allocations? 

Given that the capacity of the slice remains the same, the amount of memory allocated by the process should remain the same regardless of whether we change the length to `cap(b)` or `wantLenght`?

p2p/rlpx/rlpx.go
vpiyush: This is a very good point. I think this might happen.

p2p/rlpx/buffer.go
vpiyush: This change doesn't impact the deallocation at this point. But this is a general optimization that would aid in scenarios where we copy this slice OR if there are further allocations dependent on the length of this slice. Snappy encode calls an ASM implementation though which may OR may not benefit from this change. But this doesn't hurt anyway.

lorenzo-dev1: My comment was advocating for keeping only `cap(b) > wantLength` . This way we avoid to make again if we have sufficient capacity already.

This works as well though

p2p/rlpx/rlpx.go
```diff
@@ -238,7 +242,8 @@ func (h *sessionState) writeFrame(conn io.Writer, code uint64, data []byte) erro
 	}
 	header := h.wbuf.appendZero(16)
 	putUint24(uint32(fsize), header)
-	copy(header[3:], zeroHeader)
+	header[3] = snappyByte
```
lorenzo-dev1, line 245: Just verifying my understanding, the header is of fixed 16 bytes and we can get away with adding a snappyByte into it because:
1.   zeroHeader is just 3 bytes (`zeroHeader = []byte{0xC2, 0x80, 0x80}`)
2.  The rest of the header would be just 0x00s?

vpiyush: yes. That's right.

p2p/rlpx/buffer.go
vpiyush: Actually, you are right. This is what I intended. The only update should be to reduce the length when it is more than the wanted length.

lorenzo-dev1 (approved):

p2p/rlpx/buffer.go
```diff
@@ -117,11 +117,11 @@ func putUint24(v uint32, b []byte) {
...
 func growslice(b []byte, wantLength int) []byte {
-	if len(b) >= wantLength {
+	if len(b) == wantLength {
 		return b
 	}
 	if cap(b) >= wantLength {
-		return b[:cap(b)]
+		return b[:wantLength]
 	}
 	return make([]byte, wantLength)
```
tbssajal, line 126: Reading the description of the function, it looks like we should append some bytes at the end of `b` instead of returning a new array. So if `cap(b) < wantLength`, we can discard `b`, it does not affect anything?

tbssajal: I mean it seems like the function should add some zeros at the end of `b` in case of `cap(b) < wantLength` so that `cap(b) >= wantLength` and then return `b[:wantLength]`. Instead it returns a new array. So in case `cap(b) < wantLength`, the original array `b` is discarded. Is it okay?

vpiyush: I still don't understand, but i assume that you are hinting to expand the existing array. The way capacity expansion works is to allocate the new array and replace the old one with the new one. So no advantage of doing that.

PR #851, opened 2023-10-19
[WIP] Separate consensus msgs

m4ksio:
A WIP to remove double (or triple) encoding of messages in a consensus

### DIFF
--- a/consensus/tendermint/backend/handler.go
+++ b/consensus/tendermint/backend/handler.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"github.com/autonity/autonity/common"
 	"github.com/autonity/autonity/consensus"
 	"github.com/autonity/autonity/consensus/tendermint/core/message"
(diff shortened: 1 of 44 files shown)


consensus/tendermint/backend/handler.go
```diff
@@ -53,31 +52,140 @@ func (sb *Backend) HandleUnhandledMsgs(ctx context.Context) {
 
 // HandleMsg implements consensus.Handler.HandleMsg
 func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, errCh chan<- error) (bool, error) {
-	if msg.Code != TendermintMsg && msg.Code != SyncMsg && msg.Code != AccountabilityMsg {
+	if msg.Code != constants.TendermintMsgLightProposal &&
```
vpiyush, line 55: Is this check really needed, if we have a default statement in the switch doing same thing?

consensus/tendermint/backend/handler.go
```diff
@@ -53,31 +52,140 @@ func (sb *Backend) HandleUnhandledMsgs(ctx context.Context) {
...
+
+		var decodedMessage *message.Message
+
+		//var data message.ConsensusMsg
+
+		//var err error
+		//
+		if msg.Code == constants.TendermintMsgLightProposal {
+
+			messageToDecode := &message.MessageLightProposal{}
```
vpiyush, line 134: Suggestion - using a local function for this repeated logic for all 3 msgtypes would avoid repetition here.

m4ksio:
Superseded by #852

PR #852, opened 2023-10-19
Consensus messages refactoring

yazzaoui:
Scope: 
- Completely Flatten the consensus message objects to skip the double serialization  
- Add payload caching mechanism to avoid duplicated messages serialization
- Merge Precommit seals with the Precommit message signatures

    Currently PrepareCommittedSeal is responsible of crafting the payload for the signature that will be inserted in the block header.
 https://github.com/autonity/autonity/blob/8f10cceba9e47a014000a9de78b4ab85a91ec0ec/consensus/tendermint/core/helpers/helpers.go#L19-L28

  We are essentially serializing [round, height, hash] in a custom manner, instead of using RLP, but this is problematic because 
  - this wouldn't allow third party tools with RLP libraries to figure out the validators of this block easily.
  - endianess between round and height is different, adding even more complexity  

   Precommit message signatures of RLP[PrecommitCode, round,height,hash ] are very similar so my proposal is to simply reuse them. This would reduce complexity and improve performance as the signature verification caches for precommit signatures can be reused for verifying the header seals.  

- Merge the Light Proposal Signature with Proposal Signature
- Accountability:
    - Removal of `AccountableGarbageMsg`: Due to the flat structure, this is less required. The consensus messages that cannot be unserialized and parsed correctly are automatically disconnecting the sender and never getting processed.


consensus/tendermint/backend/handler.go
```diff
@@ -117,15 +95,61 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, errCh chan<- erro
...
-
+		go sb.Post(events.AccountabilityEvent{Sender: addr, Payload: data, ErrCh: errCh})
 	default:
 		return false, nil
 	}
 
 	return true, nil
 }
 
+func handleConsensusMsg[M message.Message](sb *Backend, addr common.Address, msg p2p.Msg, errCh chan<- error) (bool, error) {
```
vpiyush, line 106: Generic function in GO involves multiple pointer deferences when not using same GCShape object. More about this [here](https://github.com/golang/proposal/blob/master/design/generics-implementation-dictionaries-go1.18.md). 
Also with the interface it would be even more expensive because this also involves a type assertion check. 
Sharing this post [here](https://planetscale.com/blog/generics-can-make-your-go-code-slower), Deep analysis on performance impact with generics.

consensus/tendermint/core/message/message.go
```diff
@@ -19,237 +21,420 @@ var (
...
-	}
+// extPropose is the actual proposal object being exchanged on the network
+// before RLP serialization.
+type extPropose struct {
+	Code            uint8
+	Round           uint64
+	Height          uint64
+	ValidRound      uint64
+	IsValidRoundNil bool
+	ProposalBlock   *types.Block
```
vpiyush, line 77: I see that we also have an extblock object. Shouldn't the extpropse have extBlock instead of block.

consensus/tendermint/core/message/message.go
```diff
@@ -19,237 +21,420 @@ var (
...
+	if vr == -1 {
+		isValidRoundNil = true
+	} else {
+		validRound = uint64(vr)
 	}
-	signature := m.Signature
-	payload, err := m.BytesNoSignature()
-	if err != nil {
+	// Calculate signature first
+	signatureInput := []any{ProposalCode, uint64(r), h, validRound, isValidRoundNil, block.Hash()}
```
vpiyush, line 110: using a struct would be better from performance point of view here. I would argue that construction of "any" slice is expensive due to all fields being pointers to interfaces and dynamic memory allocations for each field.

consensus/tendermint/core/core.go
```diff
@@ -149,10 +147,10 @@ type Core struct {
 	lastHeader *types.Header
```
vpiyush, line 147: Can we also refactor height above to uint64. this would avoid the need of converting back and forth.
Also with that we can have height and round as atomic integers and eliminate the need of stateMu during access of these fields.

yazzaoui: Yeah that sounds like a good idea

consensus/tendermint/core/message/message.go
yazzaoui: The rlp library will call *Block's EncodeRLP() which will do the conversion to extBlock. https://github.com/autonity/autonity/blob/1dd6b8b1f0791c5efafe23ce8a5d010176a5aac9/core/types/block.go#L489-L495

consensus/tendermint/backend/handler.go
yazzaoui: Very interesting but I don't think it compares by a long shot to signature verification, deserialization, memory allocs..

m4ksio (commented):
Overall, a really nice refactor! 
One comment - some changes in this PR seems to be not strictly related to the refactor. It would be better to keep them in separate PRs for brevity.

Also, two questions:
1. Why did we bumped protocol version? Sure, we have new messages now, but we went from 2 to 5?
2. New p2p message codes were added which is great. But we will use somewhat custom RLP encoding to include this information inside the payload. Is it necessary? Seems like just using the nodes would be enough, and we could keep payload intact

yazzaoui:
Thank you @m4ksio 
> 1. Why did we bumped protocol version? Sure, we have new messages now, but we went from 2 to 5?

So I guess you are referring to https://github.com/autonity/autonity/blob/e902b9f9a529ac6b519ff062beb60cb04031bf8d/consensus/tendermint/backend/handler.go#L40-L42

The second argument is not the version but the count of message codes the protocol is using. However, after a second look,  this function is not even used. I'll keep it though and add a todo to integrate it with the wire eth handler.

> 2. New p2p message codes were added which is great. But we will use somewhat custom RLP encoding to include this information inside the payload. Is it necessary? Seems like just using the nodes would be enough, and we could keep payload intact

No, it's not necessary, we could skip the message code in the final payload. I kept it for convenience for using directly this payload as the input of the hashing function. I don't have strong feelings, at the end it's just an additional one byte.

consensus/tendermint/helpers.go
```diff
@@ -0,0 +1,28 @@
...
+	lru "github.com/hashicorp/golang-lru"
+)
+
+var (
+	signatureCacheSize = 5000 // Number of recent addresses from ecrecover
+	signatureCache, _  = lru.NewARC(signatureCacheSize)
+)
+
+func SigToAddr(hash common.Hash, sig []byte) (common.Address, error) {
+	var cacheKey [65 + 32]byte
```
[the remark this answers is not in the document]
yazzaoui: Will use the constants.

e2e_test/e2e_test.go
```diff
@@ -162,18 +162,28 @@ func TestStartingAndStoppingNodes(t *testing.T) {
 	require.NoError(t, err)
 	defer network.Shutdown()
 	n := network[0]
-
+	startTime := time.Now()
+	idump := 0
+	go func() {
+		timer := time.NewTicker(time.Second)
+		for range timer.C {
+			fmt.Println("====", idump, "TIME(", time.Now().Sub(startTime).Seconds(), ")====")
```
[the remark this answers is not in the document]
yazzaoui: Yes, removed.

consensus/tendermint/core/state_dumper.go
```diff
@@ -3,26 +3,25 @@ package core
...
 
 // State Dump is handled in the main loop triggered by an event rather than using RLOCK mutex.
-func (c *Core) handleStateDump(e tctypes.CoreStateRequestEvent) {
-	state := tctypes.TendermintState{
+func (c *Core) handleStateDump(e CoreStateRequestEvent) {
+	state := TendermintState{
 		Client:            c.address,
 		BlockPeriod:       c.blockPeriod,
-		CurHeightMessages: msgForDump(c.CurrentHeightMessages()),
+		CurHeightMessages: msgForDump(c.messages.All()),
```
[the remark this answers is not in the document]
yazzaoui: That's the same.

consensus/tendermint/core/message/set.go
```diff
@@ -0,0 +1,93 @@
...
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	if votes, ok := s.votes[h]; ok {
+		power := new(big.Int)
+		for _, v := range votes {
+			power.Add(power, v.Power())
+		}
+		return power
+	}
+	return new(big.Int)
```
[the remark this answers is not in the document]
yazzaoui: Voting power is strictly positive, so 0 power means no votes.

consensus/tendermint/backend/backend.go
```diff
@@ -132,72 +133,62 @@ func (sb *Backend) Address() common.Address {
...
-				}
+	if sb.Broadcaster == nil || len(targets) == 0 {
+		return
+	}
+	for {
+		ps := sb.Broadcaster.FindPeers(targets)
+		// If we didn't find any peers try again in 10ms or exit if we have
+		// been stopped.
+		if len(ps) == 0 {
+			t := time.NewTimer(10 * time.Millisecond)
```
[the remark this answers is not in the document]
yazzaoui: Yeah Lorenzo did that in his recently merged PR, I just rebased.

yazzaoui:
Hey everyone, I've taken off the draft status from this PR. Just a heads-up, some tests are still failing, but the main part is good to go for review. It's a bit on the larger side, and in hindsight, breaking it into smaller PRs would've been better - something for us to aim for in the future. A lot of these changes are long overdue tech debt cleanups.

Focus your attention on the changes in the non-test files. If anything's unclear or if you have questions, I'm totally up for a call to discuss.

m4ksio (approved):
This is re-review of changes made since last review. Everything looks good, and since its out of draft - approved.

m4ksio:
> However, after a second look, this function is not even used. I'll keep it though and add a todo to integrate it with the wire eth handler.

Yes, I was confused since we changed the number of messages in https://github.com/autonity/autonity/pull/852/files#diff-3210763abf230e7f0fce7f0cd739a8ffa253599c6480aa34075017a4e310cd1cR47
So seems that there's where `Protocol()` was supposed to be used

p2p/enode/node.go
```diff
@@ -169,9 +170,17 @@ func (n *Node) ResolveHost() error {
 	if len(ips) == 0 {
 		return fmt.Errorf("failed to resolve IP for host %q, no IPs returned", host)
 	}
-
+	// Use sorting to deterministicaly retrieve the final resolved ip.
+	sort.Slice(ips, func(i, j int) bool {
```
lorenzo-dev1, line 174: Just curious, why do we need determinism in the IP resolution of the host? Also, how could an host be associated to multiple IPs?

Additionally, maybe we can have a separate function to sort IPs? Using something like https://pkg.go.dev/net/netip#Addr.Less to avoid char by char comparison

p2p/enode/url4_resolve_test.go
```diff
@@ -191,7 +191,7 @@ func TestParseNodeWithDomainResolution(t *testing.T) {
 				// to allow comparing the remaining fields.
 				n.resolveFunc = nil
 				test.wantResult.resolveFunc = nil
-				assert.Equal(t, test.wantResult, n)
+				assert.Equal(t, test.wantResult, n, i)
```
lorenzo-dev1, line 194: I am not sure what does this `i` addition do, does it get printed if the assertion fails?

p2p/enode/node.go
lorenzo-dev1: Additionally, `make all` fails in this piece of code because `min` is not defined.

consensus/consensus.go
```diff
@@ -30,6 +31,15 @@ import (
 	"github.com/autonity/autonity/rpc"
 )
 
+// Place consensus message codes in protocol to break dependency circle in between tendermint and fault detectors.
```
lorenzo-dev1, line 34:
```suggestion
```

consensus/consensus.go
```diff
@@ -30,6 +31,15 @@ import (
...
+
+// setting for Autonity accountability protocol, they are a part of consensus.
```
lorenzo-dev1, line 36: Saying they are part of consensus is a bit misleading imo
```suggestion
// setting for Autonity accountability protocol
```

consensus/tendermint/helpers.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/autonity/autonity/crypto"
+	lru "github.com/hashicorp/golang-lru"
+)
+
+var (
+	signatureCacheSize = 5000 // Number of recent addresses from ecrecover
+	signatureCache, _  = lru.NewARC(signatureCacheSize)
+)
+
+func SigToAddr(hash common.Hash, sig []byte) (common.Address, error) {
```
lorenzo-dev1, line 14: Can we just use the signature as key in the cache?

Theoretically a signature on different hashes will always be different (same goes for signature from different keys for same hash).

consensus/tendermint/crypto/check_signature_test.go
```diff
@@ -1,195 +0,0 @@
-package crypto
```
lorenzo-dev1, line 1: Where these tests deleted or moved somewhere else?

consensus/tendermint/backend/engine.go
```diff
@@ -66,10 +67,9 @@ var (
 )
 
 // Author retrieves the Ethereum address of the account that minted the given
-// block, which may be different from the header's coinbase if a consensus
-// engine is based on signatures.
+// block.
 func (sb *Backend) Author(header *types.Header) (common.Address, error) {
-	return types.Ecrecover(header)
+	return header.Coinbase, nil
```
lorenzo-dev1, line 72: Is this to allow setting the block reward beneficiary to someone != proposer?

E.g. a group of stakers creating a sort of pool to minimize risk and sending all block rewards there?

p2p/enode/node.go
yazzaoui: >Just curious, why do we need determinism in the IP resolution of the host?

It was mainly because of the tests in p2p/enode/url4_resolve_test.go which were flaky as some hostnames would resolve to multiple IPs. I could have added support for multiple IPs but I felt it was overkill or definitely should be in a separate PR. The simplest solution that came to me was this one.

> Also, how could an host be associated to multiple IPs?

Having multiple A records for the same host name in DNS is a common practice to increase availability.

> Additionally, `make all` fails in this piece of code because `min` is not defined.

`min` is defined part of the go standard library 1.21 (I bumped the version).  

> Additionally, maybe we can have a separate function to sort IPs? Using something like https://pkg.go.dev/net/netip#Addr.Less to avoid char by char comparison

Agree, doesn't look great but I don't want to switch everything to `netip.Addr` or perform a conversion just for the sake of this sorting. If you can think of a minimally intrusive alternative I'm happy to take it.

p2p/enode/url4_resolve_test.go
yazzaoui: yes, it was useful for me to find which test was failing

consensus/consensus.go
yazzaoui: I didn't write this comment, I'm going to remove it.

consensus/tendermint/backend/engine.go
yazzaoui: No, coinbase is verified as part of the consensus rules in header verification to be equal to the proposer. This is an optimization to avoid ecrecover.

consensus/tendermint/helpers.go
yazzaoui: I think that introduced a vulnerability but I forgot it.

consensus/tendermint/crypto/check_signature_test.go
yazzaoui: Deleted, we're no longer using this function.

consensus/tendermint/backend/handler.go
```diff
@@ -118,15 +94,55 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg, errCh chan<- erro
...
+		go sb.Post(events.AccountabilityEvent{Sender: addr, Payload: data, ErrCh: errCh})
 	default:
 		return false, nil
 	}
 
 	return true, nil
 }
 
+func handleConsensusMsg[T any, PT interface {
+	*T
```
lorenzo-dev1, line 106: Why do we need the T generic? Is it to allow additional methods wrt the message.Msg interface (e.g. the proposal will also have the `Block()` method)

consensus/tendermint/core/message/message.go
```diff
@@ -3,253 +3,586 @@ package message
...
+		log.Crit("message returning empty payload")
 	}
-	return m.ConsensusMsg.H().Uint64()
+	return b.payload
 }
 
-func (m *Message) Sender() common.Address {
-	return m.Address
+func (b *base) Hash() common.Hash {
+	if b.payload == nil {
```
lorenzo-dev1, line 499: Should this check nil for `b.hash`?

consensus/tendermint/accountability/contracts.go
```diff
@@ -767,37 +695,38 @@ func decodeRawProof(b []byte) (*Proof, error) {
...
 
 // checkMsgSignature checks if the consensus message is from valid member of the committee.
-func checkMsgSignature(chain ChainContext, m *message.Message) error {
-	lastHeader := chain.GetHeaderByNumber(m.H() - 1)
+func verifyProofSignatures(chain ChainContext, p *Proof) error {
+	h := p.Message.H()
+	if h == 0 {
+		return errBadHeight
+	}
+	lastHeader := chain.GetHeaderByNumber(p.Message.H() - 1)
```
lorenzo-dev1, line 707: Nitpick: reuse `h`

consensus/tendermint/accountability/contracts.go
```diff
@@ -161,93 +154,58 @@ func (c *MisbehaviourVerifier) RequiredGas(input []byte) uint64 {
...
-	if evHeight == 0 {
-		return failureResult, nil
+	if err := verifyProofSignatures(c.chain, p); err != nil {
+		return failureReturn, nil
 	}
-
 	return c.validateProof(p), nil
 }
 
 // validate the Proof, if the Proof is valid, then the rlp hash of the msg payload and rlp hash of msg sender is
```
lorenzo-dev1, line 170: Might be worth to add a comment specifying that this function does not do cryptographic validation, and `verifyProofSignatures` needs to be used before.

consensus/tendermint/accountability/contracts.go
```diff
@@ -256,81 +214,82 @@ func (c *MisbehaviourVerifier) validMisbehaviourOfPN(p *Proof) bool {
...
+		return false
+	}
+	preCommit, ok := p.Evidences[0].(*message.Precommit)
+	if !ok {
 		return false
 	}
-
-	preCommit := p.Evidences[0]
 	if preCommit.Sender() == proposal.Sender() &&
-		preCommit.Type() == consensus.MsgPrecommit &&
```
lorenzo-dev1, line 269: Should we check the message code instead of the type? Or that is implicitly checked by this line:
`preCommit, ok := p.Evidences[0].(*message.Precommit)`

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -946,12 +860,12 @@ prevotesLoop:
 		// valid.
 		var prevotesProofs []*Proof
 		for _, cp := range correspondingProposals {
-			cp := cp // todo(youssef): I don't think this is necessary as nothing is async, double check
+			proposal := cp.(*message.Propose)
```
lorenzo-dev1, line 863: To understand, will this still create a copy or just assert the type?

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -1288,71 +1203,72 @@ func (fd *FaultDetector) submitMisbehavior(m *message.Message, evidence []*messa
...
+	if proposal.ValidRound() >= proposal.R() {
+		fd.submitMisbehavior(message.NewLightProposal(proposal), nil, errWrongValidRound)
 		return errWrongValidRound
 	}
 
 	// account for equivocation
-	equivocated := fd.msgStore.Get(m.H(), func(msg *message.Message) bool {
-		return msg.R() == m.R() && msg.Type() == consensus.MsgProposal && msg.Sender() == m.Sender() && msg.Value() != m.Value()
+	equivocated := fd.msgStore.Get(proposal.H(), func(msg message.Msg) bool {
+		// todo(youssef) : again validValue missing here
```
lorenzo-dev1, line 1237: ?
We need to check equivocation on valid round?

consensus/tendermint/core/message/message.go
```diff
@@ -1,255 +1,610 @@
...
+	return p.block.Hash()
+}
+
+func (p *Propose) String() string {
+	return fmt.Sprintf("{Round: %v, Height: %v, ValidRound: %v, ProposedBlockHash: %v}",
+		p.round, p.H(), p.validRound, p.block.Hash().String())
+}
+
+func (p *Propose) MustVerify(inCommittee func(address common.Address) *types.CommitteeMember) *Propose {
+	p.Validate(inCommittee)
```
lorenzo-dev1, line 133: Wouldn't it be better to panic if validate returns an error?

consensus/tendermint/core/message/message.go
```diff
@@ -1,255 +1,610 @@
...
+func (b *base) Payload() []byte {
+	return b.payload
+}
+
+func (b *base) Hash() common.Hash {
+	return b.hash
+}
+
+// Validate verify the signature and set appropriate sender / power fields
+func (b *base) Validate(inCommittee func(address common.Address) *types.CommitteeMember) error {
```
lorenzo-dev1, line 529: Maybe here we can have a shortcircuit:
```
if b.verified {
return
}
```

consensus/tendermint/core/message/set.go
```diff
@@ -0,0 +1,93 @@
+package message
+
+import (
+	"github.com/autonity/autonity/common"
+	"math/big"
+	"sync"
+)
+
+type isVote interface {
```
lorenzo-dev1, line 9: I don't understand, also the proposal satisfies this type constraint, right?

consensus/tendermint/core/message/set.go
```diff
@@ -0,0 +1,93 @@
...
+		messages: make(map[common.Address]T),
+	}
+}
+
+type Set[T isVote] struct {
+	// In some conditions we might receive prevotes or precommit before
+	// receiving a proposal, so we must save received message with different proposed block hash.
+	votes    map[common.Hash]map[common.Address]T // map[proposedBlockHash]map[validatorAddress]vote
+	messages map[common.Address]T
+	lock     sync.RWMutex
```
lorenzo-dev1, line 26: Not a big deal but we can embed the lock into the set

consensus/tendermint/core/msg_store.go
```diff
@@ -106,6 +106,31 @@ func (ms *MsgStore) Get(height uint64, query func(*message.Message) bool) []*mes
 			}
 		}
 	}
+	return result
+}
 
+func GetStore[T any, PT interface {
```
lorenzo-dev1, line 112: Why GetStore?

consensus/tendermint/core/precommit.go
```diff
@@ -1,104 +1,64 @@
...
+			// We are receiving a precommit for an old round. We need to check if we have now a quorum
+			// in this old round.
+			roundMessages := c.messages.GetOrCreate(precommit.R())
+			roundMessages.AddPrecommit(precommit)
+			oldRoundProposal := roundMessages.Proposal()
+			if oldRoundProposal != nil && roundMessages.PrecommitsPower(oldRoundProposal.Block().Hash()).Cmp(c.CommitteeSet().Quorum()) >= 0 {
+				c.logger.Info("Quorum on a old round proposal", "round", precommit.R())
+				if !roundMessages.IsProposalVerified() {
+					if _, err2 := c.backend.VerifyProposal(roundMessages.Proposal().Block()); err2 != nil {
+						// Impossible with the BFT assumptions of 1/3rd honest.
```
lorenzo-dev1, line 45: Maybe we should print a warning here? It means something is going very wrong

consensus/tendermint/core/message/message.go
yazzaoui: Depreciated, now Hash and Payload are guaranteed to be available.

consensus/tendermint/accountability/contracts.go
yazzaoui: Renamed to `validateFault` and added comment.

consensus/tendermint/accountability/fault_detector.go
yazzaoui: This creates the concrete implementation of the interface, so a copy of the underlying Propose pointer.

consensus/tendermint/accountability/fault_detector.go
yazzaoui: We have to, we should never see a validator proposing the same block with a different valid value. 
Let's discuss this at standup.

consensus/tendermint/core/message/message.go
yazzaoui: Added panic, this function is only used for testing purposes.

consensus/tendermint/core/message/message.go
yazzaoui: Yeah.. I don't see why should we re-verify an already verified message. That would seem to be a code flow issue.  I'm going to add a panic for the time being.

consensus/tendermint/core/message/message.go
```diff
@@ -1,255 +1,610 @@
...
+func PrepareCommittedSeal(hash common.Hash, round int64, height *big.Int) common.Hash {
+	// this is matching the signature input that we get from the committed messages.
+	buf, _ := rlp.EncodeToBytes([]any{PrecommitCode, uint64(round), height.Uint64(), hash})
+	return crypto.Hash(buf)
 }
 
-func (m *Message) Sender() common.Address {
-	return m.Address
+// Fake is a dummy object used for internal testing.
+type Fake struct {
```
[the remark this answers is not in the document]
yazzaoui: All those EXPECTs and mock boilerplate are making our tests brittle. Sometimes though mocks are useful it depends on the situation but not here IMO.

p2p/enode/node.go
lorenzo-dev1: > Having multiple A records for the same host name in DNS is a common practice to increase availability. 

ok but then making the address resolution deterministic does not defeat this purpose? 

> Agree, doesn't look great but I don't want to switch everything to netip.Addr or perform a conversion just for the sake of this sorting. If you can think of a minimally intrusive alternative I'm happy to take it.

ok I see

>  I could have added support for multiple IPs but I felt it was overkill or definitely should be in a separate PR. The simplest solution that came to me was this one.

I see, maybe this should be tracked as a separate issue and addressed in another PR

consensus/tendermint/core/message/set.go
yazzaoui: My bad, this is depreciated. It should be :
```
type isVote interface {
*Prevote|*Precommit
}
```

consensus/tendermint/core/message/set.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (s *Set[T]) AddVote(vote T) {
```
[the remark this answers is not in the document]
yazzaoui: Yeah, I will do some renaming as well for some other functions in this file.

consensus/tendermint/core/msg_store.go
yazzaoui: It's a typed version of Get, instead of returning an interface message I perform the type assertion and return the concrete type given the generic argument  It will simplify the codebase a lot in the fault detector part where we almost always perform a check on the type when calling Get.

consensus/tendermint/core/precommit.go
yazzaoui: Wondering if we should panic here... TBH  I think we should.

consensus/tendermint/core/msg_store.go
lorenzo-dev1: Yeah my question was more about the naming of the function. I would call it something like `TypedGet` or something like that

yazzaoui: `TypedGet` is too generic if the message store belongs to the core package. I can't define a method on a method with generic arguments this is a limitation of go.

consensus/tendermint/core/message/message.go
lorenzo-dev1: We have it for prevote and precommit too

consensus/tendermint/core/msg_store.go
lorenzo-dev1: Didn't notice it was a function, makes sense

consensus/tendermint/core/propose.go
```diff
@@ -2,111 +2,91 @@ package core
...
 }
 
-func (c *Proposer) HandleProposal(ctx context.Context, msg *message.Message) error {
-	proposal := msg.ConsensusMsg.(*message.Proposal)
+func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose) error {
 	// Ensure we have the same view with the Proposal message
-	if err := c.CheckMessage(proposal.Round, proposal.Height.Uint64(), tctypes.Propose); err != nil {
-		// If it's a future round proposal, the only upon condition
+	if err := c.checkMessageStep(proposal.R(), proposal.H(), Propose); err != nil {
+		// If it's a future round proposal2, the only upon condition
```
lorenzo-dev1, line 39: Nitpick, but there are several instances of proposal2 being mentioned, but i don't understand what is different wrt proposal

consensus/tendermint/core/message/round_messages.go
```diff
@@ -1,30 +1,30 @@
...
 	"sync"
+
+	"github.com/autonity/autonity/common"
 )
 
-type MessagesMap struct {
+type Map struct {
 	internal map[int64]*RoundMessages
-	mu       *sync.RWMutex
+	mu       sync.RWMutex
```
lorenzo-dev1, line 12: can embed

lorenzo-dev1:
Great PR, greatly streamlines and simplifies the consensus messaging code. 

In this comment are some pointers to test files where there have been deletions / tests got commented. I think it is worth to investigate the possibility of re-instating these tests, to avoid losing work already done. However I did not have time yet to dig deep into whether they are still useful or not. 

IMO this is not a blocker for merging though and it can be addressed in a separate PR. 

Tests that got removed/commented:
- `consensus/tendermint/crypto/check_signature_test.go`. File deleted. Signature verification code has been moved to `consensus/tendermint/helpers.go`. Maybe it is worth to move these tests there.
- `consensus/tendermint/core/message/consensus_msg_test.go`. File deleted. Contained some tests on encoding/decoding which might still be useful.
- `consensus/tendermint/core/message/message_test.go`. some tests there are commented out.
- `consensus/tendermint/core/message/round_messages_test.go`. some tests there are commented out.
- `consensus/tendermint/core/message/set_test.go`. Some tests there are commented out.
- `consensus/tendermint/core/prevote_test.go`. Some tests are commented out.
- `e2e_test/byzantine/accountable_garbage_test.go`. File deleted. We have no accountable garbage message anymore, so that make sense. However we might want a test to check for disconnection if someone sends garbage?
- `e2e_test/byzantine/garbage_test.go`. Some tests that were using fuzzing got removed (I suppose due to the new message format). However I think we might still need some fuzzing tests to fuzz the decoding functions.
- `e2e_test/byzantine/oversize_accountability_event_test.go`. File removed. We currently do not handle the oversized events. However I think it can be useful to already have a test that sends it, to see how our client reacts.

Commented out tests always make me a bit nervous, as it is easy to forget them and leave the commented out forever. It would be better to skip them imo (but I understand that it requires fixing the compilation errror first).

consensus/tendermint/accountability/fault_detector.go
lorenzo-dev1: Yes I think that should be considered equivocation

consensus/tendermint/core/precommit.go
lorenzo-dev1: I think that if we have a quorum on invalid proposal, it means that either:
1. we have >2/3 of malicious committee members (which is 2 times the bft limit)
2. we are on the wrong chain (e.g. long range attack, wrong genesis, etc.)

For case 1 I guess it would be better to reject the chain and start mining on previous valid block (e.g. forking the chain). But that shouldn't happen in our chain (no forks).

If we are on the wrong chain I think it is probably better to notify the user (e.g. you are under attack) and stop the node.

consensus/tendermint/core/message/message.go
yazzaoui: Ok so the fault detector is verifying messages too which is causing trouble. I'm going to add a lock until we can get your changes for parallel verification.

consensus/tendermint/core/message/message.go
yazzaoui: Added

consensus/tendermint/core/precommit.go
yazzaoui: Those are valid points, let's create a dedicated issue for it. In the meantime, I'll add a panic it is the safest for us to catch bugs.

consensus/tendermint/core/propose.go
yazzaoui: my bad

consensus/tendermint/core/precommit.go
lorenzo-dev1: issue opened: https://github.com/autonity/autonity/issues/863. resolving

Makefile
```diff
@@ -226,10 +226,7 @@ docker-e2e-test: contracts
 mock-gen:
```
vpiyush, line 226: this target also should be added to make all target,  otherwise the tests would fail, if there are changes in interfaces.

vpiyush: Also I think it's a good idea to have a fixed specific version of mockgen, either checked in or installed with make

yazzaoui: Good idea, let's have it in a separate PR.

consensus/tendermint/helpers.go
yazzaoui: Assume:
- Val A send VoteA to Val C
- Val B create VoteB with value B != value A and insert VoteA signature
- VoteB has a different hash than Val C so it gets sent to core. In SigToAddr cache is hit as hash is ignored and Val A address is returned

Successful equivocation attack from Val B.  Agree?

consensus/tendermint/accountability/fault_detector.go
yazzaoui: I'll create a separate PR for this bug and some others I found in the accountability package

Makefile
vpiyush: Tracking [here](https://github.com/autonity/autonity/issues/867)

vpiyush (approved):

yazzaoui:
Tried to fix / rewrite as many remaining tests as I could, included the commented ones that you mentioned Lorenzo, they were still work in progress. Restoring fuzz testing is a whole job by itself that would require another PR.

It's not perfect on the testing coverage but I'm happy now with the state of this PR and going to merge soon.

lorenzo-dev1 (approved):
LGTM

p2p/enode/node.go
yazzaoui: Reverted / fixed the tests instead

PR #855, opened 2023-11-01
Gossiper abstraction

lorenzo-dev1:
This PR introduces a `Gossiper` abstraction so that we can test out custom gossip logic. It also refactors the `TendermintServices` to use custom construction functions instead of reflection (less brittle).


consensus/tendermint/backend/backend.go
```diff
@@ -193,22 +197,45 @@ func (sb *Backend) AskSync(header *types.Header) {
 	}
 }
 
-// Broadcast implements tendermint.Backend.Gossip
+// Gossip implements tendermint.Backend.Gossip
 func (sb *Backend) Gossip(ctx context.Context, committee types.Committee, payload []byte) {
+	sb.gossiper.Gossip(ctx, committee, payload)
+}
+
+type Gossiper struct {
```
vpiyush, line 205: It is probably a good idea to move Gossiper and it's implementation in its own file

lorenzo-dev1: addressed in https://github.com/autonity/autonity/pull/855/commits/d026abb469d6594fa71b7eb022e8300fcd41dc0a

consensus/tendermint/backend/gossiper.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+
+	if g.broadcaster != nil && len(targets) > 0 {
+		for {
+			ps := g.broadcaster.FindPeers(targets)
+			// If we didn't find any peers try again in 10ms or exit if we have
+			// been stopped.
+			if len(ps) == 0 {
+				t := time.NewTimer(10 * time.Millisecond)
```
m4ksio, line 104: since we refactor, we can move this constant to a parameter

consensus/tendermint/core/interfaces/tendermint.go
```diff
@@ -12,4 +12,8 @@ type Tendermint interface {
 	Stop()
 	CurrentHeightMessages() []*message.Message
 	CoreState() types.TendermintState
+	Br() Broadcaster
```
m4ksio, line 15: Any reason not to simply name this method `Broadcaster()`?

m4ksio (approved):
Few minor comments

consensus/tendermint/backend/backend_test.go
```diff
@@ -60,6 +61,10 @@ func TestAskSync(t *testing.T) {
 	for _, p := range addresses {
 		m[p] = struct{}{}
 	}
+	recentMessages, err := lru.NewARC(inmemoryMessages)
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in 425708d3426c7e677d2cc52464cb4bbff09a03be

consensus/tendermint/core/interfaces/tendermint.go
lorenzo-dev1: The method in the implementation was named like this before my changes, so I just updated the interface accordingly. 

If we want we can change it to `Broadcaster()`, however I think `Br()` is good enough and gives the same information at the end of the day.

consensus/tendermint/core/interfaces/gossiper.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/autonity/autonity/common"
+	"github.com/autonity/autonity/consensus"
+	"github.com/autonity/autonity/core/types"
+	lru "github.com/hashicorp/golang-lru"
+)
+
+type Gossiper interface {
+	Gossip(committee types.Committee, payload []byte)
```
yazzaoui, line 11: Why removing the context object?  I think it's quite useful to permit gossip abortion when stopping autonity for exemple.

consensus/tendermint/core/interfaces/tendermint.go
yazzaoui: It's true that `Br()` is understandable but it's important to remain consistent with our naming conventions in the codebase. We don't use abbreviations anywhere and this violates how getters should be named according to the Go conventions ( see https://go.dev/doc/effective_go#Getters )
But it's not a big deal I can change that in my refactoring PR.

e2e_test/node.go
```diff
@@ -430,14 +433,17 @@ func (nw Network) WaitForNetworkToStartMining() error {
 
 // WaitToMineNBlocks waits for network to mine given number of
 // blocks in the given time window default value for numSec can be kept 60 seconds
-func (nw Network) WaitToMineNBlocks(numBlocks uint64, numSec int) error {
+// if verifyRate == true --> we return an error if we cannot satisfy that 1 block/s rate
```
yazzaoui, line 436: Interesting

consensus/tendermint/backend/gossiper.go
```diff
@@ -0,0 +1,131 @@
...
+				if _, k := m.Get(hash); k {
+					// This peer had this event, skip it
+					continue
+				}
+			} else {
+				m, _ = lru.NewARC(inmemoryMessages)
+			}
+
+			m.Add(hash, true)
+			g.recentMessages.Add(addr, m)
```
tbssajal, line 81: If `m` is already present, it includes additional unnecessary complexity, right? maybe we can move this line inside the `else` above?

consensus/tendermint/core/interfaces/gossiper.go
lorenzo-dev1: I removed the context object because it was not used in the Gossip function. Additionally since the Gossip function is not a long-standing function (it just fires up the Send go routines and terminates right away), do we need a termination mechanism through context?

consensus/tendermint/core/interfaces/tendermint.go
lorenzo-dev1: got it, fixed with https://github.com/autonity/autonity/pull/855/commits/8d27c44aa3d94684959511e803e91785a46f7e69

consensus/tendermint/backend/gossiper.go
lorenzo-dev1: The `Add` method of `ARCCache` contains also some promotion logic based on frequency of access (ARCCache uses both frequency and recency of use to determine cache evictions). 

Therefore doing the `Add` again with key=addr ensures that the frequency of use of `addr` gets updated.

I have simplified a lot but if you want to understand really how it works look into the `Add` function code

lorenzo-dev1: The `Add` method of ARCCache contains also some promotion logic based on frequency of access (ARCCache uses both frequency and recency of use to determine cache evictions). Therefore we should also take that into account when coding the gossip function.

I think your comment is valid though, and I improved the Gossip function in https://github.com/autonity/autonity/pull/855/commits/7e42e0ce2e9b271f30c938081d7b61e14501c5cd. Now whichever path we take, the `addr` key always gets 1 hit. Before it would get 2 hits in case of existing peer cache but no message.

lorenzo-dev1: Actually I saw other issues with the ARCCache mechanism. I will address this points in another PR.

tbssajal: Another thing is that in this part of the code, we are iterating over almost all the committee members. So almost everyone has the same frequency.

lorenzo-dev1: Let's continue the discussion in the issue, this is out of scope of this PR

consensus/tendermint/backend/gossiper.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/autonity/autonity/common"
+	"github.com/autonity/autonity/consensus"
+	"github.com/autonity/autonity/consensus/tendermint/bft"
+	"github.com/autonity/autonity/core/types"
+	"github.com/autonity/autonity/log"
+	lru "github.com/hashicorp/golang-lru"
+)
+
+type Gossiper struct {
+	recentMessages *lru.ARCCache  // the cache of peer's messages
```
Jason-Zhangxin-Chen, line 16: In the nature of gossip, a recent message is a known message that been processed by the current node. The naming confuses me, there the recentMessages is the cache of peer's message (remote peer's message?), while knownMessages is the cache of self messages.

Jason-Zhangxin-Chen (approved):
Just a comment on the confusing naming of recentMessages and the knownMessages.

consensus/tendermint/backend/gossiper.go
lorenzo-dev1: Yes those names confuse me as well. However they come from how these caches are named in the backend handler. I think I will add this comment in https://github.com/autonity/autonity/issues/859 and  address it in another PR.

yazzaoui (approved):
Looking good! Nice job.

consensus/tendermint/core/interfaces/gossiper.go
lorenzo-dev1: Ok... I am wondering if we should remove it also from `SignAndBroadcast` and `backend.Broadcast` then, as it was there just to be passed to `Gossip`. I think this break a bunch of tests so we should probably decide now if worth keeping it to avoid wasting time re-adding it. 

I guess it would be worth keeping if we might consider the idea of having a single long standing go routine / long running loop in the broadcasting functions, which needs termination via ctx.Done()

yazzaoui: Thinking more about it,  ultimately we may want to remove the use of `context` altogether in the codebase, it's well adapted in situations where we need to propagate termination across a tree of invocations. But in our case, we just need a termination signal to stop everything at once.. That can be easily achieved via a `stop` channel. `context` is overkill and clunky.

In the meantime, it's your choice..  If I'm supposing there is a long-standing go-routine (an accumulator for phase 1 gossip for example), it should be created by a dedicated "start(context)" function rather than `Gossip()` and Gossip() should communicate with this routine via a channel.. Otherwise, we'll end up creating tons of go-routines.

m4ksio: I would rather advocate addition of context anywhere it is missing. However, since this codepath leads to eth based p2p code which does not use context I think its fine to drop it. It looks like there is no actual async operation anywhere.

lorenzo-dev1: ok I will remove it from these broadcasting and gossip related codepath then.

> I would rather advocate addition of context anywhere it is missing

Why? 

Correct me if I am wrong, but from what I understood the context is useful by providing functions that allow to set a deadline / timeout on a task easily, which I can see as handy for APIs for example. However is it useful when we just want to stop all go routines when terminating the program?

lorenzo-dev1:
The failed test is not due to changes in this PR, but instead due to dependencies issues of `eth-ape`. That problem is addressed in: https://github.com/autonity/autonity/pull/861

Merging this pr.

Issue #819, opened 2023-08-18
Fairness issue after 100% slash


yazzaoui:
_Originally posted by @lorenzo-dev1 in https://github.com/autonity/autonity/issues/807#issuecomment-1651083480_

Scenario:

Supposing we have a registered validator with no self-bonded stake. Then delegator1 bonds 10 NTN to it.

At the end of the epoch we will have:
```
_validator.liquidSupply = 10
_validator.bondedStake = 10
```

Then delegator1 unbonds. The LNTN gets locked, then at the end of the epoch we will have:
```
_validator.liquidSupply = 0
 _validator.bondedStake = 0
_validator.unbondingStake = 10
_validator.unbondingShares = 10
_unbonding.unbondingShare = 10
```

Then before the unbonding period expires, the validator gets slashed due a past offence with critical severity (100% slash) and all of the unbondingStake gets slashed:
```
_validator.unbondingStake = 0
```

So delegator1 should receive 0 NTN when the unbonding period expires.

But let's assume that, always before the unbonding period expires, delegator2 bonds 10 NTN:
```
_validator.liquidSupply = 10
_validator.bondedStake = 10
```

And unbonds 10 NTN:
```
_validator.liquidSupply = 0
 _validator.bondedStake = 0
_validator.unbondingStake = 10
_validator.unbondingShares = 20
_unbonding.unbondingShare = 10
```

Now when the two unbonding periods expire, what is going to happen is that both delegator1 and delegator2 will take the hit from the slashing, instead of only delegator1.

~Agreed upon solution: Enforce minimum of `unbondingPeriod` between bond and unbond operations.~

TBD The issue is also valid for the bonding NTN -> LNTN operation

yazzaoui:
Considering to introduce the special state `killed` if slashed at 100% but still need to figure out what exactly that will involve.

tbssajal:
I think in case of 100% slash, if we jail the validator until every existing delegator (non-self-bonded) unbonds from the validator, that will solve the issue.

tbssajal:
> I think in case of 100% slash, if we jail the validator until every existing delegator (non-self-bonded) unbonds from the validator, that will solve the issue.

[...]

PR #857, opened 2023-11-06
New Validator Status Jailbound in Case of Fairness Issue

tbssajal:
Closes issue [819](https://github.com/autonity/autonity/issues/819). Following observations are taken into consideration

- Fairness issue can occur even without 100% slashing due to integer division. Currently we take floor when calculating `_unbondingSlash` and so for `delegatedSlash = _remaining - _unbondingSlash` we are taking ceil. However if slashing rate is less than 100% but very closer to 100% and `unbondingStake` is very high compared to `delegatedStake` then after slashing `delegatedStake` can become 0 while `unbondingStake` remains positive. And whenever `delegatedStake` or `unbondingStake` becomes 0 from positive value due to slashing, we face the fairness issue.
     - Changing the slashing amount calculation so that we take floor for both `_unbondingSlash` and `delegatedSlash`. This way the fairness issue can only be triggered by slashing 100%.

Based on the above observations and [to be fair for all](https://github.com/autonity/autonity/pull/857#issuecomment-1802035518), banning the validator following these rules:

- Ban if 100% slashed

Edit: Ban rule


tbssajal:
Updated the description with observation and implementation details.

tbssajal:
> On naming of the status, is Tombstone better? It's the term used by Cosmos (https://docs.cosmos.network/main/build/modules/slashing#staking-tombstone) for this scenario so already has precedence in the blockchain space as a term for this state of infinite jailing.

Sounds better than Killed.

rmsams:
> > On naming of the status, is Tombstone better? It's the term used by Cosmos (https://docs.cosmos.network/main/build/modules/slashing#staking-tombstone) for this scenario so already has precedence in the blockchain space as a term for this state of infinite jailing.
> 
> Sounds better than Killed.

Indeed. Some additional ideas: 
- _jailbound_ (implies perpetual jailing)
- _gulag_
- _alcatraz_
- _acheron_ (underworld river in Greek mythology associated with eternal suffering)

cmjc:
> * _jailbound_ (implies perpetual jailing)

That's a good alternative to _tombstone_ I think...

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1057,6 +1059,12 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         BondingRequest storage _bonding = bondingMap[id];
         Validator storage _validator = validators[_bonding.delegatee];
 
+        // killed validator is banned permanently, no new bonding can be applied for a killed validator
+        if (_validator.state == ValidatorState.killed) {
```
tbssajal, line 1063: Is there any need to add an extra field "applied" in "BondingRequest" for this change?

cmjc:
On the naming, I would like to vote for `jailbound` - I think it's descriptive and works well - jailed is temporary, jailbound is infinite...

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: I personally don't think it is needed, I don't see which value it brings

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1100,6 +1108,9 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
 
     function _releaseUnbondingStake(uint256 _id) internal virtual {
         UnbondingRequest storage _unbonding = unbondingMap[_id];
+        if (_unbonding.unbondingShare == 0) {
```
lorenzo-dev1, line 1111: How can the unbondingShare become 0, is this related to the `killed ` state?

lorenzo-dev1:
> If both delegatedStake and unbondingStake are 0 at the moment of 100% slash, it does not create any fairness issue, so not banning the validator

I think that the severity of the punishment should be the same regardless of the amount of `delegatedStake` and `unbondingStake` (even if it does not cause a fairness issue). This is because otherwise it is unfair but for the validators, i.e. 2 validators commit the same offence, but only 1 gets permanently jailed because he has delegatedStake.

autonity/solidity/contracts/Accountability.sol
```diff
@@ -374,22 +374,48 @@ contract Accountability is IAccountability {
         // Remaining stake to be slashed is split equally between the delegated
         // stake pool and the non-self unbonding stake pool.
         // As a reminder, the delegated stake pool is bondedStake - selfBondedStake.
```
lorenzo-dev1, line 376:
```suggestion
        // As a reminder, the delegated stake pool is bondedStake - selfBondedStake (which is 0 if we reached this line).
```

lorenzo-dev1: Just a clarification for why we do not have the `- selfBondedStake` in the following line

tbssajal: At this point, all selfBondedStake is 0 because penalty was greater than selfBondedState+selfUnbondingStake. So I just removed it to reduce the number of mathematical operations done (thinking from the perspective of gas cost) and also added a comment for clarity.

autonity/solidity/contracts/Autonity.sol
tbssajal: Yes, after killing validator, if someone requests unbonding, it can happen.

tbssajal:
> > If both delegatedStake and unbondingStake are 0 at the moment of 100% slash, it does not create any fairness issue, so not banning the validator
> 
> I think that the severity of the punishment should be the same regardless of the amount of `delegatedStake` and `unbondingStake` (even if it does not cause a fairness issue). This is because otherwise it is unfair but for the validators, i.e. 2 validators commit the same offence, but only 1 gets permanently jailed because he has delegatedStake.
> 
> I think that if you get slashed 100%, you should get permanently jailed regardless of your stake distribution (unless this causes some other issues that I am missing).

Yeah, it makes sense. I will change the codes and tests accordingly

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: oh ok right, it is because of this line right?
```
_newtonAmount = (_liquidAmount * _delegatedStake) / _validator.liquidSupply;
```

in case _delegatedStake = 0?

autonity/solidity/contracts/Accountability.sol
lorenzo-dev1: oh ok I had not see your new comment. then my addition is redundant, ignore it please

autonity/solidity/contracts/Autonity.sol
tbssajal: Yes

lorenzo-dev1: gotcha thanks

autonity/solidity/contracts/Accountability.sol
```diff
@@ -374,22 +374,42 @@ contract Accountability is IAccountability {
...
-        if (_remaining > 0 && (_val.unbondingStake + (_val.bondedStake - _val.selfBondedStake) > 0)) {
-            uint256 _unbondingSlash = (_remaining * _val.unbondingStake) /
-                                        (_val.unbondingStake + (_val.bondedStake - _val.selfBondedStake));
+        // if _remaining > 0 then bondedStake = delegated stake, because all selfBondedStake is slashed
+        if (_remaining > 0 && (_val.unbondingStake + _val.bondedStake > 0)) {
+            uint256 _unbondingSlash = 0;
+            uint256 _delegatedSlash = 0;
+            if (_val.unbondingStake <= _val.bondedStake) {
```
lorenzo-dev1, line 381: Could you add some comment about why this if and the consequent operations.

It is clear to me know in the context of this PR but it would probably be misterious for someone looking at the entire code.

autonity/solidity/contracts/Accountability.sol
```diff
@@ -374,22 +374,42 @@ contract Accountability is IAccountability {
...
+            _val.bondedStake -= _delegatedSlash;
+
         }
 
         _val.totalSlashed += _slashingAmount;
         _val.provableFaultCount += 1;
-        _val.jailReleaseBlock = block.number + config.jailFactor * _val.provableFaultCount * epochPeriod;
-        _val.state = ValidatorState.jailed; // jailed validators can't participate in consensus
+
+        if (_slashingAmount > 0 && _slashingAmount >= _availableFunds - 1) {
```
lorenzo-dev1, line 398: I am wondering if we can put this check at the start of the function. Meaning that we check if `_slashingAmount >= _availableFunds - 1` ---> we just set all the stake variables to 0 and kill the validator. That would save us some computation

tbssajal: In case of _slashingAmount = _availableFunds - 1, not all stake variables would be 0, there will be 1 wei remaining, but I guess it does not matter much as it is only 1 wei. Still I think others' input would be great on this.

lorenzo-dev1: yeah I think we can even round the `_slashingAmount` to `_availableFunds`

yazzaoui:
I'm okay with `jailbound` on the naming if everyone is happy but I'm a bit worried on the introduced complexity and will need some time for the review.

autonity/solidity/contracts/Accountability.sol
tbssajal: done

autonity/solidity/contracts/Accountability.sol
```diff
@@ -374,22 +374,48 @@ contract Accountability is IAccountability {
...
+            // validator is banned permanently if 100% slashed or _slashingAmount = _availableFunds-1
```
lorenzo-dev1, line 405:
```suggestion
            // validator is jailed permanently if 100% slashed or _slashingAmount = _availableFunds-1
```

I think we can say `jailed` instead of `banned`, as that's what is really happening

autonity/solidity/contracts/Accountability.sol
tbssajal: You are suggesting rounding up `_slashingAmount`? That would be good I think, as it will only increase `_slashingAmount` less than 1 wei.

autonity/solidity/contracts/Accountability.sol
tbssajal: Ah! I will need to regenerate bindings again. I think I will do it with other changes like this one: https://github.com/autonity/autonity/pull/857#discussion_r1387676441

lorenzo-dev1: The bindings shouldn't change if you change just a comment

autonity/solidity/contracts/Accountability.sol
lorenzo-dev1: correct, I think that if rounding up `_slashingAmount` to `_availableFunds` reduces the complexity, it is worth it.

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+      assert.equal((await autonity.balanceOf(delegator)).toNumber(), balance - tokenMint, "balance did not decrease after bonding request");
+      await killValidatorWithSlash(accountabilityConfig, accountability, validator, treasury);
+
+      await truffleAssert.fails(
+        autonity.bond(validator, tokenMint, {from: delegator}),
+        truffleAssert.ErrorType.REVERT,
+        "validator need to be active"
+      );
+      // 2nd bonding should not be applied
+      await utils.endEpoch(autonity, operator, deployer);
```
lorenzo-dev1, line 745: Maybe add a check also on the validator information, to see that the bonding as not been applied on that side as well

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+
+      let validatorInfo = await autonity.getValidator(validator);
+      assert.equal(validatorInfo.state, utils.ValidatorState.jailbound, "validator not jailbound");
+      await truffleAssert.fails(
+        autonity.activateValidator(validator, {from: treasury}),
+        truffleAssert.ErrorType.REVERT,
+        "validator jailed permanently"
+      );
+
+      let releaseBlock = validatorInfo.jailReleaseBlock;
```
lorenzo-dev1, line 709: Since there is not jail release for `jailbound` validators, the `jailReleaseBlock` will be 0 (every test starts with a clean slate), therefore this part of the code is not useful. I think we can remove it (it is just like calling 2 times activateValidator)

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+      valInfo = await autonity.getValidator(validator);
+      let totalStake = Number(valInfo.bondedStake) + Number(valInfo.selfUnbondingStake) + Number(valInfo.unbondingStake);
+      assert.equal(totalStake, 0, "stake remaining after 100% slash");
+    });
+
+    it('does not trigger fairness issue (unbondingStake > 0 and delegatedStake > 0)', async function () {
+      // fairness issue is triggered when delegatedStake or unbondingStake becomes 0 from positive due to slashing
+      // it can happen due to slashing rate = 100%
+      // it can also happen if both unbondingStake and delegatedStake > 0 and slashing amount >= (totalStake - 1)
+      // it should not happen for slashing amount < (totalStake - 1)
```
lorenzo-dev1, line 773: We really need to be sure of this. Do you think it is feasible to add tests for slashingAmount = totalStake - 2? And then verify that the fairness issue does not pop up even if the stake distribution is very much unbalanced between unbondingStake and delegatedStake?

I another words is it possible that we end up with 1 variable between `delegatedStake` or `unbondingStake` = 2 and the other one equal to 0?

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+        await utils.endEpoch(autonity, operator, deployer);
+        let totalStake = selfBondedStake + tokenMint;
+        let tokenUnBond = Math.ceil(tokenMint*tokenUnbondFactor[iter]);
+        await autonity.unbond(validator, tokenUnBond, {from: delegator});
+        await utils.endEpoch(autonity, operator, deployer);
+        let {txEvent, rate} = await slash(config, accountability, 1, validator, validator);
+        // checking if highest possible slashing can be done without triggering fairness issue
+        assert.equal(txEvent.amount.toNumber(), totalStake - 2);
+        valInfo = await autonity.getValidator(validator);
+        assert.equal(valInfo.state, utils.ValidatorState.jailed, "validator not jailbound");
```
lorenzo-dev1, line 801: Do you expect the validator to be jailed or jailbound here?

autonity/solidity/contracts/Accountability.sol
tbssajal: Yeah we will still need it. In my opinion rounding up `_slashingAmount` will change nothing else in the code, but it will be highly unlikely to get `_slashingAmount = _availableFunds - 1`, but still possible. I think we should round it up because if `_availableFunds < slashingRatePrecision` then `_slashingAmount` could be 0 some due to integer division. Rounding up will give us a positive amount of `_slashingAmount` if the validator is to be slashed.

autonity/solidity/contracts/Accountability.sol
tbssajal: It changes surely when I run tests as far as I remember. Although I am not sure if it affects anything or not.

autonity/solidity/test/protocol.js
tbssajal: jailed. will update

autonity/solidity/contracts/Accountability.sol
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/test/protocol.js
tbssajal: done

Jason-Zhangxin-Chen:
There may be additional edge cases to consider, given the increasing complexity of economic protocols. It is essential to establish a suitable approach for formally testing economic-related scenarios around the protocol contract.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1057,6 +1059,12 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         BondingRequest storage _bonding = bondingMap[id];
         Validator storage _validator = validators[_bonding.delegatee];
 
+        // jailbound validator is jailed permanently, no new bonding can be applied for a jailbound validator
+        if (_validator.state == ValidatorState.jailbound) {
+            accounts[_bonding.delegator] += _bonding.amount;
+            return;
```
Jason-Zhangxin-Chen, line 1065: I think it would be helpful to emit an event to tell the community that a certain bonding request is failed due to the validator is now in jailbound state.

autonity/solidity/contracts/Accountability.sol
```diff
@@ -374,22 +374,48 @@ contract Accountability is IAccountability {
...
+            _val.state = ValidatorState.jailbound;
+        } else {
+            _val.jailReleaseBlock = block.number + config.jailFactor * _val.provableFaultCount * epochPeriod;
+            _val.state = ValidatorState.jailed; // jailed validators can't participate in consensus
+        }
 
         autonity.updateValidatorAndTransferSlashedFunds(_val);
 
-        emit SlashingEvent(_val.nodeAddress, _slashingAmount, _val.jailReleaseBlock);
+        if (_val.state == ValidatorState.jailed) {
```
Jason-Zhangxin-Chen, line 414: The emit of the two events could be placed on right after the _val.state being updated at line 406 and 409.

autonity/solidity/test/protocol.js
tbssajal: done

autonity/solidity/contracts/Autonity.sol
tbssajal: done

autonity/solidity/contracts/Accountability.sol
tbssajal: updated

autonity/solidity/contracts/Accountability.sol
lorenzo-dev1: I think so, but we will probably insert it into future releases. It is out of scope of this PR.

Btw yes I think we can do these two checks at the start of the function. For example:

```
if slashingAmount == 0:
  return
```

```
if slashingAmount >= availableFunds - 1:
   set all stake to 0 and jailbound
   return
```

lorenzo-dev1: Yes, we will probably introduce it in next versions of the accountability contract. It is however out of scope of this PR.

I think we can simplify the code by pre-checking the jailbound condition. I.E. moving the check:

`_slashingAmount > 0 && _slashingAmount >= _availableFunds - 1` at the start of the function, and if valid setting all the stake to 0 and jailbounding the validator.

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+      valInfo = await autonity.getValidator(validator);
+      let totalStake = Number(valInfo.bondedStake) + Number(valInfo.selfUnbondingStake) + Number(valInfo.unbondingStake);
+      assert.equal(totalStake, 0, "stake remaining after 100% slash");
+    });
+
+    it('does not trigger fairness issue (unbondingStake > 0 and delegatedStake > 0)', async function () {
+      // fairness issue is triggered when delegatedStake or unbondingStake becomes 0 from positive due to slashing
+      // it can happen due to slashing rate = 100%
+      // it can also happen if both unbondingStake and delegatedStake > 0 and slashing amount >= (totalStake - 1)
+      // it should not happen for slashing amount < (totalStake - 1)
```
lorenzo-dev1, line 773: We figured out it can happen for slashing amount < totalStake - 1, this comment is out of date

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+      assert.equal((await autonity.balanceOf(delegator)).toNumber(), balance, "unexpected balance");
+      let newValInfo = await autonity.getValidator(validator);
+      assert.equal(newValInfo.bondedStake, oldValInfo.bondedStake, "bondedStake changed");
+      assert.equal(newValInfo.selfBondedStake, oldValInfo.selfBondedStake, "selfBondedStake changed");
+      assert.equal(newValInfo.selfUnbondingStake, oldValInfo.selfUnbondingStake, "selfUnbondingStake changed");
+      assert.equal(newValInfo.unbondingStake, oldValInfo.unbondingStake, "unbondingStake changed");
+      assert.equal(newValInfo.liquidSupply, oldValInfo.liquidSupply, "liquidSupply changed");
+    });
+
+    it('zero amount of bonded and unbonding stake after 100% slash', async function () {
```
lorenzo-dev1, line 749: Do we test something different in this test with respect to the test "kills validator for 100% slash" ?

If not, I would suggest to merge the two tests, so that the test remains as lean as possible.

autonity/solidity/test/protocol.js
lorenzo-dev1: Also could you add a comment explaining why you are setting the collusionFactor to that value, it is a bit hard to understand the test

autonity/solidity/test/protocol.js
lorenzo-dev1: Seems to me like the only difference is that the first one has unbonded stake while the second one has not?

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+      // it can happen due to slashing rate = 100%
+      // it can also happen if both unbondingStake and delegatedStake > 0 and slashing amount >= (totalStake - 1)
+      // it should not happen for slashing amount < (totalStake - 1)
+      let config = JSON.parse(JSON.stringify(accountabilityConfig));
+      config.collusionFactor = parseInt(config.slashingRatePrecision) - parseInt(config.baseSlashingRateMid) - 2;
+      accountability = await AccountabilityTest.new(autonity.address, config, {from: deployer});
+      await autonity.setAccountabilityContract(accountability.address, {from:operator});
+
+      const validatorAddresses = [ validators[0].nodeAddress, validators[1].nodeAddress];
+      const tokenUnbondFactor = [1/10, 8/10];
```
lorenzo-dev1, line 780: Maybe it would be better to have a more unbalanced stake distribution, since we want to check that our mechanism is resilient even in that case.

autonity/solidity/test/protocol.js
```diff
@@ -699,5 +684,207 @@ contract('Protocol', function (accounts) {
...
+        assert.equal(valInfo.state, utils.ValidatorState.jailbound, "validator not jailbound");
+        assert.equal(
+          parseInt(valInfo.bondedStake) + parseInt(valInfo.unbondingStake) + parseInt(valInfo.selfUnbondingStake)
+          , 0, "100% slash did not happen"
+        );
+      }
+      
+    });
+
+    it('kills validator for (slashingAmount = totalStake - 1)', async function () {
```
lorenzo-dev1, line 836: Also here, if you could just add a few hints on the reason behind operations like:
- `        const tokenMint = parseInt(config.slashingRatePrecision) - totalStake;`
- `  config.collusionFactor = parseInt(config.slashingRatePrecision) - parseInt(config.baseSlashingRateMid) - 1;`

It would be great and it will make the tests more maintenable when someone else will pick them up to update them

autonity/solidity/test/protocol.js
tbssajal: correct, will merge

autonity/solidity/contracts/Accountability.sol
tbssajal: updated

autonity/solidity/test/protocol.js
tbssajal: updated

autonity/solidity/test/protocol.js
tbssajal: tried to make it more readable.

tbssajal:
Rebased against develop branch, asm tests are still failing

lorenzo-dev1:
> Rebased against develop branch, asm tests are still failing 

yes it seems like there is another failure in the ASM tests. It is different wrt to the one of last time, so we need to investigate it.

@tbssajal  For the PR, I think that there is a modification in the golang code that is missing. It is the handling of consensus messages for jailbound validators. Basically if you look at `faultyValidatorsWatcher` in engine.go, you will see that inside the node we listen for `SlashingEvent` events to update the list of jailed validators. Then this list is used in the handler to ignore messages that come from jailed validators.

We should probably ignore messages coming from jailbound validators too, at least until the end of the epoch where they got jailbound. Then they will go out of the committee and never get re-selected, so we can "forget" that they are jailbound in the client (i.e. no need to keep a list of jailbound validators in the client).

autonity/solidity/test/utils.js
```diff
@@ -271,6 +275,65 @@ function bytesToHex(bytes) {
   return hex;
 }
 
+function randomInt() {
```
lorenzo-dev1, line 278: These utilities functions are going to be very useful for testing. I searched for them a lot in the past but never found implementations, did you find them somewhere or you just coded them up from definitions?

autonity/solidity/test/protocol.js
```diff
@@ -746,79 +746,93 @@ contract('Protocol', function (accounts) {
...
+      for (let iter = 0; iter < validatorAddresses.length; iter++) {
+        const validator = validatorAddresses[iter];
+        let {txEvent, _} = await slash(config, accountability, 1, validator, validator);
+        // checking if highest possible slashing can be done without triggering fairness issue
+        assert.equal(txEvent.amount.toNumber(), expectedSlash, "highest slash did not happen");
+        let validatorInfo = await autonity.getValidator(validator);
+        assert.equal(validatorInfo.state, utils.ValidatorState.jailed, "validator not jailed");
+        assert(parseInt(validatorInfo.bondedStake) > 0 && parseInt(validatorInfo.unbondingStake) > 0, "fairness issue triggered");
+      }
+      assert.equal((await autonity.balanceOf(delegator)).toNumber(), balance, "unbonding released");
```
lorenzo-dev1, line 808: I think if we want to check that we need first to call utils.mineTillUnbondingRelease right?

autonity/solidity/test/utils.js
tbssajal: First I was using math.random(), later found on the internet that it only gives less than 1 and also found this piece of code.

autonity/solidity/test/protocol.js
```diff
@@ -829,37 +843,40 @@ contract('Protocol', function (accounts) {
           parseInt(valInfo.bondedStake) + parseInt(valInfo.unbondingStake) + parseInt(valInfo.selfUnbondingStake)
           , 0, "100% slash did not happen"
         );
+        assert.equal((await autonity.balanceOf(delegator)).toNumber(), balances[iter], "unbonding released");
```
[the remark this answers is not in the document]
tbssajal: Oh! right.

tbssajal:
> so we can "forget" that they are jailbound in the client

Is it handled in the communication level that we only receive message from the committee members and discard the message in the low level if it is from someone outside the committee? Can someone please point me to the related codes?

autonity/solidity/test/utils.js
lorenzo-dev1: Sorry, maybe I explained myself wrongly. The previous comment was meant for the other utilities function you added (e.g. to covert pubkey to enode, priv to public, etc. etc).

autonity/solidity/contracts/Accountability.sol
```diff
@@ -333,6 +333,14 @@ contract Accountability is IAccountability {
         // last reporter is the beneficiary
         beneficiaries[_event.offender] = _event.reporter;
 
+        // if already jailbound, validator has 0 stake
+        if (_val.state == ValidatorState.jailbound) {
```
lorenzo-dev1, line 337: I would just return if this situation happens.

Provable fault counts is used for slashing, but once the validator is jailbound it will not be slashed anymore. Moreover we don't care about statistics/reputational data, as it is dead forever anyways.

autonity/solidity/test/utils.js
tbssajal: Oh! Mostly I found on the internet. I just coded the address method from definition.

autonity/solidity/contracts/Accountability.sol
tbssajal: updated

autonity/solidity/test/protocol.js
tbssajal: updated

autonity/solidity/test/protocol.js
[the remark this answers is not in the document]
tbssajal: updated

m4ksio:
Good change! However since it touches core logic of contracts I would really like to see tests for the case, especially good is we could reproduce the problem before fix

lorenzo-dev1:
> Good change! However since it touches core logic of contracts I would really like to see tests for the case, especially good is we could reproduce the problem before fix

You can find contract test in truffle under `autonity/solidity/test`. There are some test that reproduce the issue.

tbssajal:
> Good change! However since it touches core logic of contracts I would really like to see tests for the case, especially good is we could reproduce the problem before fix

Test labelled 'does not trigger fairness issue (unbondingStake > 0 and delegatedStake > 0)' in particular triggers the issue #860 which is fixed by changing the calculation of `delegatedSlash` and `unbondingSlash`.

m4ksio:
I have just reviewed PR related to this one, and found that distinction between `jailed` and `jailbound` are not very easy on the eye. I would suggest that while this is still WIP we use some other term.

tbssajal:
> I have just reviewed PR related to this one, and found that distinction between `jailed` and `jailbound` are not very easy on the eye. I would suggest that while this is still WIP we use some other term.

_gulag_ sounds nice to me originally suggested by Robert.

lorenzo-dev1:
I personally think `jailbound` is fine, but if you want to change I would take something with no political connotation

lorenzo-dev1 (approved):
LGTM, pending integration of https://github.com/autonity/autonity/pull/866

autonity/solidity/contracts/interfaces/IAccountability.sol
```diff
@@ -42,4 +42,9 @@ interface IAccountability {
     * @dev Event emitted after a successful slashing.
     */
     event SlashingEvent(address validator, uint256 amount, uint256 releaseBlock);
+
+    /**
+    * @dev Event emitted after validator is banned permanently
+    */
+    event ValidatorJailbound(address validator, uint256 amount);
```
yazzaoui, line 49: I'm not really in favor of introducing a new event, I think from the consumer point of view it would be much simpler to just add a flag to `SlashingEvent` if the validator becomes `jailbound` .

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1057,6 +1060,13 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         BondingRequest storage _bonding = bondingMap[id];
         Validator storage _validator = validators[_bonding.delegatee];
 
+        // jailbound validator is jailed permanently, no new bonding can be applied for a jailbound validator
+        if (_validator.state == ValidatorState.jailbound) {
```
tbssajal, line 1064: In light of #895 should we change the check to `if (_validator.state == ValidatorState.jailbound || _validator.state == ValidatorState.jailed)` or `if (_validator.state != ValidatorState.active)`? If we change to `if (_validator.state != ValidatorState.active)` then paused validator's bonding request will also be rejected.

autonity/solidity/contracts/interfaces/IAccountability.sol
tbssajal: Yeah, makes sense

tbssajal: done

lorenzo-dev1 (approved):

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: I think it should be addressed in a separate PR

tbssajal: Yup

yazzaoui (approved):
LGTM

PR #858, opened 2023-11-08
feature, consensus key on boarding and management

Jason-Zhangxin-Chen:
This PR contains bellow things:
- a BLS go wrapper of the BLST C lib.
- on-boarding validators from autonity contract with node's activity key and its corresponding proof.
- output the node's activity key and the corresponding signatures(proofs) via cli command: genOwnershipProof.


autonity/solidity/contracts/Autonity.sol
```diff
@@ -274,12 +277,19 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+    * @param _signatures is a combination of two ecdsa signatures, and a bls signature as the ownership proof of the
+    * activity key appended sequentially. The 1st two ecdsa signatures are in below order:
         1. a message containing treasury account and signed by validator account private key .
         2. a message containing treasury account and signed by Oracle account private key .
     * @dev Emit a {RegisteredValidator} event.
     */
-    function registerValidator(string memory _enode, address _oracleAddress, bytes memory _multisig) public {
+    function registerValidator(string memory _enode, address _oracleAddress, bytes memory _activityKey, bytes memory _signatures) public {
+        require(_signatures.length == 226, "Invalid proof length");
+        require(_activityKey.length == 48, "Invalid activity key length");
```
lorenzo-dev1, line 290: Maybe I am being too pedantic here, but I think that only the activityKeyProof should be bundled with the _signatures.

The activityKey itself should be separated in its own variable as _enode and _oracleAddress are.

Like this it works but it is not very semantically correct (activityKey is not _signature, thus it should not be part of _signatures)

Jason-Zhangxin-Chen: There is a _activityKey parameter added already, I don't know on which version are you reviewing, but it was already there, don't known if this is a issue in github.

.github/workflows/go.yml
```diff
@@ -22,7 +22,11 @@ env:
   GO_CACHE: /home/runner/.cache/go-build # This is the location of the go build cache.
   GO_MODULES_CACHE: /home/runner/go/pkg # This is where module source code is stored.
   GO_VERSION: "1.20" # 1.15 seems to cause tests to fail
-
+  # Enable a flag to build a portable version of the BLST C lib on legacy hardware without HW hash feature on GitHub CI
+  # runners. This replaces the HW instruction of sha256 with a software sha256. However, modern hardware running
+  # Autonity client does not need this flag. When building the binary or Docker image for the client, do not enable this
+  # flag.
+  CGO_CFLAGS: -O -D__BLST_PORTABLE__ # the CGO flags to compile a portable BlST C lib in CI jobs for legacy hardware.
```
yazzaoui, line 29: Seems like runtime detection is now supported in the last version of this library. See https://github.com/supranational/blst/issues/10#issuecomment-1629132181.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -43,12 +43,15 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
         uint256 provableFaultCount;
+        bytes activityKey;
+        uint256 omissionFaultCount;
         ValidatorState state;
     }
 
     struct CommitteeMember {
         address addr;
         uint256 votingPower;
+        bytes activityKey;
```
yazzaoui, line 54: This is not needed for this PR.

yazzaoui:
Sooo, I'm in the opinion that we should rename (again) `activityKey` to `consensusKey` or something else as we're no longer planning to use this key only for accountability purposes but for signing consensus messages as well.

Jason-Zhangxin-Chen:
> Sooo, I'm in the opinion that we should rename (again) `activityKey` to `consensusKey` or something else as we're no longer planning to use this key only for accountability purposes but for signing consensus messages as well.

If this key will replace the functionality of ecdsa key from the consensus perspective, then I think `validatorKey` could be a good one.

.github/workflows/go.yml
Jason-Zhangxin-Chen: > Seems like runtime detection is now supported in the last version of this library. See [supranational/blst#10 (comment)](https://github.com/supranational/blst/issues/10#issuecomment-1629132181).

Ah, I see they resolved the issue with a runtime detection in the latest release. I removed this flag via b2483457a60c68a8734b9a0aaa48818fe30b6500

autonity/solidity/contracts/Autonity.sol
```diff
@@ -43,12 +43,15 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         uint256 totalSlashed;
         uint256 jailReleaseBlock;
         uint256 provableFaultCount;
+        bytes activityKey;
+        uint256 omissionFaultCount;
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > Out of scope for this PR. Let's focus only on validator onboarding.

The counter is removed via b2483457a60c68a8734b9a0aaa48818fe30b6500 for now.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -43,12 +43,14 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         uint256 totalSlashed;
         uint256 jailReleaseBlock;
         uint256 provableFaultCount;
+        bytes validatorKey;
```
tbssajal, line 46: This is always a BLS public key?

tbssajal: Because if we maintain both bls and ecdsa, it could be confusing.

tbssajal: Got it

crypto/bls/bls.go
```diff
@@ -0,0 +1,101 @@
...
+	if !sig.Verify(pubKey, hash.Bytes()) {
+		return fmt.Errorf("cannot verify proof")
+	}
+
+	return nil
+}
+
+func GenerateValidatorKeyProof(priKey SecretKey, msg []byte) ([]byte, error) {
+	// generate bls key ownership proof
+	m := append([]byte(MsgPrefix), msg...)
```
lorenzo-dev1, line 28: I understand that here the `MsgPrefix="Auton"` is used with the intention of securing the Proof Of Possession (POP) generation. It is indeed [the recommended guideline](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-3.3.1) to use a Domain Separator Tag (DST) when generating the POP.

However I think the way it is implemented is wrong, as the DST should be passed as a parameter to the `Sign` method (see https://github.com/autonity/autonity/blob/onboarding-activity-key/crypto/bls/blst/secret_key.go#L84, third parameter) not concatenated to the message.

I think it might be useful to generate wrappers such as `PopProve` and `PopVerify` as done here: https://github.com/status-im/nim-blscurve/blob/02d4848d8444b12d6935aa57fbbb31988ba0a0b8/blscurve/bls_sig_min_pubkey.nim#L34

crypto/bls/bls.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/autonity/autonity/crypto/bls/blst"
+	"github.com/autonity/autonity/crypto/bls/common"
+	"github.com/pkg/errors"
+)
+
+const MsgPrefix = "Auton"
+
+func ValidateValidatorKeyProof(pubKey common.BLSPublicKey, sig common.BLSSignature, msg []byte) error {
+	m := append([]byte(MsgPrefix), msg...)
+	hash := crypto.Keccak256Hash(m)
```
lorenzo-dev1, line 18: Since we moved to blake2b for hashing, might as well turn this hash to blake2b as well.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -274,12 +276,19 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
         1. a message containing treasury account and signed by validator account private key .
         2. a message containing treasury account and signed by Oracle account private key .
     * @dev Emit a {RegisteredValidator} event.
     */
-    function registerValidator(string memory _enode, address _oracleAddress, bytes memory _multisig) public {
+    function registerValidator(string memory _enode, address _oracleAddress, bytes memory _validatorKey, bytes memory _signatures) public {
+        require(_signatures.length == 226, "Invalid proof length");
+        require(_validatorKey.length == 48, "Invalid validator key length");
+        bytes memory signatures = BytesLib.slice(_signatures, 0, 130);
+        bytes memory validatorKeyProof = BytesLib.slice(_signatures, 130, 96);
```
yazzaoui, line 291: better to keep those checks in `_registerValidator` since it's already the place where it is happening

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1007,12 +1021,13 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
-        for (uint i = 32; i < _multisig.length; i += 65) {
-            (r, s, v) = Helpers.extractRSV(_multisig, i);
-            signers[i / 65] = ecrecover(hashedData, v, r, s);
+        for (uint i = 32; i < _signatures.length; i += 65) {
+            (r, s, v) = Helpers.extractRSV(_signatures, i);
+            signers[i/65] = ecrecover(hashedData, v, r, s);
         }
         require(signers[0] == _validator.nodeAddress, "Invalid node key ownership proof provided");
         require(signers[1] == _validator.oracleAddress, "Invalid oracle key ownership proof provided");
+        require(Precompiled.checkValidatorKeyProof(_validatorKey, _validatorKeyProof, _validator.treasury) == 1, "Invalid validator key proof for registration");
```
yazzaoui, line 1030: Minor point (I'm guilty of it too): the correct terminology is _Signature Verification_  not "signature validation" and not "signature check".

autonity/solidity/contracts/Autonity.sol
```diff
@@ -994,8 +1009,7 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         _deployLiquidContract(_validator);
     }
 
-    function _registerValidator(Validator memory _validator, bytes memory _multisig) internal {
-        require(_multisig.length == 130, "Invalid proof length");
+    function _registerValidator(Validator memory _validator, bytes memory _signatures, bytes memory _validatorKey, bytes memory _validatorKeyProof) internal {
```
yazzaoui, line 1012: Refactoring improvement since you are working on it:
We already have a `_registerValidator(Validator memory _validator)` function and it is confusing..  If you can I would suggest, for a better description, renaming this function to `_veritfyAndRegisterValidator` and have it internally call `_registerValidator`. 
Outside the scope of this PR, the logic in `_deployLiquidContract` which is doing not only the deployment but the validator insertion in state memory need to be refactored.

crypto/bls/bls.go
```diff
@@ -0,0 +1,101 @@
...
+	return blst.PublicKeyFromBytes(pubKey)
+}
+
+// SignatureFromBytes creates a BLS signature from a LittleEndian byte slice.
+func SignatureFromBytes(sig []byte) (Signature, error) {
+	return blst.SignatureFromBytes(sig)
+}
+
+// AggregatePublicKeys aggregates the provided raw public keys into a single key.
+func AggregatePublicKeys(pubs [][]byte) (PublicKey, error) {
```
yazzaoui, line 79: It looks like most of those functions are wrappers over the `blst` package. Why do we need in this case a dedicated `bls` package. Is it possible to merge `bls` and `blst` together?

crypto/bls/bls.go
Jason-Zhangxin-Chen: > Referen

The DST is already specified in the wrapper, see my commits: 6ce810515189eb2e511ca48f36a1bffbe60dbf4b and my refine in 4608dfe0bc424daaf4cea66b65a128f3680c8160

crypto/bls/bls.go
Jason-Zhangxin-Chen: resolved in https://github.com/autonity/autonity/commit/4608dfe0bc424daaf4cea66b65a128f3680c8160

crypto/bls/common/constants.go
```diff
@@ -0,0 +1,14 @@
...
+
+const (
+	BLSSecretKeyLength       = 32
+	BLSPubkeyLength          = 48
+	BLSSignatureLength       = 96
+	BLSPubKeyHexStringLength = 98
+)
+
+// ZeroSecretKey represents a zero secret key.
+var ZeroSecretKey = [32]byte{}
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > I think that `ZeroSecretKey` and `InfinitePublicKey` should be moved to the test since they are used only there. We do not want to rely on them on any check in the library itself anyways, as comparing keys in serialized byte format can lead to security issues like the Zero Bug (https://eprint.iacr.org/2021/323.pdf)

moved in 92f8d053997a2a426c5dce20497065d8fba8cdf9

autonity/solidity/contracts/Autonity.sol
vpiyush, line 291: since there are too many numbers now, It would be a could idea to define some constants which would be meaningful

autonity/solidity/contracts/Autonity.sol
vpiyush, line 46: this can just be called "key". It's already a member of validator

autonity/solidity/contracts/Autonity.sol
```diff
@@ -43,12 +43,14 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
         uint256 jailReleaseBlock;
         uint256 provableFaultCount;
+        bytes validatorKey;
         ValidatorState state;
     }
 
     struct CommitteeMember {
         address addr;
         uint256 votingPower;
+        bytes validatorKey;
```
vpiyush, line 53: Also here, could be called key

crypto/autonity_crypto.go
```diff
@@ -46,3 +49,27 @@ func hexDecode(src []byte) ([]byte, error) {
...
+	proof := priKey.Sign(Hash(m).Bytes())
+
+	err := PopVerify(priKey.PublicKey(), proof, msg)
+	if err != nil {
+		return nil, err
+	}
+	return proof.Marshal(), nil
+}
+
+func PopVerify(pubKey common2.BLSPublicKey, sig common2.BLSSignature, msg []byte) error {
```
vpiyush, line 68: Same here

consensus/tendermint/accountability/contracts.go
```diff
@@ -22,29 +24,30 @@ import (
 // a part of consensus.
 
 var (
+	checkActivityKeyOwnershipAddress = common.BytesToAddress([]byte{251}) //0xfb
```
vpiyush, line 27: shouldn't we use the hex format as it is? This helps during grepping these addresses. I suggest we use []byte{0xfb} instead of []byte{251}.  Also applicable for other addresses i would say

consensus/tendermint/accountability/contracts.go
```diff
@@ -56,6 +59,37 @@ func LoadPrecompiles(chain ChainContext) {
...
+func (b *ValidatorKeyProofVerifier) RequiredGas(_ []byte) uint64 {
+	return params.AutonityActivityKeyCheckGas
+}
+
+func (b *ValidatorKeyProofVerifier) Run(input []byte, _ uint64) ([]byte, error) {
+	if len(input) != 196 {
+		return failure32Byte, fmt.Errorf("invalid proof - empty")
+	}
+
+	keyBytes := input[32:80]
```
vpiyush, line 73: constants in this function would help understanding these numbers

crypto/bls/blst/signature.go
```diff
@@ -0,0 +1,207 @@
+package blst
```
vpiyush, line 1: What's the idea behind blst and bls package? Why do we need 2?

crypto/bls/bls.go
```diff
@@ -0,0 +1,101 @@
...
+
+	err := ValidateValidatorKeyProof(priKey.PublicKey(), proof, msg)
+	if err != nil {
+		return nil, err
+	}
+	return proof.Marshal(), nil
+}
+
+// SecretKeyFromECDSAKey generate a BLS private key by sourcing from an ECDSA private key.
+func SecretKeyFromECDSAKey(sk *ecdsa.PrivateKey) (SecretKey, error) {
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > Maybe that works but I'm of the opinion we must follow cryptographic good practices and follow established standards. Let's use a random number generator just like `crypto.GenerateKey` is doing it, instead of using a private key as a seed.

Resolved it partially in b0288c03dfab9ed6c06451460902e47bef66548d, on top of node key management PR, I will remove this function since there is store for BLS key yet currently, thus we still derive it from ECDSA.

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > Minor point (I'm guilty of it too): the correct terminology is _Signature Verification_ not "signature validation" and not "signature check".

resolved in dd30d2336ad0fd16a3ca3cb1b77f5e0a792cdb76

autonity/solidity/test/validator_management.js
```diff
@@ -0,0 +1,468 @@
...
+function generateMultiSig(nodekey,oraclekey,treasuryAddr) {
+    let treasuryProof = web3.eth.accounts.sign(treasuryAddr, nodekey);
+    let oracleProof = web3.eth.accounts.sign(treasuryAddr, oraclekey);
+    let multisig = treasuryProof.signature + oracleProof.signature.substring(2)
+    return multisig
+}
+
+contract('Autonity', function (accounts) {
+
+    for (let i = 0; i < accounts.length; i++) {
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > This config is shared with autonity.js. Please refactor this.

resolved in 8718774377a1aa67fbe7229638483048eaa90322

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > better to keep those checks in `_registerValidator` since it's already the place where it is happening

resolved in 8fb52226f95e7dbd9157f94641cfb606780c1b68

autonity/solidity/contracts/Autonity.sol
```diff
@@ -299,10 +308,11 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
             0,                       // total slashed
             0,                       // jail release block
             0,                       // provable faults count
+            _validatorKey,            // validator key in bytes
             ValidatorState.active    // state
         );
 
-        _registerValidator(_val, _multisig);
+        _registerValidator(_val, signatures, _validatorKey, validatorKeyProof);
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > `_val` already has the validator key information, you can use it directly. I would suggest to revert to `_registerValidator(Validator memory _validator, bytes memory _signatures)`

resolved in 8fb52226f95e7dbd9157f94641cfb606780c1b68

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > Refactoring improvement since you are working on it: We already have a `_registerValidator(Validator memory _validator)` function and it is confusing.. If you can I would suggest, for a better description, renaming this function to `_veritfyAndRegisterValidator` and have it internally call `_registerValidator`. Outside the scope of this PR, the logic in `_deployLiquidContract` which is doing not only the deployment but the validator insertion in state memory need to be refactored.

resolved in 8fb52226f95e7dbd9157f94641cfb606780c1b68. The legacy _registerValidator() will be replaced by _verifyAndRegisterValidator() once we ask PoP on genesis file.

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > since there are too many numbers now, It would be a good idea to define some constants which would be meaningful

resolved in 8fb52226f95e7dbd9157f94641cfb606780c1b68

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > this can just be called "key". It's already a member of validator

resolved in 2a4c35fed4caa809051eb43cc834ee1a65bb1c78

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > Also here, could be called key

resolved in 2a4c35fed4caa809051eb43cc834ee1a65bb1c78

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -48,4 +49,26 @@ library Precompiled {
...
+
+    // @dev check the validator key proof in a precompiled contract.
+    // @param _validatorKey is a "0x" prefix hex string of the validator's BLS public key.
+    // @param _proof is a "0x" prefix hex string of the proof generated together with the bls public key.
+    // @param _treasury is a "0x" prefix hex string of the validator's treasury account.
+    // @return 0 for a failure, 1 for a successful check.
+    function checkValidatorKeyProof(bytes memory _validatorKey, bytes memory _proof, address _treasury) internal view returns (uint256) {
+        uint256[1] memory retVal;
+        bytes memory input = BytesLib.concat(BytesLib.concat(_validatorKey, _proof), abi.encodePacked(_treasury));
+        address validatorKeyVerifierAddress = address(0xfb);
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > 0xfb should be given a name (constant), these names help when we search for them.

resolved in 2a4c35fed4caa809051eb43cc834ee1a65bb1c78

crypto/autonity_crypto.go
```diff
@@ -46,3 +49,27 @@ func hexDecode(src []byte) ([]byte, error) {
 	_, err := hex.Decode(dst, src)
 	return dst, err
 }
+
+// PopProof generate the proof of possession of the validator key and the treasury account in Autonity when onboarding a
+// validator. The native input msg just contains validator's treasury address. Note: DST is already specificed at
+// bls/blst/signatures.go line: 21.
+func PopProof(priKey bls.SecretKey, msg []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > Pop is an abbreviation It should be used as POP(all caps) to reflect the correct meaning. Otherwise, it gets confusing with the word Pop.

resolved in 2a4c35fed4caa809051eb43cc834ee1a65bb1c78

autonity/solidity/contracts/Autonity.sol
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > IMO we should also define a constant for 1, as Success, in a similar manner we have it in golang

resolved in 2a4c35fed4caa809051eb43cc834ee1a65bb1c78

consensus/tendermint/accountability/contracts.go
Jason-Zhangxin-Chen: > constants in this function would help understanding these numbers

resolved in 2a4c35fed4caa809051eb43cc834ee1a65bb1c78

crypto/bls/blst/signature.go
Jason-Zhangxin-Chen: > What's the idea behind blst and bls package? Why do we need 2?

The original idea was to keep multiple implemenation of BLS, for example blst is one of them. Now they are merged into one in 098858f1b51560a20c5602f73afd26c4e46103f8

autonity/solidity/contracts/MockEnodeVerifier.sol
```diff
@@ -62,8 +62,8 @@ contract MockEnodeVerifier {
     	if(enode.length < 136) {
     		err[0] = bytes1(uint8(1));
     		assembly { mstore(add(ret, 64), mload(add(err,32))) }
-   		return ret;
-   	}
+   		    return ret;
+   	    }
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > This seems an indentation issue, also present in other places in the PR

resolve in 1c64ccfd26bc28d1cf066de9976ac5a669fe5ee4

autonity/solidity/contracts/Autonity.sol
```diff
@@ -43,12 +47,15 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         uint256 totalSlashed;
         uint256 jailReleaseBlock;
         uint256 provableFaultCount;
+        bytes consensusKey;
+        bytes pop;
```
yazzaoui, line 51: Why saving the proof of possession in contract state indefinitely? I can't think of any reason for that as it's only useful at registration time.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -21,6 +21,10 @@ uint8 constant DECIMALS = 18;
 
 contract Autonity is IAutonity, IERC20, Upgradeable {
     uint256 internal constant MAX_ROUND = 99;
+    uint256 internal constant CONSENSUS_KEY_LEN = 48;
+    uint256 internal constant SIGNATURE_LEN = 96;
```
yazzaoui, line 25:
```suggestion
    uint256 internal constant BLS_PROOF_LEN = 96;
```

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -48,4 +53,26 @@ library Precompiled {
...
     }
+
+    // @dev verify the proof of possession of validator key in a precompiled contract.
+    // @param _consensusKey is a "0x" prefix hex string of the validator's BLS public key.
+    // @param _proof is a "0x" prefix hex string of the proof generated together with the bls public key.
+    // @param _treasury is a "0x" prefix hex string of the validator's treasury account.
+    // @return 0 for a failure, 1 for a successful check.
+    function popVerification(bytes memory _consensusKey, bytes memory _proof, address _treasury) internal view returns (uint256) {
+        uint256[1] memory retVal;
+        bytes memory input = BytesLib.concat(BytesLib.concat(_consensusKey, _proof), abi.encodePacked(_treasury));
```
yazzaoui, line 64: Why not doing `abi.encode(_consensusKey,_proof,_treasury)` more simply?

params/protocol_contracts.go
```diff
@@ -131,9 +133,117 @@ type Validator struct {
 	TotalSlashed             *big.Int
 	JailReleaseBlock         *big.Int
 	ProvableFaultCount       *big.Int
+	ConsensusKey             []byte //ABI packing does not support hexutil.Bytes, thus we need to introduce customized JSON Marshal/UnMarshal methods.
+	Pop                      []byte // the proof of ownership generated by sub command: genOwnershipProof
```
yazzaoui, line 137: No need for that. GVs are not required to have a proof-of-possession.

cmd/autonity/misccmd.go
```diff
@@ -199,45 +237,40 @@ func genOwnershipProof(ctx *cli.Context) error {
...
+		if err == hexutil.ErrMissingPrefix {
+			utils.Fatalf("Failed to decode: hex string without 0x prefix")
+		}
+		utils.Fatalf("Failed to generate Autonity POP: %v", err)
 	}
-	multisig := append(nodeSignature[:], oracleSignature[:]...)
-	hexStr := hexutil.Encode(multisig)
-	fmt.Println("Signature hex:", hexStr)
+
+	fmt.Println("Node consensus key:", consensusKey.PublicKey().Hex())
```
yazzaoui, line 248:
```suggestion
	fmt.Println("Consensus public key: ", consensusKey.PublicKey().Hex())
```

cmd/autonity/misccmd.go
```diff
@@ -199,45 +237,40 @@ func genOwnershipProof(ctx *cli.Context) error {
...
+	fmt.Println("Signatures:", hexutil.Encode(signatures))
```
yazzaoui, line 249:
```suggestion
	fmt.Println(hexutil.Encode(signatures))
```
Make it easier to script.

cmd/autonity/misccmd.go
```diff
@@ -199,45 +237,40 @@ func genOwnershipProof(ctx *cli.Context) error {
...
+	if err = crypto.SaveNodeKey(outKeyFile, nodeKey, consensusKey); err != nil {
 		utils.Fatalf("could not save key %v", err)
 	}
+
 	writeAddr := ctx.GlobalBool(utils.WriteAddrFlag.Name)
 	if writeAddr {
-		fmt.Printf("%x\n", crypto.FromECDSAPub(&nodeKey.PublicKey)[1:])
+		fmt.Printf("Node address: %s\n", crypto.PubkeyToAddress(nodeKey.PublicKey).String())
+		fmt.Printf("Node public key: 0x%x\n", crypto.FromECDSAPub(&nodeKey.PublicKey)[1:])
+		fmt.Println("Node consensus key:", consensusKey.PublicKey().Hex())
```
yazzaoui, line 273:
```suggestion
	fmt.Println("Consensus public key: ", consensusKey.PublicKey().Hex())
```

yazzaoui:
Points to discuss with the team, not strictly part of this PR:
- rename the `genOwnershiproof` command to the more succinct `genpop`.
- add the public consensus key to the enode format such that `enode://NODEPUBKEY+CONSENSUSPUBKEY@IP`  
- have the consensus public key appearing with `aut node info`

params/protocol_contracts.go
lorenzo-dev1: We shall not forget to ask PoPs from non AGF controlled GVs. Even offline verification would be fine, but we still need a PoP.

consensus/tendermint/accountability/contracts.go
```diff
@@ -56,6 +63,41 @@ func LoadPrecompiles(chain ChainContext) {
 	setPrecompiles(vm.PrecompiledContractsBLS)
 }
 
+// POPVerifier verifies the proof of possession of validator key.
+type POPVerifier struct{}
+
+func (b *POPVerifier) RequiredGas(_ []byte) uint64 {
+	return params.AutonityActivityKeyCheckGas
```
lorenzo-dev1, line 70: Here and in this file in general the bls key is still referred to as `activityKey`, but I think now we are calling it consensus key? @yazzaoui

consensus/tendermint/accountability/contracts.go
```diff
@@ -22,29 +24,34 @@ import (
 // a part of consensus.
 
 var (
+	checkActivityKeyOwnershipAddress = common.BytesToAddress([]byte{0xfb})
```
lorenzo-dev1, line 27: I am not sure that the bls ownership proof precompiled belongs to `/accountability/contracts.go`. I think it could go into the general precompiled under `/vm/contracts.go`

crypto/autonity_crypto.go
```diff
@@ -46,3 +132,125 @@ func hexDecode(src []byte) ([]byte, error) {
...
+	blsPOPProof, err := BLSPOPProof(consensusKey, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	signatures := append(append(nodeSignature[:], oracleSignature[:]...), blsPOPProof[:]...)
+	return signatures, nil
+}
+
+func AutonityPOPVerify(signatures []byte, treasuryHex string, nodeAddress, oracleAddress common.Address, consensusKey []byte) error {
```
lorenzo-dev1, line 188: Is this function used?

params/protocol_params.go
```diff
@@ -146,6 +146,7 @@ const (
 	Bn256PairingPerPointGasIstanbul  uint64 = 34000  // Per-point price for an elliptic curve pairing check
 
 	AutonityEnodeCheckGas       uint64 = 5000 // Price for enode checker precompiled contracts
+	AutonityActivityKeyCheckGas uint64 = 5000 // Price for the bls key ownership verifier precompiled contract
```
lorenzo-dev1, line 149: same here, not called activity key anymore I think

params/config.go
```diff
@@ -494,8 +504,9 @@ var (
 
 func init() {
 	TestAutonityContractConfig.Prepare()
-	PiccaddillyChainConfig.AutonityContractConfig.Prepare()
-	BakerlooChainConfig.AutonityContractConfig.Prepare()
+	// todo: ask Raj to generate POP for Piccaddily and Bakerloo network since we don't know the secret keys.
+	//PiccaddillyChainConfig.AutonityContractConfig.Prepare()
```
lorenzo-dev1, line 508: Maybe it would be better to set up some fake keys for piccadilly and bakerlooo so that we do not have to comment out this code. And maybe print a warning like "Using test keys" or something similar just to inform the operator...

consensus/tendermint/accountability/contracts.go
yazzaoui: Yeah let's rename that to `params.POPVerifierGas` and make sure all references to "activityKey" are renamed to "consensusKey".

consensus/tendermint/accountability/contracts.go
yazzaoui: Agree, let's move the `POPVerifier` precompile to `core/vm/contracts.go`

crypto/autonity_crypto.go
```diff
@@ -1,12 +1,98 @@
...
+// LoadNodeKey loads a secp256k1 private key and a derived validator BLS private key from the given file.
+func LoadNodeKey(file string) (*ecdsa.PrivateKey, blst.SecretKey, error) {
+	fd, err := os.Open(file)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer fd.Close()
+
+	r := bufio.NewReader(fd)
+	buf := make([]byte, AutonityNodeKeyLenInChar)
```
lorenzo-dev1, line 43: If I understand correctly `AutonityNodeKeyLenInChar` now is the length of the nodekey + the bls key. Therefore I think it should be renamed. Also the name of the method is not really appropriate (`LoadNodeKey`).

I think we should find a term to refer to node key + consensus key (bls key). Something like `Client Keys` maybe? or `Autonity Keys`.

@Jason-Zhangxin-Chen @yazzaoui  ideas?

node/config.go
```diff
@@ -393,19 +397,27 @@ func (c *Config) NodeKey() *ecdsa.PrivateKey {
...
+			log.Crit(fmt.Sprintf("Failed to generate ephemeral node key: %v", err))
+		}
+
 		return key
 	}
 
 	keyfile := c.ResolvePath(datadirPrivateKey)
-	if key, err := crypto.LoadECDSA(keyfile); err == nil {
+	if key, consensusKey, err := crypto.LoadNodeKey(keyfile); err == nil {
+		c.ConsensusKey = consensusKey
```
lorenzo-dev1, line 417: Just to understand, why do we need to set `consensus key` here? Since the ecdsa key is not set in config in this function

e2e_test/node.go
```diff
@@ -169,14 +172,8 @@ func (n *Node) Start() error {
...
 	}()
-	// Provide a copy of the config to node.New, so that we can rely on
-	// Node.Config field not being manipulated by node and hence use our copy
-	// for black box testing.
-	nodeConfigCopy := &node.Config{}
-	if err = copyNodeConfig(n.Config, nodeConfigCopy); err != nil {
-		return err
-	}
 
+	nodeConfigCopy := *n.Config
```
lorenzo-dev1, line 176: I think here we do not do anymore a deep copy without using `copyNodeConfig`.

crypto/autonity_crypto.go
cmjc: They work. I think `AutonityKeys` is nicest. Something descriptive like `nodeBlsKey` is a bit clunky.

cmd/autonity/misccmd.go
```diff
@@ -159,31 +166,62 @@ along with autonity. If not, see <http://www.gnu.org/licenses/>.`)
...
+		if s.Size() >= crypto.AutonityNodeKeyLenInChar {
+			// load key from the node key file
+			nodePrivateKey, consensusKey, err = crypto.LoadNodeKey(nodeKeyFile)
+			if err != nil {
+				utils.Fatalf("Failed to load the node private key: %v", err)
+			}
+		}
+	} else if privateKeysHex := ctx.GlobalString(utils.NodeKeyHexFlag.Name); privateKeysHex != "" {
+		// if the consensus key is missing from the input hex string, terminate the execution.
+		nodePrivateKey, consensusKey, err = crypto.HexToNodeKey(privateKeysHex)
```
lorenzo-dev1, line 216: So the nodekeyhex now contains both the nodekey and the bls consensus key. This makes me even more convinced that we should agree on a name to refer to both. relates to https://github.com/autonity/autonity/pull/858#discussion_r1453248705, let's reach consensus on the name before any modifications.

cmd/gengen/gengen/gengen_test.go
```diff
@@ -79,9 +80,8 @@ func TestEncodeDecodeConsistency(t *testing.T) {
 	require.NoError(t, err)
 	// Set one of the validators to have a publick key, just to cover more code
 	// branches.
-	k, ok := validators[0].Key.(*ecdsa.PrivateKey)
-	require.True(t, ok, "key should be an *ecdsa.PrivateKey")
-	validators[0].Key = &k.PublicKey
+	k := validators[0].NodeKey
```
lorenzo-dev1, line 83: Is this useful?

cmd/utils/flags.go
```diff
@@ -616,11 +617,11 @@ var (
 	}
 	NodeKeyFileFlag = cli.StringFlag{
 		Name:  "nodekey",
-		Usage: "P2P node key file",
+		Usage: "Autonity P2P node key file generated by sub command genNodeKey",
 	}
 	NodeKeyHexFlag = cli.StringFlag{
 		Name:  "nodekeyhex",
-		Usage: "P2P node key as hex (for testing)",
+		Usage: "Autonity P2P node key as hex (for testing)",
```
lorenzo-dev1, line 624: same as https://github.com/autonity/autonity/pull/858#discussion_r1453365913

cmd/utils/flags.go
```diff
@@ -784,29 +785,32 @@ func MakeDataDir(ctx *cli.Context) string {
 	return ""
 }
 
-// setNodeKey creates a node key from set command line flags, either loading it
-// from a file or as a specified hex value. If neither flags were provided, this
-// method returns nil and an emphemeral key is to be generated.
-func setNodeKey(ctx *cli.Context, cfg *p2p.Config) {
+// setNodeKey creates a node key and its derived validator key from set command line flags, either loading it
```
lorenzo-dev1, line 788: bls key is not derived anymore

crypto/autonity_crypto.go
```diff
@@ -1,12 +1,98 @@
...
+const AutonityNodeKeyLenInChar = 128
+const AutonityNodeKeyLen = 64
+const ECDSAKeyLen = 32
+
+var (
+	ErrorInvalidPOP    = errors.New("invalid Autonity POP")
+	ErrorInvalidSigner = errors.New("mismatched Autonity POP signer")
+)
+
+// SaveNodeKey saves a secp256k1 private key and its derived validator BLS
```
lorenzo-dev1, line 26: nitpick: consensus key not derived anymore

crypto/autonity_crypto.go
```diff
@@ -1,12 +1,98 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return ecdsaKey, consensusKey, nil
+}
+
+// HexToNodeKey parse the hex string into a secp256k1 private key and a derived BLS private key.
+func HexToNodeKey(hexKeys string) (*ecdsa.PrivateKey, blst.SecretKey, error) {
+	b, err := hex.DecodeString(hexKeys)
+	if byteErr, ok := err.(hex.InvalidByteError); ok {
```
lorenzo-dev1, line 74: Maybe be better to check if err!= nil before casting the error

crypto/autonity_crypto.go
```diff
@@ -1,12 +1,98 @@
...
 	"github.com/autonity/autonity/common"
+	"github.com/autonity/autonity/common/hexutil"
+	"github.com/autonity/autonity/crypto/blst"
 	"golang.org/x/crypto/blake2b"
+	"os"
 )
 
+const AutonityPOPLen = 226
+const AutonityNodeKeyLenInChar = 128
+const AutonityNodeKeyLen = 64
```
lorenzo-dev1, line 18: nitpick: instead of defining constants as absolute numbers, define as sum of the composing parts. E.G.

AutonityNodeKeyLen = ECDSAKeyLen + BLSSecretKeyLength

This reduces the probability of errors and also makes it clearer to the reader

crypto/blst/common.go
```diff
@@ -0,0 +1,59 @@
...
+import "errors"
+
+const (
+	BLSSecretKeyLength       = 32
+	BLSPubkeyLength          = 48
+	BLSSignatureLength       = 96
+	BLSPubKeyHexStringLength = 98
+)
+
+// ErrSecretConvert describes an error which happens when generating a BLS secret key with ECDSA secret key.
```
lorenzo-dev1, line 12: unused

params/protocol_contracts.go
yazzaoui: Agreed, @cmjc we need to include this for the onboarding process for non-AGF GV's.

params/config.go
yazzaoui: I think this comment is depreciated anyways

crypto/autonity_crypto.go
yazzaoui: Happy with `AutonityKeys` = `NodeKey` + `ConsensusKey`

crypto/autonity_crypto.go
```diff
@@ -1,12 +1,98 @@
...
+		return nil, nil, err
+	} else if n != len(buf) {
+		return nil, nil, fmt.Errorf("key file too short, want 128 hex characters")
+	}
+
+	if err = checkKeyFileEnd(r, AutonityNodeKeyLenInChar); err != nil {
+		return nil, nil, err
+	}
+
+	ecdsaKey, err := HexToECDSA(string(buf[0 : AutonityNodeKeyLenInChar/2]))
```
yazzaoui, line 55: I think we should have a constant for `AutonityNodeKeyLenInChar/2`

crypto/autonity_crypto.go
vpiyush: It's actually not needed. From [here](https://go.dev/ref/spec#Type_assertions):

> For an expression x of [interface type](https://go.dev/ref/spec#Interface_types), but not a [type parameter](https://go.dev/ref/spec#Type_parameter_declarations), and a type T, the primary expression
> 
> `x.(T)`
> asserts that x is not nil and that the value stored in x is of type T. The notation x.(T) is called a type assertion.

crypto/blst/secret_key.go
```diff
@@ -0,0 +1,106 @@
...
+//	a deterministic signature given a secret key SK and a message.
+func (s *bls12SecretKey) Sign(msg []byte) Signature {
+	signature := new(blstSignature).Sign(s.p, msg, generalDST)
+	return &BlsSignature{s: signature}
+}
+
+// Marshal a secret key into a LittleEndian byte slice.
+func (s *bls12SecretKey) Marshal() []byte {
+	keyBytes := s.p.Serialize()
+	if len(keyBytes) < BLSSecretKeyLength {
```
lorenzo-dev1, line 97: Why do we need this padding left?

There are some secret keys who do not serialize to 32 bytes? Prysm doesn't have that https://github.com/prysmaticlabs/prysm/blob/develop/crypto/bls/blst/secret_key.go#L81

crypto/blst/secret_key.go
```diff
@@ -0,0 +1,106 @@
...
+	return wrappedKey, nil
+}
+
+// BlsPublicKey obtains the public key corresponding to the BLS secret key.
+func (s *bls12SecretKey) PublicKey() PublicKey {
+	return &BlsPublicKey{p: new(blstPublicKey).From(s.p)}
+}
+
+// IsZero checks if the secret key is a zero key.
+func (s *bls12SecretKey) IsZero() bool {
```
lorenzo-dev1, line 73: I am not sure this is the correct way to check for a zero private key. Prysm uses a different method, but I am not sure about theirs too https://github.com/prysmaticlabs/prysm/blob/develop/crypto/bls/blst/secret_key.go#L81.

We should probably write a test and deserialize some known 0 private key, and see if our implementation catches it. Some 0 keys can be found in the papers explaining the splitting zero attack.

crypto/autonity_crypto.go
```diff
@@ -46,3 +132,125 @@ func hexDecode(src []byte) ([]byte, error) {
...
 	return dst, err
 }
+
+// BLSPOPProof generate POP of BLS private key of Autonity protocol, the msg start with a prefix of treasury address and
+// ended with the public key of the secrete key, since we don't want the POP being cloned during the propagation of the
+// on-boarding TX. Thus, this POP generation is different from the spec of BLS, which means we have a compatibility issue
+// with a standard POP generation/verification implementation.
+func BLSPOPProof(priKey blst.SecretKey, msg []byte) ([]byte, error) {
+	// the msg contains treasury address and the public key of private key.
+	m := append(msg, priKey.PublicKey().Marshal()...)
```
yazzaoui, line 142: `msg` is too abstract - can we rename this to `treasuryPubKey` or something more specific.
Also, the `msg` buffer is going to be modified by this function. As a general rule, let's try to avoid unnecessary side effects, this goes a long way toward reducing chances for bugs. 
Let's have a new local buffer created here for those operations.

crypto/blst/signature.go
```diff
@@ -0,0 +1,223 @@
...
+	rawKeys := make([]*blstPublicKey, len(pubKeys))
+	for i := 0; i < len(pubKeys); i++ {
+		rawKeys[i] = pubKeys[i].(*BlsPublicKey).p
+	}
+
+	return s.s.FastAggregateVerify(true, rawKeys, msg[:], generalDST)
+}
+
+// NewAggregateSignature creates a blank aggregate signature.
+func NewAggregateSignature() Signature {
```
lorenzo-dev1, line 128: THis can be moved to the tests I think

crypto/blst/signature.go
```diff
@@ -0,0 +1,223 @@
...
+	rawMsgs := make([]blst.Message, length)
+
+	for i := 0; i < length; i++ {
+		mulP1Aff[i] = pubKeys[i].(*BlsPublicKey).p
+		rawMsgs[i] = msgs[i][:]
+	}
+	// Secure source of RNG
+	randLock := new(sync.Mutex)
+
+	randFunc := func(scalar *blst.Scalar) {
```
lorenzo-dev1, line 189: In Prysm they updated this `randFunc` with a check that we might want to import (to avoid rbytes being 0) https://github.com/prysmaticlabs/prysm/blob/develop/crypto/bls/blst/signature.go#L264

params/protocol_contracts.go
cmjc: Absolute.

>Even offline verification would be fine, but we still need a PoP.

Yes, so the functional RQ here is perhaps a function to verify an ownership proof offline. e.g. a `verifyOwnershipProof` command that:

- takes as input the `proof` generated by `genOwnershipProof`
- prints as output the string of the decrypted `proof` to terminal
- prints if the decrypted `proof` matches the `treasury account address` to terminal.

Just a starter for 10 :)

params/protocol_contracts_test.go
```diff
@@ -13,29 +14,42 @@ import (
...
 
 func TestPrepareAutonityContract(t *testing.T) {
-	key1, _ := crypto.GenerateKey()
-	addr1 := crypto.PubkeyToAddress(key1.PublicKey)
-	node1 := enode.NewV4(&key1.PublicKey, net.ParseIP("127.0.0.1"), 30303, 0)
 
-	key2, _ := crypto.GenerateKey()
-	addr2 := crypto.PubkeyToAddress(key2.PublicKey)
-	node2 := enode.NewV4(&key2.PublicKey, net.ParseIP("127.0.0.1"), 30303, 0)
+	treasury1, _ := crypto.GenerateKey()
```
yazzaoui, line 18: Maybe  we can have that in a range loop

autonity/solidity/contracts/Autonity.sol
```diff
@@ -990,34 +1006,37 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         validators[_validator.nodeAddress] = _validator;
     }
 
+    function _verifyAndRegisterValidator(Validator memory _validator, bytes memory _signatures) internal {
+        require(_signatures.length == POP_LEN, "Invalid proof length");
+        require(_validator.consensusKey.length == CONSENSUS_KEY_LEN, "Invalid consensus key length");
```
lorenzo-dev1, line 1011: Maybe we can add the same check for oracle address as well, to make the function more robust

autonity/solidity/contracts/Autonity.sol
```diff
@@ -990,34 +1006,37 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+        // verify proof of possessions.
         bytes memory prefix = "\x19Ethereum Signed Message:\n";
         bytes memory treasury = abi.encodePacked(_validator.treasury);
         bytes32 hashedData = keccak256(abi.encodePacked(prefix, Helpers.toString(treasury.length), treasury));
         address[] memory signers = new address[](2);
         bytes32 r;
         bytes32 s;
         uint8 v;
+        // 1st batch bytes are signatures generated by node key and oracle node key.
+        bytes memory ecdsaSignatures = BytesLib.slice(_signatures, 0, 130);
```
lorenzo-dev1, line 1025: Maybe express 130 as `ECDSA_SIGNATURE_LEN * 2` with `ECDSA_SIGNATURE=65`

autonity/solidity/test/utils.js
```diff
@@ -288,12 +292,6 @@ function randomPrivateKey() {
   return bytesToHex(key).substring(2);
 }
 
-function privateKeyToEnode(privateKey) {
```
lorenzo-dev1, line 291: why removing this utils function? they are useful and I always forget how to do this conversions tbh

autonity/solidity/test/utils.js
```diff
@@ -307,11 +305,6 @@ function publicKey(privateKey, hex = true) {
   return (hex == true) ? publicKey.encode("hex") : new Uint8Array(publicKey.encode());
 }
 
-function publicKeyCompressed(privateKey, hex = true) {
```
lorenzo-dev1, line 310: same as https://github.com/autonity/autonity/pull/858/files#r1455712753

autonity/solidity/test/utils.js
```diff
@@ -323,18 +316,62 @@ function generateMultiSig(nodekey, oraclekey, treasuryAddr) {
...
 }
 
-async function registerValidator(autonity, validatorPrivateKey, treasuryAddr) {
-  let multisig = generateMultiSig(validatorPrivateKey, validatorPrivateKey, treasuryAddr);
-  let oracleAddress = address(publicKey(validatorPrivateKey, false));
-  let enode = privateKeyToEnode(validatorPrivateKey);
-  await autonity.registerValidator(enode, oracleAddress, multisig, {from: treasuryAddr});
-  return oracleAddress;
+async function generateAutonityPOP(nodeKeyFile, oracleKeyHex, treasuryAddress) {
+  const command = `../../../build/bin/autonity genOwnershipProof --nodekey ${nodeKeyFile} --oraclekeyhex ${oracleKeyHex} ${treasuryAddress}`;
```
lorenzo-dev1, line 320: Probably the best would be to not rely on the autonity binary in truffle tests... So that they can be ran in isolation. But I think we can improve this later on, for the moment I think this approach is good.

crypto/autonity_crypto_test.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+
+	blsSigOffset := common.SealLength * 2
+	if err = ecdsaPOPVerify(signatures[common.SealLength:blsSigOffset], hash, oracleAddress); err != nil {
+		return err
+	}
+
+	// check zero signature.
+	validatorSig, err := blst.SignatureFromBytes(signatures[blsSigOffset:])
```
lorenzo-dev1, line 166: The `SignatureFromBytes` method does not check for zero signature. This is correct (at least for the moment) because an aggregate signature could be 0.

However I think we should add a call to `SigValidate(true)` after the deserialization, to prevent validators from registering with 0 signature as POP.

consensus/tendermint/accountability/contracts.go
```diff
@@ -56,6 +63,41 @@ func LoadPrecompiles(chain ChainContext) {
...
+	treasuryOffset := signatureOffset + blst.BLSSignatureLength
+	keyBytes := input[ArrayLenBytes:signatureOffset]
+	sigBytes := input[signatureOffset:treasuryOffset]
+	treasuryBytes := input[treasuryOffset:]
+
+	key, err := blst.PublicKeyFromBytes(keyBytes)
+	if err != nil {
+		return failure32Byte, err
+	}
+	sig, err := blst.SignatureFromBytes(sigBytes)
```
lorenzo-dev1, line 89: same as https://github.com/autonity/autonity/pull/858/files#r1455991642

I think we need a call to `SigValidate(true)` after the deserialization, to prevent validator registering with 0 signature as POP.

crypto/autonity_crypto.go
```diff
@@ -13,51 +13,52 @@ import (
 	"os"
 )
 
-const AutonityPOPLen = 226
-const AutonityNodeKeyLenInChar = 128
-const AutonityNodeKeyLen = 64
 const ECDSAKeyLen = 32
+const AutonityPOPLen = 226
+const ECDSAKeyLenInChar = 64
+const AutonityKeysLenInChar = 128
```
lorenzo-dev1, line 19: This is AutonityKeysLen*2

crypto/autonity_crypto.go
```diff
@@ -13,51 +13,52 @@ import (
 	"os"
 )
 
-const AutonityPOPLen = 226
-const AutonityNodeKeyLenInChar = 128
-const AutonityNodeKeyLen = 64
 const ECDSAKeyLen = 32
+const AutonityPOPLen = 226
+const ECDSAKeyLenInChar = 64
```
lorenzo-dev1, line 18: In general all LEN in char is bytes * 2 (assuming hexadecimal chars)

e2e_test/node.go
lorenzo-dev1: Is using copyNodeConfig is a problem? 

It seems to me that there are other fields that are pointer (like the tendermintServices). 

But if you are sure this copy is enough I trust you

node/config.go
```diff
@@ -389,13 +389,14 @@ func (c *Config) instanceDir() string {
...
-	if c.P2P.PrivateKey != nil {
-		return c.P2P.PrivateKey
+// AutonityKeys retrieves the currently configured private key of the node and the consensus key of the node, checking
+// first any manually set keys, falling back to the one found in the configured  data folder. If no keys can be found,
+// new ones are generated.
+func (c *Config) AutonityKeys() (*ecdsa.PrivateKey, blst.SecretKey) {
+	// Use any specifically configured key, keys are configured via CLI by flags in key file or in key hex string, at
+	// this point, if any of the keys are not nil, it indicates they were parsed on the flag loading phase, return them.
+	if c.P2P.PrivateKey != nil || c.ConsensusKey != nil {
+		return c.P2P.PrivateKey, c.ConsensusKey
```
lorenzo-dev1, line 399: Is it possible that only one of them is != nil?

params/protocol_contracts.go
```diff
@@ -316,10 +315,14 @@ func (v *Validator) Validate() error {
 	if v.CommissionRate != nil && v.CommissionRate.Cmp(big.NewInt(0)) != 0 {
 		return fmt.Errorf("commission rate for enode %q not allowed", nodeAddr.String())
 	}
-	_, err = blst.PublicKeyFromBytes(v.ConsensusKey)
-	if err != nil {
-		return errors.New("cannot decode bls public key")
-	}
+
+	// todo: check below logic for the consensusKey of GVs once we use them in consensus engine, it requires update in
```
lorenzo-dev1, line 319: Is this because we do not have set the keys for piccadilly and bakerloo configs?

cmd/gengen/gengen/gengen_test.go
```diff
@@ -20,33 +20,13 @@ var (
...
-
-	keyfile3, cleanup := tempFile(t)
-	defer cleanup()
-	err = ioutil.WriteFile(keyfile3, crypto.PubECDSAToHex(&k.PublicKey), os.ModePerm)
-	require.NoError(t, err)
-
-	validator1 := "1e12,v,1,:6789," + keyfile1
-	validator2 := "1e12,s,1,:6799," + keyfile2
-	validator3 := "1e12,p,1,:6780," + keyfile3
+	validator1 := "1e12,v,1,:6789,/tmp/autonityKeys1,/tmp/oracleKey1,/tmp/treasuryKey1"
```
lorenzo-dev1, line 27: Are the /tmp keys generated somewhere in the code?

crypto/autonity_crypto.go
```diff
@@ -1,14 +1,97 @@
...
+	"github.com/autonity/autonity/common/hexutil"
+	"github.com/autonity/autonity/crypto/blst"
 	"golang.org/x/crypto/blake2b"
+	"os"
 )
 
-func Hash(data []byte) common.Hash {
-	return blake2b.Sum256(data)
+const ECDSAKeyLen = 32
+const AutonityPOPLen = 226
```
lorenzo-dev1, line 17: Sorry to be annoying but also this should be decomposed in `2*ECDSA_SIGNATURE_LEN+BLS_SIGNATURE_LENGTH`. BLS signature length can be taken from the blst pakcage.

crypto/blst/common.go
```diff
@@ -0,0 +1,56 @@
...
+
+// ErrSecretUnmarshal describes an error which happens during unmarshalling
+// a secret key.
+var ErrSecretUnmarshal = errors.New("could not unmarshal bytes into secret key")
+
+// ErrInfinitePubKey describes an error due to an infinite public key.
+var ErrInfinitePubKey = errors.New("received an infinite public key")
+
+// ErrInvalidSecreteKey describes an invalid length of bytes of secret blst key
+var ErrInvalidSecreteKey = errors.New("secret key must be 32 bytes")
```
lorenzo-dev1, line 26: Spelling error "ErrInvalidSecreteKey" --> "ErrInvalidSecretKey"

crypto/blst/secret_key_test.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			res, err := SecretKeyFromBytes(test.input)
+			if test.err != nil {
+				require.NotEqual(t, nil, err, "No error returned")
+				require.Error(t, test.err, err.Error())
```
lorenzo-dev1, line 65: This test actually only checks the generic presence of an error. Whether the actual error is the same as the specified one is not taken into consideration (try substituting `ErrInvalidSecreteKey` with `ErrSecretUnmarshal`, tests will pass anyways).

cmd/autonity/genesis_test.go
```diff
@@ -53,8 +53,11 @@ var genesisTest = struct {
 					"epochPeriod": 30,
 					"validators" : [ 
 						{
-							"enode" : "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@localhost:3",
-							"treasury": "0x373bf7359fc85Df6A3Cd1726bef4edDa0460b3F3",
+							"enode": "enode://395f3b74b236bccde1684d50a715c2349ee66741d7a69a571c0b176d129ac4ee33acbc85456f7ada3cdc87e1a56f591ba624870a1381c266d41b34c9476c5bc4@172.25.0.11:30303",
+							"treasury": "0x4b07239bd581d21aefcdee0c6db38070f9a5fd2d",
+							"consensusKey": "0xa5b7ba3a6abd4e6d2859733447e39996b8bb4d284aefc8e0f7a909d51873c13fa5c60246c96705a5ae239153f366f188",
+							"pop": "0x4cb3af04550c08dbb7dd0553fd735e3aed6d122dd9a2bf527bbf3af2354add5743f0ff6a6b6f27d03fb01a0cc3cdd390ea38fdb40192f0ef1a8386d0538bb2df01ac0555893eaedf6f605c178878492c7d95f26d706c0b84c5bb96b77f7e8428ef6ad143121317d5f48101d476d818911636eb0f5c25758932f8feddd61682f8d50096a0ffdc430d230f29db673f5e0eb6c726a8525bd2e997a92a2466b0d4f7d346874f22299edb072659de8b0f6186086113f221fa0d8ec30040db87dbbe9b5276c82dd2f11a2f9d286896e1181adb6f4744b9258c6ba5935ef1f8148271c5d313",
```
yazzaoui, line 59: don't forget to remove

consensus/tendermint/accountability/contracts.go
```diff
@@ -56,6 +63,45 @@ func LoadPrecompiles(chain ChainContext) {
 	setPrecompiles(vm.PrecompiledContractsBLS)
 }
 
+// POPVerifier verifies the proof of possession of validator key.
+type POPVerifier struct{}
```
yazzaoui, line 67: Please move this into `core/vm/contracts.go`, this does not belong with the accountability package.

params/protocol_contracts.go
```diff
@@ -131,9 +133,112 @@ type Validator struct {
 	TotalSlashed             *big.Int
 	JailReleaseBlock         *big.Int
 	ProvableFaultCount       *big.Int
+	ValidatorKey             []byte //ABI packing does not support hexutil.Bytes, thus we need to introduce customized JSON Marshal/UnMarshal methods.
 	State                    *uint8
 }
 
+// UnmarshalJSON and MarshalJSON are customized marshal and unmarshal methods to parse validators with activity key in
+// hex string from genesis file, the raw type []byte is replaced by hexutil.Bytes.
+func (v *Validator) UnmarshalJSON(input []byte) error {
```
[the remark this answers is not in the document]
yazzaoui: Ok got it, because you want `ConsensusKey` to be of type []byte. Would that be possible to have it with type `string` here and we do the conversion in deployAutonityContract, so to avoid all this boilerplate?

node/config.go
yazzaoui: I think it's a fatal issue if one of these is missing.

yazzaoui (approved):
LGTM :raised_hands:
Just please don't forget https://github.com/autonity/autonity/pull/858/files#r1458029013

node/config.go
```diff
@@ -385,27 +389,39 @@ func (c *Config) instanceDir() string {
...
-	// Use any specifically configured key.
+// AutonityKeys retrieves the currently configured private key of the node and the consensus key of the node, checking
+// first any manually set keys, falling back to the one found in the configured  data folder. If no keys can be found,
+// new ones are generated.
+func (c *Config) AutonityKeys() (*ecdsa.PrivateKey, blst.SecretKey) {
+	// Use any specifically configured key, keys are configured via CLI by flags in key file or in key hex string, at
+	// this point, if any of the keys are not nil, it indicates they were parsed on the flag loading phase, return them.
 	if c.P2P.PrivateKey != nil {
-		return c.P2P.PrivateKey
+		if c.ConsensusKey == nil {
```
lorenzo-dev1, line 399: I think the best would be just to enter this if only if `c.P2P.PrivateKey != nil && c.ConsensusKey != nil`. Or you think it is better to crash? @Jason-Zhangxin-Chen @yazzaoui 

If we want to crash, I think we should crash also if only consensusKey != nil and PrivateKey == nil

lorenzo-dev1:
Looks great @Jason-Zhangxin-Chen 

On my side I have just 2 remaining comments which I would like to address before approving:
- I do not understand why this part of the `Validate()` function is commented out. Not saying it should not be, but I would like to understand why it is https://github.com/autonity/autonity/pull/858#discussion_r1456132123
- Question regarding autonity keys handling, https://github.com/autonity/autonity/pull/858/files#r1459361326

Additionally, these are some other comments regarding tests and minor issues (not merge blocking imo):
- question about deep copy in tests, https://github.com/autonity/autonity/pull/858#discussion_r1453264846
- more of question, there is a back and forth copy of the node key which I do not understand. Maybe it has a purpose but I don't get it. https://github.com/autonity/autonity/pull/858#discussion_r1453386503
- comment still mentions consensus key being derived, https://github.com/autonity/autonity/pull/858#discussion_r1453446091
- The method `autonityPOPVerify` in `autonity_crypto_test.go` still misses the 0 signature check, https://github.com/autonity/autonity/pull/858#discussion_r1455991642
- not sure where these paths come from, https://github.com/autonity/autonity/pull/858#discussion_r1456134338

lorenzo-dev1 (approved):
LGTM, let's not forget about uncommenting this`Validate` part later on when we have bls keys of genesis validators https://github.com/autonity/autonity/pull/858#discussion_r1456132123


it might be good to fix these unchecked comments here, but not merge blocking https://github.com/autonity/autonity/pull/858#issuecomment-1900782429

Jason-Zhangxin-Chen:
> LGTM, let's not forget about uncommenting this`Validate` part later on when we have bls keys of genesis validators [#858 (comment)](https://github.com/autonity/autonity/pull/858#discussion_r1456132123)
> 
> it might be good to fix these unchecked comments here, but not merge blocking [#858 (comment)](https://github.com/autonity/autonity/pull/858#issuecomment-1900782429)

@lorenzo-dev1 , I left responses for each of them days ago, it is wired that the review board does not present them to you. Or please double check them, thanks.

lorenzo-dev1:
> > LGTM, let's not forget about uncommenting this`Validate` part later on when we have bls keys of genesis validators [#858 (comment)](https://github.com/autonity/autonity/pull/858#discussion_r1456132123)
> > it might be good to fix these unchecked comments here, but not merge blocking [#858 (comment)](https://github.com/autonity/autonity/pull/858#issuecomment-1900782429)
> 
> @lorenzo-dev1 , I left responses for each of them days ago, it is wired that the review board does not present them to you. Or please double check them, thanks.

Please do not modify my comments. If you think that the highlighted issues are already solved, let me know. However I will mark the issues as solved in my own comments once I am convinced they are solved.

I do not see any reply to my specific questions for:
- https://github.com/autonity/autonity/pull/858#discussion_r1453264846. I saw you modified the code here, but it does not actually answer my question which is: why not keep using the `copyNodeConfig` function that we were using before? The deep copy you substituted it with does not do the same thing.
- https://github.com/autonity/autonity/pull/858#discussion_r1455991642. You did add the check in the precompiled contract, but not in the method autonityPOPVerify in autonity_crypto_test.go. That is why I think the issue is still valid.
- https://github.com/autonity/autonity/pull/858#discussion_r1456134338, I do not see any reply here.

Issue #850, opened 2023-10-12
Port ASM tests to use latest Ape Framework version


aiman:
The ASM tests were working fine until Ape `0.6.19`. However, after `0.6.19` there seem to be some changes to the log format that cause the `test_acu.py` test `test_modify_basket` to fail for different reasons.

There seem to be different failures due to the log format in `0.6.20` and `0.6.21`+, both are described below in case it offers some more clues.

I suspect the `eth-ape` package as the cause because I have tested with eth-ape versions `0.6.19 - 0.6.22`, keeping Web3 and Ape plugin versions fixed, so the only difference in my tests has been `eth-ape` itself.

But these changes do not seem to have been announced or documented in `eth-ape`. The only related changes I can find in the [eth-ape release notes](https://github.com/ApeWorX/ape/releases) are described as "fixes":

- https://github.com/ApeWorX/ape/pull/1659
- https://github.com/ApeWorX/ape/pull/1662

### `0.6.20`

Environment:

```console
$ pipx runpip eth-ape freeze  | egrep "(ape|web3)"
ape-hardhat==0.6.13
ape-solidity==0.6.9
eth-ape==0.6.20
web3==6.11.0
```

Failure:

```console
$ make test-contracts-asm
...
_______________________________________________________ test_modify_basket ________________________________________________________
/host/autonity/solidity/test/asm/test_acu.py:223: in test_modify_basket
    assert event.symbols == tuple(new_basket)
E   AssertionError: assert [['X', 'Y']] == ('X', 'Y')
E     At index 0 diff: ['X', 'Y'] != 'X'
E     Right contains one more item: 'Y'
E     Full diff:
E     - ('X', 'Y')
E     + [['X', 'Y']]
        acu_basic  = ACUTestData(contract=<ACU 0xBcF7FFFD8B256Ec51a36782a52D0c34f6474D951>, oracle=<Oracle 0x274b028b03A250cA03644E6c578D81f019eE1323>, symbols=['AAA', 'BBB', 'CCC'], quantities=[0, 100000, 200000], prices=None, scale_factor=100000)
        event      = <BasketModified symbols=[['X', 'Y']] quantities=[[1, 2]] scale=18>
        new_basket = ['X', 'Y']
        new_quantities = [1, 2]
        new_scale  = 18
        new_scale_factor = 1000000000000000000
        receipt    = <Receipt 0xaeb961ccc3f02c0f5ec9e28aaba981104a1c2f601709c282f5a6463212069f35>
[...]

PR #861, opened 2023-11-15
pin ape-solidity plugin version to 0.6.9

lorenzo-dev1:
Version 0.6.10 of ape-solidity requires eth-ape version >=0.6.25, which causes issues for us (see https://github.com/autonity/autonity/issues/850). 

Due to this ACU tests [were failing again](https://github.com/autonity/autonity/actions/runs/6875042982/job/18700681721?pr=855). This version pinning solves the issue.

### DIFF
--- a/autonity/solidity/ape-config.yaml
+++ b/autonity/solidity/ape-config.yaml
@@ -4,7 +4,7 @@ contracts_folder: contracts
 
 plugins:
   - name: hardhat
-  - name: solidity
+  - name: solidity==0.6.9
 
 # Ignore `bindings.sol` to speed up compilation. This is because compiling it
 # doesn't result in a cached JSON file under `.build/`, and so it must be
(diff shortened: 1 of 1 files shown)


yazzaoui (approved):

aiman:
Great detective work Lorenzo. Can I suggest then to also pin the hardhat plugin version in the `ape-config.yml` to avoid a similar situation when the latest hardhat plugin requires a newer version of `eth-ape` than what we fix.

lorenzo-dev1:
> Great detective work Lorenzo. Can I suggest then to also pin the hardhat plugin version in the `ape-config.yml` to avoid a similar situation when the latest hardhat plugin requires a newer version of `eth-ape` than what we fix.

Makes sense! pinned hardhat plugin version in https://github.com/autonity/autonity/pull/861/commits/c1c9822c15925af0492c751d04bb83d3a99b00b9

aiman (approved):
LGTM

PR #866, opened 2023-11-24
Autonity validator status jailbound client update

tbssajal:


consensus/tendermint/backend/engine.go
```diff
@@ -554,6 +559,10 @@ func (sb *Backend) faultyValidatorsWatcher(ctx context.Context) {
 				}
 			}
 			sb.jailedLock.Unlock()
+
+			// TODO: delete jailbound validator from map after they are kicked out from committee
```
tbssajal, line 563: Should it be implemented already? [Comment](https://github.com/autonity/autonity/pull/857#issuecomment-1825366169)

consensus/tendermint/backend/engine.go
```diff
@@ -546,6 +547,10 @@ func (sb *Backend) faultyValidatorsWatcher(ctx context.Context) {
 			sb.jailedLock.Lock()
 			sb.jailed[ev.Validator] = ev.ReleaseBlock.Uint64()
```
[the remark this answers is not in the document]
lorenzo-dev1: The events are read by the client in the same order as the blockchain produces them. Why would the max be needed?

consensus/tendermint/core/handler.go
```diff
@@ -251,6 +254,10 @@ func (c *Core) SendEvent(ev any) {
 // handleMsg assume msg has already been decoded
 func (c *Core) handleMsg(ctx context.Context, msg *message.Message) error {
 	msgHeight := new(big.Int).SetUint64(msg.H())
+	if c.backend.IsJailbound(msg.Address) {
```
lorenzo-dev1, line 257: I think this needs to be moved next to the `IsJailed` check. This is because we cannot be sure that msg.Address corresponds to the real sender until we validate the message.

However note that these flows will change once we merge https://github.com/autonity/autonity/pull/852

consensus/tendermint/backend/engine.go
```diff
@@ -513,6 +515,7 @@ func (sb *Backend) faultyValidatorsWatcher(ctx context.Context) {
 	var subscriptions event.SubscriptionScope
 	newFaultProofCh := make(chan *autonity.AccountabilityNewFaultProof)
 	slashingEventCh := make(chan *autonity.AccountabilitySlashingEvent)
+	jailboundEventCh := make(chan *autonity.AccountabilityValidatorJailbound)
```
lorenzo-dev1, line 518: I think you are missing the call to do the actual subscription to the event. Look at how the other channel are subscribed and tracked by the object `subscriptions`

lorenzo-dev1: It would also be good to add a golang test for this behaviour (ignore and disconnect jailbound)

consensus/tendermint/backend/engine.go
lorenzo-dev1: I think so

consensus/tendermint/backend/engine.go
```diff
@@ -546,8 +549,13 @@ func (sb *Backend) faultyValidatorsWatcher(ctx context.Context) {
 			sb.jailedLock.Lock()
 			sb.jailed[ev.Validator] = ev.ReleaseBlock.Uint64()
 			sb.jailedLock.Unlock()
+		case ev := <-jailboundEventCh:
+			sb.jailedLock.Lock()
+			sb.jailed[ev.Validator] = jailboundReleaseHeight
```
m4ksio, line 554: I just wonder - since we never release jailbound validators. So we will perpetually store those entries in the map. I'm worried it might simply accumulate overtime, not only using memory (we store umaxint N times) but also adding to every iteration over this array.
Maybe separate map like `jailbound map[common.Address]struct{}` to store those item would be a bit more performant

m4ksio (approved):

consensus/tendermint/backend/engine.go
lorenzo-dev1: We can delete jailbound validators from the map once the epoch in which they were jailbound ends. After that they will not get selected in the committee anymore.

consensus/tendermint/backend/engine.go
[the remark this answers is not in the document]
tbssajal: In that case we don't need it.

consensus/tendermint/core/handler.go
tbssajal: Right, updated

yazzaoui:
Probably part of a separate PR, we should think about saving those entries in the local DB / or retrieving them from state at startup in case the node shuts down for some reason.

tbssajal:
> I think so, as of now if we receive a msg is from a validator that is not part of the committee, the msg will fail validation.

Now looking at the codes again, a validator is jailed or jailbound at the end of an epoch. And at the same time committee is changes, so jailed or jailbound validators will not be in the committee any more. In that case do we need to keep track of jailed or jailbound validators in the client? I think we can remove the logic of keeping track of jailed/jailbound validators.

cmjc:
> > I think so, as of now if we receive a msg is from a validator that is not part of the committee, the msg will fail validation.
> 
> Now looking at the codes again, a validator is jailed or jailbound at the end of an epoch. And at the same time committee is changes, so jailed or jailbound validators will not be in the committee any more. In that case do we need to keep track of jailed or jailbound validators in the client? I think we can remove the logic of keeping track of jailed/jailbound validators.

That's made me think about `activateValidator` in Autonity.sol.  Should this have a check for `jailbound` added, e.g.

```
require(_val.state != ValidatorState.jailbound, "validator is permanently jailed!");
```

https://github.com/autonity/autonity/blob/develop/autonity/solidity/contracts/Autonity.sol#L346C1-L354C6

tbssajal:
> > > I think so, as of now if we receive a msg is from a validator that is not part of the committee, the msg will fail validation.
> > 
> > 
> > Now looking at the codes again, a validator is jailed or jailbound at the end of an epoch. And at the same time committee is changes, so jailed or jailbound validators will not be in the committee any more. In that case do we need to keep track of jailed or jailbound validators in the client? I think we can remove the logic of keeping track of jailed/jailbound validators.
> 
> That's made me think about `activateValidator` in Autonity.sol. Should this have a check for `jailbound` added, e.g.
> 
> ```
> require(_val.state != ValidatorState.jailbound, "validator is permanently jailed!");
> ```
> 
> https://github.com/autonity/autonity/blob/develop/autonity/solidity/contracts/Autonity.sol#L346C1-L354C6

Yeah, the check is added in the PR #857. [Code](https://github.com/autonity/autonity/pull/857/files#diff-a4ed163d1032b7ac9e50c385517d2389adb36caac0e45853d32a4345b676766bR353)

lorenzo-dev1:
> Now looking at the codes again, a validator is jailed or jailbound at the end of an epoch. And at the same time committee is changed, so jailed or jailbound validators will not be in the committee any more. In that case do we need to keep track of jailed or jailbound validators in the client? I think we can remove the logic of keeping track of jailed/jailbound validators.

That's a good point. It is useful to distinguish jailed and jailbound:
- **jailing** happens at the end of the epoch. However whenever there is a `NewFaultProof` event (which will lead to jailing at the end of the epoch) we already insert the offender in the `jailed` map and ignore its messages. So we effectively ignore him at consensus layer before he is actually jailed in the smart contract.
- **jailbound** happens at the end of the epoch, and we have no way to detect it beforehand. So actually yes, I think we do not need to track jailbound validators in the client side.

Now a second point is: is the current behaviour for the jailed validators what we really want? Thinking forward, this might be a problem for when we will introduce omission failure. Because a node which commits a fault will be ignored, thus he will then also be considered offline. However this is out of scope of this PR.

tbssajal:
> (they got assigned 0 when the fault is reported, but then the value is never updated because they never get jailed)

You mean because maybe they got jailbound instead?

consensus/tendermint/backend/engine.go
tbssajal: added subscription

tbssajal: Not adding test for now, as we need test for jailed validator logic as well and those tests could be quite big and out of the scope of this PR. Will create issue to track this.

PR #881, opened 2023-11-29
[WIP] Signature parallelization rebased

### LINKED ISSUE #878: Parallel signature verification
Currently, consensus message signature verification happens twice, one time during core processing and one time in the fault detector. This is very resource-costly, the signature verification needs to happen ideally upfront in the sender's message handler go-routine.

lorenzo-dev1:
Closes #878

Changes:
- no accountability for messages with: invalid round or step, wrong valid round (e.g. valid round >= round), garbage message. All these conditions are catch at decode time in the peer session handler. If decode fails, the sending peer is disconnected.
- future messages are now handled in the peer session handler rather than in core (and in the fault detector). Signature verification is postponed to when they are re-injected.
   - as a consequence backlog tests about future messages have been moved to the backend handler tests.
- messages sent to self are considered trusted, no signature verification is performed.


lorenzo-dev1:
Closing, this PR won't be merged as we will be dealing with signature verification using a BLS aggregator.

Issue #883, opened 2023-11-30
Benchmark optimisitc BLS aggregation and verifcation


m4ksio:
### Description

Initial benchmarks shows that we can gain significant improvement by using BLS signature aggregation and single verification compared to iterative verification using ECDSA. However, in case of a failed verification we are unable to easily identify offenders with this approach. A separate probably slower backup/emergency approach is required in such case. 
We want to test and benchmark this procedure.

### Rationale

We assume that failed validation is uncommon, as it triggers slashing. Therefore optimistic approach with occasional longer alternative path is considered 

### Implementation

Given set of signatures and the information that some of them are invalid, we need to identify one or more offenders.
One obvious approach is to simply iterate all and verify one by one.
Second is divide and conquer approach, recursively splitting the set, aggregating and checking the aggregated signature for correctness.

m4ksio:
Benchmark results:

Note, there are two cases - `simple` and `fast`.
`Simple` is using `AggregateVerify` for distinct messages. There are N different messages with N keys and N signatures. 
`Fast` uses `FastAggregateVerify`, for the same message, so there is 1 message, signed by N different keys producing N signatures.




```
cpu: AMD Ryzen 7 3700X 8-Core Processor             
BenchmarkAggregateVerifyInvalid
BenchmarkAggregateVerifyInvalid/1_in_500
BenchmarkAggregateVerifyInvalid/1_in_500/simple
BenchmarkAggregateVerifyInvalid/1_in_500/simple-16  	      21	  59233909 ns/op
BenchmarkAggregateVerifyInvalid/1_in_500/fast
BenchmarkAggregateVerifyInvalid/1_in_500/fast-16    	      69	  17402256 ns/op
BenchmarkAggregateVerifyInvalid/1_in_1000
BenchmarkAggregateVerifyInvalid/1_in_1000/simple
BenchmarkAggregateVerifyInvalid/1_in_1000/simple-16 	      10	 100154733 ns/op
BenchmarkAggregateVerifyInvalid/1_in_1000/fast
BenchmarkAggregateVerifyInvalid/1_in_1000/fast-16   	      58	  22356562 ns/op
BenchmarkAggregateVerifyInvalid/5_in_1000
BenchmarkAggregateVerifyInvalid/5_in_1000/simple
BenchmarkAggregateVerifyInvalid/5_in_1000/simple-16 	       6	 168475760 ns/op
BenchmarkAggregateVerifyInvalid/5_in_1000/fast
[...]

PR #893, opened 2023-12-04
Add Find(Fast)InvalidSignatures which locates invalid BLS signatures

m4ksio:


crypto/bls/bls.go
```diff
@@ -72,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+	if end-start <= 1 {
+
+		var ret []int
+
+		valid := signatures[start].Verify(pks[start], msgs[start][:])
+		if !valid {
+			ret = append(ret, start)
+		}
+
+		if start-end == 1 {
```
tbssajal, line 116: Is it `end-start == 1`. Or in this case you can simply write `start < end`

crypto/bls/bls.go
```diff
@@ -72,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+
+		go func() {
+			//func() {
+
+			aggSig := AggregateSignatures(signatures[pivot:end])
+			verified := aggSig.AggregateVerify(pks[pivot:end], msgs[pivot:end])
+
+			if !verified {
+
+				rightInvalid, rightErr = findInvalidSignaturesRecursive(signatures, pks, msgs, pivot, end)
```
tbssajal, line 163: I think it should be pivot+1

crypto/bls/bls.go
```diff
@@ -72,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+
+		}()
+
+		//right
+		wg.Add(1)
+
+		go func() {
+			//func() {
+
+			aggSig := AggregateSignatures(signatures[pivot:end])
```
tbssajal, line 158: I think it should be `[pivot+1][end+1]`

crypto/bls/bls.go
```diff
@@ -72,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+			verified := aggSig.AggregateVerify(pks[pivot:end], msgs[pivot:end])
```
tbssajal, line 159: I think it is `[pivot+1][end+1]`

crypto/bls/bls.go
```diff
@@ -72,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+		//left
+
+		var leftErr error
+		var rightErr error
+
+		wg.Add(1)
+
+		go func() {
+			//func() {
+			aggSig := AggregateSignatures(signatures[start:pivot])
```
tbssajal, line 142: I think it should be `pivot+1`

crypto/bls/bls.go
```diff
@@ -72,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+			verified := aggSig.AggregateVerify(pks[start:pivot], msgs[start:pivot])
```
tbssajal, line 143: I think it is `pivot+1`

tbssajal (changes requested):
According to the logic of the base case (2 signatures), you are handling both start and end signatures. Then `pivot` is repeated in both left and right functions.

crypto/bls/bls.go
m4ksio: Hmm, why?
Here we check `<start, pivot)` and next half checks `<pivot, end)` which covers entire range (since end is len of array so last index + 1.

crypto/bls/bls.go
m4ksio: Funny thing. 
This if is not really used anyway, since there is a bug in line 107 (or maybe not bug, since overall algorithms works OK). It only gets true for a single element slice. If it changes to match the comment and trigger for two items - than indeed, line 116 is wrong and tests no longer pass.
So since line 107 should be `end-start <= 2` this cannot be simply changed to `start < end` but it should be `end-start == 2` indeed. 
Having those both changed now add a nice speedup - worst case scenario in benchmark went from 634ms to 593ms! Really good find!

crypto/bls/bls.go
tbssajal: Okay I see, start is included and end is excluded in your function.

crypto/bls/bls.go
```diff
@@ -99,3 +74,191 @@ func NewAggregateSignature() common.BLSSignature {
...
+func findInvalidSignaturesRecursive(
+	signatures []common.BLSSignature,
+	pks []common.BLSPublicKey,
+	msgs [][32]byte,
+	start, end int,
+) ([]int, error) {
+
+	// if we have two elements, no point in further splitting since we need to do
+	// two verifications anyway
+	if end-start <= 2 {
```
tbssajal, line 107: So in this start is included and end is excluded, that case there should be another case before this one just to be safe: `if start >= end` it should return an empty list.

tbssajal (changes requested):
Just to be safe, maybe add another case.

crypto/bls/bls.go
m4ksio: I don't think its necessary. This is an internal function and we control how parameters are passed. 
But even if for whatever reason `start` will become larger than `end` the worst that can happen is simply an extra verification. I don't think this justified extra safety check.

However, since start and end are indices, and we also return indices of invalid signatures, I have changed the type to `uint` so we can't even have negative numbers

lorenzo-dev1:
@m4ksio Is there somewhere where I can see the results of this benchmark?

crypto/bls/bls.go
tbssajal: Yeah, I see you already put a check on the array size before calling the function

m4ksio:
> @m4ksio Is there somewhere where I can see the results of this benchmark?

In the issue https://github.com/autonity/autonity/issues/883
Updated description with this link

crypto/blst/find_invalid_benchmark_test.go
```diff
@@ -0,0 +1,460 @@
...
+	// start the actual aggregation benchmarking
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+
+		invalid, err := FindFastInvalidSignatures(sigs, pks, msg)
+		if err != nil {
+			b.Fatalf("not expected err %s", err)
+		}
+		if len(invalid) != len(failedNs) {
```
tbssajal, line 455: maybe we could do this sort of checks in separate tests and keep the benchmark free of extra work that may add to runtime. Besides I think any concerns should be addressed in tests, because benchmarks are not run in checks.

crypto/blst/find_invalid_benchmark_test.go
```diff
@@ -0,0 +1,460 @@
...
+
+import (
+	mrand "math/rand"
+	"testing"
+)
+
+// benchmarks
+
+// benchmark bls public key aggregation
+func benchmarkAggregatePK(seed int64, n int, b *testing.B) {
```
tbssajal, line 11: It seems like `seed` and `n` are unnecessary input parameters.

tbssajal (approved):
Just some minor comments. Except that LGTM

crypto/blst/find_invalid_benchmark_test.go
m4ksio: They are rater miniscule in this case, but I agree with principle, removed.

crypto/blst/find_invalid_benchmark_test.go
m4ksio: `n` is used, `seed` is not indeed, Removed `seed`

lorenzo-dev1:
I cherry picked the functions to find invalid signatures and the related benchmarks and tests on top of https://github.com/autonity/autonity/pull/966 (BLS aggregator PR)

lorenzo-dev1:
I think this PR can be closed. I cherry-picked all the useful code in https://github.com/autonity/autonity/pull/957

lorenzo-dev1:
Closing as obsolete

PR #896, opened 2023-12-07
Optimize Sorting

tbssajal:
Regarding #901
Implemented an optimization of `computeCommittee` method in solidity named `computeCommitteeOptimized`. Cannot figure out anything else, maybe we need to try precompiled contract. Kept both `computeCommittee` and `computeCommitteeOptimized` function for now to compare their performance. For me it took around 8 seconds and 1.4 seconds respectively for 100,000 validators and 0.7 and 0.3 seconds for 10,000 validators. Note that there needs to be some sort of tuning of the new parameter `minimumBondedStake` in Autonity.sol to get good performance.


tbssajal:
Added details about optimization in #901, might help review the PR more easily.

tbssajal:
It is unnecessary as the function can be moved to precompiled contract

PR #908, opened 2023-12-19
Optimize sorting precompiled

tbssajal:


tbssajal:
Unnecessary as the whole function `computeCommittee` can be moved to precompiled contract

PR #909, opened 2023-12-20
[Rebasing] Save&Query committee with Epoch Header

Jason-Zhangxin-Chen:
This pull request introduce a set of features:
-  save the committee only in the epoch headers, rather than storing the committee at every block header. 
-  query the committee via new chain interface.
-  epoch rotation event subscriber is used for those committee watcher routines.


Jason-Zhangxin-Chen:
Replaced with https://github.com/autonity/autonity/pull/1009

PR #910, opened 2023-12-20
Fix consensus issues

### LINKED ISSUE #907: Investigation of Unexplained Node Slashing Incidents
## Description 

A bunch of validators have reported unexpected jailing of their nodes. Initially, I considered this might be due to operational errors on their part, such as accidentally creating a twin setup. However, the increasing frequency of these incidents suggests a potential bug that we haven't previously encountered in our testing and development environments. I've examined one of the systems in question and found no issues from my perspective. I've circulated a [report form](https://docs.google.com/forms/d/10K8o4K_SeDTitIaAu7S3bh6j_w4lyh6JBKS_J8hvSg4), but the responses received so far lack detail.

Despite my efforts to replicate the issue yesterday, I was unable to do so. Here are some key observations:

- The majority of professional operators haven't reported any jailing.
- One validator observed that upgrading to a more powerful server resolved the issue of being slashed.
- Another reported incidents of slashing during periods of stress testing.
- Slashed machines seem to be using shared vCPUs or low on cores.

This sounds very very most likely to be a race condition due to resource constants. From the on-chain data I got, it's all prevotes equivocation (prevote nil + prevote non-nil). After hurting my brain yesterday the only theoretical scenario I could think of is :

1. Loss of sync: local core at block h, receives an announcement for block h + 1
2. Downloader is triggered
3. Core get back on sync, managed to include block h and h+1 
4. Validator vote on h+2
5. Core is stopped because of Downloader signal
6. Downloader has nothing to do, reenable core
7. Core go back to h+2 with empty internal state

## Next steps

We need to be able to consistently replicate. Once it will be just a matter of time for identifying the problem and solving it but replication is the hard part. I've created a [branch](https://github.com/autonity/autonity/tree/v0.12.2-investigation) which contain code to help investigation, see here:

- Local e2e test replication. This is the most useful for us for debugging so focus should be on finding `e2e_tests` scenarios which could trigger it.
- Piccadilly testnet deployment: GCP deployment of a resource constrained node and monitoring.
- Deployment on a multi-slashed machine from one of the victim of the incident. Last option.

yazzaoui:
closes #907 

Fixes the two issues found:

1- Cache size increase.
2- Propose Flow reworked to discard additional proposals received and to make sure no conflicting prevote can ever be sent.  

I'll add tests in this PR or in a subsequent one depending on the situation.

### DIFF
--- a/consensus/tendermint/core/propose.go
+++ b/consensus/tendermint/core/propose.go
@@ -117,9 +117,6 @@ func (c *Proposer) HandleProposal(ctx context.Context, msg *message.Message) err
 	}
 
 	if err != nil {
-		if timeoutErr := c.proposeTimeout.StopTimer(); timeoutErr != nil {
-			return timeoutErr
-		}
 		// if it's a future block, we will handle it again after the duration
 		// TODO: implement wiggle time / median time
 		if err == consensus.ErrFutureTimestampBlock {
@@ -157,10 +154,6 @@ func (c *Proposer) HandleProposal(ctx context.Context, msg *message.Message) err
 	}
 
 	if c.step == tctypes.Propose {
-		if err := c.proposeTimeout.StopTimer(); err != nil {
-			return err
-		}
-
 		vr := proposal.ValidRound
 		h := proposal.ProposalBlock.Hash()
 
--- a/consensus/tendermint/core/propose.go
+++ b/consensus/tendermint/core/propose.go
@@ -61,17 +61,18 @@ func (c *Proposer) HandleProposal(ctx context.Context, msg *message.Message) err
 
 			roundMsgs := c.messages.GetOrCreate(proposal.Round)
 
-			// if we already have a proposal then it must be different than the current one
-			// it can't happen unless someone's byzantine.
-			if roundMsgs.ProposalDetails != nil {
-				return err // do not gossip, TODO: accountability
+			// if we already have a proposal for this old round - ignore
+			// the first proposal sent by the sender in a round is always the only one we consider.
+			if roundMsgs.ProposalMsg != nil {
+				return constants.ErrAlreadyProcessed
 			}
 
 			if !c.IsFromProposer(proposal.Round, msg.Address) {
 				c.logger.Warn("Ignoring proposal from non-proposer")
 				return constants.ErrNotFromProposer
 			}
-			// We do not verify the proposal in this case.
+
+			// Save it, but do not verify the proposal yet unless we have enough precommits for it.
 			roundMsgs.SetProposal(proposal, msg, false)
 
 			if roundMsgs.PrecommitsPower(roundMsgs.GetProposalHash()).Cmp(c.CommitteeSet().Quorum()) >= 0 {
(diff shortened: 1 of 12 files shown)


consensus/tendermint/core/propose.go
```diff
@@ -71,7 +71,12 @@ func (c *Proposer) HandleProposal(ctx context.Context, msg *message.Message) err
 				c.logger.Warn("Ignoring proposal from non-proposer")
 				return constants.ErrNotFromProposer
 			}
-			// We do not verify the proposal in this case.
+			// if we already have a proposal for this old round - ignore
+			// the first proposal sent by the sender in a round is always the only one we consider.
+			if roundMsgs.ProposalMsg != nil {
```
lorenzo-dev1, line 76: Wouldn't this be already caught by the check on line 66:
```
if roundMsgs.ProposalDetails != nil {
    return err // do not gossip, TODO: accountability
}
```

lorenzo-dev1: Actually since `NewRoundMessages` does:
```
ProposalDetails:  new(Proposal),
```
wouldn't it always be != nil even if we did not receive an old proposal?

consensus/tendermint/core/propose.go
```diff
@@ -85,6 +90,12 @@ func (c *Proposer) HandleProposal(ctx context.Context, msg *message.Message) err
 		}
 		return err
 	}
+	// if we already have processed a proposal in this round we ignore.
+	if c.curRoundMessages.ProposalMsg != nil {
```
lorenzo-dev1, line 94: I think this is good, however there is a weird edge case in the tendermint algo that I think it is worth noting. 

I think that if we would  strictly follow the algo we would actually process 2 proposal in the same round.

The scenario is:
- we receive a valid proposal with vr != -1, but we do not have a quorum of prevotes at vr. Therefore we do not send a prevote and we remain in propose phase (because line 28 does not get triggered)
- the proposer is equivocating, and we receive another valid proposal with vr=-1. Since we are still in propose step, line 22 gets triggered and we send a prevote for this proposal.

However since this scenario can happen only if the proposer is equivocating, I am inclined to think that we can just enforce the 1 proposal per round check. But we are deviating from the tendermint algo imo.

consensus/tendermint/core/propose.go
yazzaoui: Yeah I can see that, this check is duplicated and wrong because `ProposalDetails` is always non nil. 
Thankfully we don't have this issue on the main branch since the refactoring. 
I'll remove this check and verify everywhere else we use ProposalDetails if it's done properly.

consensus/tendermint/core/propose.go
yazzaoui: I see what you mean but we are anyways fully ignoring byzantine validators as soon as an equivocation proof is processed, so ignoring the second equivocated proposal is line with our current approach. Our ethos is to fully sideline byzantine entities rather than trying to reach consensus at all cost with the garbage we're receiving from them.

lorenzo-dev1 (approved):

PR #911, opened 2023-12-21
fix invalid metric due to hyphen in protocol name

vpiyush:
fix invalid metric due to hyphen in protocol name
Avoid warnings on buffered gauge on prometheus scrapping.


lorenzo-dev1 (approved):
LGTM, maybe it would be good to track the missing prometheus code for the buffered gauge in an issue to avoid forgetting

yazzaoui:
Can you disable the buffered gauge tests so we can have the e2e tests running on CI

vpiyush:
> Can you disable the buffered gauge tests so we can have the e2e tests running on CI

Thanks for taking care of it.

Issue #919, opened 2023-12-28
Incorrect or unexpected value for `treasury_account` protocol parameter


lennylessizmor:
```
$ aut protocol config
{
  "policy": {
    "treasury_fee": 10000000000000000,
    "min_basefee": 500000000,
    "delegation_rate": 1000,
    "unbonding_period": 21600,
    "treasury_account": "0xd32C0812Fa1296F082671D5Be4CbB6bEeedC2397"
  },
  "contracts": {
    "accountability_contract": "0x5a443704dd4B594B382c22a083e2BD3090A6feF3",
    "oracle_contract": "0x47e9Fbef8C83A1714F1951F142132E6e90F5fa5D",
    "acu_contract": "0x8Be503bcdEd90ED42Eff31f56199399B2b0154CA",
    "supply_control_contract": "0x47c5e40890bcE4a473A49D7501808b9633F29782",
    "stabilization_contract": "0x29b2440db4A256B0c1E6d3B4CDcaA68E2440A08f"
  },
  "protocol": {
    "operator_account": "0xd32C0812Fa1296F082671D5Be4CbB6bEeedC2397",
    "epoch_period": 1800,
    "block_period": 1,
    "committee_size": 100
  },
  "contract_version": 1
}
```

shows the same account for `treasury_account` and `operator_account`. But `treasury_account` in v0.12.2 config [appears to be ](https://github.com/autonity/autonity/blob/2495b37ae4aacc6505f6287cafe19cfbb0b7f17b/params/config.go#L95) different (`0xF74c34Fed10cD9518293634C6f7C12638a808Ad5`).

Is this a bug in `aut` or are these values changeable post-genesis?

PR #920, opened 2024-01-03
fix autonity contract deployment, use treasury instead of operator

lorenzo-dev1:
closes https://github.com/autonity/autonity/issues/919


vpiyush (approved):

lorenzo-dev1:
> LGTM but I'm confused how did we let that slip through our e2e tests.

Yes, we should catch that I think. I saw you added this to the testing workflow issue, so I will merge this in the meantime.

Issue #901, opened 2023-12-13
Committee selection: Proposal and design of a new approach


tbssajal:
Proposing some optimization and using some threshold on stakes to optimize committee selection. Implemented in #896.
Optimization done:
- Using CommitteeMember struct to create copies in memory because it has the minimum information needed to sort. Creating unnecessary copies of information in memory takes a lot of time and gas.
- Removing array `_committeeList` as it is unnecessary
- Create a storage pointer of validator before querying its information, because using `validators[address].item` every time to query the item does lot of operations to locate the struct first in the map, and then get the `item`. Creating a storage pointer will locate the struct in the map only once, so it saves time and gas where there are more than one queries. However if there is only 1 query, then creating storage pointer costs more gas and runtime because it creates a pointer in memory.
- Optimize quick sort: sort the left or right sub segment only if not sorted already.
- Threshold on stakes: Using a threshold `minimumBondedStake` and taking validators which have more than `minimumBondedStake` stakes. This allows to minimize extra memory usage and reduces the runtime and gas usage.

Except the last one, all other optimizations have been benchmarked against the current implementation and give better runtime. Share any thoughts of improving committee selection here.

tbssajal:
Latest solution implemented in #921. It does the sorting of validators according to their stake and committee selection and updates the committee in persistent storage, all in precompiled contract. For 100,000 validators, all with positive stakes, it takes 70-80 ms time. Machine specification:
```
cpu: AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx 
ram: 16 GB
```

Issue #904, opened 2023-12-13
Remove useless validators from validatorList array


tbssajal:

tbssajal:
I suggest we don't put a newly registered validator's address in `validatorList`, we just put it in `validators` map. When the newly registered validator has a `minimum amount` of stake by bonding, we can put its address in `validatorList`. I would argue this `minimum amount` to be somewhat big to prevent missuse, otherwise someone can bond some wei and put the address in `validatorList` array.

PR #921, opened 2024-01-03
Committee selection in precompiled contract

### LINKED ISSUE #922: Implementation of Committee Selection via Precompiled Contract

tbssajal:
Closes #922 and #901.
- Moved committee selection process in precompiled contract. Reading from `StateDB`, sorting the validators and writing to `StateDB`, all are done in precompiled contract.


tbssajal:
Updated benchmark time here: https://github.com/autonity/autonity/issues/901#issuecomment-1875345079

core/vm/contracts.go
```diff
@@ -163,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+
+type CommitteeSelector struct{}
+
+func (a *CommitteeSelector) RequiredGas(input []byte) uint64 {
+	times := uint64(len(input)/KB + 1)
+	return params.AutonityAFDContractGasPerKB * times
+}
+
+func (a *CommitteeSelector) Run(input []byte, _ uint64, stateDB StateDB, caller common.Address) ([]byte, error) {
+	// input
```
yazzaoui, line 194: You need to make a check on the caller to ensure only the autonity contract can call this function, otherwise it will panic with malformed input.

core/vm/contracts.go
```diff
@@ -163,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	offset += 32
+	configCommitteeLen := int(big.NewInt(0).SetBytes(input[offset : offset+32]).Int64())
+
+	// validatorListSlot has the size of the array validatorList
+	validatorListSize := int(stateDB.GetState(caller, common.BytesToHash(validatorListSlot)).Big().Uint64())
+	baseOffsetArray := crypto.Keccak256Hash(validatorListSlot).Big()
+	threshold := big.NewInt(0)
+
+	// get validators from DB concurrently
+	threadCount := 10000
```
yazzaoui, line 212: Too many go routines, there are no real gains to go above the CPU core count.

core/state/state_object.go
```diff
@@ -148,21 +151,31 @@ func (s *stateObject) touch() {
 }
 
 func (s *stateObject) getTrie(db Database) Trie {
+	if s.trie != nil {
+		return s.trie
+	}
+	return s.tryFetchTrie(db)
+}
+
+func (s *stateObject) tryFetchTrie(db Database) Trie {
```
vpiyush, line 160: what's the need to introduce a separate function, Unless we have to call only this?

core/vm/contracts.go
```diff
@@ -162,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+			})
+		}
+	}
+	return validators
+}
+
+// sorts validators according to their VotingPower in descending order
+// in case more than one valdiator have same VotingPower, the output could be different
+// depending on the positions of the validators, even if the set of validators is same each time
+func quickSort(validators []*types.CommitteeMember, low int32, high int32, task *sync.WaitGroup) {
```
vpiyush, line 354: did you try the go's inbuilt sort. It is optimized for performance and works very well with already sorted datasets. (pdq-sort).

core/vm/contracts.go
tbssajal: Could you help how to check if it is autonity contract?

core/vm/contracts.go
```diff
@@ -162,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		// We need reference of validator mapping + relative offset of bondedStake + relavtive offset of state
+		key := make([]byte, 64)
+		copy(key[12:32], address.Bytes())
+		copy(key[32:64], validatorsSlot)
+		baseOffsetMap := crypto.Keccak256Hash(key).Big()
+		// bondedStake is at slot 5
+		baseOffsetMap.Add(baseOffsetMap, big.NewInt(5))
+		bondedStake := stateDB.GetState(caller, common.BytesToHash(baseOffsetMap.Bytes())).Big()
+		// status is at slot 18
+		baseOffsetMap.Add(baseOffsetMap, big.NewInt(13))
```
vpiyush, line 338: bondedStake and Status can be fetched concurrently.

core/state/state_object.go
tbssajal: I separated it because fetching the trie seems to be some extra work which is not needed once it fetched. And due to concurrent reading, it was being fetched multiple times. Also I was getting error while running test with -race because s.trie was being written cocurrently.

core/vm/contracts.go
```diff
@@ -162,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+			}
+			if isRightSorted && j+1 < high {
+				isRightSorted = validators[j+1].VotingPower.Cmp(validators[j+2].VotingPower) >= 0
+			}
+		}
+	}
+
+	if !isLeftSorted && low < j {
+		task.Add(1)
+		go func() {
```
vpiyush, line 393: Does this improve the execution time? I assume the go routine creation and scheduling cost would itself overshadow the performance gain here.

core/vm/contracts.go
```diff
@@ -162,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+			// Recursion call in the left partition of the array
+			quickSort(validators, low, j, task)
+			task.Done()
+		}()
+	}
+	if !isRightSorted && i < high {
+		task.Add(1)
+		go func() {
```
vpiyush, line 401: same here

core/vm/contracts.go
vpiyush: do you have the test which you ran with inbuilt sort?

core/state/state_object.go
vpiyush: I think those are wrong reasons to separate it in a different function. also if you use the lock in getTrie itself you would not see the issues with -race. and using a lock in one function and not in another defeats the purpose of locks. doesn't it?

core/vm/contracts.go
vpiyush, line 1: What is the Gas cost of this precompiled contract for the 10k and 100k validator tests you did? Would like to know those if you could provide them. There is another idea if you haven't explored it already is to use index based sorting to just sort the index of the validator array by voting power.  I would hope that it would be more gas-efficient. Do give it a though t if not explored already.

core/vm/contracts.go
vpiyush: I agree that It wouldn't be wise to create too many goroutines. IMHO limiting you could still limit the number of go routines with this concurrency in place. Also, a side note - you could still get the advantage of having more than the number of cores if you are dealing with I/O or any blocking operation(but don't opt for very large numbers). Have a look at Concurrency vs Parallelism (In your case I see the whole getTrie is locked so that part of the code is as good as single-threaded)

core/vm/contracts.go
tbssajal: As you can see QuickSortFast has the best performance. So I used QuickSortFast.

core/vm/contracts.go
tbssajal: Benchmarks on computeCommittee take around around 8.7 seconds and 41 gwei of gas for 100,000 validators and 0.8 seconds and 0.5 gwei of gas for 10,000 validators. You can find them in PR #892 in description.

> There is another idea if you haven't explored it already is to use index based sorting

Yeah I thought about it. But if everything is done in precompiled, I guess gas is not an issue. In terms of performance, it seems to be not too much of an optimization because it works with pointers only, like swap or anything. But it will occupy less memory so it could be an optimization, not sure really. And if we do index based sorting, then we need to read the addresses from DB again for writing committee. I can give it a try and see how much is improved.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -131,7 +131,6 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     CommitteeMember[] internal committee;
     uint256 public totalRedistributed;
     uint256 public epochReward;
-    string[] internal committeeNodes;
```
yazzaoui, line 134: revert

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1226,39 +1173,39 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
-            if (_i <= _j) {
-                (_users[uint(_i)], _users[uint(_j)]) = (_users[uint(_j)], _users[uint(_i)]);
-                _i++;
-                _j--;
+     * @dev Sends necessary slots to precompiled contract.
+     * Committee selection and storing the committee and writing it in persistent storage are done in precompiled contract
+     */
+    function computeCommitteePrecompiled(uint256 _committeeSize) internal returns (address[] memory) {
+        require(_committeeSize <= 1000, "hardcoded array size 1002");
+        address[1002] memory _returnData;
```
tbssajal, line 1181: This is for the oracle addresses. But we can get the oracle address from `committee` array after `computeCommitteePrecompiled` is done. Which one is better?

core/vm/contracts.go
```diff
@@ -162,24 +169,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		key := make([]byte, 64)
+		copy(key[12:32], address.Bytes())
+		copy(key[32:64], validatorsSlot)
+		baseOffsetMap := crypto.Keccak256Hash(key).Big()
+		// bondedStake is at slot 5
+		baseOffsetMap.Add(baseOffsetMap, big.NewInt(5))
+		bondedStake := stateDB.GetState(caller, common.BytesToHash(baseOffsetMap.Bytes())).Big()
+		// status is at slot 18
+		baseOffsetMap.Add(baseOffsetMap, big.NewInt(13))
+		state := stateDB.GetState(caller, common.BytesToHash(baseOffsetMap.Bytes())).Big()
```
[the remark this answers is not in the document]
vpiyush: Ideally every field fo statedb which is concurrently being accessed should be protected. e.g. setStateObject is being done concurrently without any safety this can cause panic. Also metrics are getting updated which are not atomic. there are other members which are being accessed too. getTrie needs protection too.  Also the TryGet method. The Trie object itself is not thread-safe.

core/vm/contracts.go
vpiyush: I was curious on these results so i checked it on my machine. I did a minor change on the library sort to not use the sorting interface instead the sort.slice with comparison func. It should avoid the interface conversion and performance cost associated with it. Here are the results on your tests, for various validator size set:

```
**val = 100**
BenchmarkSortingAlgo/SortLibrarySort
BenchmarkSortingAlgo/SortLibrarySort-8         	  716415	     20365 ns/op	   15608 B/op	     304 allocs/op
BenchmarkSortingAlgo/QuickSortFast
BenchmarkSortingAlgo/QuickSortFast-8           	  396495	     30065 ns/op	   18640 B/op	     367 allocs/op

**val = 1000**
BenchmarkSortingAlgo/SortLibrarySort
BenchmarkSortingAlgo/SortLibrarySort-8         	   51724	    254471 ns/op	  153017 B/op	    3004 allocs/op
BenchmarkSortingAlgo/QuickSortFast
BenchmarkSortingAlgo/QuickSortFast-8           	   58646	    278887 ns/op	  183891 B/op	    3647 allocs/op

**val=10000**
BenchmarkSortingAlgo/SortLibrarySort
BenchmarkSortingAlgo/SortLibrarySort-8         	    4942	   2560870 ns/op	 1529668 B/op	   30004 allocs/op
BenchmarkSortingAlgo/QuickSortFast
BenchmarkSortingAlgo/QuickSortFast-8           	    5833	   4473629 ns/op	 1841508 B/op	   36500 allocs/op

**val=100000**
BenchmarkSortingAlgo/SortLibrarySort
BenchmarkSortingAlgo/SortLibrarySort-8         	     399	  29556865 ns/op	15205959 B/op	  300004 allocs/op
BenchmarkSortingAlgo/QuickSortFast
BenchmarkSortingAlgo/QuickSortFast-8           	     487	  30708891 ns/op	18333052 B/op	  365149 allocs/op
```

From these results, I would say we use the library sort function for following reasons:
a) less code maintainence for us,
b) goroutine variant does more allocation which incurs GC pressure,
c) in an isolated environment it's just slightly bad but with a running process when there are other goroutines fighting for CPU usage it's going to be even worse.

tbssajal: It says sort.Slice is not stable, that is why I was using sort.SliceStable with compare function.
```
sort.SliceStable(validators, func(i, j int) bool {
			return validators[i].Stake.Cmp(validators[j].Stake) == 1
		})
```
As long as the output is same for the same input array, it is not an issue, but I think we need to be sure if sort.Slice has any randomness, because if it has then  I think we cannot use it.

tbssajal: I will test sort.Slice, thanks @vpiyush. But if you know anything about the unstable behavior please share.

autonity/solidity/contracts/Autonity.sol

core/vm/contracts.go
tbssajal: updated

core/vm/contracts.go
vpiyush: [Here](https://pkg.go.dev/sort#Slice) it is:

> The sort is not guaranteed to be stable: equal elements may be reversed from their original order. For a stable sort, use SliceStable.

It basically means if 2 elements are equal, they may or may not preserve their original order form the input list. QuickSort by nature is not stable. Your own implementation is also not as far as i understand. But that really doesn't matter for us.

core/state/state_object.go
tbssajal: > using a lock in one function and not in another defeats the purpose of locks. doesn't it?

No I think because getTrie will read s.trie if it is already set, but if not set then it will call tryFetchTrie and will write to s.trie. That is also the reason I separated it into 2 functions, if reading concurrently then the writing part should be in separate function I think. Or maybe I am missing something.

core/vm/contracts.go
tbssajal: > In your case I see the whole getTrie is locked so that part of the code is as good as single-threaded

Yeah. There is also the part where data is fetched from `s.db.snap`. I am not sure when it uses trie or when it uses `s.db.snap`.

core/state/state_object.go
vpiyush: The s.Trie access (checking if it is nil or not itself) is a shared access since you are calling this concurrently. a lock is only effective if it's protected on all access (read or write).

core/vm/contracts.go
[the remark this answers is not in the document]
tbssajal: > The Trie object itself is not thread-safe

Yeah I addressed that by putting lock in the whole trie section.

> setStateObject is being done concurrently without any safety this can cause panic. Also metrics are getting updated which are not atomic.

I did not notice that. Thanks! So I will try to update them then. I will also measure the performance again with and without concurrency and see if we really gain much improvement.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -619,65 +618,7 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
-        epochTotalBondedStake = 0;
-        for (uint256 _k = 0; _k < _committeeLength; _k++) {
-            CommitteeMember memory _member = CommitteeMember(_committeeList[_k].nodeAddress, _committeeList[_k].bondedStake);
-            committee.push(_member);
-            committeeNodes.push(_committeeList[_k].enode);
-            _voterList[_k] = _committeeList[_k].oracleAddress;
-            epochTotalBondedStake += _committeeList[_k].bondedStake;
-        }
-        return _voterList;
+        return computeCommitteePrecompiled(config.protocol.committeeSize);
```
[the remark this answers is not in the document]

core/state/state_object.go
tbssajal: I see. So I need to use a read lock as well. Also it can be done in a single function, I see.

yazzaoui:
The failing slashing test is due to the combinations of the following factors:
- Committee is now sorted at genesis - which is good.
- Reporting is based on the current validator index - That is also fine. See https://github.com/autonity/autonity/blob/ee8c11f0b2a0070b05812e6259868579c9d11036/consensus/tendermint/accountability/fault_detector.go#L380-L398
- The slashing test is only waiting for one epoch before checking if a fault has been registered . https://github.com/autonity/autonity/blob/ee8c11f0b2a0070b05812e6259868579c9d11036/e2e_test/byzantine/slashing_test.go#L122
This can be I think safely extended to have enough time for a validator to report the fault.

tbssajal:
> This can be I think safely extended to have enough time for a validator to report the fault.

Extending timeout doesn't fix the tests actually. But extending the epoch period fixes the test.

core/vm/contracts.go
```diff
@@ -163,24 +175,178 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+
+type CommitteeSelector struct{}
+
+func (a *CommitteeSelector) RequiredGas(input []byte) uint64 {
+	times := uint64(len(input)/KB + 1)
+	return params.AutonityAFDContractGasPerKB * times
+}
+
+func (a *CommitteeSelector) Run(input []byte, _ uint64, stateDB StateDB, caller common.Address) ([]byte, error) {
+	if caller != params.AutonityContractAddress {
```
yazzaoui, line 211: The least "intrusive" way to bypass the check i could think of would be to :
- Add a new genesis configuration option "testMode" , in `param.ChainConfig`
- Pass the the whole evm context in place of StateDB in the input argument of `Run`
- Bypass this check if `evm.chainConfig.testMode` is true

tbssajal: Or it it possible to read the contract address from genesis block? In case genesis block does not exist we can use the hardcoded `params.AutonityContractAddress`.

core/vm/contracts.go
tbssajal: using library function `slices.SortStableFunc`

core/vm/contracts.go
tbssajal: using library function `slices.SortStableFunc`

core/vm/contracts.go
tbssajal: I see it does not work because in truffle tests we deploy new contract everytime.

Jason-Zhangxin-Chen (commented):
This is a very interesting approach to implement the commitee election. As all the precompiled contracts of now are state readers, they don't write state update back to stateDB, thus we have to be very careful to it. I will analyze the execution context to double check if this approach is safe or not.

core/state/state_object.go
tbssajal: removed concurrency

core/vm/contracts.go
tbssajal: removed concurrency

core/vm/contracts.go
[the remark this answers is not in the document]
tbssajal: removed concurrency

yazzaoui (approved):
LGTM

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -25,6 +26,22 @@ library Precompiled {
...
+     * @dev Sends necessary slots to precompiled contract.
+     * Committee selection and storing the committee and writing it in persistent storage are done in precompiled contract
+     */
+    function computeCommitteePrecompiled(uint256[5] memory input) internal {
+        address to = COMPUTE_COMMITTEE_CONTRACT;
+        uint256 _length = 32*5;
+        assembly {
+            //delegatecall(gasLimit, to, inputOffset, inputSize, outputOffset, outputSize)
+            if iszero(delegatecall(gas(), to, input, _length, 0, 0)) {
+                returndatacopy(0, 0, returndatasize())
```
lorenzo-dev1, line 39: Just curious, why do we need this code if the computeCommittee reverts?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	validatorListSlot         []byte
+	validatorsSlot            []byte
+	committeeSlot             []byte
+	epochTotalBondedStakeSlot []byte
+	configCommitteeLen        int
+}
+
+func (a *CommitteeSelector) RequiredGas(input []byte) uint64 {
+	times := uint64(len(input)/KB + 1)
+	return params.AutonityAFDContractGasPerKB * times
```
lorenzo-dev1, line 233: I think this is fine but I would define a new constant for the GasPerKb that is standalone for the committee selector, instead of using the one defined for Accountability precompiles

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+var errUnauthorized = errors.New("caller address not authorize")
+
+type CommitteeSelector struct{}
+
+type CommitteeInput struct {
+	validatorListSlot         []byte
+	validatorsSlot            []byte
+	committeeSlot             []byte
+	epochTotalBondedStakeSlot []byte
+	configCommitteeLen        int
```
lorenzo-dev1, line 228: maybe MaxCommitteeSize would be more meaningful?

core/vm/contracts.go
tbssajal: Agreed. The value should be same?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	return params.AutonityAFDContractGasPerKB * times
+}
+
+func (a *CommitteeSelector) Run(input []byte, _ uint64, evm *EVM, caller common.Address) ([]byte, error) {
+	// skip auth check if run in test mode
+	if !evm.chainConfig.TestMode && caller != params.AutonityContractAddress {
+		return nil, errUnauthorized
+	}
+	// input
+	if len(input) < 5*DataLen {
```
lorenzo-dev1, line 242: I don't understand this, `DataLen = lenght of an hash`. Is it because the slots are 32 byte hashes? And since the uint256 for the committee size is also 32 bytes we just say `5*DataLen` instead of `4 * hashlength + uint256 lenght?`

lorenzo-dev1: If that is the case I think it is fine but I would add a comment explaining it just for clarity sake

core/vm/contracts.go
lorenzo-dev1: Also in solidity it is a uint256... so I would store it in a big.Int here to avoid any possible overflow/underflow issues... Or at least as a uint64 (since we will probably never exceed 2^64 committee members anyways)

core/vm/contracts.go
lorenzo-dev1: For the moment I would say yes

core/vm/contracts.go
lorenzo-dev1: You also have the `IsUint64` to check if a big int can be safely casted to uint64... if not we should revert

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	inputs.committeeSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	inputs.epochTotalBondedStakeSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	inputs.configCommitteeLen = int(big.NewInt(0).SetBytes(input[offset : offset+DataLen]).Int64())
+
+	stateDB := evm.StateDB
+	// detailed specifications on how to read state variable from slot: https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#storage-inplace-encoding
+	// validatorListSlot has the size of the array validatorList
+	validatorListSize := int(stateDB.GetState(caller, common.BytesToHash(inputs.validatorListSlot)).Big().Uint64())
```
lorenzo-dev1, line 260: Why converting to int?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		committeeOffset.Add(committeeOffset, increment)
+	}
+}
+
+// solidity storage layout: https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#storage-inplace-encoding
+func (a *CommitteeSelector) getValidatorInfo(
+	mapKey []byte, addressSlot []byte, stakeThreshold *big.Int, expectedState *big.Int, caller common.Address, stateDB StateDB,
+) *types.CommitteeMember {
+
+	address := common.BytesToAddress(stateDB.GetState(caller, common.Hash(addressSlot)).Bytes())
```
lorenzo-dev1, line 377: If I understood correctly the solidity storage layout, this should be the `nodeAddress` of a validator. If so, I would suggest to rename `address` to `nodeAddress` to be more specific and help comprehension since it is quite a low-level piece of code.

core/vm/contracts.go
tbssajal: > Is it because the slots are 32 byte hashes?

Yes. I will add comments.

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	inputs.epochTotalBondedStakeSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	inputs.configCommitteeLen = int(big.NewInt(0).SetBytes(input[offset : offset+DataLen]).Int64())
+
+	stateDB := evm.StateDB
+	// detailed specifications on how to read state variable from slot: https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#storage-inplace-encoding
+	// validatorListSlot has the size of the array validatorList
+	validatorListSize := int(stateDB.GetState(caller, common.BytesToHash(inputs.validatorListSlot)).Big().Uint64())
+	addressOffset := crypto.Keccak256Hash(inputs.validatorListSlot).Big()
+	stakeThreshold := big.NewInt(StakeThreshold)
```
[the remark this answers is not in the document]
tbssajal: `StakeThreshold` is constant so it is int.

core/vm/contracts.go
tbssajal: because we need validatorListSize for iteration, easy to compare in int. But I will add a check so it is not negative.

lorenzo-dev1: Also I am thinking that nothing stops a malicious party to register tons of validator, potentially more than 2^64, which might cause problem here when casting to uint64 (behaviour is undefined if it cannot be casted). Maybe we should add a check on this at registration time? Or at least handle it gracefully here.

@yazzaoui

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		// goto next slot
+		addressOffset.Add(addressOffset, increment)
+	}
+
+	if len(validators) == 0 {
+		return nil, fmt.Errorf("no active validators with stake more than %v", stakeThreshold.String())
+	}
+
+	var committeeSize int
+	if inputs.configCommitteeLen > len(validators) {
```
lorenzo-dev1, line 286: If possible use the golang `min` builtin function, it improves readability

tbssajal: tried it, but common/math already imported so skipped that. Should I add a function Min() in common/math or keep it as it is?

core/vm/contracts.go
tbssajal: I addressed this potential issue in #904 with gas usage. Maybe we don't need to put a newly registered validator into validatorList before it has any bonding request applied. Please share any thoughts in this issue. Because if we have so many validators registered even with 0 stakes, it will take too much time to complete the process and it has 0 gas cost because it is in precompiled contract.

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		// save address
+		stateDB.SetState(caller, common.Hash(committeeOffset.Bytes()), validators[i].Address.Hash())
+		// goto next slot
+		committeeOffset.Add(committeeOffset, increment)
+		// save voting power
+		stateDB.SetState(caller, common.Hash(committeeOffset.Bytes()), common.BytesToHash(validators[i].VotingPower.Bytes()))
+		// goto next slot
+		committeeOffset.Add(committeeOffset, increment)
+
+		// read and then save consensus key (bytes)
```
lorenzo-dev1, line 328: Is this not done in getValidatorInfo for performance improvement?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		copy(mapKey[DataLen-common.AddressLength:DataLen], validators[i].Address.Bytes())
+		mapItemSlot := crypto.Keccak256Hash(mapKey).Big()
+		mapItemSlot.Add(mapItemSlot, big.NewInt(ConsensusKeyOffset))
+		data := stateDB.GetState(caller, common.BigToHash(mapItemSlot)).Bytes()
```
lorenzo-dev1, line 332: `data` --> `consensusKeyBytes` ?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		// if last bit is not set, then data is short and contains all the info
+		// otherwise we have (2*length + 1) at the slot and the bytes is saved as bytes1[]
+		stateDB.SetState(caller, common.Hash(committeeOffset.Bytes()), common.BytesToHash(data))
+		if (data[DataLen-1] & 1) == 1 {
```
lorenzo-dev1, line 336: Why do we need this? Shouldn't consensusKey be of same size for all validators?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	// delete old committee members, if any
+	for i := committeeSize; i < oldCommitteeSize; i++ {
+		// delete address
+		stateDB.SetState(caller, common.Hash(committeeOffset.Bytes()), common.Hash{})
+		// goto next slot
+		committeeOffset.Add(committeeOffset, increment)
+		// delete voting power
+		stateDB.SetState(caller, common.Hash(committeeOffset.Bytes()), common.Hash{})
+		// goto next slot
+		committeeOffset.Add(committeeOffset, increment)
```
lorenzo-dev1, line 368: we don't delete the consensuskey?

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
 	suppliedGas -= gasCost
-	output, err := p.Run(input, blockNumber)
+	output, err := p.Run(input, blockNumber, evm, caller)
 	return output, suppliedGas, err
 }
 
+const (
+	KB = 1024
+	// offset constants of Validator struct from Autonity.sol
+	OracleOffset       = 2
```
lorenzo-dev1, line 209: I guess there is now way to get these offsets dynamically from the contract ABI?

autonity/solidity/test/validator_management.js
```diff
@@ -9,6 +9,10 @@ const config = require('./config.js')
 
 contract('Autonity', function (accounts) {
 
+    before(async function () {
```
lorenzo-dev1, line 12: I don't this is needed because validator_management only runs against autonity at the moment (we do not have POPverifier precompile solidity mock)

tests/fuzzers/bls12381/precompile_fuzzer.go
```diff
@@ -101,3 +109,25 @@ func fuzz(id byte, data []byte) int {
 	}
 	return 1
 }
+
+func testEVM() (*vm.EVM, error) {
```
lorenzo-dev1, line 113: Since I see this `testEVM()` function repeated in multiple test files, maybe it is worth to just add it to the evm package and avoid code repetition. It will make tests less brittle

lorenzo-dev1: Also for this test (fuzzing bls precompiles) I guess you could just pass nil as as EVM? in any case it will be ignored.

core/vm/contracts.go
tbssajal: Following [this](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#bytes-and-string) to do it appropriately. I think it is better this way because I remember a conversation from standup about short and long bls public key and which one to use. So maybe they could be different for different users? Either way it should not cost anything at runtime.

core/vm/contracts.go
tbssajal: right, forgot to update.

autonity/solidity/test/autonity.js
```diff
@@ -1306,4 +1304,23 @@ contract('Autonity', function (accounts) {
         // check that blockEpochMap and getEpochFromBlock return the numbers we expect. Terminate a couple epochs and check the variables.
       });
   });
+
+  describe('Test functions', function () {
```
lorenzo-dev1, line 1308:
```suggestion
  describe('Test computeCommittee', function () {
```

core/vm/contracts.go
tbssajal: I could not find any way and I don't think it is possible for a particular reason. We can discuss it after standup.

tests/fuzzers/bls12381/precompile_fuzzer.go
tbssajal: Actually, it is better to use nil I think. Because Youssef is also updating the test framework.

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -49,4 +53,57 @@ contract AutonityTest is Autonity {
...
 
+   function testComputeCommittee() public {
+      _stakingOperations();
+      address[] memory voters = computeCommittee();
+      address[] memory addresses = new address[](voters.length);
+      uint256 totalStake = 0;
+      uint256 minStake = 0;
+      uint256 lastStake = 0;
+      for (uint256 i = 0; i < committee.length; i++) {
+        address memberAddress = committee[i].addr;
```
lorenzo-dev1, line 64: Maybe check also that the consensus key is not empty

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -49,4 +53,57 @@ contract AutonityTest is Autonity {
...
+        require(memberAddress != address(0), "invalid address");
+        addresses[i] = memberAddress;
+        uint256 stake = committee[i].votingPower;
+        require(stake > 0, "0 stake in committee");
+        totalStake += stake;
+        if (i == 0) {
+          minStake = stake;
+        }
+        else if (stake < minStake) {
```
lorenzo-dev1, line 73: shouldn't minstake always be the last committee member?

autonity/solidity/contracts/AutonityTest.sol
lorenzo-dev1: Nevermind I saw you check that later

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -49,4 +53,57 @@ contract AutonityTest is Autonity {
...
    }
 
+   function testComputeCommittee() public {
+      _stakingOperations();
+      address[] memory voters = computeCommittee();
+      address[] memory addresses = new address[](voters.length);
+      uint256 totalStake = 0;
+      uint256 minStake = 0;
+      uint256 lastStake = 0;
+      for (uint256 i = 0; i < committee.length; i++) {
```
lorenzo-dev1, line 63: maybe add check that committee.length is < maxocmmittesize

autonity/solidity/contracts/AutonityTest.sol
```diff
@@ -49,4 +53,57 @@ contract AutonityTest is Autonity {
...
+        else if (stake < minStake) {
+          minStake = stake;
+        }
+
+        if (i > 0) {
+          require(lastStake >= stake, "committee members not sorted");
+        }
+        lastStake = stake;
+        Validator storage validator = validators[memberAddress];
+        require(validator.nodeAddress == memberAddress, "validator does not exist");
```
lorenzo-dev1, line 82: maybe add check that state is active

autonity/autonity_test.go
```diff
@@ -332,7 +376,126 @@ func testEVMProvider() func(header *types.Header, origin common.Address, statedb
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, gasLeft, err := evmContract.CallContractFunc(stateDB, header, contractAddress, packedArgs)
+		require.NoError(b, err)
+		gasUsed += gas - gasLeft
+	}
+	b.Log(1.0 * gasUsed / uint64(b.N))
+}
+
+func isVotersSorted(voters []common.Address, committeeMembers []types.CommitteeMember, validators []params.Validator, enodes []string, totalStake *big.Int) error {
```
lorenzo-dev1, line 401: The same comments I left for AutonityTest.sol are valid here since it seems to me that it replicates the same logic

core/vm/contracts.go
yazzaoui: Since the protocol address is the only caller accepted for this precompile, accurate `RequiredGas` is unecessary, as we are giving practically an infinite amount of gas to run `finalize()`. We still have to make sure it can't be a vector for spamming transactions targetting this precompile and reverting so it can't be absolutely free.
On my upgrade contract branch I've set a constant amount of gas to be returned here.

core/vm/contracts.go
yazzaoui: Getting into an overflow situation with more than 2^64 committee members is a very theoretical situation. It will never happen. It would cost too much in gas for an attacker to execute that and we will have a liveness issue long before then because sorting would take too long.
Back-of-the-envelope calculation is roughly one year at 500 billion TPS in order to register  2^64 validators.

core/vm/contracts.go
```diff
@@ -179,24 +191,215 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+}
+
+func (a *CommitteeSelector) Run(input []byte, _ uint64, evm *EVM, caller common.Address) ([]byte, error) {
+	// skip auth check if run in test mode
+	if !evm.chainConfig.TestMode && caller != params.AutonityContractAddress {
+		return nil, errUnauthorized
+	}
+	// input
+	if len(input) < 5*DataLen {
+		return nil, fmt.Errorf("not enough input")
```
yazzaoui, line 244: Returning `fmt.Errorf("......")` is bad pattern, it doesn't let the caller perform a check on the error type easily using `errors.Is`. Better to create a new global variable for errors `errBadInput = errors.New("....")`.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -631,73 +635,29 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
-                _validatorList[j] = _user;
-                j++;
-            }
+        uint256[5] memory input;
+        input[4] = config.protocol.committeeSize;
+        assembly {
+            mstore(input, validatorList.slot)
+            mstore(add(input, 0x20), validators.slot)
+            mstore(add(input, 0x40), committee.slot)
+            mstore(add(input,0x60), epochTotalBondedStake.slot)
```
lorenzo-dev1, line 644: Unless I am missing something I think it is not necessary to pass the epochTotalBondedStake slot to the precompiled. It could be computed when iterating over the committee to generate the voters list. However it is not a big deal

core/vm/contracts.go
```diff
@@ -179,24 +191,245 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+	offset := 0
+	inputs.validatorListSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	inputs.validatorsSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	inputs.committeeSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	inputs.epochTotalBondedStakeSlot = input[offset : offset+DataLen]
+	offset += DataLen
+	maxCommitteeSize := common.Big0.SetBytes(input[offset : offset+DataLen])
```
lorenzo-dev1, line 256: I think this will modify common.Big0

core/vm/contracts.go
lorenzo-dev1: Now I see why you had called it data, is it because it could be either the consensusKeyBytes or the length of it? (In our case it will always be the length of it since consensus key is always > 32bytes)

autonity/solidity/contracts/Autonity.sol
tbssajal: You are right. Actually I was trying to do everything in the precompiled contract. Later I had to do some part in solidity and did not update `epochTotalBondedStake` part.

tbssajal: I could not return the voters from precompiled contract using dynamic array. If someone figures it out, maybe it can be modified later to do everything in the precompiled part.

autonity/solidity/contracts/AutonityTest.sol
tbssajal: done

autonity/solidity/contracts/AutonityTest.sol
tbssajal: done

core/vm/contracts.go
```diff
@@ -179,24 +191,251 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+		copy(mapKey[DataLen-common.AddressLength:DataLen], validators[i].Address.Bytes())
+		mapItemSlot := crypto.Keccak256Hash(mapKey).Big()
+		mapItemSlot.Add(mapItemSlot, big.NewInt(ConsensusKeyOffset))
+		consensusKeyBytes := stateDB.GetState(caller, common.BigToHash(mapItemSlot)).Bytes()
+		// if last bit is not set, then data is short and contains all the info
+		// otherwise we have (2*length + 1) at the slot and the bytes is saved as bytes1[]
+		stateDB.SetState(caller, common.BigToHash(committeeOffset), common.BytesToHash(consensusKeyBytes))
+		if (consensusKeyBytes[DataLen-1] & 1) == 1 {
+			keyLenBig := new(big.Int).SetBytes(consensusKeyBytes)
+			if !keyLenBig.IsUint64() {
```
lorenzo-dev1, line 348: Here you should check `isInt64()` since you are casting to `Int64()`. Also here this check is not really needed because this variable is not under the control of an attacker (attacker cannot register bls keys which are longer than what we expect). 

It was needed only for the validatorlist lenght, because that is actually unbounded and under attacker control.

lorenzo-dev1: Same is valid for the deletion part

core/vm/contracts.go
tbssajal: done

Issue #544, opened 2020-05-05
Remove use of custom resolve functions in p2p/enode/urlv4.go


piersy:
# Problem

We allow custom resolve functions only so that we can use fake hosts in the tests of the form 'VA',  'SB', 'PX' standing for Validator, A Stakeholder B  and Participant X.  We shouldn't be hacking our real code to support tests, instead we should be constructing our tests so that they work with the real code. In this case I guess the 'VA',  'SB', 'PX' makes it easy to spot in logs what type a node is, but we can instead provide a function that pretty prints a node, including all the relevant information, which would free us from needing to use fake hostnames in our tests and hence remove the need for custom resolve functions in `enode`. 

# Implementation

Create a function that pretty prints a node with relevant information, use that in place of fake host names in enodes. Then remove abilty to set custom resolve functions from `p2p/enode/urlv4.go` 

# Acceptance Criteria

The `enode` package should have no mechanism for changing the host resolution.

lorenzo-dev1:
reopening, this issue is still present.

PR #923, opened 2024-01-05
Consensus channel separation - introducing ACN protocol

vpiyush:
#  Description:

For a comprehensive understanding of the context and motivation behind this pull request, please refer to the discussion at [this link](https://github.com/clearmatics/autonity-protocol/discussions/250).

In this PR, a dedicated network for committee members, interconnected in a mesh network, is introduced. To enable this separate network, two new command line options have been added:

1. **Consensus Listener Port:**
   - Name: "consensus.port"
   - Usage: "Network listening port for consensus channel"
   - Default Value: 20203

2. **Consensus endpoint NAT Mechanism:**
   - Name: "consensus.nat"
   - Usage: "NAT port mapping mechanism for consensus channel (any|none|upnp|pmp|extip:<IP>)"
   - Default Value: "any"

Additionally, the enode URL for the validator is structured as follows:

`pub_key@127.0.0.1:30303?discPort=30303&acn=<acn_ip>:<acn_port>`

(Acnep is an abbreviation for Autonity Consensus Network Endpoint)
- `acn_ip`: The IP on which the Autonity client's consensus network is running. It can be the same as the execution network IP or different, at the discretion of the validator. The default value is the same as the execution network IP.
- `acn_port`: The port for the consensus endpoint. The default value is 20203.

Valid examples include:
- `pub_key@127.0.0.1:30303?discPort=30303&acn=127.0.0.1:20203`
- `pub_key@127.0.0.1:30303?discPort=30303&acn=:20203`
- `pub_key@127.0.0.1:30303?discPort=30303&acn=127.0.0.1`
- `pub_key@127.0.0.1:30303?acnep=127.0.0.1:20203`

Following items are still pending and are being worked on :

- [x] Fix tests 
- [x] Address discussion points marked as TODO in the PR - WIP
- [x] review for optimizations and code refactoring - WIP
- [x] Fix metric for consensus server


atc/atc.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/autonity/autonity/consensus"
+	"github.com/autonity/autonity/core"
+	"github.com/autonity/autonity/core/forkid"
+	"github.com/autonity/autonity/crypto"
+	"github.com/autonity/autonity/log"
+	"github.com/autonity/autonity/node"
+	"github.com/autonity/autonity/p2p"
+	"github.com/autonity/autonity/p2p/enode"
+	"math"
+	"sync"
```
yazzaoui, line 33: Let's start the habit of sorting imports. If using Goland, you can configure it to do it for you :+1:

atc/committeeWatcher.go
yazzaoui, line 1: Let's use the project conventions for file names.

atc/atc.go
```diff
@@ -0,0 +1,136 @@
...
+	if handler, ok := chain.Engine().(consensus.Handler); ok {
+		handler.SetBroadcaster(atc)
+	}
+	return atc
+}
+
+func (atc *ATC) Start() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	atc.watchCommittee(ctx)
+	protocol.StartENRUpdater(ctx, atc.chain, atc.server.LocalNode())
```
yazzaoui, line 71: Do we need that?

atc/atc.go
vpiyush: can you share the setting, So we keep these identical. This is how i am setting the imports settings now.  
image

Makefile
```diff
@@ -162,8 +162,7 @@ test-contracts-pre:
 	@npm list ganache > /dev/null || npm install ganache
 	@npx truffle version
 
-APE_VERSION := 0.6.26
```
lorenzo-dev1, line 165: Is this change of version wanted? I think version in develop works

p2p/server.go
```diff
@@ -172,6 +189,7 @@ type Server struct {
 	// Config fields may not be modified while the server is running.
 	Config
 
+	Typ ServerTyp
```
yazzaoui, line 192: Would prefer renaming `ServerTyp` to `Network` and `Typ` to `Net`

p2p/server.go
```diff
@@ -41,6 +41,23 @@ import (
...
+	EthTx = iota + 1
+	Consensus
+)
+
+func (ser ServerTyp) String() string {
+	switch ser {
+	case Consensus:
+		return "Consensus"
+	case EthTx:
+		return "ethereum"
```
yazzaoui, line 56: `Execution` network

atc/committeeWatcher.go
vpiyush: Fixed

p2p/server.go
vpiyush: Yes, Updated this

Makefile
vpiyush: Not intentional, I will revert this

cmd/utils/flags.go
```diff
@@ -607,7 +619,12 @@ var (
 	ListenPortFlag = cli.IntFlag{
 		Name:  "port",
 		Usage: "Network listening port",
```
Jason-Zhangxin-Chen, line 621: I suggest to update the usage to: P2P network listening port or Network lisening port for P2P protocol channel.

cmd/utils/flags.go
```diff
@@ -607,7 +619,12 @@ var (
...
-		Value: 30303,
+		Value: node.DefaultETHPortInt,
+	}
+	NATFlag = cli.StringFlag{
+		Name:  "nat",
+		Usage: "NAT port mapping mechanism (any|none|upnp|pmp|extip:<IP>)",
```
Jason-Zhangxin-Chen, line 626: It would be better to add `P2P` to make more sense.

cmd/utils/flags.go
```diff
@@ -1166,9 +1178,45 @@ func SetP2PConfig(ctx *cli.Context, cfg *p2p.Config) {
 	}
 }
 
+func SetConsensusP2PConfig(ctx *cli.Context, cfg *p2p.Config) {
+	setNodeKey(ctx, cfg)
```
Jason-Zhangxin-Chen, line 1182: You'd need to rebase it with develop, the setNodeKey() was replaced with setAutonityKeys()

consensus/network/acn/acn.go
```diff
@@ -0,0 +1,143 @@
...
+	backend.FD().SetBroadcaster(acn)
+	return acn
+}
+
+func (acn *ACN) Start() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	acn.watchCommittee(ctx)
+	protocol.StartENRUpdater(ctx, acn.chain, acn.server.LocalNode())
+	acn.cancel = cancel
+	return nil
```
Jason-Zhangxin-Chen, line 78: As it always returns nil, how about to remove the return error in the function spec.

consensus/network/acn/committee_watcher.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+
+	"github.com/autonity/autonity/core"
+	"github.com/autonity/autonity/core/types"
+)
+
+func (acn *ACN) watchCommittee(ctx context.Context) {
+	acn.wg.Add(1)
+
```
Jason-Zhangxin-Chen, line 12:
```suggestion
// TODO: watch for epoch rotation event instead.
```

consensus/network/acn/acn.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := peer.Handshake(acn.networkID, td, hash, genesis.Hash(), forkID, acn.forkFilter); err != nil {
+		peer.Log().Debug("Consensus handshake failed", "err", err)
+		return err
+	}
+
+	if err := acn.peers.register(peer); err != nil {
+		peer.Log().Error("Snapshot extension registration failed", "err", err)
+		return err
+	}
+	//TODO: checkpoint hash and required blocks check not done on consensus channel
```
Jason-Zhangxin-Chen, line 114: The checkpoint hash is used for light syncing, I am wondering if the validator node can do such kind of sync? However, if we do not do the checkpoint hash and the required blocks check here, consensus engine might votes for consensus at an old height, while the native P2P sync side might already aborted due to this checker. A testcase is worth to test this scenario.

node/node.go
```diff
@@ -520,6 +543,16 @@ func (n *Node) Server() *p2p.Server {
 	return n.server
 }
 
+// Server retrieves the currently running P2P network layer. This method is meant
+// only to inspect fields of the currently running server. Callers should not
+// start or stop the returned server.
+func (n *Node) ConsensusServer() *p2p.Server {
```
Jason-Zhangxin-Chen, line 549: As this unlocks the lock and returns the pointer of the server, any reference of this point should be careful in case of race condition

Jason-Zhangxin-Chen (commented):
The design and structure of this ACN LGTM, just left few comments on it, but this PR is great one!

cmd/utils/flags.go
```diff
@@ -1610,6 +1658,10 @@ func RegisterEthService(stack *node.Node, cfg *ethconfig.Config) (ethapi.Backend
 	return backend.APIBackend, backend
 }
 
+func RegisterConsensusService(stack *node.Node, backend *eth.Ethereum, netID uint64) *acn.ACN {
```
lorenzo-dev1, line 1661: It is a bit weird that the return of the `acn.New` and of this function are actually ignored by the caller. I guess that we don't need it since we do the protocol registration inside the `New` (which is also the same approach used by `eth.New`).

consensus/acn/peerset.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// errPeerNotRegistered is returned if a peer is attempted to be removed from
+	// a peer set, but no peer with the given id exists.
+	errPeerNotRegistered = errors.New("peer not registered")
+)
+
+// PeerSet represents the collection of active peers currently participating in consensus
+type peerSet struct {
+	peers  map[string]*protocol.Peer // Peers connected on the `eth` protocol
+	lock   sync.RWMutex
```
lorenzo-dev1, line 47: This lock can be embedded into the struct

consensus/acn/protocol/protocol.go
```diff
@@ -0,0 +1,54 @@
...
+// devp2p capability negotiation.
+const ProtocolName = "acn"
+
+// ProtocolVersions are the supported versions of the `snap` protocol (first
+// is primary).
+var ProtocolVersions = []uint{ACNv1}
+
+// protocolLengths are the number of implemented message corresponding to
+// different protocol versions.
+var protocolLengths = map[uint]uint64{ACNv1: 22}
```
lorenzo-dev1, line 26: I think this should be 5 (propose, prevote, precommit, sync, off chain accusation).

vpiyush: we already have code defined for all the consensus messages, which starts from 0x11, If we change the length here we also end up changing those codes which i really don't have a strong opinion on. Above the acn and execution layer, those codes would conflict with other messages. Which is not really a problem if we know which protocol the message belongs too. 
Another point is if we have any other dependency on these codes which is not so obvious from the code.

I am open to changing these but it's slightly risky at this point. Let's hear from others.

p2p/server.go
vpiyush: Fixed

cmd/utils/flags.go
vpiyush: done

consensus/network/acn/acn.go
vpiyush: this implements the lifecycle interface, so we can't really change the function spec.

consensus/network/acn/acn.go
vpiyush: I would like to understand this better, In what scenario this is possible? Do we have such usecases? We could introduce a similar check during consensus handshake but would like to understand this further before that. We can discuss this in standup though, @yazzaoui inputs on this plz.

node/node.go
vpiyush: this lock is to protect node. not intended for server.

autonity/autonity.go
```diff
@@ -194,8 +194,8 @@ func (c *Cache) MinimumBaseFee() *big.Int {
 	return new(big.Int).Set(c.minBaseFee.Load())
 }
 
-func (c *AutonityContract) CommitteeEnodes(block *types.Block, db *state.StateDB) (*types.Nodes, error) {
-	return c.callGetCommitteeEnodes(db, block.Header())
+func (c *AutonityContract) CommitteeEnodes(block *types.Block, db *state.StateDB, asATC bool) (*types.Nodes, error) {
```
lorenzo-dev1, line 197: nitpick: as ACN

eth/peerset_eth.go
```diff
@@ -98,9 +99,23 @@ func (ps *peerSet) registerSnapExtension(peer *snap.Peer) error {
...
 
+// EthHandshakeStatus returns if eth handshake was successfully done
+func (ps *ethPeerSet) EthHandshakeStatus(peer *eth.Peer) bool {
+	// Ensure nobody can double connect
+	ps.lock.Lock()
+	defer ps.lock.Unlock()
+
+	id := peer.ID()
+	if _, ok := ps.peers[id]; ok {
+		ps.lock.Unlock()
```
lorenzo-dev1, line 110: Why do we need this if we have already the defer?

p2p/enode/urlv4.go
```diff
@@ -156,11 +165,69 @@ func isNewV4(n *Node) bool {
...
+	// set same ip as eth for acn protocol
+	acnIP, _, _, err = ethProtoParams(u)
+	if err != nil {
+		return "", 0, 0, err
+	}
+	acnPort = uint64(DefaultACNPortInt)
+	return acnIP, acnPort, 0, nil
+}
+
+func parseComplete(rawurl string, _ func(host string) ([]net.IP, error),
```
lorenzo-dev1, line 226: Not a big deal but it is a bit weird that we pass the resolver function down the chain of enode parsing functions, but then here we ignore it. And later when needing to do host resolution we directly call the resolve function that is defined as global variable. At this point let's not pass it down all the function calls? Or does it creates other issues?

Also related I think this issue is actually still valid https://github.com/autonity/autonity/issues/544. Tests are the underlying reason we need a variable for the host resolution function and we cannot simply call the `net.LookupIP`. But we should try to minimize code changes to make tests work.

p2p/server.go
```diff
@@ -386,10 +405,13 @@ func (srv *Server) RemoveTrustedPeer(node *enode.Node) {
 // UpdateConsensusEnodes is responsible to make sure that
 // a node belonging to the consensus committee is fully connected
 // to the other consensus committee nodes.
-func (src *Server) UpdateConsensusEnodes(enodes []*enode.Node) {
-
+func (srv *Server) UpdateConsensusEnodes(enodes []*enode.Node) {
+	var currEnodes []*enode.Node
+	srv.enodeMu.RLock()
```
lorenzo-dev1, line 410: Is read lock enough here?

lorenzo-dev1:
Do we have anywhere tests that check committee members entering/exiting the committee (and consequent timely setup/teardown of their related consensus p2p connection)? I think it would be a good addition if we have time for it.

consensus/acn/protocol/protocol.go
lorenzo-dev1: Ok I guess for the moment is no big deal since we have only the consensus protocol running on the consensus server (and probably it will remain always like this?). Also the backend/handler filters out anything which is not within the correct bounds.

I would still suggest to track it with a comment, it is kinda confusing if someone arrives on the code and doesn't know the transition we made from single server.

consensus/acn/protocol/handler.go
```diff
@@ -0,0 +1,163 @@
...
+					return Handle(backend, peer)
+				})
+			},
+			NodeInfo: func() interface{} {
+				return nodeInfo(backend.Chain(), network)
+			},
+			PeerInfo: func(id enode.ID) interface{} {
+				return backend.PeerInfo(id)
+			},
+			Attributes:     []enr.Entry{currentENREntry(backend.Chain())},
```
lorenzo-dev1, line 105: What is the purpose of this ENR entry?

vpiyush: It sets the enrkey for localnode object.

autonity/autonity.go
```diff
@@ -194,8 +194,8 @@ func (c *Cache) MinimumBaseFee() *big.Int {
 	return new(big.Int).Set(c.minBaseFee.Load())
 }
 
-func (c *AutonityContract) CommitteeEnodes(block *types.Block, db *state.StateDB) (*types.Nodes, error) {
-	return c.callGetCommitteeEnodes(db, block.Header())
+func (c *AutonityContract) CommitteeEnodes(block *types.Block, db *state.StateDB, asACN bool) (*types.Nodes, error) {
```
Jason-Zhangxin-Chen, line 197: Just wondering if we need to keep the legacy enode patterns.

cmd/autonity/config.go
```diff
@@ -224,7 +224,8 @@ func makeFullNode(ctx *cli.Context) (*node.Node, ethapi.Backend) {
 	if ctx.GlobalIsSet(utils.OverrideTerminalTotalDifficulty.Name) {
 		cfg.Eth.OverrideTerminalTotalDifficulty = new(big.Int).SetUint64(ctx.GlobalUint64(utils.OverrideTerminalTotalDifficulty.Name))
 	}
-	backend, _ := utils.RegisterEthService(stack, &cfg.Eth)
+	backend, ethBackend := utils.RegisterEthService(stack, &cfg.Eth)
+	utils.RegisterConsensusService(stack, ethBackend, cfg.Eth.NetworkID)
```
Jason-Zhangxin-Chen, line 228: If one runs just a full node client, not intend to be a validator node of the consensus, do we need to config the ACN then? We will need to create an issue to consider this.

cmd/utils/flags.go
```diff
@@ -1170,10 +1184,44 @@ func setP2PConfig(ctx *cli.Context, cfg *p2p.Config) {
...
+		}
+		cfg.NAT = natif
+	}
+	// set listener
+	if ctx.GlobalIsSet(ConsensusListenPortFlag.Name) {
+		cfg.ListenAddr = fmt.Sprintf(":%d", ctx.GlobalInt(ConsensusListenPortFlag.Name))
+	}
+
+	cfg.MaxPeers = math.MaxInt
+	cfg.MaxPendingPeers = 100 // current max committee size
```
Jason-Zhangxin-Chen, line 1201: Let's move `100` to a predefined variable.

consensus/acn/acn.go
```diff
@@ -0,0 +1,142 @@
...
+		td      = acn.chain.GetTd(hash, number)
+	)
+	forkID := forkid.NewID(acn.chain.Config(), acn.chain.Genesis().Hash(), acn.chain.CurrentHeader().Number.Uint64())
+	if err := peer.Handshake(acn.networkID, td, hash, genesis.Hash(), forkID, acn.forkFilter); err != nil {
+		peer.Log().Debug("Consensus handshake failed", "err", err)
+		return err
+	}
+
+	if err := acn.peers.register(peer); err != nil {
+		peer.Log().Error("Snapshot extension registration failed", "err", err)
```
Jason-Zhangxin-Chen, line 110: The error msg `Snapshot extension` is confusing to me, what does it mean?

autonity/autonity.go
vpiyush: Yes, The legacy enode pattern is still used by the execution channel.

cmd/autonity/config.go
vpiyush: Yes, We have discussed this, We should still start the ACN because we can not anticipate if a node would in the future want to register as a validator or not. From the performance point of view this should not be a problem because if not a validator the endpoint will not be used.

consensus/acn/acn.go
vpiyush: Will fix this. it's wrong

consensus/acn/protocol/handler.go
lorenzo-dev1: I don't really understand what the purpose of setting it, since it is used only in discovery and we do not have discovery for ACN. But I guess setting an additional record in the localnode does not hurt either

consensus/acn/protocol/peer.go
```diff
@@ -0,0 +1,111 @@
...
+// Peer is a collection of relevant information we have about a `acn` peer.
+type Peer struct {
+	id      string // Unique ID for the peer, cached
+	address common.Address
+
+	*p2p.Peer                   // The embedded P2P package peer
+	rw        p2p.MsgReadWriter // Input/output streams for snap
+	version   uint              // Protocol version negotiated
+
+	head common.Hash // Latest advertised head block hash
```
Jason-Zhangxin-Chen, line 22: Regarding to the head and TD, it seems to be initialized only in ACN handshake phase, while they are not updated during runtime. In my opinion, they are execution layer's info, not sure if we need them in consensus channel.

consensus/acn/protocol/handler.go
lorenzo-dev1: I am not sure still why we need to set this entry in the localnode. I am not sure 100% but it seems to me like it is used only for discovery (which if I remember correctly it is not enabled for ACN?).

Also the ETH protocol has a routine that continuously updates this value to the latest chain head. Do we not need it because this value is actually not used?

consensus/tendermint/backend/backend.go
```diff
@@ -100,6 +100,9 @@ type Backend struct {
 
 	// interface to enqueue blocks to fetcher and find peers
```
Jason-Zhangxin-Chen, line 101: This comment shall be updated with just find peers.

consensus/tendermint/backend/backend.go
```diff
@@ -100,6 +100,9 @@ type Backend struct {
...
 	Broadcaster consensus.Broadcaster
+
+	// interface to enqueue blocks to fetcher and find peers
```
Jason-Zhangxin-Chen, line 104: Please remove find peers in the comment for the consensus.Enqueuer.

consensus/tendermint/backend/backend.go
```diff
@@ -143,6 +146,11 @@ func (sb *Backend) Gossip(committee types.Committee, msg message.Msg) {
 	sb.gossiper.Gossip(committee, msg)
 }
 
+// Gossip implements tendermint.Backend.Gossip
```
Jason-Zhangxin-Chen, line 149: This comment does not reflect the function's feature.

eth/backend.go
```diff
@@ -582,20 +581,20 @@ func (s *Ethereum) newCommitteeWatcher() {
...
 		s.p2pServer.UpdateConsensusEnodes(enodesList.List)
 	}
-
 	wasValidating := false
 	currentBlock := s.blockchain.CurrentBlock()
 	if currentBlock.Header().CommitteeMember(s.address) != nil {
-		updateConsensusEnodes(currentBlock)
 		s.miner.Start()
 		s.log.Info("Starting node as validator")
+		updateConsensusEnodes(currentBlock)
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Leave some comments on why we change the execution orders of the `updateConsensusEnodes()` function?

p2p/peer.go
```diff
@@ -58,6 +59,11 @@ const (
 	pongMsg      = 0x03
 )
 
+type ConsensusInfo struct {
```
Jason-Zhangxin-Chen, line 62: Please remember to remove this structure, it is not used now.

consensus/acn/protocol/protocol.go
```diff
@@ -0,0 +1,54 @@
...
+
+// StatusPacket is the network packet for the status message for eth/64 and later.
+type StatusPacket struct {
+	ProtocolVersion uint32
+	NetworkID       uint64
+	TD              *big.Int
+	Head            common.Hash
+	Genesis         common.Hash
+	ForkID          forkid.ID
+}
```
Jason-Zhangxin-Chen, line 54: Not a big deal, please add some unit tests for acn package, but I would trust you if you feel confident on it.

p2p/server.go
```diff
@@ -398,29 +420,42 @@ func (src *Server) UpdateConsensusEnodes(enodes []*enode.Node) {
 			}
 		}
 		if !found {
-			log.Debug("Dropping node from static peers", "enode", connectedPeer.String())
-			src.RemoveTrustedPeer(connectedPeer)
-			src.dialsched.removeStatic(connectedPeer)
+			log.Debug("Dropping node from static peers", "enode", connectedPeer.String(), "server", srv.Net.String())
+			srv.RemoveTrustedPeer(connectedPeer)
+			srv.RemovePeer(connectedPeer)
```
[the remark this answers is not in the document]
vpiyush: It does wait on channels, but I can't see a blocking scenario happening. I think it's safe.

atc/atc.go
vpiyush: removed this.

vpiyush:
> Do we have anywhere tests that check committee members entering/exiting the committee (and consequent timely setup/teardown of their related consensus p2p connection)? I think it would be a good addition if we have time for it.

We do have tests which checks the mining status of member going out of committee.  But there is nothing that checks members going in and out of committee multiple times. But as far as i know that's what Jason has tested with docker e2e tests. But I agree that such test would be helpful i will see if we can fit that in else take it up later as a  todo in the tests.

eth/backend.go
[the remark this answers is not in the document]
vpiyush: Actually, It was an attempt to fix the shutdown issue which i was facing. But in the end this was not the reason it was happening. But i believed it was better to stop the miner as soon as possible as we are already out of committee. also, i have updated the order so when we are in committee we connect first and start miner and when we go out of the committee we stop miner first and then disconnect peers.

consensus/acn/protocol/peer.go
vpiyush: I have removed these i kept it with the intention of using them for info, But now i have decided we would add them back if needed

lorenzo-dev1 (approved):
LGTM

p2p/server.go
```diff
@@ -386,41 +406,71 @@ func (srv *Server) RemoveTrustedPeer(node *enode.Node) {
 // UpdateConsensusEnodes is responsible to make sure that
 // a node belonging to the consensus committee is fully connected
 // to the other consensus committee nodes.
-func (src *Server) UpdateConsensusEnodes(enodes []*enode.Node) {
-
+func (srv *Server) UpdateConsensusEnodes(committeeSubset []*enode.Node, committee []*enode.Node) {
+	srv.enodeMu.Lock()
+	currentlyRequested := make([]*enode.Node, len(srv.committeeSubset))
+	copy(currentlyRequested, srv.committeeSubset)
```
lorenzo-dev1, line 412: Is this correct? we copy in the currentlyRequested the old ones and not the new ones passed by params

eth/backend.go
```diff
@@ -62,6 +62,12 @@ import (
 	"github.com/autonity/autonity/rpc"
 )
 
+const (
+	// txChanSize is the size of channel listening to NewTxsEvent.
```
lorenzo-dev1, line 66: copy-paste error

eth/topology.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/autonity/autonity/p2p/enode"
+)
+
+type NetworkTopology struct {
+	diameter uint
+	minNodes int
+}
+
+// Only diameter = 2, to support diameter > 2, ComputeBase function has to be modified
+func NewGraphTopology(diameter uint, minNodes int) *NetworkTopology {
```
lorenzo-dev1, line 15: Not a big deal but imo we should panic if diameter != 2 then

lorenzo-dev1: Also I think we are actually not using this func and instantiating directly the struct with `&NetworkTopology{...}`

p2p/server.go
vpiyush: This is correct the naming is confusing, I will rename this

p2p/peer_error.go
```diff
@@ -71,6 +71,8 @@ const (
 	DiscReadTimeout
 	DiscJailed
 	DiscPeerNotInCommittee
+	DiscMaxDiscoveredPeers
```
lorenzo-dev1, line 74: I think this is unused

p2p/peer_error.go
```diff
@@ -87,6 +90,9 @@ var discReasonToString = [...]string{
 	DiscSelf:                "connected to self",
 	DiscReadTimeout:         "read timeout",
 	DiscJailed:              "jailed node",
+	DiscPeerNotInCommittee:  "validator is not part of committee",
+	DiscMaxDiscoveredPeers:  "reached maximum discoverable peers",
```
lorenzo-dev1, line 94: I think this is unused

yazzaoui (approved):
LGTM

PR #924, opened 2024-01-08
Fix consensus issues

lorenzo-dev1:


miner/miner.go
```diff
@@ -112,34 +134,42 @@ func (miner *Miner) update() {
...
-				canStart = true
-				if shouldStart {
-					miner.worker.start()
+				miner.syncFailures++
+				miner.eth.Logger().Info("Chain syncing failed", "#failures", miner.syncFailures, "shouldStart", miner.shouldStart, "canStart", miner.canStart)
+				// TODO(lorenzo) discuss
+				/* if we fail more than maxSyncFailures times consequently, assume we are under attack and start consensus anyways:
+				*  - if we are actually synced, consensus will run fine
+				*  - if we are actually behind, we might send messages for old heights and equivocate (until we implement a WAL)
+				 */
```
lorenzo-dev1, line 145: Please if you have any feedback on this approach let me know, I am not 100% sure this is the best way to go about it.

I am thinking about scenarios like:
- epoch period 1800
- we are at height=4000
- our node $x$ was offline and he is at height = 100. He was in the committee at that time. He is still in committee at height 4000 (but he doesn't know it yet).
- node $x$ comes back online, but he is attacked by malicious nodes which advertise height >> 4000 as their chain head
- node $x$ will keep trying to sync with the malicious nodes (since they claim to have the higher total difficulty chain). Once a sync fails, the malicious peer gets dropped. However the attacker keep spinning up fresh nodes (they don't have to be validators) and connecting them to $x$, thus keeping the attack going.
- node $x$ will stay stuck at height=100, without possibility of syncing.

After several failures, we have two possibilities:
- we do not start consensus, the attacker succeeds. He can stop our consensus until he manages to keep connecting nodes to $x$
- we start consensus anyways, hoping that we are synced.
  - if we are synced, all good  
  -  If we are not like in the example, we will start asking sync to other nodes, which will reply with consensus messages for height = 4000. However we cannot even verify those (we do not know the committee).

Block and block hashes announcements will not help, since they will be discarded because too far (blockwise). see:
- https://github.com/autonity/autonity/blob/develop/eth/fetcher/block_fetcher.go#L399 for hashes
- https://github.com/autonity/autonity/blob/develop/eth/fetcher/block_fetcher.go#L757 for blocks

Full block propagations

lorenzo-dev1: I am wondering if the best way if we keep failing to sync would be to crash the client and log a message to the operator explaining what is likely happening. Then the operator can connect only to a set of trusted peers to sync and restart consensus.

tbssajal: > - epoch period 1800
> - we are at height=4000
> - our node $x$ was offline and he is at height = 100. He was in the committee at that time. He is still in committee at height 4000 (but he doesn't know it yet).

It doesn't sound right. So the epochs are 0 -> for blocks [0, 1800), 1 -> for blocks [1800, 3600), 2 -> for blocks [3600, 5400). If $x$ is at height 100 and we are at height 4000, that means $x$ was inactive the whole epoch 1 -> for blocks [1800, 3600). So $x$ should not be selected as a committee member for epoch 2 by the protocol as he was inactive in in the previous epoch and if some sort of mechanism punishes inactivity then $x$ would probably be in jail for epoch 2.

lorenzo-dev1: At the moment inactive validators are not punished, so this is possible. Until we have the omission failure in place, I don't think we can assume this scenario cannot happen.

tbssajal: At the moment, yes we cannot. But I was thinking there should be a simple mechanism that does not punishes or rewards anyone but takes some sort of attendance before selecting committee members. For example right now validators need to be in `Active` state to be in committee. There can be a simple mechanism via contract that will change validator state as a form of attendance and only validators who submit attendance in the previous epoch can be considered as a candidate for the committee of the next epoch. There should be a separate issue for this I suppose.

cmd/utils/flags.go
```diff
@@ -1597,6 +1597,7 @@ func RegisterEthService(stack *node.Node, cfg *ethconfig.Config) (ethapi.Backend
 		return backend.ApiBackend, nil
 	}
 	backend, err := eth.New(stack, cfg)
+
```
vpiyush, line 1600: redundant

yazzaoui:
Looking fine so far
> removes the backlogging of FutureStepMessages and processes them as soon as they arrive. This requires some small modification in the core logic (namely, checking line 34 and line 44 in the setStep function, and checking line 36 when receiving a valid current proposal), but it solves the issue of line 47 not being checked if we are in Propose step.

This is the bit that is worrying me - I have to spend more time on the review

consensus/tendermint/core/backlog.go
```diff
@@ -17,12 +18,7 @@ type backlogUntrustedMessageEvent struct {
...
-// checkMessageStep checks the message step
-// return errInvalidMessage if the message is invalid
-// return errFutureHeightMessage if the message view is larger than curRoundMessages view
-// return errOldHeightMessage if the message view is smaller than curRoundMessages view
-// return errFutureStepMessage if we are at the same view but at the propose step and it's a voting message.
-func (c *Core) checkMessageStep(round int64, height uint64, step Step) error {
+func (c *Core) checkMessage(round int64, height uint64) error {
 	h := new(big.Int).SetUint64(height)
 	switch {
 	case round < 0 || round > constants.MaxRound:
```
vpiyush, line 24: shouldn't we also reject such message during decoding itself? Is there a usecase for this?

consensus/tendermint/core/core.go
```diff
@@ -409,27 +409,53 @@ func (c *Core) SetStep(step Step) {
...
+		c.logger.Debug("Cannot stop prevote timer", "c.step", c.step, "step", step, "err", err)
+	}
+	if err := c.precommitTimeout.StopTimer(); err != nil {
+		c.logger.Debug("Cannot stop precommit timer", "c.step", c.step, "step", step, "err", err)
+	}
+
+	// we might have already received a quroum of prevotes while still in propose step.
+	// therefore once changing step to prevote, we need to check line 34 and 44
+
+	if c.step == Prevote {
```
yazzaoui, line 444: I don't like to see this core logic in setStep, it's not the right place for that. 
I'm still unsure if it's a bug or wanted but in the propose flow you are not returning after calling setStep - meaning the code path starting L155 will be executed after changing to Prevote step.

consensus/tendermint/core/core.go
```diff
@@ -409,27 +409,53 @@ func (c *Core) SetStep(step Step) {
...
 			PrevoteStepTimer.Update(now.Sub(c.stepChange))
 			PrevoteStepBg.Add(now.Sub(c.stepChange).Nanoseconds())
-		case c.step == PrecommitDone && step == PrecommitDone:
-			//this transition can also happen when we already received 2f+1 precommits but we did not start the new round yet.
-			PrecommitDoneStepTimer.Update(now.Sub(c.stepChange))
-			PrecommitDoneStepBg.Add(now.Sub(c.stepChange).Nanoseconds())
 		default:
-			// TODO(lorenzo) this ideally should be a .Crit but these transitions do actually happen.
-			// see: https://github.com/autonity/autonity/issues/803
+			// Ideally should be a .Crit, however it does not seem right to me because in the same sceneario the node would:
```
yazzaoui, line 420: To discuss at standup

consensus/tendermint/core/core.go
lorenzo-dev1: Do you believe that for better code quality we should move this outside of the setStep function, or you think this code should not executed at setStep? I do agree that probably it is better to move it in a separate function, however I think it is logically correct to execute these checks at step change time.

For L155 of propose.go, the lack of return in the previous code block is wanted. It covers this scenario:
- we already have a quorum of prevotes for a proposal we did not receive yet
- we receive the proposal and prevote it, move to prevote step
- since we already have a quorum of prevotes, we can execute L36 of tendermint and precommit
-

cmd/utils/flags.go
lorenzo-dev1: solved in https://github.com/autonity/autonity/pull/924/commits/7fd972d7e3ea9cca9177eae6c121eda8ae49fec2

consensus/tendermint/core/core.go
```diff
@@ -409,33 +406,56 @@ func (c *Core) SetStep(step Step) {
...
 	c.stepChange = now
-	c.processBacklog()
+
+	// stop consensus timeouts
+	c.stopAllTimeouts()
+
+	// if we are moving from propose to prevote step we need to check again line 34,36 and 44
+	// NOTE: this call to stepChangeChecks can cause recursion in the SetStep function.
+	// This can happen if the checks cause a transition to Precommit step. It is expected behaviour.
+	// If we want to remove this recursion possibility, we could post an Event that signals a step change,
```
Jason-Zhangxin-Chen, line 433: The recursion do happen when the stepChangeChecks trigger new steps, I think it would be better to post event to remove the recursion.

lorenzo-dev1: Why do you think it would be better?

I am not sure which one is better because:
- recursion increases code complexity but it ensures that the stepChangeChecks() are done as soon as we change step
- if we post an event, there will be no recursion (code will be simpler) but now the stepChangeChecks() will be triggered asynchronously. So following situation becomes possible:
  - setStep(prevote)
  - process some other message m
  - do stepChangeChecks()

So I think the triggering of some upon conditions might get delayed if we post an event?

yazzaoui: Yeah I think it's fine to keep the recursive call - it's easy to see that it's bounded. I don't like the complication involved with posting events.

yazzaoui (approved):
LGTM. You've done a good job with the rule isolation - much cleaner now. Let's get it merged.

consensus/tendermint/core/propose.go
```diff
@@ -18,60 +18,69 @@ type Proposer struct {
...
-	// Check if the message comes from curRoundMessages proposer
-	if !c.IsFromProposer(c.Round(), proposal.Sender()) {
-		c.logger.Warn("Ignore proposal messages from non-proposer")
+	// proposal is either for current round or old round
+	roundMessages := c.messages.GetOrCreate(proposal.R())
+
+	// if we already have a proposal for this round - ignore
+	// the first proposal sent by the sender in a round is always the only one we consider.
+	if roundMessages.Proposal() != nil {
+		return constants.ErrAlreadyProcessed
```
Jason-Zhangxin-Chen, line 60: It may be an equivocated proposal that haven't been processed before. The ErrAlreadyProcessed seems to be inaccurate.

Jason-Zhangxin-Chen (approved):
LGTM, the upon conditions were unified in the rule set in rules.go, which is making the code lean and clean. Left few minor comments on it. Thank you for that.

consensus/tendermint/core/propose.go
lorenzo-dev1: `ErrAlreadyProcessed` is returned whenever we already processed *a proposal* for the round, but it does not mean that we already processed the specific proposal that we are handling.

But I agree that the naming is a bit ambiguous, I changed it to a more accurate name in https://github.com/autonity/autonity/pull/924/commits/a8390bfa34c0214401b6da53415c058e883109b4

PR #925, opened 2024-01-09
Separate Liquid Newton Logic into a separate contract

m4ksio:


autonity/solidity/contracts/Autonity.sol
```diff
@@ -204,6 +213,8 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     ) internal {
         config = _config;
 
+        liquidLogicAddress = new LiquidLogic(address(this));
```
tbssajal, line 216: `address(this)` is basically the address of Autonity contract, right? Then this is incorrect, no?

autonity/solidity/contracts/LiquidState.sol
```diff
@@ -0,0 +1,496 @@
...
+    * @dev Restricted to the autonity contract.
+    */
+    function redistribute() external payable onlyAutonity returns (uint256)
+    {
+        (bool success, bytes memory data) = liquidLogic.delegatecall(
+            abi.encodeWithSignature("redistribute()")
+        );
+
+        if (success == false) {
+            // if there is a return reason string
```
tbssajal, line 114: This part looks repetitive. Maybe add a function for it?

autonity/solidity/contracts/LiquidLogic.sol
```diff
@@ -343,6 +241,32 @@ contract Liquid is IERC20
         supply -= _value;
     }
 
+    /**
+* @dev See {IERC20-allowance}.
+    */
+    function allowance(address _owner, address _spender)
```
tbssajal, line 247: Why do we need these getters here if we implemented them in LiquidState.sol?

autonity/solidity/contracts/LiquidLogic.sol
```diff
@@ -343,6 +241,32 @@ contract Liquid is IERC20
...
+    public view override returns (uint256)
+    {
+        return allowances[_owner][_spender];
+    }
+
+    /**
+* @notice Returns the amount of liquid newtons held by the account (ERC-20).
+    */
+    function balanceOf(address _delegator)
```
tbssajal, line 256: Why do we need these getters here if we implemented them in LiquidState.sol?

autonity/solidity/contracts/LiquidLogic.sol
```diff
@@ -343,6 +241,32 @@ contract Liquid is IERC20
...
+    external view override returns (uint256)
+    {
+        return balances[_delegator];
+    }
+
+    /**
+* @notice Returns the total amount of stake token issued.
+    */
+    function totalSupply() public view override returns (uint256)
```
tbssajal, line 265: Why do we need these getters here if we implemented them in LiquidState.sol?

autonity/solidity/contracts/LiquidLogic.sol
```diff
@@ -161,88 +117,45 @@ contract Liquid is IERC20
...
     /**
-    * @notice Moves `_amount` LNEW tokens from the caller's account to the recipient `_to`.
+* @notice Moves `_amount` LNEW tokens from the caller's account to the recipient `_to`.
     *
     * @return _success a boolean value indicating whether the operation succeeded.
     *
     * @dev Emits a {Transfer} event. Implementation of {IERC20 transfer}
     */
     function transfer(address _to, uint256 _amount)
-        public override returns (bool _success)
```
tbssajal, line 214: Why remove `override`? Isn't it overriding IERC20 function?

tbssajal: Got it

autonity/solidity/contracts/LiquidLogic.sol
m4ksio: We want to emit IERC20 events, and this requires us to implement the interface

autonity/autonity.go
```diff
@@ -372,9 +372,11 @@ func (c *EVMContract) DeployContractWithValue(header *types.Header, origin commo
 
 	evm := c.evmProvider(header, origin, statedb)
 
-	_, _, _, vmerr := evm.Create(vm.AccountRef(DeployerAddress), data, gas, value)
+	ret, contractAddr, _, vmerr := evm.Create(vm.AccountRef(DeployerAddress), data, gas, value)
+	log.Warn("contract deployed to" + contractAddr.String())
```
lorenzo-dev1, line 376: Probably info level is enough? Warning tend to scare users

lorenzo-dev1:
Although the scope of this PR is just the LNTN state separation and not upgradability, I think it is worth to build this with upgradability in mind already. It will save us hassle later on.

The current implementation works, but it has some problems when taken into consideration together with upgradability. Mainly:
1. The logic contract cannot be easily extended. Since we need a 1:1 mapping between methods of the state contract and of the logic contract, we cannot easily add methods to the liquid logic contract. This can be solved by doing the redirection via the `fallback` function, which is the industry standard. Something [like this](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.2.0/contracts/proxy/Proxy.sol).
2. We cannot add storage variables to the logic contract, since they would cause a storage collision and overwrite the `liquidLogic` address in the state contract. This is where [EIP-1967 Proxy Storage Slots](https://eips.ethereum.org/EIPS/eip-1967) comes to help. See also [this open zeppelin article](https://docs.openzeppelin.com/upgrades-plugins/1.x/proxies) for a very clear explaination of this issue.

Generally speaking I am strongly in favour of using a standardized battle tested solution like [UUPS proxies](https://proxies.yacademy.dev/pages/proxies-list/#universal-upgradeable-proxy-standard-uups) (if we want upgradability baked in) or [immutable clones](https://docs.openzeppelin.com/contracts/4.x/api/proxy#Clones) if we only want state separation.

It might seem easy to build upgradable proxied contracts but **it is NOT**. There are a lot of subtleties and sneaky errors that are not intuitive at all. That is why the community iterated through multiple designs and did several mistake in the process.

Useful references which give an idea of the existing solutions and their issues:
- https://blog.openzeppelin.com/the-state-of-smart-contract-upgrades
- https://proxies.yacademy.dev/pages/proxies-list

autonity/solidity/contracts/LiquidState.sol
yazzaoui, line 1: Please have a look at https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/proxy/Proxy.sol
It's the correct way to implement the proxy contract.
Can't be directly imported though, it will need a few adjustments.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -37,7 +38,7 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         uint256 selfUnbondingStake;
         uint256 selfUnbondingShares; // not effective - used for accounting purposes
         uint256 selfUnbondingStakeLocked;
-        Liquid liquidContract;
+        LiquidState liquidContract;
```
Jason-Zhangxin-Chen, line 41: name the variable with `liquidSate` would make more sense.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -95,6 +99,7 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         IACU acuContract;
         ISupplyControl supplyControlContract;
         IStabilization stabilizationContract;
+//        address liquidLogicAddress;
```
Jason-Zhangxin-Chen, line 102: Please remember to remove this.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -149,6 +153,11 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     */
     address public deployer;
 
+    /*
+        Address of current logic contract for Liquid Newton
+    */
+    LiquidLogic public liquidLogicAddress;
```
Jason-Zhangxin-Chen, line 159: Put liquid logic contract here is good, I think in the future, the system operator can upgrade the contract and relace the its address at here.

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > `address(this)` is basically the address of Autonity contract, right? Then this is incorrect, no?

In such way the liquid logic contract can limit its caller to AC.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -981,7 +999,9 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     function _deployLiquidContract(Validator memory _validator) internal {
         if (address(_validator.liquidContract) == address(0)) {
             string memory stringLength = Helpers.toString(validatorList.length);
-            _validator.liquidContract = new Liquid(_validator.nodeAddress,
+            _validator.liquidContract = new LiquidState(
```
Jason-Zhangxin-Chen, line 1002: It would be better to name the function: `_deployLiquidStateContract`

autonity/solidity/contracts/LiquidLogic.sol
```diff
@@ -51,42 +12,36 @@ contract Liquid is IERC20
     uint256 public constant FEE_FACTOR_UNIT_RECIP = 1_000_000_000;
     uint256 public constant COMMISSION_RATE_PRECISION = 10_000;
 
+    // storage layout block - must be compatible with LiquidState
+    address autonityContract; //not hardcoded for testing purposes
+
```
Jason-Zhangxin-Chen, line 17: To my understanding, those `balances` and `lockedBalances` etc... should belong to the liquid state contracts. Am I miss understanding this feature? To me, all the states of liquid contract should be moved into liquid state contract, and we just leave those compute logics in liquid logic contract. We may need a ADR doc for this first.

cmjc: They are defined in LiquidState, line 53 on?

Jason-Zhangxin-Chen:
As we have a PR which do the same thing: https://github.com/autonity/autonity/pull/1015. Let's close this one.

PR #926, opened 2024-01-12
Reject bonding to jailed validator

### LINKED ISSUE #895: Bonding to Jailed Validator
Not sure if it is a bug or intentional, but when applying bonding at epoch end, we apply all the pending bonding request. When receiving a bonding request we check if the validator was active or not and accept the request only if the validator was active. But the validator can be jailed after the bonding request is accepted and before the bonding is applied. In that case, I think we should check if the validator is still active before applying bonding. From the point of view of the delegator, he would not want to bond to a jailed validator and he may not know that the validator was going to be jailed when requesting for bonding, if he only listens to `SlashingEvent`. I think we can reject bonding request at `_applyBonding` function if the validator is jailed.

tbssajal:
Closes #895


yazzaoui (approved):
LGTM

Jason-Zhangxin-Chen (approved):
LGTM

PR #927, opened 2024-01-12
start consensus only after syncing with the network

lorenzo-dev1:
This PR aims at providing a quick mitigation for one slashing issue that we saw on Piccadilly. More specifically the issue happens when we start the consensus engine before syncing to the current chain head. This can cause equivocation to be done by the node, because he is going to vote for an old height in which he might have already voted.

Please note that this is not a final fix, it just aims at preventing the most trivial edge case. The question of figuring out whether we are synced or not is not trivial at all, and requires more work to be implemented properly.

Moreover this solution will not be effective if the validator is under attack by malicious node who wants to impede its syncing. This specific problem requires modification on the sync mechanism, exploiting the existance of a segregated consensus network (https://github.com/autonity/autonity/pull/923)

The consensus engine now will start only if:
- we are already in sync with the network
- we completed a chain sync with a peer
- we forced mining start via:
  - devmode
  - call to miner api
  - using the `--mine` flag

### DIFF
--- a/eth/sync.go
+++ b/eth/sync.go
@@ -62,11 +62,12 @@ func (h *handler) syncTransactions(p *eth.Peer) {
 
 // chainSyncer coordinates blockchain sync components.
 type chainSyncer struct {
-	handler     *handler
-	force       *time.Timer
-	forced      bool // true when force timer fired
-	peerEventCh chan struct{}
-	doneCh      chan error // non-nil when sync is running
+	handler         *handler
+	force           *time.Timer
+	forced          bool // true when force timer fired
+	peerEventCh     chan struct{}
+	doneCh          chan error // non-nil when sync is running
+	syncedEventSent bool       // true when the downloader.SyncedEvent has been sent (we are synced with our current peerset)
 }
 
 // chainSyncOp is a scheduled sync operation.
@@ -80,8 +81,9 @@ type chainSyncOp struct {
 // newChainSyncer creates a chainSyncer.
 func newChainSyncer(handler *handler) *chainSyncer {
 	return &chainSyncer{
-		handler:     handler,
-		peerEventCh: make(chan struct{}),
+		handler:         handler,
+		peerEventCh:     make(chan struct{}),
+		syncedEventSent: false,
 	}
 }
 
@@ -145,6 +147,7 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
 	if cs.doneCh != nil {
 		return nil // Sync already running.
 	}
+
 	// Ensure we're at minimum peer count.
 	minPeers := defaultMinSyncPeers
 	if cs.forced {
@@ -158,15 +161,27 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
 	// We have enough peers, check TD
 	peer := cs.handler.peers.peerWithHighestTD()
 	if peer == nil {
+		// we should never end up here
 		return nil
 	}
 	mode, ourTD := cs.modeAndLocalHead()
 
 	op := peerToSyncOp(mode, peer)
 	if op.td.Cmp(ourTD) <= 0 {
-		return nil // We're in sync.
+		// we are synced with our current peerset
+		// notify miner about it if not yet done (so consensus can be started if needed)
+		cs.postSyncedEvent()
+		return nil
 	}
 	return op
+}
+
+func (cs *chainSyncer) postSyncedEvent() {
+	if cs.syncedEventSent {
+		return
+	}
+	cs.handler.eventMux.Post(downloader.SyncedEvent{})
+	cs.syncedEventSent = true
 }
 
 func peerToSyncOp(mode downloader.SyncMode, p *eth.Peer) *chainSyncOp {
--- a/eth/sync.go
+++ b/eth/sync.go
@@ -161,7 +161,6 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
 	// We have enough peers, check TD
 	peer := cs.handler.peers.peerWithHighestTD()
 	if peer == nil {
-		// we should never end up here
 		return nil
 	}
 	mode, ourTD := cs.modeAndLocalHead()
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -74,10 +74,8 @@ type Miner struct {
 	wg sync.WaitGroup
 
 	// used in the miner update loop
-	syncFailures uint64 // counts how many times chain sync failed with remote peers
-	canStart     bool
-	shouldStart  bool
-	events       *event.TypeMuxSubscription // subscription to downloader events()
+	canStart    bool
+	shouldStart bool
 }
 
 func New(eth Backend, config *Config, chainConfig *params.ChainConfig, mux *event.TypeMux, engine consensus.Engine, isLocalBlock func(header *types.Header) bool) *Miner {
@@ -90,7 +88,6 @@ func New(eth Backend, config *Config, chainConfig *params.ChainConfig, mux *even
 		stopCh:       make(chan struct{}),
 		forceStartCh: make(chan struct{}),
 		worker:       newWorker(config, chainConfig, engine, eth, mux, isLocalBlock, true),
-		syncFailures: 0,
 		shouldStart:  false,
 		canStart:     false,
 	}
@@ -106,9 +103,11 @@ func New(eth Backend, config *Config, chainConfig *params.ChainConfig, mux *even
 func (miner *Miner) update() {
 	defer miner.wg.Done()
 
-	miner.events = miner.mux.Subscribe(downloader.StartEvent{}, downloader.DoneEvent{}, downloader.SyncedEvent{}, downloader.FailedEvent{})
+	events := miner.mux.Subscribe(downloader.StartEvent{}, downloader.DoneEvent{}, downloader.SyncedEvent{}, downloader.FailedEvent{})
 	defer func() {
-		miner.unsubscribe()
+		if !events.Closed() {
+			events.Unsubscribe()
+		}
 	}()
 
 	// miner.shouldStart is set at initialization to false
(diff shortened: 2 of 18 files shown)


miner/miner.go
```diff
@@ -112,34 +134,42 @@ func (miner *Miner) update() {
...
-				if shouldStart {
-					miner.worker.start()
-				}
-			case downloader.DoneEvent:
-				canStart = true
-				if shouldStart {
-					miner.worker.start()
+				miner.syncFailures++
+				miner.eth.Logger().Info("Chain syncing failed", "#failures", miner.syncFailures, "shouldStart", miner.shouldStart, "canStart", miner.canStart)
+				// TODO(lorenzo) discuss
```
lorenzo-dev1, line 141: Please if you have any feedback on this approach let me know, I am not 100% sure this is the best way to go about it.

I am thinking about scenarios like:
- epoch period 1800
- we are at height=4000
- our node $x$ was offline and he is at height = 100. He was in the committee at that time. He is still in committee at height 4000 (but he doesn't know it yet).
- node $x$ comes back online, but he is attacked by malicious nodes which advertise height >> 4000 as their chain head
- node $x$ will keep trying to sync with the malicious nodes (since they claim to have the higher total difficulty chain). Once a sync fails, the malicious peer gets dropped. However the attacker keep spinning up fresh nodes (they don't have to be validators) and connecting them to $x$, thus keeping the attack going.
- node $x$ will stay stuck at height=100, without possibility of syncing.

After several failures, we have two possibilities:
- we do not start consensus, the attacker succeeds. He can stop our consensus until he manages to keep connecting nodes to $x$
- we start consensus anyways, hoping that we are synced.
  - if we are synced, all good  
  -  If we are not like in the example, we will start asking sync to other nodes, which will reply with consensus messages for height = 4000. However we cannot even verify those (we do not know the committee).

Block and block hashes announcements will not help, since they will be discarded because too far (blockwise). see:
- https://github.com/autonity/autonity/blob/develop/eth/fetcher/block_fetcher.go#L399 for hashes
- https://github.com/autonity/autonity/blob/develop/eth/fetcher/block_fetcher.go#L757 for blocks

So basically they will stop us from syncing. We can start consensus and hope we are synced (risking equivocating until we have a WAL) or just do not start consensus.

lorenzo-dev1: I am wondering if the best solution if we keep failing to sync would be to crash the client and log a message to the operator explaining what is likely happening. Then the operator can connect only to a set of trusted peers to sync and restart consensus.

Another option would be to feed additional information from:
- consensus (askSync)
- announcement

To understand if we are actually synced or not.

lorenzo-dev1: Originally posted by @tbssajal here https://github.com/autonity/autonity/pull/924#discussion_r1450409852
 
> - epoch period 1800
> - we are at height=4000
> - our node $x$ was offline and he is at height = 100. He was in the committee at that time. He is still in committee at height 4000 (but he doesn't know it yet).

It doesn't sound right. So the epochs are 0 -> for blocks [0, 1800), 1 -> for blocks [1800, 3600), 2 -> for blocks [3600, 5400). If $x$ is at height 100 and we are at height 4000, that means $x$ was inactive the whole epoch 1 -> for blocks [1800, 3600). So $x$ should not be selected as a committee member for epoch 2 by the protocol as he was inactive in in the previous epoch and if some sort of mechanism punishes inactivity then $x$ would probably be in jail for epoch 2.

lorenzo-dev1: At the moment inactive validators are not punished, so this is possible. Until we have the omission failure in place, I don't think we can rule out this scenario.

lorenzo-dev1: Originally posted by @tbssajal here https://github.com/autonity/autonity/pull/924#discussion_r1450575225

At the moment, yes we cannot. But I was thinking there should be a simple mechanism that does not punishes or rewards anyone but takes some sort of attendance before selecting committee members. For example right now validators need to be in `Active` state to be in committee. There can be a simple mechanism via contract that will change validator state as a form of attendance and only validators who submit attendance in the previous epoch can be considered as a candidate for the committee of the next epoch. There should be a separate issue for this I suppose.

tbssajal: > There can be a simple mechanism via contract that will change validator state as a form of attendance and only validators who submit attendance in the previous epoch can be considered as a candidate for the committee of the next epoch.

I created a discussion in response: https://github.com/clearmatics/autonity-protocol/discussions/269

lorenzo-dev1: I replied to your proposal in the protocol repo. I will mark it this discussion as solved for the moment, since this PR scope has been redefined and now it only tries to provide a quick fix for cases where there is no attack going on.

miner/miner.go
```diff
@@ -68,40 +71,56 @@ type Miner struct {
 	stopCh  chan struct{}
 
 	wg sync.WaitGroup
+
+	// used in the miner update loop
+	syncFailures uint64 // counts how many times chain sync failed with remote peers
+	canStart     bool
```
lorenzo-dev1, line 77: Still need to figure out whether `canStart` and `events` should be protected for concurrent access. They are both accessed by the `update()` loop and by the `ForceStartMining` method (which is called by the API or when starting the client).

I think that theoretically there is a concurrent access possibility, but practically it should not be a problem. Still thinking about it though

eth/sync.go
```diff
@@ -62,11 +62,12 @@ func (h *handler) syncTransactions(p *eth.Peer) {
...
-	force       *time.Timer
-	forced      bool // true when force timer fired
-	peerEventCh chan struct{}
-	doneCh      chan error // non-nil when sync is running
+	handler         *handler
+	force           *time.Timer
+	forced          bool // true when force timer fired
+	peerEventCh     chan struct{}
+	doneCh          chan error // non-nil when sync is running
+	syncedEventSent bool       // true when the downloader.SyncedEvent has been sent (we are synced with our current peerset)
```
Jason-Zhangxin-Chen, line 70: How is the lifecycle of the chainSyncer object managed? Is it created for each sync procedure? In this case, the flag should work.

miner/miner.go
Jason-Zhangxin-Chen: I think we need to use RWMutex, if the two routine are just readers of them, just apply readLock while the writter should apply writeLock.

miner/miner.go
```diff
@@ -68,40 +71,56 @@ type Miner struct {
...
+	// it will be true when node is a committee member, therefore consensus engine should be started if possible.
+
+	/* miner.canStart is set at initialization to false
+	* miner.canStart = true when consensus engine can be safely started OR if we forced mining start.
+	* It is safe to start the consensus engine when we are reasonably sure to be synced with the head of the chain:
+	* - the first chain sync with our peers terminates, and we conclude that we are synced to the chain head.
+	* - the first chain sync with our peer terminates, and we successfully imported blocks till the head of the chain
+	*
+	* NOTE: This mechanism does not give 100% guarantee that we are synced to the head of the chain before starting consensus. There is always the possibility that we are connected to peers which are behind the "global" chain head.
+	 */
```
Jason-Zhangxin-Chen, line 122: Is it possilble for a connected malicious peer who announces a block been mined in the future height can cause the stop of the mining in local node?

miner/miner.go
```diff
@@ -112,34 +131,39 @@ func (miner *Miner) update() {
...
+				miner.eth.Logger().Info("Chain syncing failed", "#failures", miner.syncFailures, "shouldStart", miner.shouldStart, "canStart", miner.canStart)
+				// if we fail more than maxSyncFailures times consequently, assume we are under attack
+				if miner.syncFailures >= maxSyncFailures {
+					miner.eth.Logger().Warn("************************** SYNC ATTACK DETECTED **************************")
+					miner.eth.Logger().Warn("Multiple sequential chain sync failures detected", "sync failures", miner.syncFailures)
+					miner.eth.Logger().Warn("Your node is probably under attack by malicious peers, which are preventing your node from syncing")
+					miner.eth.Logger().Warn("Try restarting your node and connecting to a trusted set of peers")
+					miner.eth.Logger().Warn("Reach out to Autonity social media channels for support and additional informations")
+					miner.eth.Logger().Warn("**************************************************************************")
+					panic("sync attack detected")
```
Jason-Zhangxin-Chen, line 146: Do we really panic when we are under sync attack?

eth/sync.go
lorenzo-dev1: There is only one chainSyncer that is created at node initialization. The syncedEventSent is not accessed concurrently however, only a single go routine ever reads/writes it.

miner/miner.go
lorenzo-dev1: Yeah that is possible, because we always try to sync with the peer which advertises the highest height. However once we managed to sync successfully once we will stop reacting to downloader events.

miner/miner.go
lorenzo-dev1: Yes, I think it is the safest way to handle this scenario currently. This is because if we are under a sync attack (nodes advertising height >> actual current height, and then failing sync) we will not be able to actually sync the chain. Therefore we cannot start consensus anyways.

The proper solution will require modifications in the syncing mechanism (e.g. relying on 2/3 honesty assumption of committee) but it is too complicated to be included into this release. This fix just aims at avoiding node getting slashed because of stop-start.

At the moment I feel like the only way to mitigate the issue is to crash the node (since we cannot start consensus anyways) and notify the operator, which can then connect to some trusted peers to actually make its node sync and start consensus.

If any better suggestions let me know.

lorenzo-dev1: Please note that we just need to sync successfully once and this attack will not work anymore. The attacker needs to basically eclipse us at startup for this to work.

miner/miner.go
lorenzo-dev1: Solved by https://github.com/autonity/autonity/pull/927/commits/d2b3c521c61561c79e41228a412649a9d11f5f8a via code reorganization. Now concurrent access to canStart and events cannot happen anymore.

yazzaoui (approved):
LGTM !

vpiyush (approved):
LGTM

PR #928, opened 2024-01-17
Consensus small improvements

yazzaoui:
I think it would be great to have those integrated in https://github.com/autonity/autonity/pull/924
- Reduce a bit the cyclomatic complexity in the handlePropose/Prevote/Precommit flow
- Add info logs when sending votes

### DIFF
--- a/consensus/tendermint/core/handler.go
+++ b/consensus/tendermint/core/handler.go
@@ -292,12 +292,18 @@ func (c *Core) handleFutureRoundMsg(ctx context.Context, msg message.Msg, sender
 func (c *Core) handleValidMsg(ctx context.Context, msg message.Msg) error {
 	logger := c.logger.New("from", msg.Sender())
 
+	if c.Height().Uint64() > msg.H() {
+		return constants.ErrOldHeightMessage
+	}
+
+	if c.Height().Uint64() < msg.H() {
+		panic("Processing future message")
+	}
+
 	// Store the message if it's a future message
 	testBacklog := func(err error) error {
 		// We want to store only future messages in backlog
 		switch {
-		case errors.Is(err, constants.ErrFutureHeightMessage):
-			panic("Processed future message as a valid message")
 		case errors.Is(err, constants.ErrFutureRoundMessage):
 			logger.Debug("Storing future round message in backlog")
 			c.storeBacklog(msg, msg.Sender())
--- a/consensus/tendermint/core/precommit.go
+++ b/consensus/tendermint/core/precommit.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"context"
-	"errors"
 	"math/big"
 
 	"github.com/autonity/autonity/common"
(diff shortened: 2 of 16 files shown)


consensus/tendermint/core/precommit.go
```diff
@@ -23,57 +22,68 @@ func (c *Precommiter) SendPrecommit(ctx context.Context, isNil bool) {
...
+	if precommit.R() < c.Round() {
+		// We are receiving a precommit for an old round. We need to check if we have now a quorum
+		// in this old round.
+		roundMessages := c.messages.GetOrCreate(precommit.R())
+		roundMessages.AddPrecommit(precommit)
+		oldRoundProposal := roundMessages.Proposal()
+		if oldRoundProposal == nil {
+			return constants.ErrOldRoundMessage
+		}
+		if roundMessages.PrecommitsPower(oldRoundProposal.Block().Hash()).Cmp(c.CommitteeSet().Quorum()) >= 0 {
```
lorenzo-dev1, line 49: It should be < 0 here I think

consensus/tendermint/core/precommit.go
```diff
@@ -23,57 +22,68 @@ func (c *Precommiter) SendPrecommit(ctx context.Context, isNil bool) {
...
 	c.LogPrecommitMessageEvent("MessageEvent(Precommit): Received", precommit, precommit.Sender().String(), c.address.String())
 	if curProposalHash != (common.Hash{}) && c.curRoundMessages.PrecommitsPower(curProposalHash).Cmp(c.CommitteeSet().Quorum()) >= 0 {
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		default:
-			c.Commit(ctx, c.Round(), c.curRoundMessages)
-		}
-
+		// Other places where Commit is called do not have a prior check on context.
```
lorenzo-dev1, line 72: agree

consensus/tendermint/core/prevote.go
```diff
@@ -30,66 +32,67 @@ func (c *Prevoter) SendPrevote(ctx context.Context, isNil bool) {
...
-				}
-			}
+	if prevote.R() > c.Round() {
+		return constants.ErrFutureRoundMessage
+	}
+	if prevote.R() < c.Round() {
+		// We only process old rounds while future rounds messages are pushed on to the backlog
+		oldRoundMessages := c.messages.GetOrCreate(prevote.R())
+		oldRoundMessages.AddPrevote(prevote)
+		if c.step == Prevote || c.step == Precommit {
```
lorenzo-dev1, line 42: Theoretically this should be fine because we do not process new messages once in `PrecommitDone` step, but maybe I would do `if c.step != Propose` just to be more safe.

lorenzo-dev1:
LGTM outside https://github.com/autonity/autonity/pull/928#discussion_r1455193015 

@yazzaoui could you re-run tests it seemed like the CI failed

consensus/tendermint/core/precommit.go
yazzaoui: Yes absolutely

consensus/tendermint/core/handler.go
```diff
@@ -292,12 +292,18 @@ func (c *Core) handleFutureRoundMsg(ctx context.Context, msg message.Msg, sender
 func (c *Core) handleValidMsg(ctx context.Context, msg message.Msg) error {
 	logger := c.logger.New("from", msg.Sender())
 
+	if c.Height().Uint64() > msg.H() {
```
lorenzo-dev1, line 295: Do we need these checks also here because of the backlogged messages calling directly `handleValidMsg`?

yazzaoui: Yes

lorenzo-dev1 (approved):

PR #930, opened 2024-01-22
Add Magic header for Discovery v4 udp packets

yazzaoui:
This adds a simple magic header to discovery packets to ensure proper isolation from the Ethereum discovery network. 
That should speed up time to find an autonity peer significantly.


p2p/discover/v4wire/v4wire.go
```diff
@@ -210,8 +214,11 @@ func Decode(input []byte) (Packet, Pubkey, []byte, error) {
 	if len(input) < headSize+1 {
 		return nil, Pubkey{}, nil, ErrPacketTooSmall
 	}
-	hash, sig, sigdata := input[:macSize], input[macSize:headSize], input[headSize:]
-	shouldhash := crypto.Keccak256(input[macSize:])
+	magic, hash, sig, sigdata := input[:magicSize], input[magicSize:macSize+magicSize], input[macSize+magicSize:headSize], input[headSize:]
+	if !bytes.Equal(magic, AutMagic) {
+		return nil, Pubkey{}, nil, ErrBadHash
```
Jason-Zhangxin-Chen, line 219: ErrBadMagic should be returned.

p2p/discover/v4wire/v4wire.go
```diff
@@ -44,6 +44,8 @@ const (
 	ENRResponsePacket
 )
 
+var AutMagic = []byte{'a', 'u', 't'}
```
Jason-Zhangxin-Chen, line 47: Do we need to add the `aut` magic number to the v5 discovery protocol?

vpiyush (commented):
these looks good in addition to this we should probably have same checks for V5 too. We do support for v5

Jason-Zhangxin-Chen (approved):
LGTM now.

lorenzo-dev1 (approved):
LGTM

FYI I saw that discovery v5 should have something that allows to do the same without modifying the protocol (using topics, which map to the services offered by the node). See https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#topic-advertisement

However I am not sure if using the ETH DHT to advertise autonity nodes would bring benefits, maybe enhanced network robustness?

yazzaoui:
Very interesting video you linked @lorenzo-dev1. I can definitely see a benefit of sharing the DHT if Autonity was more integrated with Ethereum but that's not really the case for now and we can easily restore that possibility without hardforking.
We'll get back to this topic in the future, for now, I think we're good.

PR #931, opened 2024-01-30
Contract Upgrade Mechanism

yazzaoui:
Add a protocol contract, the `UpgradeManager`, deployed at `0x3C368B86AF00565Df7a3897Cfa9195B9434A59f9`, for the upgrade mechanism. It is precompile-based and wrapping the autonity's EVM replace instruction.
The Upgrade Managercontract exposes now a function restricted to the operator with the following signature:
 ```solidity
    /** @dev Call the in-protocol EVM replace mechanism. Requires specific tool to interact.
    * Restricted to the operator account.
    *  @param _target is the target contract address to be updated.
    *  @param _data is the contract creation code.
    */
    function upgrade(address _target, string memory _data) external onlyOperator
```

Additionally, I established the foundations for an in-house contract testing framework using Go with low-level EVM functions. The vision is to be:
- Lean: Only the essential boilerplate required.
- Fast: as efficient as Go can be, enabling fuzzing due to its speed. No more 2 hours wait.
- Integrated with our protocol, eliminating the need for mocks for precompiles.
- Type-safe through generated bindings using a new abigen flag -test.

### CODE REVIEW

File: autonity/solidity/contracts/Autonity.sol [hunk 1]
```diff
@@ -658,6 +658,25 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+        // copy in memory the calldata. Why? Is it a bug or what?
+        address precompile = Precompiled.UPGRADER_CONTRACT;
+        assembly {
+            let result := delegatecall(gas(), precompile, add(_data,32), mload(_data), 0, 0)
+            returndatacopy(0, 0, returndatasize())
+            switch result
+            case 0 { revert(0, returndatasize()) }
+            default { return(0, returndatasize()) }
+        }
+    }
```
Jason-Zhangxin-Chen, line 678: It would be very helpful to emit an event after a successful upgrade.

cmd/abigen/main.go
```diff
@@ -254,7 +260,7 @@ func abigen(c *cli.Context) error {
 		}
 	}
 	// Generate the contract binding
-	code, err := bind.Bind(types, abis, bins, sigs, c.GlobalString(pkgFlag.Name), lang, libs, aliases)
+	code, err := bind.Bind(types, abis, bins, sigs, c.GlobalString(pkgFlag.Name), lang, libs, aliases, c.GlobalBool("test"))
```
lorenzo-dev1, line 263:
```suggestion
	code, err := bind.Bind(types, abis, bins, sigs, c.GlobalString(pkgFlag.Name), lang, libs, aliases, c.GlobalBool(testFlag.Name))
```

autonity/solidity/contracts/Autonity.sol
```diff
@@ -658,6 +658,25 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+    *  @param _data is the concatenation of the target contract address and the create transaction calldata
+    */
+    function replaceContract(string memory _data) public onlyOperator {
+        // Youssef: I've tried hard to use the "bytes" type but couldn't make it works. According
+        // to the solidity reference https://docs.soliditylang.org/en/latest/types.html#the-functions-bytes-concat-and-string-concat
+        // string and bytes types should be equivalent BUT if using the bytes type then somehow the solidity compiler doesn't
+        // copy in memory the calldata. Why? Is it a bug or what?
+        address precompile = Precompiled.UPGRADER_CONTRACT;
+        assembly {
+            let result := delegatecall(gas(), precompile, add(_data,32), mload(_data), 0, 0)
```
lorenzo-dev1, line 672: This construction assumes that _data is longer than 31 bytes (otherwise it would be stored in a single slot without length at the start). But I think that is fine, since address is 20 bytes and I guess contract deployment code cannot be <12 bytes

autonity/solidity/contracts/Autonity.sol
```diff
@@ -658,6 +658,25 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+            returndatacopy(0, 0, returndatasize())
```
lorenzo-dev1, line 673: I just saw that these area of memory that we are using for the revert reason is reserved (https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html). We are doing the same in the other precompiles.

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: Or we can add an explicit check that require _data.length >= 32

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: I guess when we revert that is not an issue, but if we just return to another function it could be maybe?

autonity/solidity/contracts/Tests.sol
```diff
@@ -0,0 +1,20 @@
...
+// Used for testing contract upgrade mechanism.
+contract TestBase {
+    string public Foo;
+    constructor(string memory _foo){
+        Foo = _foo;
+    }
+}
+
+contract TestUpgraded is TestBase {
+    string public Bar;
```
lorenzo-dev1, line 13: Here Bar is going to be in the same storage of `Foo` is that wanted?

autonity/solidity/contracts/Tests.sol
```diff
@@ -0,0 +1,20 @@
...
+    constructor(string memory _bar, string memory _foo) TestBase(_foo) {
+        _bar = Bar;
```
lorenzo-dev1, line 15: The opposite?

lorenzo-dev1:
@yazzaoui will this new upgrade mechanism cohexist with the old one (`Upgradable.sol`)? Is it supposed to be used for all other contracts except `Autonity.sol`?

autonity/solidity/contracts/Autonity.sol
```diff
@@ -631,72 +635,47 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
         }
-        return _voterList;
+        return _voters;
     }
 
+    /** @dev Call the in-protocol EVM replace mechanism. Requires specific tool to interact.
+    * Restricted to the operator account.
+    *  @param _data is the concatenation of the target contract address and the create transaction calldata
+    */
+    function replaceContract(string memory _data) public onlyOperator {
```
lorenzo-dev1, line 665: Do we assume that the deployment bytecode is going to fit in a single tx ? I remember that the `Upgradable.sol` had a chunking mechanism to work around this issue (since protocol contracts are deployed at genesis without a tx)

autonity/solidity/contracts/Autonity.sol
yazzaoui: > This construction assumes that _data is longer than 31 bytes (otherwise it would be stored in a single slot without length at the start). But I think that is fine, since address is 20 bytes and I guess contract deployment code cannot be <12 bytes

Can you point me where this is specified? I looked both at the ABI packing code and the solidity ABI spec and couldn't find anything - admittedly this is for calldata but I would be very surprised if the layout in memory wasn't the same.

autonity/solidity/contracts/Tests.sol
yazzaoui: Notice inheritance of `TestBase` - TestUpgraded's `Bar` will have the next slot.

autonity/solidity/contracts/Tests.sol
yazzaoui: See above.

autonity/solidity/contracts/Autonity.sol
yazzaoui: Making a great point here @lorenzo-dev1 this is violating Solidity memory safety - I'll fix that and check the other precompile calls.

autonity/solidity/contracts/Autonity.sol
yazzaoui: A lot of governance operator functions are not emitting events. Which use case are you thinking about?

yazzaoui:
> @yazzaoui will this new upgrade mechanism cohexist with the old one (`Upgradable.sol`)? Is it supposed to be used for all other contracts except `Autonity.sol`?

Yes. The old one is the native Autonity contract upgrade mechanism and was only built for the Autonity contract.

> In any case I think it would be better to have a special protocol contract that acts as an upgrade manager for all the other ones. This is because if we upgrade the autonity contract and we brick it for some reason (e.g. wrong deployment bytecode), then we cannot access anymore the `replaceContract` function and we cannot fix it.

I see your point but the Autonity contract is a special case scenario - we should not update it using this mechanism. We have the proper checks in place with the native Autonity contract upgrade mechanism - it reverts if deployment is not successful.

> Instead I think that with a separated contract we could still issue a tx, revert back to the old autonity contract (not bricked) and finalize the block. Then we can try again to upgrade

Yes, that's the logic implemented with the native mechanism.

> Also, if we upgrade the autonity contract using a function of the autonity contract itself, I think it might mess things up (based on how the program counter is manipulated when returning from a function in the evm interpreter. I still have to check this).

Agree, we should not do that. I'm hesitant to put a require statement here to completely exclude it - if there is an absolute emergency situation where the native upgrade version cannot be executed, but I doubt it. I leave it like that because anyway, this is an admin tool that is only meant to be used with expert-level understanding of the protocol with prior local tests.

autonity/solidity/contracts/Autonity.sol
yazzaoui: Yes, all the other contracts with the exclusion of the Autonity contract can be deployed within a single tx so an upgrade can be made with a single tx too.

autonity/solidity/contracts/Autonity.sol
tbssajal: > This construction assumes that _data is longer than 31 bytes (otherwise it would be stored in a single slot without length at the start). But I think that is fine, since address is 20 bytes and I guess contract deployment code cannot be <12 bytes

@lorenzo-dev1  I think it is true only for storage layout to optimize the space used. I think it is different for layout in memory.

autonity/solidity/contracts/Autonity.sol
tbssajal: I also did something similar in compute committee precompiled. I was inspired from this one: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/proxy/Proxy.sol

tbssajal: I did that only to get the full revert reason in case the call failed, because in case of a successful call for compute committee, there was no need of any return data.

core/vm/contracts.go
```diff
@@ -179,24 +199,231 @@ func ActivePrecompiles(rules params.Rules) []common.Address {
...
+}
+
+func (u *Upgrader) Run(input []byte, _ uint64, evm *EVM, caller common.Address) ([]byte, error) {
+	if caller != params.AutonityContractAddress {
+		return nil, errUnauthorized
+	}
+	if len(input) < common.AddressLength+1 {
+		return nil, errBadInput
+	}
+	targetContract := common.Address(input[:common.AddressLength])
```
lorenzo-dev1, line 240: What about limiting the targetContract to be one of the protocol contracts? I don't think we want mutability for third party contracts (at least not yet)

autonity/solidity/contracts/Autonity.sol
vpiyush: > I just saw that these area of memory that we are using for the revert reason is reserved (https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html). We are doing the same in the other precompiles calls.

I think it's fine and cheaper to use 0 memory location, for functions. where we are not returning to solidity again. Also in my understanding, the initial 2 words are the scratchpad location which is safe even if we are returning to solidity so it also depends on the return data size

lorenzo-dev1: Intuitively I would say it is fine if we revert, but what if we just return the function and jump to another one?

lorenzo-dev1: > 0x00 - 0x3f (64 bytes): scratch space for hashing methods

I think the only risk here is that if we return to another function, the return value might get overwritten by some hashing function (however it depends on how the return value is managed). 

> also depends on the return data size

I agree, indeed if the return value is > 64 bytes it could overwrite the free memory pointer, which would cause serious issues

autonity/solidity/contracts/Tests.sol
lorenzo-dev1: I get the inheritance now but here aren't we assigning the parameter `_bar` with the uninitialized storage variable `Bar`?

Shouldn't it be `Bar = _bar`?

lorenzo-dev1:
> Yes, that's the logic implemented with the native mechanism.

I saw the logic it implements, however it will still succeed if we upgrade to a, say, empty contract by mistake. The contract is valid, so the upgrade will go through. But then you will have a bricked AC. It is true though that this is supposed to be tested locally before production

> Agree, we should not do that. I'm hesitant to put a require statement here to completely exclude it - if there is an absolute emergency situation where the native upgrade version cannot be executed, but I doubt it. I leave it like that because anyway, this is an admin tool that is only meant to be used with expert-level understanding of the protocol with prior local tests.

I guess it depends on whether we want to encode all the safety checks in the code, or if we just say "it is an admin tool and should be used carefully" and perform checks by hand when we will eventually use it.

autonity/solidity/contracts/Tests.sol
yazzaoui: My bad !!

core/vm/contracts.go
yazzaoui: I don't know... This logic should belong to the governance operator contract anyway.

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: Solving as not an issue

autonity/autonity.go
```diff
@@ -27,14 +26,8 @@ import (
 )
 
 var (
-	DeployerAddress               = params.DeployerAddress
-	AutonityContractAddress       = params.AutonityContractAddress
-	AccountabilityContractAddress = params.AccountabilityContractAddress
-	OracleContractAddress         = params.OracleContractAddress
-	ACUContractAddress            = params.ACUContractAddress
-	SupplyControlContractAddress  = params.SupplyControlContractAddress
-	StabilizationContractAddress  = params.StabilizationContractAddress
```
yazzaoui, line 36: I removed those variables, we can just use the `params.` prefixed ones.

lorenzo-dev1: not a big deal but we stil have `DeployerAddress = params.DeployerAddress`

autonity/solidity/contracts/UpgradeManager.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "./Precompiled.sol";
+
+contract UpgradeManager {
+    address public autonity;
```
lorenzo-dev1, line 7: I had a quick look at the other contract and it seems to me like our convention is to keep `autonity` and `operator` private

core/vm/contracts.go
lorenzo-dev1: In my opinion whether third party contracts are mutable (without the use of a proxy) should be something defined at protocol layer that cannot be changed by the operator

autonity/solidity/contracts/UpgradeManager.sol
```diff
@@ -0,0 +1,55 @@
...
+            case 0 { revert(0, returndatasize()) }
+            default { return(0, returndatasize()) }
+        }
+    }
+
+    /*
+    * @notice Set the Operator account. Restricted to the Operator account.
+    * @param _account the new operator account.
+    */
+    function setOperator(address _account) public onlyAutonity {
```
lorenzo-dev1, line 36: `external` instead of `public`?

core/vm/contracts.go
cmjc: Good quesiton. I think it should be limited to the protocol contracts. Rationale: in a decentralised network the operator governance contract upgrade scope is the protocol contract. 

Why would the governance operator want to upgrade a third party contract deployed by a third party?  surely only if it was tightly bound to the protocol, in which case wouldn't it be a protocol not third party contract?

autonity/solidity/contracts/UpgradeManager.sol
yazzaoui: I leave it like that for the time being, it's not a big deal and it's useful for e2e testing to check if the deploy configuration was okay. We may want to generalize that across all contracts to make those fields public, I don't see why not.

core/vm/contracts.go
yazzaoui: This is prone to debate and can be discussed later. I don't have a view on it. However, this restriction logic can be implemented in the governance contract which can be argued is a protocol contract too, so under our control. Anyways in practical terms It would take me too long to implement restrictions at this stage because we would need a registry contract for protocol contracts: it's unclear if protocol contracts == contracts deployed at genesis time. It is the case right now but chances it may change. E.g. if a genesis contract deploy itself a contract.

lorenzo-dev1: Then I think we should probably open a discussion about it. For the time being I am fine with the current implementation.

lorenzo-dev1 (approved):
LGTM once CI is green

core/vm/contracts.go
```diff
@@ -227,6 +232,25 @@ var (
...
+func (u *Upgrader) Run(input []byte, _ uint64, evm *EVM, caller common.Address) ([]byte, error) {
+	if caller != params.UpgradeManagerContractAddress {
+		return nil, errUnauthorized
+	}
+	if len(input) < common.AddressLength+1 {
+		return nil, errBadInput
+	}
+	targetContract := common.Address(input[:common.AddressLength])
+	// input is TARGET_ADDRESS + DEPLOY_CALLDATA
+	_, _, _, err := evm.Replace(AccountRef(params.DeployerAddress), input[common.AddressLength:], targetContract)
```
vpiyush, line 250: I remember looking into evm.Replace. It doesn't check whether the targetContract already exists or not? So replacing a contract with non-existent targetContract is repairable? what happens in this case?

PR #932, opened 2024-01-30
Port v0.12.3 fixes

lorenzo-dev1:
Ports v0.12.3 temporary fixes on:
- prometheus incompatibility with buffered gauges
- message cache size

### DIFF
--- a/consensus/tendermint/backend/engine.go
+++ b/consensus/tendermint/backend/engine.go
@@ -29,8 +29,8 @@ import (
 
 const (
 	inmemorySnapshots = 128 // Number of recent vote snapshots to keep in memory
-	inmemoryPeers     = 40
-	inmemoryMessages  = 1024
+	inmemoryPeers     = 150
+	inmemoryMessages  = 4096
 )
 
 // ErrStartedEngine is returned if the engine is already started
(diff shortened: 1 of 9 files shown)


yazzaoui (approved):
Thanks for that 👍

Jason-Zhangxin-Chen (approved):
LGTM

consensus/tendermint/backend/engine.go
```diff
@@ -29,8 +29,8 @@ import (
 
 const (
 	inmemorySnapshots = 128 // Number of recent vote snapshots to keep in memory
-	inmemoryPeers     = 40
-	inmemoryMessages  = 1024
+	inmemoryPeers     = 150
+	inmemoryMessages  = 8192
```
vpiyush, line 33: For my curiousity, Why do we need this so high? this is going to increase live heap size.

lorenzo-dev1: I leave this one for @yazzaoui but I think it is just trial and error to see how big we need to make it to not have the repeated messages issue.

yazzaoui: TBH even we those values we still have duplicate message warnings. I don't think we have time to solve the caching for the next release but we should think about a new approach after. For now let's leave it like that.

PR #934, opened 2024-02-01
ACN refinements

yazzaoui:
Currently WIP, made some refinements here and there. Please have a look at it @vpiyush.

Trying to figure out how to run large-scale internal testing network, I think it should be possible...


e2e_test/e2e_test.go
```diff
@@ -32,6 +33,36 @@ func TestSendingValue(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestLargeNetwork(t *testing.T) {
+	//t.Skip("only on demand")
+	// Kernel doesn't allow to have that many TCP connections required for running locally large networks
```
vpiyush, line 38: Theoretically, we can have socket connections as many as 65536 (barring the few reserved ones and already allotted). But there are soft limits on the terminal (execution environment) which is limited to 1024. This can be overridden in the test itself using rlimit (provided we have no permission issues). I tried the test with 75 nodes setting rlimit and also reducing caches (to avoid OOM) and did not face connection related issues. 

So probably we don't need in-memory dialing

p2p/server.go
```diff
@@ -406,12 +406,13 @@ func (srv *Server) RemoveTrustedPeer(node *enode.Node) {
 // a node belonging to the consensus committee is fully connected
 // to the other consensus committee nodes.
 func (srv *Server) UpdateConsensusEnodes(enodes []*enode.Node) {
-	var currEnodes []*enode.Node
-	srv.enodeMu.RLock()
-	currEnodes = append(currEnodes, srv.consensusNodes...)
-	srv.enodeMu.RUnlock()
+	currentNodes := make([]*enode.Node, len(srv.consensusNodes))
```
vpiyush, line 409: Thanks, agree with this change, I guess we could also improve this further by not doing any allocation/updation if there is no change in the committee, however when we have it only on epoch blocks that's going to be insignificant.

lorenzo-dev1: Maybe is not a problem but aren't we reading the `srv.consensusNodes` in an unprotected way here?

vpiyush: you have a valid point here( i missed this), we should have it within lock

yazzaoui: Good point!

vpiyush:
Is this still needed. Can we close this?

yazzaoui:
Closing as depreciated

PR #935, opened 2024-02-05
handle errKnownBlock properly on proposal verification

vpiyush:
handle ErrknownBlock from proposal verification properly. Additionally, We check if the height in proposal Message is same as the proposedBlock number to avoid processing the old valid proposal again.

### DIFF
--- a/consensus/tendermint/core/propose.go
+++ b/consensus/tendermint/core/propose.go
@@ -110,8 +110,12 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
 				})
 			})
 			return err
-		} else if errors.Is(err, core.ErrKnownBlock) {
-			return constants.ErrAlreadyHaveProposal
+		}
+		// if the proposal block is already in the chain, no need to prevote for nil
+		if errors.Is(err, core.ErrKnownBlock) || errors.Is(err, constants.ErrAlreadyHaveBlock) {
+			c.logger.Info("Verified proposal that was already in our local chain", "err", err, "duration", duration)
+			c.SetStep(ctx, PrecommitDone) // we do not need to process any more consensus messages for this height
+			return constants.ErrAlreadyHaveBlock
 		}
 		// Proposal is invalid here, we need to prevote nil.
 		// However, we may have already sent a prevote nil in the past without having processed the proposal
--- a/consensus/tendermint/core/propose.go
+++ b/consensus/tendermint/core/propose.go
@@ -9,6 +9,7 @@ import (
 	"github.com/autonity/autonity/consensus"
 	"github.com/autonity/autonity/consensus/tendermint/core/constants"
 	"github.com/autonity/autonity/consensus/tendermint/core/message"
+	"github.com/autonity/autonity/core"
 	"github.com/autonity/autonity/core/types"
 	"github.com/autonity/autonity/metrics"
 )
@@ -109,6 +110,8 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
 				})
 			})
 			return err
+		} else if errors.Is(err, core.ErrKnownBlock) {
+			return constants.ErrAlreadyHaveProposal
 		}
 		// Proposal is invalid here, we need to prevote nil.
 		// However, we may have already sent a prevote nil in the past without having processed the proposal
--- a/consensus/tendermint/core/rules.go
+++ b/consensus/tendermint/core/rules.go
@@ -3,7 +3,6 @@ package core
 import (
 	"context"
 	"errors"
-	"fmt"
 	"math/big"
 
 	"github.com/autonity/autonity/common"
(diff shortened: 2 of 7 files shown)


Jason-Zhangxin-Chen (approved):
LGTM

consensus/tendermint/core/rules.go
```diff
@@ -123,6 +125,9 @@ func (c *Core) quorumPrecommitsCheck(ctx context.Context, proposal *message.Prop
 	// if there is a quorum, verify the proposal if needed
 	if !verified {
 		if _, err := c.backend.VerifyProposal(proposal.Block()); err != nil {
+			if errors.Is(err, core.ErrKnownBlock) {
```
lorenzo-dev1, line 128: Could you add a comment for why this can happen (insertion of block from p2p layer due to block propagation) ?

This is one of the pieces of code that my future self will spend 2 hours reverse engineering why this can happen otherwise ahah

consensus/tendermint/core/propose.go
```diff
@@ -109,6 +110,8 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
 				})
 			})
 			return err
+		} else if errors.Is(err, core.ErrKnownBlock) {
+			return constants.ErrAlreadyHaveProposal
```
lorenzo-dev1, line 114: I think we can even return nil. `ErrAlreadyHaveProposal` should be returned only when we are receiving a proposal but we already previously processed another one.

Here we are actually processing our first proposal, but while we are processing it the same block is inserted from p2p propagation. So it is slightly different

lorenzo-dev1: Or just return ErrKnownBlock and add it to the shouldDisconnect list to avoid disconnection

lorenzo-dev1:
I am thinking that we should add an explicit check in `VerifyProposal` to check if we already have the header of the proposed block. 

Right now we rely on other checks done in, for example, `ValidateBody` but they are not really made on purpose for our proposal verification flow (e.g. the `ValidateBody` checks that we have the block AND the related state. We actually want to reject a proposal if we just have the header related to it, even without state).

@vpiyush if you are busy with ACN I can take over these changes (if we agree on them), let me know

consensus/tendermint/core/propose.go
vpiyush: Or we can define a new error specific for this scenario?

vpiyush:
> I am thinking that we should add an explicit check in `VerifyProposal` to check if we already have the header of the proposed block.
> 
> Right now we rely on other checks done in, for example, `ValidateBody` but they are not really made on purpose for our proposal verification flow (e.g. the `ValidateBody` checks that we have the block AND the related state. We actually want to reject a proposal if we just have the header related to it, even without state).
> 
> @vpiyush if you are busy with ACN I can take over these changes (if we agree on them), let me know

Please go ahead, Thanks.

consensus/tendermint/core/rules.go
lorenzo-dev1: Solved in https://github.com/autonity/autonity/pull/935/commits/b4f00000118ace6ffe31e84010a5ceeabe14a878

lorenzo-dev1:
> > I am thinking that we should add an explicit check in `VerifyProposal` to check if we already have the header of the proposed block.
> > Right now we rely on other checks done in, for example, `ValidateBody` but they are not really made on purpose for our proposal verification flow (e.g. the `ValidateBody` checks that we have the block AND the related state. We actually want to reject a proposal if we just have the header related to it, even without state).
> > @vpiyush if you are busy with ACN I can take over these changes (if we agree on them), let me know
> 
> Please go ahead, Thanks.

done in https://github.com/autonity/autonity/pull/935/commits/b4f00000118ace6ffe31e84010a5ceeabe14a878

lorenzo-dev1:
I also added some code to improve how we deal with the `ErrPrunedAncestor` when verifying a proposal --> https://github.com/autonity/autonity/pull/935/commits/b611f45538b998171ecc8235ad308cb991ab6133

consensus/tendermint/core/rules.go
```diff
@@ -123,7 +126,16 @@ func (c *Core) quorumPrecommitsCheck(ctx context.Context, proposal *message.Prop
...
+			// This can happen if while we are processing the proposal,
+			// we actually receive the finalized proposed block from p2p block propagation (other peers already reached quorum on it)
+			// In this case we can just consider the proposal as committed.
+			if errors.Is(err, core.ErrKnownBlock) || errors.Is(err, constants.ErrAlreadyHaveBlock) {
+				c.logger.Info("Verified proposal that was already in our local chain", "err", err)
+				c.SetStep(ctx, PrecommitDone) // we do not need to process any more consensus messages for this height
+				return true
+			}
 			// Impossible with the BFT assumptions of 1/3rd honest.
+			c.logger.Error("Failed verification of committed proposal", "err", err)
```
vpiyush, line 138: better to print the err with panic itself. So you get the info at one shot with stack trace. Also you could just avoid this error log completely

vpiyush: Yes, It should be. errr implements string. you can also concateneate

lorenzo-dev1: right I had forgot about that lol

fixed in https://github.com/autonity/autonity/pull/935/commits/1af2a7b838bb526386d5312e0064d8198f24652d

vpiyush (commented):
LGTM!

yazzaoui (approved):

PR #936, opened 2024-02-07
Execution layer network topology

tbssajal:
proposal [here](https://github.com/clearmatics/autonity-protocol/discussions/267)


eth/backend.go
```diff
@@ -658,3 +659,58 @@ func (s *Ethereum) Stop() error {
 func (s *Ethereum) Logger() log.Logger {
 	return s.log
 }
+
+func (s *Ethereum) squareRootUInt(n uint) uint {
+	if n == 0 {
+		return 0
+	}
+	// return x (uint) such that x*x >= n and (x-1)*(x-1) < n where x > 0
```
vpiyush, line 667: For value 5, this returns 2 and for 10 it would be 3. which fails against this constraint. If thats the expectation better to have a unit test which verfiy's against this constraint.

vpiyush, line 667: can you elaborate on why math.ceil(math.sqrt(n)) wouldn't work here?

eth/backend.go
```diff
@@ -658,3 +659,58 @@ func (s *Ethereum) Stop() error {
...
+		a /= base
+		b /= base
+		digitCount--
+	}
+	return count
+}
+
+// Returns the list of adjacentNodes to connect with localNode. Given that the order of the input array nodes is same
+// for everyone, connecting to only adjacentNodes will create a connected graph with diameter = 2
+func (s *Ethereum) AdjacentNodes(nodes []*enode.Node, localNode *enode.LocalNode) []*enode.Node {
```
vpiyush, line 696: All 3 fuctions have nothing to do with ethereum object. So either they go in helper OR you can make them independent locals

eth/backend.go
```diff
@@ -658,3 +659,58 @@ func (s *Ethereum) Stop() error {
...
+		if mid >= n/mid {
+			high = mid
+		} else {
+			low = mid + 1
+		}
+	}
+	return high
+}
+
+func (s *Ethereum) similarDigitCount(a, b, base, digitCount uint) uint {
```
vpiyush, line 681: can you add some comments here, what's the objective of this method

eth/backend.go
vpiyush, line 696: Add some more comments on the graph construction. it's fine to have elaborative comments i'd say

eth/backend.go
tbssajal: Thanks for catching it!

tbssajal: I am just wary of math.sqrt due to past experience. I once got $\text{math.sqrt}(10^{18}) > 10^9$, although it should work for small numbers but I was just trying to play safe. Also this particular experience was in c++, no golang. If it is okay to use math.ceil(math.sqrt(n)), then I will use it.

core/types/nodes.go
```diff
@@ -30,23 +28,23 @@ func NewNodes(strList []string, asACN bool) *Nodes {
 		make([]string, len(strList)),
 	}
 
-	for _, enodeStr := range strList {
+	for i, enodeStr := range strList {
+		idx := i
```
lorenzo-dev1, line 32: If you do this you can avoid passing `idx` to the go routine later, since a new memory allocation will be done at every iteration, therefore the go routine will always take the closure for the correct variable.

To understand I used this:
```
package main

import (
  "fmt"
  "sync"
)

func main() {
  var wg sync.WaitGroup
  for i := 0; i < 100; i++ {
    idx := i
    fmt.Println(&idx)
    wg.Add(1)
    go func() {
      defer wg.Done()
      fmt.Println(idx)
    }()
  }
  wg.Wait()
}
```

core/types/nodes.go
```diff
@@ -14,32 +13,38 @@ type Nodes struct {
...
+	}
 
 	n := &Nodes{
 		make([]*enode.Node, len(strList)),
 		make([]string, len(strList)),
 	}
 
-	for _, enodeStr := range strList {
+	for i, enodeStr := range strList {
+		idx := i
```
lorenzo-dev1, line 32: If you do this you can avoid passing `idx` to the go routine later, since a new memory allocation will be done at every iteration, therefore the go routine will always take the closure for the correct variable.

To understand I used this:
```
package main

import (
  "fmt"
  "sync"
)

func main() {
  var wg sync.WaitGroup
  for i := 0; i < 100; i++ {
    idx := i
    fmt.Println(&idx)
    wg.Add(1)
    go func() {
      defer wg.Done()
      fmt.Println(idx)
    }()
  }
  wg.Wait()
}
```

tbssajal: done. thanks

eth/backend.go
tbssajal: done

eth/backend.go
tbssajal: done

eth/backend.go
```diff
@@ -191,6 +198,7 @@ func New(stack *node.Node, config *Config) (*Ethereum, error) {
 		bloomRequests:     make(chan chan *bloombits.Retrieval),
 		bloomIndexer:      core.NewBloomIndexer(chainDb, params.BloomBitsBlocks, params.BloomConfirms),
 		p2pServer:         stack.ExecutionServer(),
+		topologySelector:  NetworkTopology{2, maxInCommitteePeers},
```
tbssajal, line 201: I think we should call topologySelector.SetMinNodes(1) for testing, otherwise tests will not active the logic of graph construction in the topology or we will need to add tests for more than 20 nodes.

eth/backend.go
```diff
@@ -581,20 +589,27 @@ func (s *Ethereum) newCommitteeWatcher() {
...
-		s.p2pServer.UpdateConsensusEnodes(enodesList.List)
+		s.p2pServer.UpdateConsensusEnodes(s.topologySelector.RequestSubset(committee.List, s.p2pServer.LocalNode()), committee.List)
 	}
 	wasValidating := false
 	currentBlock := s.blockchain.CurrentBlock()
 	if currentBlock.Header().CommitteeMember(s.address) != nil {
-		updateConsensusEnodes(currentBlock)
 		s.miner.Start()
 		s.log.Info("Starting node as validator")
+		updateConsensusEnodes(currentBlock)
```
vpiyush, line 612: please undo this change. Let this be same as consensus_channel_enode branch

eth/backend.go
```diff
@@ -610,17 +625,17 @@ func (s *Ethereum) newCommitteeWatcher() {
...
 				}
 				continue
 			}
-			updateConsensusEnodes(ev.Block)
 			// if we were not committee in the past block we need to enable the mining engine.
 			if !wasValidating {
 				s.log.Info("Local node detected part of the consensus committee, mining started")
 				s.miner.Start()
 			}
+			updateConsensusEnodes(ev.Block)
```
vpiyush, line 638: same here

eth/protocols/eth/peer.go
```diff
@@ -79,6 +79,8 @@ type Peer struct {
 	head common.Hash // Latest advertised head block hash
 	td   *big.Int    // Latest advertised head block total difficulty
 
+	//TODO: should be removed from here, these will be handled through
```
vpiyush, line 82: remove this todo

p2p/server.go
```diff
@@ -405,17 +406,18 @@ func (srv *Server) RemoveTrustedPeer(node *enode.Node) {
 // UpdateConsensusEnodes is responsible to make sure that
 // a node belonging to the consensus committee is fully connected
 // to the other consensus committee nodes.
-func (srv *Server) UpdateConsensusEnodes(enodes []*enode.Node) {
-
+// Note that peers is a subset of committee.
```
vpiyush, line 409: this comment is not needed anymore

eth/backend.go
tbssajal: using math.sqrt

tbssajal: using math.sqrt

eth/backend.go
vpiyush: We can have a separate test which can do that. but default behaviour is 25 nodes.

eth/backend.go
```diff
@@ -62,6 +62,12 @@ import (
 	"github.com/autonity/autonity/rpc"
 )
 
+const (
+	// txChanSize is the size of channel listening to NewTxsEvent.
+	// The number is referenced from the size of tx pool.
+	maxInCommitteePeers = 20
```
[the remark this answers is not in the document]
tbssajal: aw, I said yesterday at standup that I got max degree 24 but that was due to an error in binary search in the squareRootUInt function. After fixing it, I got max degree <= 18 for nodes <= 100

vpiyush (approved):
LGTM

PR #942, opened 2024-02-14
Optimize execution layer topology

tbssajal:
- Runtime optimization of execution layer network topology to support $\text{MaxCommitteeSize} = 1000$
- Improve execution layer network topology construction to have $\text{diameter} \leq 4 \ and \ \text{degree} \leq 24$

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -595,7 +595,8 @@ func (s *Ethereum) newCommitteeWatcher() {
 			return
 		}
 
-		s.p2pServer.UpdateConsensusEnodes(s.topologySelector.RequestSubset(committee.List, s.p2pServer.LocalNode()), committee.List)
+		idx := s.topologySelector.MyIdx(committee.List, s.p2pServer.LocalNode())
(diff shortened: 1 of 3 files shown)


lorenzo-dev1 (approved):
LGTM!

Is there a test that checks that no node remains isolated even when we have multiple components?

yazzaoui (approved):
LGT

tbssajal:
> LGTM!
> 
> Is there a test that checks that no node remains isolated even when we have multiple components?

I think I have it in tests, but I will confirm!

tbssajal:
> LGTM!
> 
> Is there a test that checks that no node remains isolated even when we have multiple components?

TestGraphDiamter does that by checking the shortest distance between each pair of nodes. If some node is isolated, then the shortest distance for some pair will not be determined.

tbssajal:
[degreeCount.txt](https://github.com/user-attachments/files/15884829/degreeCount.txt)
Degree count for all nodes for committee size = 50, 64,65,100,200,500,1000

PR #943, opened 2024-02-14
Various fixes

yazzaoui:
- Genesis header correct committee assignment if the genesis validator config is above max consensus meembers
- Add a test supporting large internal network using pipes
- Fix a potential panic in the fault detector
- Rework consensus message handling flow to avoid decoding if message was already received

### DIFF
--- a/cmd/ethkey/autinspect.go
+++ b/cmd/ethkey/autinspect.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	"gopkg.in/urfave/cli.v1"
+
+	"github.com/autonity/autonity/cmd/utils"
+	"github.com/autonity/autonity/common/hexutil"
+	"github.com/autonity/autonity/crypto"
+)
+
+type outputAutInspect struct {
+	NodePublicKey       string
+	NodeAddress         string
+	NodePrivateKey      string `json:",omitempty"`
+	ConsensusPublicKey  string
+	ConsensusPrivateKey string `json:",omitempty"`
+}
+
+var commandAutInspect = cli.Command{
+	Name:      "autinspect",
+	Usage:     "inspect autonity keys file",
+	ArgsUsage: "<keyfile>",
+	Description: `
+Print various information about the autonity keyfile.
+
+Private key information can be printed by using the --private flag;
+make sure to use this feature with great caution!`,
+	Flags: []cli.Flag{
+		jsonFlag,
+		cli.BoolFlag{
+			Name:  "private",
+			Usage: "include the private keys in the output",
+		},
+	},
+	Action: func(ctx *cli.Context) error {
+		if ctx.NArg() != 1 {
+			fmt.Println("Incorrect number of arguments, 1 expected.")
+			cli.ShowCommandHelpAndExit(ctx, "autinspect", 1)
+		}
+		nodeKey, consensusKey, err := crypto.LoadAutonityKeys(ctx.Args().Get(0))
+		if err != nil {
+			utils.Fatalf("error opening key file %v", err)
+		}
+		out := outputAutInspect{
+			NodePublicKey:      hexutil.Encode(crypto.FromECDSAPub(&nodeKey.PublicKey)[1:]),
+			NodeAddress:        crypto.PubkeyToAddress(nodeKey.PublicKey).String(),
+			ConsensusPublicKey: consensusKey.PublicKey().Hex(),
+		}
+		showPrivate := ctx.Bool("private")
+		if showPrivate {
+			out.ConsensusPrivateKey = hex.EncodeToString(consensusKey.Marshal())
+			out.NodePrivateKey = hex.EncodeToString(crypto.FromECDSA(nodeKey))
+		}
+		if ctx.Bool(jsonFlag.Name) {
+			mustPrintJSON(out)
+		} else {
+			fmt.Println("Node Address:          ", out.NodeAddress)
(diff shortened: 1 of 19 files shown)


cmd/ethkey/verifypop.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/autonity/autonity/accounts"
+	"github.com/autonity/autonity/cmd/utils"
+	"github.com/autonity/autonity/common"
+	"github.com/autonity/autonity/common/hexutil"
+	"github.com/autonity/autonity/crypto"
+	"github.com/autonity/autonity/crypto/blst"
+	enodes "github.com/autonity/autonity/p2p/enode"
+)
+
+const popSize = 226
```
lorenzo-dev1, line 17: we have `AutonityPOPLen` in autonity_crypto.go

consensus/tendermint/backend/handler.go
```diff
@@ -108,23 +109,16 @@ func handleConsensusMsg[T any, PT interface {
 	*T
 	message.Msg
 }](sb *Backend, sender common.Address, p2pMsg p2p.Msg, errCh chan<- error) (bool, error) {
+	buffer := bytes.NewBuffer(make([]byte, 0, p2pMsg.Size))
+	if _, err := io.Copy(buffer, p2pMsg.Payload); err != nil {
+		return true, err
+	}
+	p2pMsg.Payload = bytes.NewReader(buffer.Bytes())
 	if !sb.coreStarted {
-		// We copy the message here as it can't be saved directly due
```
lorenzo-dev1, line 112: Not a big deal but I think this comment was actually useful, otherwise the reason why we reassign the payload is a bit obscure if you are reading only the handler

consensus/tendermint/backend/handler.go
```diff
@@ -108,23 +109,16 @@ func handleConsensusMsg[T any, PT interface {
...
-		// to a call to Discard in the eth handler which is going to empty this buffer.
-		buffer := bytes.NewBuffer(make([]byte, 0, p2pMsg.Size))
-		if _, err := io.Copy(buffer, p2pMsg.Payload); err != nil {
-			return true, errDecodeFailed
-		}
-		savedMsg := p2pMsg
```
lorenzo-dev1, line 118: Maybe I am missing something but I think we need this copy?

Because otherwise the ACN handler is going to call `Discard()` on the same p2pMsg and delete the payload even if we cloned it? Since it is going to be able to access the new reference.

Or does the copy happen when we `Enqueue`?

core/tx_list.go
```diff
@@ -194,83 +194,83 @@ func (m *txSortedMap) Remove(nonce uint64) bool {
...
+	}
+	m.cache = nil
+
+	return ready
 }
 
 // Len returns the length of the transaction map.
 func (m *txSortedMap) Len() int {
-    return len(m.items)
+	return m.index.Len()
```
lorenzo-dev1, line 215: Is this change because retrieving the index length is supposed to be faster?

lorenzo-dev1 (approved):
LGTM

consensus/tendermint/backend/handler.go
```diff
@@ -108,23 +109,16 @@ func handleConsensusMsg[T any, PT interface {
 	*T
 	message.Msg
 }](sb *Backend, sender common.Address, p2pMsg p2p.Msg, errCh chan<- error) (bool, error) {
+	buffer := bytes.NewBuffer(make([]byte, 0, p2pMsg.Size))
```
vpiyush, line 112: It's not clear to me as why should we create the new buffer everytime when we are not enqueuing the message for handling it later

yazzaoui: The hashing function needs access to a byte slice. There are probably ways to optimize that later if needed.

core/tx_list.go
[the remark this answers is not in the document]
yazzaoui: Indeed, I checked the internal of the golang map and it saves a counter so I will revert that

yazzaoui: This is what I thought but I was wrong, the golang map saves internally a counter too.

cmd/ethkey/autinspect.go
```diff
@@ -0,0 +1,70 @@
...
+		jsonFlag,
+		cli.BoolFlag{
+			Name:  "private",
+			Usage: "include the private keys in the output",
+		},
+	},
+	Action: func(ctx *cli.Context) error {
+		if ctx.NArg() != 1 {
+			fmt.Println("Incorrect number of arguments, 5 expected.")
+			cli.ShowCommandHelpAndExit(ctx, "verifypop", 1)
```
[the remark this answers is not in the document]
yazzaoui: Oops!

consensus/tendermint/backend/handler.go
```diff
@@ -134,12 +130,17 @@ func handleConsensusMsg[T any, PT interface {
...
 	}
-	m.Add(msg.Hash(), true)
+	m.Add(hash, true)
 	// Mark the message known for ourselves
-	if _, ok := sb.knownMessages.Get(msg.Hash()); ok {
+	if _, ok := sb.knownMessages.Get(hash); ok {
 		return true, nil
 	}
-	sb.knownMessages.Add(msg.Hash(), true)
+	sb.knownMessages.Add(hash, true)
```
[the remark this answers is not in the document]
vpiyush: created a separate ticket for this as it's OOS for this PR

vpiyush (approved):
LGTM

PR #945, opened 2024-02-20
Various fixes gogc

vpiyush:
Improvements On
1. GOGC setting
2. proposal verification 
3. keep a counter to avoid iterating the pending map every time we reorg

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -486,10 +486,7 @@ func (pool *TxPool) Stats() (int, int) {
 // stats retrieves the current pool stats, namely the number of pending and the
 // number of queued (non-executable) transactions.
 func (pool *TxPool) stats() (int, int) {
-	pending := 0
-	for _, list := range pool.pending {
-		pending += list.Len()
-	}
+	pending := int(pool.totalPending.Load())
 	queued := 0
 	for _, list := range pool.queue {
 		queued += list.Len()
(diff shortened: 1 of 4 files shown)


core/tx_pool.go
```diff
@@ -1385,10 +1388,7 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) []*types.Trans
 // pending limit. The algorithm tries to reduce transaction counts by an approximately
 // equal number for all for accounts with many pending transactions.
 func (pool *TxPool) truncatePending() {
-	pending := uint64(0)
-	for _, list := range pool.pending {
-		pending += uint64(list.Len())
-	}
+	pending := uint64(pool.totalPending.Load())
```
lorenzo-dev1, line 1391: I would be more confident in approving this change if we can have a test that ensures that `pool.totalPending` is always equal to the length computed via the legacy iteration method

lorenzo-dev1: Also the `pool.stats()` function still uses the legacy method

lorenzo-dev1: Another idea would be to keep both methods for now, and compare the values when the len is computed. If they mismatch print a log. That way we can monitor if the two values ever mismatch and eventually update for next release.

Jason-Zhangxin-Chen (approved):
LGTM.

core/tx_pool.go
vpiyush: updated pool.stats to use the totalPending counter, All of the tests use to validate the pool internals . this should automatically test the totalPending counter

lorenzo-dev1:
The txpool change looks definitely better now that `Stats()` is modified and therefore we are testing the change. However tbh I am not very familiar with the txpool logic, therefore I don't really feel super confident about it. I would just rely on hoping that the go-ethereum tests (which I never looked at) are properly written to approve...

What is your opinion @yazzaoui ?

yazzaoui (approved):

Issue #873, opened 2023-11-28
Testing workstream


yazzaoui:
This meta-issue is created to serve as a central hub for coordinating and tracking all testing-related efforts and discussions.

- [ ] #806

lorenzo-dev1:
I think we are also missing tests for self incriminating messages in the accountability package (`fault_detector_test.go`).

lorenzo-dev1:
Another thing to improve would be the precompiled contract testing from truffle.

At the moment we are using mocks to run test that use precompiled contracts on ganache, however it is very cumbersome and wasteful to have to mock every precompile in solidity.

It would be create to switch to a testing framework that allows instamine (thus fast test times) and as well the runtime addition of precompiled contracts.

yazzaoui:
On that note @lorenzo-dev1 I'm currently investigating the possibility to adopt Go with an in-house framework for testing protocol contracts. It solves many issues that you are mentioning and more.
See the work @tbssajal has been doing https://github.com/autonity/autonity/blob/develop/autonity/autonity_test.go , we can call protocol contracts functions with regular Go Test* functions.

Jason-Zhangxin-Chen:
This one: Refine the tendermint upon condittion test by creating common funtion to setup committee, consensus state, propser. could be closed by PR: https://github.com/autonity/autonity/pull/950 as it refine the upon_condition_test.go

Jason-Zhangxin-Chen:
- `Fuzz Testing` is resolved by commits: https://github.com/autonity/autonity/pull/950/commits/9ed002c069a7c2f5bce846b7272c4d0089dd7c83, https://github.com/autonity/autonity/pull/950/commits/c7ce5b78b1637a7b9dadb141f672a5af866c8b97, https://github.com/autonity/autonity/pull/950/commits/b393491843ae19392bc4533e5be4bb0ba4aa8d7b, https://github.com/autonity/autonity/pull/950/commits/d97be968659038071b1de57bdf44dd37e7880587

Jason-Zhangxin-Chen:
[...]

Issue #806, opened 2023-07-17
Fix flaky tests


yazzaoui:
Opening this issue to centralize issues we're having with flaky tests:

-  FAIL: TestQuorumPrevote/receive_quroum_prevote_for_proposal_block_when_in_step_>=_prevote (0.00s)

```
panic: invalid argument to Intn [recovered]
	panic: invalid argument to Intn

goroutine 149 [running]:
testing.tRunner.func1.2({0x993e00, 0xbdab30})
	/opt/hostedtoolcache/go/1.[20](https://github.com/autonity/autonity/actions/runs/5579173061/jobs/10194380373#step:5:21).5/x64/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.20.5/x64/src/testing/testing.go:1529 +0x39f
panic({0x993e00, 0xbdab30})
	/opt/hostedtoolcache/go/1.20.5/x64/src/runtime/panic.go:884 +0x[21](https://github.com/autonity/autonity/actions/runs/5579173061/jobs/10194380373#step:5:22)3
math/rand.(*Rand).Intn(0xc990e6?, 0xc990e6?)
	/opt/hostedtoolcache/go/1.20.5/x64/src/math/rand/rand.go:179 +0x65
math/rand.Intn(...)
	/opt/hostedtoolcache/go/1.20.5/x64/src/math/rand/rand.go:358
github.com/autonity/autonity/consensus/tendermint/core.TestQuorumPrevote.func1(0x5c46f88ca79a792d?)
	/home/runner/work/autonity/autonity/consensus/tendermint/core/upon_condition_test.go:914 +0x1b5
testing.tRunner(0xc0001b[24](https://github.com/autonity/autonity/actions/runs/5579173061/jobs/10194380373#step:5:25)e0, 0xc00007a300)
	/opt/hostedtoolcache/go/1.20.5/x64/src/testing/testing.go:1576 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.20.5/x64/src/testing/testing.go:16[29](https://github.com/autonity/autonity/actions/runs/5579173061/jobs/10194380373#step:5:30) +0x3ea
```

Jason-Zhangxin-Chen:
> TestMultiSyncManyUselessWithLowTimeout

@lorenzo-dev1 About this flaky test, do you have any failed call trace or log context available?

Jason-Zhangxin-Chen:
[...]

Issue #799, opened 2023-07-12
accountability : add support for PVN and PVO e2e tests


yazzaoui:
PVN and PVO e2e testing is currently disabled as those tests can not be supported with the current testing framework.
Indeed, with only one byzantine validator one cannot produce these kind of faults without triggering PN and PO faults as well.

To fix that, we'll need to add support for multiple byzantine validators.

Jason-Zhangxin-Chen:
To be closed by PR: https://github.com/autonity/autonity/pull/950 via commit https://github.com/autonity/autonity/pull/950/commits/318bdfbb495225f47d21ae628184a28f2e6a790e

lorenzo-dev1:
Thanks to @Jason-Zhangxin-Chen we now have the collusion framework, so this can be closed.

Issue #751, opened 2021-01-21
TestTendermintNoQuorum is unreliable


piersy:
# Problem

The test framework provides a parameter `noQuorumAfterBlock` that is used to check that no blocks are mined by after the given block, but this relies on the assumption that a node is stopped at a particular block, because node stopping happens in a different go-routine to block creation and there is no direct synchronisation between them, we can make no assumption about how many blocks will have been mined by the time a node stops. As such this test is unreliable.

PR #950, opened 2024-02-28
Test Flow Improvements

Jason-Zhangxin-Chen:
There are a lot of issues on the board pointing to the tests on top of legacy integration test framework are flaky, for example timeout and un-clean shutdowns. Thus we moved those integration tests to the new test framework, to maintain the tests on top of a more stable test framework, making the engineering to be more eficient.     

This PR is to close issues regarding the tests on top of the legacy test framework, and those flaky ones of unit tests:
- https://github.com/autonity/autonity/issues/750
- https://github.com/autonity/autonity/issues/751.
- https://github.com/autonity/autonity/issues/711.
- https://github.com/autonity/autonity/issues/653
- https://github.com/autonity/autonity/issues/645
- https://github.com/autonity/autonity/issues/641
- https://github.com/autonity/autonity/issues/637
- https://github.com/autonity/autonity/issues/624
- https://github.com/autonity/autonity/issues/806
- https://github.com/autonity/autonity/issues/659
- https://github.com/autonity/autonity/issues/799
- it also closes sub tasks of https://github.com/autonity/autonity/issues/873

Up stream flaky test: "TestMultiSyncManyUselessWithLowTimeout" was fixed in commit: https://github.com/autonity/autonity/commit/2c65d23d9b7cbe720f811103bd1d408bf4a69730

Up stream flaky test: " TestWalletNotifications" was fixed in commit: https://github.com/autonity/autonity/commit/ee3037f1f51f2488a5c88da93997d74eb98660ac

For now, we deprecated the legacy test framework, and rewrite those tests base on the new e2e test framework.


consensus/tendermint/core/upon_condition_test.go
```diff
@@ -667,798 +505,625 @@ func TestOldProposal(t *testing.T) {
...
+		e := NewConsensusEnv(t, customizer)
+
 		// create random prevote or precommit from 2 different
-		msg1 := message.NewPrevote(currentRound+1, currentHeight.Uint64(), common.Hash{}, makeSigner(privateKeys[sender1.Address], sender1.Address)).
+		msg1 := message.NewPrevote(e.curRound+1, e.curHeight.Uint64(), common.Hash{}, signer(e, 1)).
 			MustVerify(func(address common.Address) *types.CommitteeMember {
 				return &types.CommitteeMember{
 					Address:     address,
-					VotingPower: new(big.Int).Sub(roundChangeThreshold, common.Big1),
+					VotingPower: e.committee.F(),
```
lorenzo-dev1, line 1119: I think here the power we levels we were using are different wrt to the new refactored version. Is this test still correct?

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -1467,66 +1132,51 @@ func TestFutureRoundChange(t *testing.T) {
...
+		e := NewConsensusEnv(t, customizer)
+
 		// The collective power of the 2 messages  is more than roundChangeThreshold
-		prevoteMsg := message.NewPrevote(currentRound+1, currentHeight.Uint64(), common.Hash{}, makeSigner(privateKeys[sender1.Address], sender1.Address)).
+		prevoteMsg := message.NewPrevote(e.curRound+1, e.curHeight.Uint64(), common.Hash{}, signer(e, 1)).
 			MustVerify(func(address common.Address) *types.CommitteeMember {
 				return &types.CommitteeMember{
 					Address:     address,
-					VotingPower: new(big.Int).Sub(roundChangeThreshold, common.Big1),
+					VotingPower: e.committee.F(),
```
lorenzo-dev1, line 1157: same here

consensus/tendermint/accountability/fault_detector_test.go
```diff
@@ -498,760 +499,1001 @@ func TestRuleEngine(t *testing.T) {
...
-		assert.Equal(t, autonity.Accusation, onChainProofs[0].Type)
-		assert.Equal(t, autonity.PO, onChainProofs[0].Rule)
-		assert.Equal(t, message.NewLightProposal(oldProposal).Signature(), onChainProofs[0].Message.Signature())
+		proofs := fd.oldProposalsAccountabilityCheck(height, quorum)
+		require.Equal(t, 1, len(proofs))
+		actualProof := proofs[0]
+		require.Equal(t, expectedProof, actualProof)
 	})
 
-	t.Run("RunRule address the Accusation of PO rule, no quorum preVotes presented on the valid round BUT proposal was committed", func(t *testing.T) {
```
lorenzo-dev1, line 700: I think that in the refactor we lost the test that were checking that no accusation is raised if the block has been committed. It is important that we re-add this kind of tests, we want to avoid that nodes raise useless accusations. 

see https://github.com/autonity/autonity/blob/develop/consensus/tendermint/accountability/fault_detector.go#L827

lorenzo-dev1 (changes requested):
Awesome work! `upon_condition` tests looks much better and the collusion framework looks awesome!

I left few comments, the only one blocking for merging is [this one](https://github.com/autonity/autonity/pull/950#discussion_r1602203595) related to some important accountability tests that were lost while refactoring.

Additionally, I think that we need a better way to check that misbehaving nodes get disconnected at p2p level. A lot of our DoS defence relies on this. I left a comment [here](https://github.com/autonity/autonity/issues/873#issuecomment-2113411585) to track this, it can be done in a separate PR.

.github/workflows/go.yml
```diff
@@ -156,7 +154,7 @@ jobs:
             ${{ env.GO_CACHE }}
             ${{ env.GO_MODULES_CACHE }}
           key: ${{ runner.os }}-gocache-${{ env.GO_VERSION }}${{ hashFiles('go.mod', 'go.sum') }}
-      - run: go test -v -run ${{ matrix.tests }} ./e2e_test/byzantine/...
+      - run: go test -v -run ${{ matrix.tests }} ./e2e_test/byzantine/... -timeout 30m
```
vpiyush, line 157: Isn't the 30 min timeout too high for this considering we have whole ci running under 10m now

vpiyush, line 157: Also what do you thing about adding a race flag to all of these tests. I think it's of value to do that.

consensus/tendermint/accountability/fault_detector_test.go
```diff
@@ -498,760 +499,1001 @@ func TestRuleEngine(t *testing.T) {
...
-		// If one send a maliciousProposal for a new V, then all preCommits for previous rounds from this sender are nil.
-		ctrl := gomock.NewController(t)
-		defer ctrl.Finish()
-		chainMock := NewMockChainContext(ctrl)
-		var blockSub event.Subscription
-		chainMock.EXPECT().SubscribeChainEvent(gomock.Any()).AnyTimes().Return(blockSub)
-		chainMock.EXPECT().Config().AnyTimes().Return(&params.ChainConfig{ChainID: common.Big1})
-		accountability, _ := autonity.NewAccountability(proposer, backends.NewSimulatedBackend(ccore.GenesisAlloc{proposer: {Balance: big.NewInt(params.Ether)}}, 10000000))
+func TestNewProposalAccountabilityCheck(t *testing.T) {
+	height := uint64(0)
```
vpiyush, line 505: It's worth adding some comments in the test case to indicate what we are doing

consensus/tendermint/accountability/fault_detector_test.go
```diff
@@ -498,760 +499,1001 @@ func TestRuleEngine(t *testing.T) {
...
-		// simulate there were quorum preVotes for initProposal at init round 0, and save them.
-		for i := 0; i < len(committee); i++ {
-			preVote := message.NewPrevote(0, height, initProposal.Value(), makeSigner(keys[i], committee[i])).MustVerify(stubVerifier)
-			fd.msgStore.Save(preVote)
+	newProposal1 := newProposalMessage(height, 5, -1, signer, committee, nil).MustVerify(stubVerifier)
+	nilPrecommit1 := message.NewPrecommit(3, height, common.Hash{}, signer).MustVerify(stubVerifier)
+
+	newProposal0E := newProposalMessage(height, 3, 1, signer, committee, nil).MustVerify(stubVerifier)
+
+	t.Run("misbehaviour when pi has sent a non-nil precommit in a previous round", func(t *testing.T) {
```
vpiyush, line 516: what's pi, It will be helpful to use non-abbreviated terms in messages.

e2e_test/byzantine/accountability_network_test.go
```diff
@@ -303,18 +324,13 @@ func runDropPeerConnectionTest(t *testing.T, handler *interfaces.Services, testP
...
 	require.NoError(t, err)
-	defer network.Shutdown()
+	defer network.Shutdown(t)
 
 	// network should be up and continue to mine blocks
-	err = network.WaitToMineNBlocks(testPeriod, 20, false)
+	err = network.WaitToMineNBlocks(testPeriod, numSec, false)
 	require.NoError(t, err)
 
-	// the challenger should get no peer connection left.
```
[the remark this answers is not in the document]
vpiyush: Here is my understanding of this: 
IMO We don't have a logic to wait for 30 second for the peer we disconnected. We do have inbound connection attempt logic but that's for DDOS of connection attempts . Also we have dial checks which are done every 5 seconds, these are also for dial attempts. In this test case what could happen is we disconnect the peer and either we wait for less than 5 seconds if last dial was not expired OR we immediately dial again. and get connected back.
So, I would think we need to add Jail time here. whenever we disconnect a peer which should ensure 30 second (Or Whatever jailing time we choose) delay in next successful connection.
In conlusion - in this test we should bring back the client connection count logic and rely on server jailing for peer to remain disconnected.

e2e_test/byzantine/garbage_and_fuzz_test.go
```diff
@@ -0,0 +1,282 @@
...
+	*core.Core
+	interfaces.Precommiter
+}
+
+func (c *fuzzPrecommitSender) SendPrecommit(_ context.Context, isNil bool) {
+	var precommit *message.Precommit
+	r := rand.Int63()
+	h := rand.Uint64()
+	if isNil {
+		precommit = message.NewPrecommit(r, h, common.Hash{}, c.Backend().Sign)
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > These tests do not fuzz the signature field. Is that covered in some other test?

I will add a todo to track the missing fuzz of signature field in consensus messages.

e2e_test/byzantine/accountability_network_test.go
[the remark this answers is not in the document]
vpiyush: > 1. The disconnection doesn't happen if the malicious sender is locally connected (i.e. has a LAN address such as 127.0.0.0)

Also, I don't see this. We do disconnect irrespetive of whether we are connected on loopback. Though this is applicable to inbound connection attempts.

consensus/tendermint/core/upon_condition_test.go
```diff
@@ -28,593 +28,431 @@ const minSize, maxSize = 4, 100
...
-		if currentRound > 0 {
-			core.height = proposalHeight
-		}
-		core.pendingCandidateBlocks[proposalHeight.Uint64()] = proposal.Block()
-
-		if currentRound == 0 {
+		backendMock.EXPECT().Sign(gomock.Any()).DoAndReturn(e.clientSigner)
+		backendMock.EXPECT().SetProposedBlockHash(proposal.Block().Hash())
+		backendMock.EXPECT().Broadcast(e.committee.Committee(), proposal)
+		if e.curRound == 0 {
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > I think we don't need this if condition since the `curRound` is not randomized anymore in this test?

Yes, the default value is 0 for curRound unless we set a random one. In this context, we can remove the checker now.

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: > I think here the power we levels we were using are different wrt to the new refactored version. Is this test still correct?

Hi Lorenzo, could you be more specific? Do you the new version with msg aggregation?

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: > same here

I guess you mean the voting power is resloved by the preValidate() and Validate(), thus we'd need a new way to level the voting power in our bls-aggregation branch.

autonity/solidity/ape-config.yaml
```diff
@@ -2,6 +2,9 @@
 
 contracts_folder: contracts
 
+solidity:
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > is this branch not rebased correctly, these changes were also part of other PR??

It haven't been rebased yet.

e2e_test/byzantine/accountability_network_test.go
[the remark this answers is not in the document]
lorenzo-dev1: we do disconnect but then the inbound check is skipped: https://github.com/autonity/autonity/blob/develop/p2p/server.go#L1057

So basically the malicious peer can re-connect right away

consensus/tendermint/core/upon_condition_test.go
lorenzo-dev1: No, I mean the new power is F while in the old code (before your changes) it was F-1

e2e_test/byzantine/accountability_network_test.go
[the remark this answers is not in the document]
lorenzo-dev1: In any case yes I think we need a more reliable disconnection mechanism through jailing, it should be a high priority feature imo.

consensus/tendermint/accountability/fault_detector_test.go
Jason-Zhangxin-Chen: `pi` in the context stands for a consensus participant with an unique identity `i`, it is adopt from the tendermint paper and the accountability paper work. I will add this in the context explaination comment in rule engine.

consensus/tendermint/accountability/innocence_proof_handler.go
```diff
@@ -261,6 +261,7 @@ func (fd *FaultDetector) handleOffChainProofOfInnocence(proof *Proof, sender com
 	}
 	// the proof is valid, withdraw the off chain challenge.
 	fd.removeOffChainAccusation(proof)
+	fd.logger.Info("Withdraw accusation", "rule", proof.Rule, "suspected account", proof.Message.Sender())
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: resolved in a9708ffe0daf2adb42668b9b066df55d82d43ca0

e2e_test/byzantine/accountability_network_test.go
```diff
@@ -116,16 +117,32 @@ func (s *C1OffChainAccusation) Broadcast(msg message.Msg) {
...
 
-func (s *GarbageOffChainAccusation) Broadcast(msg message.Msg) {
+func fuzzedMessages() []message.Msg {
+	num := rand.Intn(10) + 1
+	var msgs []message.Msg
+	f := fuzz.New()
+	for i := 0; i < num; i++ {
+		m := &message.Fake{}
+		f.Fuzz(m)
+		// since our RLP encoding of accusation msg does not allow a full proposal, thus we skip the proposal code.
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: resolved in a9708ffe0daf2adb42668b9b066df55d82d43ca0 by adding todo to track it.

e2e_test/byzantine/accountability_network_test.go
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: resolved in a9708ffe0daf2adb42668b9b066df55d82d43ca0 by adding todo to track this, in that todo, I suggested to create a framework to test the peer drop/reconection events, for the time being, we don't have such framework yet.

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: Oh, I see. The new voting power leveled is correct with the new refacted test framework, we have to make sure the total power of a new round is over F().

consensus/tendermint/core/upon_condition_test.go
Jason-Zhangxin-Chen: See my last reply.

.github/workflows/go.yml
Jason-Zhangxin-Chen: Resolved in a9708ffe0daf2adb42668b9b066df55d82d43ca0

Jason-Zhangxin-Chen: Resolved in a9708ffe0daf2adb42668b9b066df55d82d43ca0

consensus/tendermint/accountability/fault_detector_test.go
Jason-Zhangxin-Chen: Resolved in a9708ffe0daf2adb42668b9b066df55d82d43ca0, by adding some comments.

consensus/tendermint/core/message/message.go
```diff
@@ -590,7 +590,7 @@ func PrepareCommittedSeal(hash common.Hash, round int64, height *big.Int) common
 // Fake is a dummy object used for internal testing.
 type Fake struct {
 	FakeCode      uint8
-	FakeRound     int64
+	FakeRound     uint64
```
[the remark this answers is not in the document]
lorenzo-dev1: yeah, actually I realize now that this is `round`, not `validRound`. So it is actually correct for it to be uint

e2e_test/byzantine/accountability_network_test.go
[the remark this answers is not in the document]
vpiyush: > we do disconnect but then the inbound check is skipped: https://github.com/autonity/autonity/blob/develop/p2p/server.go#L1057
> 
> So basically the malicious peer can re-connect right away

Inbound check is only for connection attempts, so whether it get's skipped or not. For a peer who was connected to us more than 30 seconds(or we will dial to it if it was connected for more than 5 seconds) will always get connected even if not on LAN.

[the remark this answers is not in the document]
lorenzo-dev1: opened issue to track this https://github.com/autonity/autonity/issues/983, it is important

lorenzo-dev1 (approved):

vpiyush (approved):

yazzaoui (approved):
I haven’t delved into the details of this PR, but at a glance, it looks good. We might need some refinements later—I have a few ideas. For now, it’s a solid improvement to our testing suite. LGTM.

PR #957, opened 2024-03-11
BLS aggregation

lorenzo-dev1:
Replace ECDSA signatures with BLS signatures (aggregatable)


core/types/bft.go
```diff
@@ -103,27 +104,21 @@ func WriteRound(h *Header, round int64) error {
...
-		if len(seal) != BFTExtraSeal {
-			return ErrInvalidCommittedSeals
-		}
-	}
-	h.CommittedSeals = make([][]byte, len(committedSeals))
-	for i, val := range committedSeals {
-		h.CommittedSeals[i] = make([]byte, len(val))
-		copy(h.CommittedSeals[i], val)
+	h.CommittedSeals = make(Signatures)
+	for addr, val := range committedSeals {
```
Jason-Zhangxin-Chen, line 112: Rename val to seal?

core/types/block.go
```diff
@@ -96,19 +98,91 @@ type Header struct {
...
-	Round          uint64   `json:"round"               gencodec:"required"`
-	CommittedSeals [][]byte `json:"committedSeals"      gencodec:"required"`
+	ProposerSeal   []byte     `json:"proposerSeal"        gencodec:"required"`
+	Round          uint64     `json:"round"               gencodec:"required"`
+	CommittedSeals Signatures `json:"committedSeals"      gencodec:"required"`
 }
 
+type Signatures map[common.Address]blst.Signature
+
+// serialize the map as an bytes key1|value1|key2|value2|...
```
Jason-Zhangxin-Chen, line 108: Looping the underlying map to serialize the data may lead to un-deterministic bytes output, if we can add a comment to warn the caller of it to avoid usage mistake?

Jason-Zhangxin-Chen (approved):
LGTM, just left minor comments on logging and comments.

core/types/block.go
```diff
@@ -96,19 +98,91 @@ type Header struct {
...
+	ConsensusKeyBytes hexutil.Bytes
 }
 
 type Committee []CommitteeMember
 
+func (c Committee) DeserializeConsensusKeys() error {
+	for i := range c {
+		consensusKey, err := blst.PublicKeyFromBytes(c[i].ConsensusKeyBytes)
+		if err != nil {
+			return fmt.Errorf("Error when decoding bls key with index %d, err: %w", i, err)
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/957/commits/800f6cbb7febe5bf0ce444619cea609b95760ae7

autonity/calls.go
```diff
@@ -389,9 +389,16 @@ func (c *AutonityContract) callGetCommitteeEnodes(state vm.StateDB, header *type
...
-	var committee []types.CommitteeMember
-	err := c.AutonityContractCall(state, header, "getCommittee", &committee)
-	return committee, err
+	var committee types.Committee
+	if err := c.AutonityContractCall(state, header, "getCommittee", &committee); err != nil {
+		return nil, err
+	}
+
+	if err := committee.DeserializeConsensusKeys(); err != nil {
+		panic("Committee member has invalid consensus key")
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/957/commits/800f6cbb7febe5bf0ce444619cea609b95760ae7

autonity/calls.go
```diff
@@ -409,6 +416,11 @@ func (c *AutonityContract) callFinalize(state vm.StateDB, header *types.Header)
 	if err := c.AutonityContractCall(state, header, "finalize", &[]any{&updateReady, &committee}); err != nil {
 		return false, nil, err
 	}
+
+	if err := committee.DeserializeConsensusKeys(); err != nil {
+		panic("Committee member has invalid consensus key")
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/957/commits/800f6cbb7febe5bf0ce444619cea609b95760ae7

consensus/tendermint/backend/engine.go
```diff
@@ -241,20 +240,19 @@ func (sb *Backend) verifyCommittedSeals(header, parent *types.Header) error {
...
+	for addr, signedSeal := range header.CommittedSeals {
 		member := parent.CommitteeMember(addr)
 		if member == nil {
 			sb.logger.Error(fmt.Sprintf("block had seal from non committee member %q", addr))
 			return types.ErrInvalidCommittedSeals
 		}
 
+		valid := signedSeal.Verify(member.ConsensusKey, headerSeal[:])
+		if !valid {
+			sb.logger.Error("block had invalid committed seal")
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/957/commits/800f6cbb7febe5bf0ce444619cea609b95760ae7

core/types/bft.go
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/957/commits/800f6cbb7febe5bf0ce444619cea609b95760ae7

core/types/block.go
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/957/commits/800f6cbb7febe5bf0ce444619cea609b95760ae7

consensus/tendermint/accountability/contracts.go
```diff
@@ -279,14 +280,14 @@ func (c *MisbehaviourVerifier) validMisbehaviourOfPO(p *Proof) bool {
 	switch vote := p.Evidences[0].(type) {
 	case *message.Precommit:
 		if vote.R() == proposal.ValidRound() &&
-			vote.Sender() == p.Message.Sender() &&
+			//	vote.Sender() == p.Message.Sender() && //TODO(lorenzo) fix this
```
vpiyush, line 283: this needs to be addressed

consensus/tendermint/accountability/contracts.go
```diff
@@ -279,14 +280,14 @@ func (c *MisbehaviourVerifier) validMisbehaviourOfPO(p *Proof) bool {
...
 			vote.Value() != nilValue &&
 			vote.Value() != proposal.Value() {
 			return true
 		}
 		if vote.R() > proposal.ValidRound() &&
 			vote.R() < proposal.R() &&
-			vote.Sender() == p.Message.Sender() &&
+			//vote.Sender() == p.Message.Sender() && //TODO(lorenzo) fix this
```
vpiyush, line 290: same here - address this

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	MessageBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/message", nil)     // time it takes to process a single message as received by backend
+	RoundBg       = metrics.NewRegisteredBufferedGauge("aggregator/bg/round", nil)       // time it takes to process a round change event
+	PowerBg       = metrics.NewRegisteredBufferedGauge("aggregator/bg/power", nil)       // time it takes to process a power change event
+	FuturePowerBg = metrics.NewRegisteredBufferedGauge("aggregator/bg/futurepower", nil) // time it takes to process a future power change event
+	ProcessBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/process", nil)     // time it takes to process batches of messages
+	BatchesBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/batches", nil)     // size of batches (aggregated together with a single fastAggregateVerify)
+	InvalidBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/invalid", nil)     // number of invalid sigs
+)
+
+type eventerFn func(msg message.Msg, errCh chan<- error) interface{}
```
vpiyush, line 39: a messageDispatcher OR eventEmitter should make better naming sense here

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+
+// function to create the event for current height messages (they get picked up by Core and by the FD)
+func currentHeightEventer(msg message.Msg, errCh chan<- error) interface{} {
```
vpiyush, line 42: accordingly adjust this name

consensus/tendermint/backend/aggregator.go
vpiyush, line 39: use a generic here as return type(with constraint of oldMessage | MessageEvent) for better performance and the readability. also makes it possible to catch code errors at compile time

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+}
+
+type aggregator struct {
+	backend interfaces.Backend
+	core    interfaces.Core
+	// old height messages
+	oldMessages map[common.Hash][]events.UnverifiedMessageEvent
+	// TODO(lorenzo) might be worth to re-use the set logic that is used in Core (without locks)
+	// map[height][round][code][value][]msgs
+	messages  map[uint64]map[int64]map[uint8]map[common.Hash][]events.UnverifiedMessageEvent
```
vpiyush, line 76: Though there may be better alternatives here, but for starters a struct based approach would be cleaner and improve the code readability. for eg. type MessageSet struct {
Messages map[common.Hash][]events.UnverifiedMessageEvent
}

type CodeMessageSets struct {
CodeSets [3]*MessageSet
}

type RoundMessageSets struct {
RoundSets [maxRounds]*CodeMessageSets
}

type HeightMessageSets struct {
HeightSets [maxHeights]*RoundMessageSets
}
This also avoids the lock on the top level. and hopefully reduces the memory allocations

vpiyush: this should also allow for better encapsulation of logic if needed

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	votesFrom map[common.Address][]common.Hash
+	toIgnore  map[common.Hash]struct{}
+	// TODO(lorenzo) can one sub starve the other?
```
vpiyush, line 79: no, they don't

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	sub     *event.TypeMuxSubscription // backend events
```
vpiyush, line 80: needs a better name, perhaps: backendEventSub  OR UnverfiedMessageSub

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	coreSub *event.TypeMuxSubscription // core events
```
vpiyush, line 81: could be renamed as coreEventSub

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	cancel  context.CancelFunc
+	wg      sync.WaitGroup
+	logger  log.Logger
+}
+
+func (a *aggregator) start(ctx context.Context) {
+	a.logger.Info("Starting the aggregator routine")
+	a.sub = a.backend.Subscribe(events.UnverifiedMessageEvent{})
```
vpiyush, line 89: these should also be unsubscribed on aggregator stop.

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	a.coreSub = a.backend.Subscribe(events.RoundChangeEvent{}, events.PowerChangeEvent{}, events.FuturePowerChangeEvent{})
```
vpiyush, line 90: same

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+type aggregator struct {
+	backend interfaces.Backend
+	core    interfaces.Core
+	// old height messages
+	oldMessages map[common.Hash][]events.UnverifiedMessageEvent
+	// TODO(lorenzo) might be worth to re-use the set logic that is used in Core (without locks)
+	// map[height][round][code][value][]msgs
+	messages  map[uint64]map[int64]map[uint8]map[common.Hash][]events.UnverifiedMessageEvent
+	votesFrom map[common.Address][]common.Hash
+	toIgnore  map[common.Hash]struct{}
```
vpiyush, line 78: This could be avoided if we utilize the votesFrom itself by extending it to contain a validity Flag in the stored struct.

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+	for _, batch := range batches {
+		for _, e := range batch {
+			messages = append(messages, e.Message)
+		}
+	}
+
+	return message.Power(messages)
+}
+
+func (a *aggregator) batches(h uint64, r int64, codeAndValue []interface{}) [][]events.UnverifiedMessageEvent {
```
vpiyush, line 179: this seems  complicated, These alternatives could be better:
- Using a query struct:  
type MessageQuery struct {
Code  *uint8
Value *common.Hash
} 
So that we could set in it with what we want to filter and organize the logic accordingly.
- Or use a different function altogether for e.g. - ByCode, ByCodeAndValue and All. Then pass these as function objects to power.

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,637 @@
...
+		Message: msg,
+		ErrCh:   errCh,
+	}
+}
+
+func newAggregator(backend interfaces.Backend, core interfaces.Core, logger log.Logger) *aggregator {
+	return &aggregator{
+		backend:     backend,
+		core:        core,
+		oldMessages: make(map[common.Hash][]events.UnverifiedMessageEvent),
```
vpiyush, line 61: We should think about initializing these maps with a minimum capacity, to avoid frequent map resizing which is a costly operation

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: This function just creates the event, it does not send it anywhere. So if you prefer another name I would go for something like `eventBuilder`

consensus/tendermint/core/message/message.go
```diff
@@ -490,93 +808,31 @@ func (p *Precommit) DecodeRLP(s *rlp.Stream) error {
...
+	if encoded.Height == 0 {
 		return constants.ErrInvalidMessage
 	}
-	p.signature = encoded.Signature
 	if encoded.Round > constants.MaxRound {
 		return constants.ErrInvalidMessage
 	}
-	p.round = int64(encoded.Round)
-	if p.round < 0 {
+	if encoded.Senders == nil || encoded.Senders.Bits == nil || len(encoded.Senders.Bits) == 0 {
```
yazzaoui, line 822: A check on `encoded.Senders.Coefficients != nil` is not needed here?

core/types/senders_info.go
```diff
@@ -0,0 +1,414 @@
...
+		powers:        make(map[int]*big.Int),
+		power:         new(big.Int),
+	}
+}
+
+type validatorBitmap []byte
+
+func NewValidatorBitmap(committeeSize int) validatorBitmap {
+	bitLength := committeeSize * bitsPerValidator
+	byteLength := int(math.Ceil(float64(bitLength) / float64(bitsInByte)))
```
yazzaoui, line 72: `byteLength := (committeeSize * bitsPerValidator + bitsInByte - 1) / bitsInByte`

consensus/tendermint/core/message/message.go
```diff
@@ -350,12 +389,67 @@ type extVote struct {
...
+	power := new(big.Int)
+
+	for i, index := range indexes {
+		member := header.Committee[index]
+
+		keys[i] = member.ConsensusKeyBytes
+
+		_, alreadyPresent := powers[index]
+		if !alreadyPresent {
+			powers[index] = new(big.Int).Set(member.VotingPower)
```
yazzaoui, line 429: This will put a lot of pressure on the GC. It's probably safe to just use a reference as `member.VotingPower` should not be modified.

core/types/senders_info.go
```diff
@@ -0,0 +1,414 @@
...
+	// gather data about senders bits
+	countNonZero := 0
+	countLong := 0
+	sum := 0
+	for i := 0; i < committeeSize; i++ {
+		value := s.Bits.Get(i)
+		if value > 0 { // 01 10 11
+			countNonZero++
+		}
+		if value == 3 { // 11
```
yazzaoui, line 129: Please define a const for those special values

core/types/block.go
```diff
@@ -96,19 +98,59 @@ type Header struct {
...
 
-	ProposerSeal   []byte   `json:"proposerSeal"        gencodec:"required"`
-	Round          uint64   `json:"round"               gencodec:"required"`
-	CommittedSeals [][]byte `json:"committedSeals"      gencodec:"required"`
+	ProposerSeal   []byte             `json:"proposerSeal"        gencodec:"required"`
+	Round          uint64             `json:"round"               gencodec:"required"`
+	CommittedSeals AggregateSignature `json:"committedSeals"      gencodec:"required"`
 }
 
+type AggregateSignature struct {
```
yazzaoui, line 106: Suggest renaming to `QuorumCertificate`

core/types/senders_info.go
```diff
@@ -0,0 +1,414 @@
...
+* 11 --> look at the number of signatures in the `Coefficients` array
+*
+* example:
+* indexes 			 0  1  2  3  4  5
+* bits    			 00 11 10 01 00 11
+* coefficients   17 170
+* #sigs          0  17 2  1  0  170
+ */
+
+type SendersInfo struct {
```
yazzaoui, line 45: Why not
```suggestion
type Signers struct {
```

consensus/tendermint/core/message/message.go
```diff
@@ -350,12 +389,67 @@ type extVote struct {
...
+			power.Add(power, member.VotingPower)
+		}
+	}
+	aggregatedKey, err := blst.AggregatePublicKeys(keys)
+	if err != nil {
+		panic("Error while aggregating keys from committee: " + err.Error())
+	}
+
+	v.senders.SetPowers(powers)
+	v.senders.SetPower(power)
```
yazzaoui, line 439: Code smell: Too many mutations of the object `senders` is required to get one valid after construction. This is the kind of thing that bites back later down the road when wanting to refactor or test.
Can a valid `senders` object exist without those power field assigned? If not the case then try to do it a construction time.

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: I refactored the way msgs are stored

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: why?

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: done

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: done

consensus/tendermint/core/message/message.go
lorenzo-dev1: I generally assume the power fields to be assigned in the code.

The power assignments could easily be done at conthstruction time for senders object created locally, however it is trickier for the ones received from the wire. In that case I cannot access the `header` object when decoding from RLP, thus why I decided to go for a separate `PreValidate` function to be called after decoding. Any suggestions?

core/types/block.go
lorenzo-dev1: What about leaving the type as `AggregateSignature` (since that is what a quorum certificate is) and instead changing the *name* of the field from `committedSeals` to `quorumCertificate`?

consensus/tendermint/core/message/message.go
lorenzo-dev1: @yazzaoui Thinking again about this what about adding a `preverified` flag that we can then check at runtime to make sure we do not use un-initialized sendersinfo/msgs ?

core/types/block.go
yazzaoui: Sounds good as well

lorenzo-dev1: done in https://github.com/autonity/autonity/pull/957/commits/5245a321cfda17d2156fbb65547016e80d56e223

@cmjc this might need some updates on the docs and on the aut tool. In general every tool fetching from RPC will be broken because:
1. `committedSeals` are now renamed to `quorumCertificate`
2. the format of the data is different (a single aggregated BLS signature instead of a list of ECDSA sigs)

core/types/senders_info.go
lorenzo-dev1: done in https://github.com/autonity/autonity/commit/5245a321cfda17d2156fbb65547016e80d56e223

core/types/signers.go
```diff
@@ -0,0 +1,465 @@
...
+	shift := validatorsPerByte - 1 - bitIndex
+	valueShifted := value << (shift * bitsPerValidator)
+
+	byteIndex := validatorIndex / validatorsPerByte
+	vb[byteIndex] = vb[byteIndex] & setMasks[bitIndex]
+	vb[byteIndex] = vb[byteIndex] | valueShifted
+}
+
+// validates the sender info, used to ensure received aggregates have correctly sized buffers
+func (s *Signers) Valid(committeeSize int) error {
```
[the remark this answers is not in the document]
lorenzo-dev1: done in https://github.com/autonity/autonity/pull/957/commits/57911f4a07ed0c9faf231c407da70bad10aef4d3

core/types/senders_info.go
lorenzo-dev1: I renamed all the instances of `sender` with `signer` because it is more appropriate. Now the term `sender` is only used for `p2pSender`

core/types/senders_info.go
lorenzo-dev1: done

consensus/tendermint/core/message/message.go
lorenzo-dev1: done

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,856 @@
...
+
+const (
+	aggregationPeriod            = 100 * time.Millisecond
+	oldMessagesAggregationPeriod = 2 * time.Second
+)
+
+// aggregator metrics
+var (
+	// metrics for different aggregator flows
+	MessageBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/message", nil, nil)     // time it takes to process a single message as received by backend
```
vpiyush, line 29: you should use higher capacity for this gauge. could be 1K

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,856 @@
...
+	PowerBg       = metrics.NewRegisteredBufferedGauge("aggregator/bg/power", nil, nil)       // time it takes to process a power change event
+	FuturePowerBg = metrics.NewRegisteredBufferedGauge("aggregator/bg/futurepower", nil, nil) // time it takes to process a future power change event
+	BatchesBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/batches", nil, nil)     // size of batches (aggregated together with a single fastAggregateVerify)
+	InvalidBg     = metrics.NewRegisteredBufferedGauge("aggregator/bg/invalid", nil, nil)     // number of invalid sigs
+)
+
+type eventBuilder func(msg message.Msg, errCh chan<- error) interface{}
+
+// function to create the event for current height messages (they get picked up by Core and by the FD)
+func currentHeightEventBuilder(msg message.Msg, errCh chan<- error) interface{} {
```
vpiyush, line 40: Minor:  Height builder could be merged into one. If you prefer:
type EventType interface {
	events.MessageEvent | events.OldMessageEvent
}

func HeightEventBuilder[T EventType](msg message.Msg, errCh chan<- error) interface{} {
	return T{
		Message: msg,
		ErrCh:   errCh,
	}
}

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,856 @@
...
+	if corePower.Cmp(quorum) < 0 && corePower.Add(corePower, aggregatorPower).Cmp(quorum) >= 0 {
+		a.processVotes(height, round, code)
+	}
+}
+
+// checks if a message is already in core
+// TODO(lorenzo) performance, do something more efficient without iterating over all messages of Core
+// Depending on how implemented, it might be possible to remove core.futureRoundLock
+func (a *aggregator) alreadyProcessed(msg message.Msg) bool {
+	for _, m := range a.core.CurrentHeightMessages() {
```
vpiyush, line 609: The current implementation collects all messges into a new allocated slice and iterates over it, the new slice is un-necessary. We should be able to directly check the msg existence while we iterated over to messages.

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,856 @@
...
+			a.votesFrom[p2pSender] = append(a.votesFrom[p2pSender], msg.Hash())
+
+			// NOTE: Aggregator and Core run asynchronously. The code needs to take into account that Core can change state at any point here.
+			// This also implies that height checks still needs to be done in Core.
+			coreHeight := a.core.Height().Uint64()
+			if msg.H() < coreHeight {
+				a.logger.Debug("Storing old height message in the aggregator", "msgHeight", msg.H(), "coreHeight", coreHeight)
+				signatureInput := msg.SignatureInput()
+				a.staleMessages[signatureInput] = append(a.staleMessages[signatureInput], event)
+				MessageBg.Add(time.Since(start).Nanoseconds())
```
vpiyush, line 656: These readings of MessageBg will be quite small, than the ones who are actually getting processed in aggregator. To get a better view of time spent in aggregator we should only consider to record the processing time which is being done after these special conditions are checked. Or if we also want to capture this perhaps we can have separate metrics for the normal case

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,856 @@
...
+				break
+			}
+			if msg.H() > coreHeight {
```
vpiyush, line 659: This shoud be removed because this is by design.

consensus/tendermint/core/handler.go
```diff
@@ -133,45 +157,76 @@ eventLoop:
...
 					if shouldDisconnectSender(err) {
 						tryDisconnect(e.ErrCh, err)
 					}
-					continue
+					break
 				}
+
 				if !c.noGossip {
-					c.backend.Gossip(c.CommitteeSet().Committee(), e.Message)
+					if !hadQuorum {
```
vpiyush, line 181: So if we already have quorum, shouldn't we still broadcast the complex aggregate. Is there any advantage of broadcasing the votes when we already have quorum?

consensus/tendermint/core/handler.go
```diff
@@ -133,45 +157,76 @@ eventLoop:
...
 			switch e := ev.Data.(type) {
 			case events.MessageEvent:
+				msg := e.Message
 
-				// At this stage, a message is parsed and all the internal fields must be accessible
-				if err := c.handleMsg(ctx, e.Message); err != nil {
+				var hadQuorum bool
+				if !c.noGossip {
+					// check if we have quorum for message type for this round
+					hadQuorum = c.quorumFor(msg.Code(), msg.R(), msg.Value())
```
vpiyush, line 168: Once we have the quorum for value this can be saved so we don't have compute for next message processing

consensus/tendermint/core/message/message.go
```diff
@@ -393,90 +480,311 @@ func (p *Precommit) Value() common.Hash {
...
 }
 
-func NewPrecommit(r int64, h uint64, value common.Hash, signer Signer) *Precommit {
-	return newVote[Precommit](r, h, value, signer)
+// NOTE: this function assumes that:
+// 1. all votes are for the same signature input (code,h,r,value)
+// 2. all votes have previously been preverified and cryptographically verified
+func AggregateVotes[
+	E Prevote | Precommit,
+	PE interface {
```
vpiyush, line 554: There is no use of PE in this function, this can be removed

core/types/signers.go
```diff
@@ -0,0 +1,478 @@
...
+		// we are adding a new signer, update the length cache
+		s.length++
+	case oneSignature:
+		value = twoSignatures // 10
+	case twoSignatures:
+		value = multipleSignatures // 11
+		// add a new uint16 into the Coefficients array
+		count := 0
+		for i := 0; i < index; i++ {
+			if s.Bits.Get(i) == byte(multipleSignatures) {
```
vpiyush, line 277: typecasting is redundant here

core/types/signers.go
```diff
@@ -0,0 +1,479 @@
...
+	shift := validatorsPerByte - 1 - bitIndex
+	valueShifted := value << (shift * bitsPerValidator)
+
+	byteIndex := validatorIndex / validatorsPerByte
+	vb[byteIndex] = vb[byteIndex] & setMasks[bitIndex]
+	vb[byteIndex] = vb[byteIndex] | valueShifted
+}
+
+// validates the sender info, used to ensure received aggregates have correctly sized buffers
+func (s *Signers) Validate(committeeSize int) error {
```
vpiyush, line 125: It's a general observation: All function comments should start with the function name. It helps for any tool to parse the comments as Docs. Also, IDEs flag them as warning otherwise, which is annoying,

vpiyush: e.g.: this comment would be modified as following: 
// Validate validates the sender info, used to ensure received aggregates have correctly sized buffers

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,850 @@
...
+		return
+	}
+
+	// we are processing an individual vote or a simple aggregate vote
+	if add {
+		a.saveMessage(voteEvent)
+	}
+
+	// check if we reached quorum voting power on a specific value
+	corePower := a.core.VotesPowerFor(height, round, code, value)
```
vpiyush, line 587: Why do we compute this again in the same flow? we computed this at line 570

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: Yeah but then we would have to loop through all the hashes to verify if we have to ignore a message I think?

consensus/tendermint/core/message/message.go
lorenzo-dev1: Theoretically no because rlp decode should not leave it nil, but I am going to add it for good measure

consensus/tendermint/backend/aggregator.go
```diff
@@ -0,0 +1,856 @@
...
+		case ev, ok := <-a.backendSub.Chan():
+			start := time.Now()
+			if !ok {
+				break loop
+			}
+			event := ev.Data.(events.UnverifiedMessageEvent)
+			msg := event.Message
+			p2pSender := event.P2pSender
+
+			a.votesFrom[p2pSender] = append(a.votesFrom[p2pSender], msg.Hash())
```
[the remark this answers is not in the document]
lorenzo-dev1: right! renamed

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: I would keep it for the moment just to see if there are any bugs. Then when we are sure that everything works as expected we can indeed remove it

consensus/tendermint/core/message/message.go
```diff
@@ -393,90 +480,311 @@ func (p *Precommit) Value() common.Hash {
...
+	// compute new aggregated signature and related signers information
+	var signatures []blst.Signature
+	var publicKeys [][]byte
+	for _, vote := range votes {
+		// do not aggregate votes if they do not add any useful information
+		// e.g. signers contains already at least 1 signature for all signers of vote.Signers()
+		// we would just create and gossip new aggregates that would uselessly flood the network
+		// additionally, we also check if the resulting aggregate respects the coefficient boundaries.
+		// this avoids that we aggregate two complex aggregates together, which can lead to coefficient breaching.
+		if signers.AddsInformation(vote.Signers()) && signers.RespectsBoundaries(vote.Signers()) {
```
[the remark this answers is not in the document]
lorenzo-dev1: No these two functions are doing different check

core/types/signers.go
lorenzo-dev1: fixed

core/types/signers.go
lorenzo-dev1: I saw that when using golang, but tbh it does not make sense to me. If I put a comment on top of a function it is implied that it refers to that function. Not sure if it is something that has to do with GoDoc

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: Because things could have changed in the meantime as core is running in another go routine

consensus/tendermint/core/handler.go
lorenzo-dev1: yeah that's a good point and I am thinking about it too, my thinking was:
1. once we reach quorum and broadcasting, that should bring everyone else to quorum as well and we should move to next step
2. However I still broadcast the additional votes we get to make sure we are still compliant with the tendermint specification of the paper

lorenzo-dev1: I think we can optimize further but I want first to dig into the performance reports and understand precisely what is going on on a real network

consensus/tendermint/core/message/message.go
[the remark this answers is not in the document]
vpiyush: I meant the inner safetyCheck(). called by both the function

core/types/signers.go
vpiyush: It's a convention to write comments in a certain way. Though you are right that it is enforced in this manner because fo GoDoc but also other tools. Check here: https://go.dev/wiki/CodeReviewComments#comment-sentences

consensus/tendermint/core/message/message.go
```diff
@@ -393,90 +481,309 @@ func (p *Precommit) Value() common.Hash {
...
+				preverified:    true,
+				signerKey:      aggregatedPublicKey, // this is not strictly necessary since the vote is already verified
+			},
+		},
+	}
+	return &aggregateVote
+}
+
+// NOTE: these functions will aggregate votes as much as possible without creating complex aggregates
+func AggregatePrevotesSimple(votes []Vote) []*Prevote {
```
vpiyush, line 628: These functions seems un-necessary. can be removed

lorenzo-dev1: This function is actually very important, why do you say it is un-necessary?

It is this that allows us to reduce the number of msgs gossiped

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: Addressed in https://github.com/autonity/autonity/pull/993

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: addressed in https://github.com/autonity/autonity/pull/993

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: You mean initializing the map or the arrays in it? or both?

consensus/tendermint/backend/aggregator.go
lorenzo-dev1: I agree, however this func doesn't pop up in the cpu profile, so I will leave this for a later PR improvement

consensus/tendermint/core/handler.go
lorenzo-dev1: True, however it is just reading a cached value. I will add the cache of the quorum flag in a following PR

consensus/tendermint/core/message/message.go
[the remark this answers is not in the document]
lorenzo-dev1: Ok I see, however the safetyCheck is there to avoid misuse of those functions for the future as well. Maybe in the future we will use those function in separate places

consensus/tendermint/core/message/message.go
lorenzo-dev1: I see, however I followed the same pattern we used for `newVote` for example

yazzaoui (approved):
LGTM

PR #961, opened 2024-03-19
[ReadyForReview] Repair APE framework for ASM tests.

Jason-Zhangxin-Chen:
This PR present the repairing of APE framework for ASM tests. As current HardHat is not compatible with the latest version of solc, 0.8.25, thus we have to specify the solidity version in the ape-config file.


yazzaoui (approved):

vpiyush (approved):
LGTM

PR #963, opened 2024-03-20
Metrics update

vpiyush:


vpiyush:
depreciated in favour of #974 and #980

PR #966, opened 2024-03-22
WIP - bls aggregation

lorenzo-dev1:
First implementation.

There is still a lot of improvements I want to implement but the code is decent and all the main logic is there. You can start to have a look at it if you want.

Only big pieces missing:
1. the fault detector and accountability
2. improving the disconnection mechanism (current one is not sufficient, we want something more harsh for validators sending invalid bls sigs. At least 1 epoch disconnect)


consensus/tendermint/core/message/set.go
```diff
@@ -7,88 +7,104 @@ import (
...
-	if _, ok := s.votes[value]; !ok {
-		s.votes[value] = make(map[common.Address]T)
+	// aggregate previous votes and vote
+	//TODO(lorenzo) performance, verify that this doesn't create too much memory
+	switch vote.(type) {
+	case *Prevote:
+		aggregatedVotes := AggregatePrevotesSimple(append(previousVotes, vote))
+		s.votes[value] = make([]Vote, len(aggregatedVotes))
+		for i, aggregatedVote := range aggregatedVotes {
+			s.votes[value][i] = aggregatedVote
```
Jason-Zhangxin-Chen, line 44: Why the merge of previousVotes and the vote to be adding leads to multiple votes? In my understanding, they should be merged into a single one, since the votes are for the same value.

lorenzo-dev1: We only allow the creation of complex aggregates (with at least 1 coefficient > 1) if they carry quorum.

I.E. If we are trying to merge two aggregates containing an overlapping signature, whose total power is < quorum --> keep them separated as two simple aggregates.

Issue #812, opened 2023-08-02
Fetching RoundData from Oracle Contract for an invalid symbol


aiman:
There is an issue when calling the Oracle Contract functions to fetch the aggregated prices:

- `getRoundData(uint256 _round, string memory _symbol) external view returns (RoundData)`
- `latestRoundData(string memory _symbol) external view returns (RoundData)`

In both of these functions, the Price data is fetched from the `prices` mapping for the appropriate `round`

```solidity
// latestRoundData
Price memory _p = prices[round-1][_symbol];
RoundData memory _d = RoundData(round-1, _p.price, _p.timestamp, uint(_p.status));

// getRoundData
Price memory _p = prices[_round][_symbol];
RoundData memory _d = RoundData(_round, _p.price, _p.timestamp, uint(_p.status));
```

But this means if the symbol is not found in mapping, then the function will get back a zero-initialized `Price` struct:

```
enum PriceStatus {SUCCESS, FAILURE}

struct Price {
    int256 price;
    uint timestamp;
    PriceStatus status;
}
```

Note the zero value for `PriceStatus` is `PriceStatus.SUCCESS`, which is `0`. This value is propagated into the `RoundData.status` that is returned from these functions. As a result, querying a symbol that was not found will return a `RoundData` with a `RoundData.status == 0`, which is not correct. This may happen if the symbol is invalid, or no longer part of the symbols list.

It would be more robust if the `status` fields in `Price` and `RoundData` was a `bool` that takes the value `false` in case of failure. For example, the field could be named `ok`, `done`, `success`, `completed` to avoid any ambiguity.

An open question is whether the calls to these two functions should revert in case the symbol is invalid for that round, instead of returning a `RoundData` value.

PR #967, opened 2024-03-28
Replace the enum of PriceStatus with boolean in Oracle.

Jason-Zhangxin-Chen:
To close https://github.com/autonity/autonity/issues/812.


autonity/solidity/contracts/Oracle.sol
```diff
@@ -38,8 +37,7 @@ contract Oracle is IOracle {
     struct Price {
         int256 price;
         uint timestamp;
-        PriceStatus status; // Do we get back a status code if couldn't compute last
-        // price in time ?
+        bool success;
```
vpiyush, line 40: the variable name "success" somehow indicates already a success, I think the variable name should be changed here. A possible alternative could be "valid" or "isValid"

vpiyush (approved):
LGTM, except a minor comment about the variable name.

PR #970, opened 2024-04-17
Adds an option to disable gossip from command line

vpiyush:
this is to help in debugging and to be easily able to switch gossip on/off from command line.
new option:  --nogossip

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -595,6 +595,10 @@ var (
 		Usage: "Allow for unprotected (non EIP155 signed) transactions to be submitted via RPC",
 	}
 
+	NoGossip = cli.BoolFlag{
+		Name:  "nogossip",
+		Usage: "disable consensensus message gossip",
+	}
 	//Consensus Network settings
 	ConsensusListenPortFlag = cli.IntFlag{
 		Name:  "consensus.port",
@@ -1230,10 +1234,12 @@ func SetNodeConfig(ctx *cli.Context, cfg *node.Config) {
 	setDataDir(ctx, cfg)
 	setSmartCard(ctx, cfg)
 
+	if ctx.GlobalIsSet(NoGossip.Name) {
+		cfg.NoGossip = ctx.GlobalBool(NoGossip.Name)
+	}
 	if ctx.GlobalIsSet(ExternalSignerFlag.Name) {
 		cfg.ExternalSigner = ctx.GlobalString(ExternalSignerFlag.Name)
 	}
-
 	if ctx.GlobalIsSet(KeyStoreDirFlag.Name) {
 		cfg.KeyStoreDir = ctx.GlobalString(KeyStoreDirFlag.Name)
 	}
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -597,7 +597,7 @@ var (
 
 	NoGossip = cli.BoolFlag{
 		Name:  "nogossip",
-		Usage: "disable consensensus message gossip",
+		Usage: "disable consensus message gossip",
 	}
 	//Consensus Network settings
 	ConsensusListenPortFlag = cli.IntFlag{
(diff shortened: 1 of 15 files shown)


Jason-Zhangxin-Chen (approved):
LGTM.

cmd/utils/flags.go
```diff
@@ -595,6 +595,10 @@ var (
 		Usage: "Allow for unprotected (non EIP155 signed) transactions to be submitted via RPC",
 	}
 
+	NoGossip = cli.BoolFlag{
+		Name:  "nogossip",
+		Usage: "disable consensensus message gossip",
```
lorenzo-dev1, line 600:
```suggestion
		Usage: "disable consensus message gossip",
```

consensus/tendermint/core/handler.go
```diff
@@ -130,15 +130,19 @@ eventLoop:
 					}
 					continue
 				}
-				c.backend.Gossip(c.CommitteeSet().Committee(), e.Message)
+				if !c.noGossip {
```
lorenzo-dev1, line 133: I think it would be better to do the nogossip check inside the gossiper itself. Since we have a single component who handles gossip it makes sense to call it and let it handle the gossip/nogossip. 

Then when we are instead broadcasting for the first time a vote to the entire committee we can call a different function that bypasses this check and gossips no matter what. 

This would be better because we will probably modify the gossip strategy in the future (I already did it in the BLS pr) and therefore we will otherwise to always copy also the nogossip check everytime

What do you think?

lorenzo-dev1 (commented):
Left an observation on the code structure and a typo fix.

Another observation from me is: do we really want to expose such feature to node operators? I understand that it is useful for testing, however I am not sure it is realy good to expose it to the "general public".

vpiyush:
> Another observation from me is: do we really want to expose such feature to node operators? I understand that it is useful for testing, however I am not sure it is realy good to expose it to the "general public". Thoughts?

It's a good point. My opinion is since it's an opt-in feature with a descriptive name(**nogossip**, which clearly indicates it's use), discourages accidental use by those unfamiliar with its purpose. which aligns with the idea that users who choose to enable it likely understand the implications.

vpiyush:
> > Another observation from me is: do we really want to expose such feature to node operators? I understand that it is useful for testing, however I am not sure it is realy good to expose it to the "general public". Thoughts?
> 
> It's a good point. My opinion is since it's an opt-in feature with a descriptive name(**nogossip**, which clearly indicates it's use), discourages accidental use by those unfamiliar with its purpose. which aligns with the idea that users who choose to enable it likely understand the implications.

@Jason-Zhangxin-Chen @yazzaoui @tbssajal and @cmjc What do you all think about this? any thoughts?

yazzaoui:
It's fair a point Lorenzo but that is just a non-documented feature of the system that we will most likely remove with the mainnet version of Autonity. I wouldn't stress too much about that.

consensus/tendermint/core/handler.go
yazzaoui: I would tend to agree with @vpiyush here

cmjc:
> It's fair a point Lorenzo but that is just a non-documented feature of the system that we will most likely remove with the mainnet version of Autonity. I wouldn't stress too much about that.

It seems like a useful debug option to me as @vpiyush states above. So, I'd think here:

- the default value for `--nogossip` is `false`
- it goes in the command group `LOGGING AND DEBUGGING OPTIONS` rather than `NETWORKING OPTIONS`


So, you have to explicitly set it and it's clear it's for debugging.

vpiyush:
> * it goes in the command group `LOGGING AND DEBUGGING OPTIONS` rather than `NETWORKING OPTIONS`

Do we all (@yazzaoui  @lorenzo-dev1 @Jason-Zhangxin-Chen @tbssajal) agree on this? I am okay with either, having it as it is or moving it to debugging options.

yazzaoui (approved):
I'm fine with either.

PR #971, opened 2024-04-19
prioritize candidate receiver, move subscription to channel

vpiyush:
We priortize candidate channel read in core handler loop to avoid any delays introduced due to other messages. We have also moved it out of the event multiplexer to make it simple


vpiyush:
moved it to draft becuse this needs an update

vpiyush:
this is deprecated by #980 . closing it

PR #972, opened 2024-04-19
Remove core mutex

vpiyush:
We make use of 2 atomic variables to remove core mutex.
CoreStarting: this is to prevent double start/stop.
CoreRunning: this only gets set when core is properly initialized and running.

Also, SetResultCh is now invoked from worker(taskLoop). Thereby eliminating the need of setting resultChannel on every sealing operation.

### DIFF
--- a/consensus/tendermint/backend/engine.go
+++ b/consensus/tendermint/backend/engine.go
@@ -7,20 +7,19 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/autonity/autonity/consensus/tendermint"
-	"github.com/autonity/autonity/consensus/tendermint/core/message"
-	"github.com/autonity/autonity/crypto"
-
 	"github.com/autonity/autonity/autonity"
 	"github.com/autonity/autonity/common"
 	"github.com/autonity/autonity/consensus"
 	"github.com/autonity/autonity/consensus/misc"
+	"github.com/autonity/autonity/consensus/tendermint"
 	"github.com/autonity/autonity/consensus/tendermint/bft"
 	"github.com/autonity/autonity/consensus/tendermint/core/constants"
+	"github.com/autonity/autonity/consensus/tendermint/core/message"
 	"github.com/autonity/autonity/consensus/tendermint/events"
 	"github.com/autonity/autonity/core"
 	"github.com/autonity/autonity/core/state"
 	"github.com/autonity/autonity/core/types"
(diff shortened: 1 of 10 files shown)


yazzaoui:
Looking good,
Is it possible though to set the result channel directly at construction time of the worker?

vpiyush:
> Looking good, Is it possible though to set the result channel directly at construction time of the worker?

Yes, That's fine. update the change

consensus/tendermint/backend/engine.go
```diff
@@ -474,20 +459,16 @@ func (sb *Backend) Start(ctx context.Context) error {
...
 	return nil
 }
 
 // Close signals core to stop all background threads.
 func (sb *Backend) Close() error {
-	// the mutex along with coreStarted should prevent double stop
-	sb.coreMu.Lock()
-	if !sb.coreStarted {
-		sb.coreMu.Unlock()
+	if !sb.coreStarting.CompareAndSwap(true, false) {
```
lorenzo-dev1, line 468: This logic LGTM but I am curious about why a single atomic bool is not enough.

Is it to address the edge case where  `Start()` is called and then `Stop()` is called before we finished the startup sequence?

lorenzo-dev1 (approved):
LGTM

consensus/tendermint/backend/engine.go
vpiyush: Yes, Exactly. It is to prevent concurrent start-stop races.

tbssajal: I am not sure how it prevents concurrent start-stop races. Don't you want to check
 `if !sb.coreRunning.CompareAndSwap(true, false)` instead of `coreStarting` here to prevent concurrent start-stop races?

tbssajal: I guess the flow should be like this:
```
func Close() {
      set running = false
      // code
      set starting = false
}
```
exactly the opposite of Start(), or am I missing something?

vpiyush: It's a valid point @tbssajal, It should be exactly opposite of start, Though it got merged now i have added this change in the cache-custom PR [here](https://github.com/autonity/autonity/blob/fdf913ffadb8e95045ad82afd161fd267d3574a4/consensus/tendermint/backend/engine.go#L469)

Issue #947, opened 2024-02-22
We are probably caching the same message twice, Once while receiving handler and second time in the gossiper.


vpiyush:
I saw the part in gossiper, where we add the message's again in the knownmessages cache. For the messages which are  simply gossiped  should we get rid of that. here https://github.com/autonity/autonity/blob/8b4a17c18951088c3af01c06da707a87d7887971/consensus/tendermint/backend/gossiper.go#L62

_Originally posted by @vpiyush in https://github.com/autonity/autonity/pull/943#discussion_r1497468487_

vpiyush:
This needs to be analyzed further and fixed

PR #973, opened 2024-04-22
expirable cache

vpiyush:
Replace ARC cache to expirable LRU cache. Prior to this, we didnot have a period cleanup mechanisim for knownCache and recentMessages cache. Which always caused an eviction for new entries when cache was full. an Add & Eviction in ARC cache cacuses multiple deletes and addition because by desing it maintains 4 lists of entries.

This change avoids all that. All expired entries gets deleted periodically in a goroutine, which avoids eviction on every add when we receive/gossip messages.

Another fix is to delete Peer cache on unregister peer in ACN handler


vpiyush:
This also fixes #947

vpiyush:
moving to draft, Needs an update

vpiyush:
this is deprecated by #980 . closing it

PR #974, opened 2024-04-23
Fixed buffered gauge, added new metrics and other improvements

vpiyush:
In This PR, following changes are done:
- Fixed buffered metrics to avoid memory leak, fixed buffer sizes for the counters
- Added new metrics in system 
- optimized memory intensive metrics (removed sprintf)
- Lazy logging for expensive logs

### DIFF
--- a/consensus/acn/protocol/peer.go
+++ b/consensus/acn/protocol/peer.go
@@ -1,10 +1,20 @@
 package protocol
 
 import (
+	"time"
+
 	"github.com/autonity/autonity/crypto"
+	"github.com/autonity/autonity/metrics"
 
 	"github.com/autonity/autonity/common"
 	"github.com/autonity/autonity/p2p"
+)
+
+var (
+	ProposalWriteBg  = metrics.NewRegisteredBufferedGauge("acn/proposal/write", nil, metrics.GetIntPointer(1000))  // time to write proposal to wire
+	PrevoteWriteBg   = metrics.NewRegisteredBufferedGauge("acn/prevote/write", nil, metrics.GetIntPointer(5000))   // time to write prevote to wire
+	PrecommitWriteBg = metrics.NewRegisteredBufferedGauge("acn/precommit/write", nil, metrics.GetIntPointer(5000)) // time to write precommit to wire
+	DefaultWriteBg   = metrics.NewRegisteredBufferedGauge("acn/any/write", nil, nil)
 )
 
 // Peer is a collection of relevant information we have about a `acn` peer.
@@ -53,11 +63,33 @@ func (p *Peer) Address() common.Address {
 }
 
 func (p *Peer) Send(msgcode uint64, data interface{}) error {
+	if metrics.Enabled {
+		defer func(start time.Time) {
+			getWriteMetric(msgcode).Add(time.Since(start).Nanoseconds())
+		}(time.Now())
+	}
 	return p2p.Send(p.rw, msgcode, data)
 }
 
 func (p *Peer) SendRaw(msgcode uint64, data []byte) error {
+	if metrics.Enabled {
+		defer func(start time.Time) {
+			getWriteMetric(msgcode).Add(time.Since(start).Nanoseconds())
+		}(time.Now())
+	}
 	return p2p.SendRaw(p.rw, msgcode, data)
+}
+
+func getWriteMetric(msgCode uint64) metrics.BufferedGauge {
+	switch msgCode {
+	case 0x11:
+		return ProposalWriteBg
+	case 0x12:
+		return PrevoteWriteBg
+	case 0x13:
+		return PrecommitWriteBg
+	}
+	return DefaultWriteBg
 }
 
 // Version retrieves the peer's negoatiated `acn` protocol version.
--- a/consensus/tendermint/backend/engine.go
+++ b/consensus/tendermint/backend/engine.go
@@ -7,6 +7,8 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/autonity/autonity/metrics"
+
 	"github.com/autonity/autonity/autonity"
 	"github.com/autonity/autonity/common"
 	"github.com/autonity/autonity/consensus"
@@ -64,6 +66,7 @@ var (
 	nilUncleHash                  = types.CalcUncleHash(nil) // Always Keccak256(RLP([])) as uncles are meaningless outside of PoW.
 	emptyNonce                    = types.BlockNonce{}
 	now                           = time.Now
+	sealDelayBg                   = metrics.NewRegisteredBufferedGauge("work/seal/delay", nil, nil) // injected sleep delay before producing new candidate block
 )
 
 // Author retrieves the Ethereum address of the account that minted the given
@@ -371,6 +374,9 @@ func (sb *Backend) Seal(chain consensus.ChainReader, block *types.Block, _ chan<
 
 	// wait for the timestamp of header, use this to adjust the block period
 	delay := time.Unix(int64(block.Header().Time), 0).Sub(now())
+	if metrics.Enabled {
+		sealDelayBg.Add(delay.Nanoseconds())
+	}
 	select {
 	case <-time.After(delay):
 		// nothing to do
@@ -383,6 +389,7 @@ func (sb *Backend) Seal(chain consensus.ChainReader, block *types.Block, _ chan<
 	// post block into BFT engine
 	sb.Post(events.NewCandidateBlockEvent{
 		NewCandidateBlock: *block,
+		CreatedAt:         time.Now(),
 	})
 
 	return nil
--- a/consensus/tendermint/core/metrics.go
+++ b/consensus/tendermint/core/metrics.go
@@ -42,20 +42,36 @@ var (
 
 	// Instant metrics
 
-	ProposeBg   = metrics.NewRegisteredBufferedGauge("tendermint/bg/propose", nil)
-	PrevoteBg   = metrics.NewRegisteredBufferedGauge("tendermint/bg/prevote", nil)
-	PrecommitBg = metrics.NewRegisteredBufferedGauge("tendermint/bg/precommit", nil)
+	ProposeBg   = metrics.NewRegisteredBufferedGauge("tendermint/bg/propose", nil, nil)
+	PrevoteBg   = metrics.NewRegisteredBufferedGauge("tendermint/bg/prevote", nil, nil)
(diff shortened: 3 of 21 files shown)


metrics/buffered_gauge.go
```diff
@@ -129,7 +135,14 @@ func (g *StandardBufferedGauge) SnapshotAndClear() BufferedGauge {
 func (g *StandardBufferedGauge) Add(v int64) {
 	g.Lock()
 	defer g.Unlock()
-	g.values = append(g.values, GaugeValue{value: v, timestamp: time.Now()})
+
+	if len(g.values) < g.capacity {
+		g.values = append(g.values, GaugeValue{value: v, timestamp: time.Now()})
+	} else {
+		g.rIndex = g.rIndex % g.capacity
```
lorenzo-dev1, line 142: does `rIndex` stand for `replaceIndex`? Because now the bufferedgauge is a circular buffer?

consensus/acn/protocol/peer.go
```diff
@@ -53,13 +67,35 @@ func (p *Peer) Address() common.Address {
...
 func (p *Peer) SendRaw(msgcode uint64, data []byte) error {
+	if metrics.Enabled {
+		defer func(start time.Time) {
+			getWriteMetric(msgcode).Add(time.Since(start).Nanoseconds())
+		}(time.Now())
+	}
 	return p2p.SendRaw(p.rw, msgcode, data)
 }
 
+func getWriteMetric(msgCode uint64) metrics.BufferedGauge {
```
lorenzo-dev1, line 87: Why `Write`? I would rename to `getMetric` or even just `metric`.

consensus/tendermint/core/metrics.go
```diff
@@ -42,20 +42,40 @@ var (
...
+	MsgProposalBg  = metrics.NewRegisteredBufferedGauge("core/handler/msg/proposal.bg", nil, nil)                 // time to handle proposal
+	MsgPrevoteBg   = metrics.NewRegisteredBufferedGauge("core/handler/msg/prevote.bg", nil, getIntPointer(256))   // time to handle prevote
+	MsgPrecommitBg = metrics.NewRegisteredBufferedGauge("core/handler/msg/precommit.bg", nil, getIntPointer(256)) // time to handle precommit
+
+	MsgProposalPackets  = metrics.NewRegisteredMeter("core/handler/msg/proposal/packets", nil)  //nolint:goconst
+	MsgPrevotePackets   = metrics.NewRegisteredMeter("core/handler/msg/prevote/packets", nil)   //nolint:goconst
+	MsgPrecommitPackets = metrics.NewRegisteredMeter("core/handler/msg/precommit/packets", nil) //nolint:goconst
 )
+
+func getIntPointer(val int) *int {
```
lorenzo-dev1, line 79: Not a big deal but since I saw already two declaration of this func it might be worth to put it into a `helpers` function somewhere. Or maybe in the metrics package itself.

consensus/tendermint/core/precommit.go
```diff
@@ -76,22 +84,21 @@ func (c *Precommiter) HandleCommit(ctx context.Context) {
...
+		"currentHeight", log.Lazy{Fn: c.Height},
 		"msgHeight", precommit.H(),
-		"currentRound", c.Round(),
+		"currentRound", log.Lazy{Fn: c.Round},
 		"msgRound", precommit.R(),
 		"currentStep", c.step,
-		"isProposer", c.IsProposer(),
-		"currentProposer", c.CommitteeSet().GetProposer(c.Round()),
+		"isProposer", log.Lazy{Fn: c.IsProposer},
+		"currentProposer", log.Lazy{Fn: func() types.CommitteeMember { return c.CommitteeSet().GetProposer(c.Round()) }},
```
lorenzo-dev1, line 96: Just curious, is the function wrapping needed to defer also the inner `c.Round()` call?

metrics/exp/exp.go
```diff
@@ -170,7 +170,6 @@ func (exp *exp) syncToExpvar() {
 		case metrics.Gauge:
 			exp.publishGauge(name, i)
 		case metrics.BufferedGauge:
-			//TODO:
```
lorenzo-dev1, line 173: Why don't need  anything here?

metrics/prometheus/prometheus.go
```diff
@@ -48,7 +48,6 @@ func Handler(reg metrics.Registry) http.Handler {
 			case metrics.Gauge:
 				c.addGauge(name, m.Snapshot())
 			case metrics.BufferedGauge:
-				// TODO: need to add support for bufferedGauge to correct prometheus type
```
lorenzo-dev1, line 51: We don't need anything here?

lorenzo-dev1 (commented):
Everything looks good, except that maybe there is some missing code for Prometheus export?

metrics/buffered_gauge.go
vpiyush: yes, essentially its a circular buffer now. We have a capacity requirement when creating the metric at that point it's user responsibility to allocate the max capacity for the metric,

consensus/acn/protocol/peer.go
vpiyush: this is because getWriteMetric is only fetching the socket write related metrics.

consensus/tendermint/core/metrics.go
vpiyush: sure

metrics/exp/exp.go
vpiyush: Ok, the idea is to not throw any errors on buffered gauge when metrics are being fetched with proemtheus client. we established that it's ok to only support the buffered gauge with influx db but we must not throw any errors on prometheus

metrics/prometheus/prometheus.go
vpiyush: see response above

vpiyush:
> Everything looks good, except that maybe there is some missing code for Prometheus export?

Chec response [here](https://github.com/autonity/autonity/pull/974#discussion_r1595365112)

consensus/tendermint/core/precommit.go
vpiyush: no, because c.Round() is only evaluated when outer function is invoked and which would be invoked on log construction.

consensus/acn/protocol/peer.go
lorenzo-dev1: It is not very clear, because the reader doesn't know that Write is an adjective. I thought this function would Write something because I interpreted it as a verb

consensus/tendermint/core/precommit.go
lorenzo-dev1: Then why the wrapping into a `func()`

vpiyush: If you directly pass it would become function call, and evaluate immediately. So when there is an argument we need to wrap it under func to defer execution.

consensus/acn/protocol/peer.go
vpiyush: Since it's a local function it won't matter much, what do you think about getMsgWriteMetric OR getSockeWriteMetric?

lorenzo-dev1: `getSocketWriteMetric`? I would still remove the `get` since it shouldn't be used in go I think

Anyways not a big deal you can even leave it like this

lorenzo-dev1 (approved):
LGTM

consensus/tendermint/backend/engine.go
```diff
@@ -371,6 +374,9 @@ func (sb *Backend) Seal(chain consensus.ChainReader, block *types.Block, _ chan<
 
 	// wait for the timestamp of header, use this to adjust the block period
 	delay := time.Unix(int64(block.Header().Time), 0).Sub(now())
+	if metrics.Enabled {
+		sealDelayBg.Add(delay.Nanoseconds())
```
tbssajal, line 378: Can delay be negative here?

tbssajal: Assuming it was created before `now()`, I guess it can be negative.

tbssajal (changes requested):

consensus/tendermint/backend/engine.go
vpiyush: Yes, It can be -ve

tbssajal: In that case, should we add it to the metrics? We are not delaying anything, instead we are late I guess.

consensus/tendermint/core/precommit.go
tbssajal: > no, because c.Round() is only evaluated when outer function is invoked and which would be invoked on log construction.

Shouldn't we pass `c.Round()` as an argument to the wrapper `func()` for this to be true?

tbssajal: Also, just curious, this lazy functions would invoke later, right? In that case is it possible that some data might change before the function can be invoked, like c.Round() or c.Height() for example?

metrics/buffered_gauge.go
tbssajal, line 142: Instead of doing a mod each time, we can set `g.rIndex = 0` when we have `g.rIndex == g.capacity` and it will be the same as doing mod, because we start from `g.rIndex = 0` and we always increase it by 1.

tbssajal (changes requested):

metrics/buffered_gauge.go
tbssajal: And also because doing mod operation could be quite heavy in run time.

consensus/tendermint/backend/engine.go
vpiyush: Yes, we also want to see that if we are late

metrics/buffered_gauge.go
vpiyush: I don't think we need to micro-optimize here. we will be essentially replacing  mod to a branch statement so no real gain IMO going to be worse. Having said that, It is so small that we don't really care.

tbssajal (approved):
Left two comments here: https://github.com/autonity/autonity/pull/974#discussion_r1598544199
Except that LGTM.

consensus/tendermint/core/precommit.go
vpiyush: > Shouldn't we pass `c.Round()` as an argument to the wrapper `func()` for this to be true?
> Also, just curious, this lazy functions would invoke later, right? In that case is it possible that some data might change before the function can be invoked, like c.Round() or c.Height() for example?

There are no goroutines involved here, they are invoked synchronously, what you said would be true only if they were asynchronous. it just changes the order of invokation.

Issue #802, opened 2023-07-12
accountability: remove dependency on chain history


yazzaoui:
The fault detector logic relies on the chain history to retrieve past consensus committees for various rules. This is implicitly converting the block state transition from `past_state, block -> new_state` to `past_state,chain_history, block -> new_state`

That is not a good design and will probably create a bunch of issues in the long run. We need to remove this dependency by saving past committee in the current state and refactor the fault detector accordingly.

tbssajal:
Isn't `chain_history` included in the `past_state`? Or maybe I am misunderstanding the meaning of `chain_history` here.

tbssajal:
@yazzaoui please let me if I am mistaken.

yazzaoui:
In order to compute the state at block H , the evm only needs the state at block H-1 (`past_state`) and  block H.
With the changes we made, the accountability module need the ability to retrieve past consensus committee members from the block headers 0 to H-1, and this is what I refer to as `chain_history`. 
This is bad design as with this approach we wouldn't be able to prune the history from old blocks.

lorenzo-dev1:
I noticed that in the `vm.BlockContext` we have a `GetHashFn` which already provides EVM access to the chain object. It is used by the `BLOCKHASH` opcode to retrieve hashes of a certain block number.

Maybe we can adapt this and use it for the accountability precompiles as well.

Jason-Zhangxin-Chen:
@yazzaoui @lorenzo-dev1, EVM `vm.BlockContext` depends on the chain-history to get block hash for BLOCKHASH opcode from header chain, thus, I think we can query the committee and proposer for consensus instance messages in the same way from header chain. I am not sure if this is okay for @yazzaoui 's idea `the removal of chain history`. 

BTW, we can still keep the proposer election without refering to state db for better gain of performance. However as last time you mentioned to place proposer election only in protocol contract due to the election outputs mismatched between the two sides, that was a bug introduced by `computeCommitteePrecompiled` since the sortings of committee members are different.

lorenzo-dev1:
I believe that this issue was about the accountability precompiles containing a chain reference. If that's correct, I think we can close this issue. @yazzaoui @Jason-Zhangxin-Chen what do you think?

The accountability precompiles now fetch the committee from the state.

Jason-Zhangxin-Chen:
I think so, we can close it.

PR #975, opened 2024-04-23
[Rebasing] Remove chain history dependency in precompiles

Jason-Zhangxin-Chen:
This PR is created to close issue #802, it was built on top of the epoch header PR which introduce big update in the client, thus let's remove it from the content of the comming release. It depends on the PR of epoch header: https://github.com/autonity/autonity/pull/909 to be reviewed first.


autonity/solidity/contracts/Autonity.sol
```diff
@@ -726,6 +732,26 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
         return committee;
     }
 
+    /**
+     * @notice Returns the committee of a specific height.
+     * @param _height the input block number.h
+     */
+    function getCommitteeOfHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory) {
```
vpiyush, line 739: naming suggestion ==> getCommitteeByHeight would be better

autonity/solidity/contracts/Autonity.sol
```diff
@@ -726,6 +732,26 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+        require(_height <= block.number, "cannot get committee for a future height");
+        if (_height == 0) {
+            return committee;
+        }
+        uint256 eID = blockEpochMap[_height];
+
+        // if current epoch haven't been finalized, then return current committee.
+        CommitteeMember[] memory members = epochCommittees[eID];
```
vpiyush, line 747: If the current epoch haven't finalized, Shouldn't the eID itself would be 0. we can return there itself.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -815,10 +841,12 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     * state.
     */
     function getProposer(uint256 height, uint256 round) external view virtual returns (address) {
+        CommitteeMember[] memory members = getCommitteeOfHeight(height);
+
```
vpiyush, line 845: the comment needs to be updated? this doesn't have to be current committee

consensus/tendermint/accountability/contracts.go
```diff
@@ -64,7 +57,7 @@ func LoadPrecompiles(chain ChainContext) {
 
 // AccusationVerifier implemented as a native contract to validate if an accusation is valid
 type AccusationVerifier struct {
-	chain ChainContext
+	//chain ChainContext
```
vpiyush, line 60: can be removed

consensus/tendermint/accountability/contracts.go
```diff
@@ -88,16 +81,12 @@ func preVerifyAccusation(chain ChainContext, m message.Msg, currentHeight uint64
 		return errTooOldAccusation
 	}
 
-	// if the suspicious message is for a value that got committed in the same height --> reject accusation
```
vpiyush, line 91: Is this condition not possible anymore? how do we do this now?

consensus/tendermint/accountability/contracts.go
```diff
@@ -730,11 +738,45 @@ func maxEvidenceMessages(committee *types.Committee) int {
...
-	committee, err := chain.CommitteeOfHeight(height)
-	if err != nil {
-		return false
-	}
+func overQuorumVotes(committee *types.Committee, msgs []message.Msg) bool {
 	q := bft.Quorum(committee.TotalVotingPower())
 	return engineCore.OverQuorumVotes(msgs, q) != nil
 }
+
+func committeeOfHeight(height uint64, evm *vm.EVM) (*types.Committee, error) {
```
vpiyush, line 746: Probably it's not directly related to this change, But i have noticed that the height here is a uint64 type but in solidity it's uint256, Shouldn't we keep it same both the places. Also i believe instead of change it here. We should probably use uint64 in solidity, That's a big enough number and should save us some space

core/types/bft.go
```diff
@@ -302,3 +303,29 @@ func (c *Committee) Sort() {
 		})
 	}
 }
+
+func (c *Committee) Proposer(height uint64, round int64) common.Address {
+	totalVotingPower := c.TotalVotingPower()
+	c.lock.RLock()
```
vpiyush, line 309: We are taking up lock too early, this should be taken just before the member's access, to keep the critical section short

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > naming suggestion ==> getCommitteeByHeight would be better

resolve with b6ca941f04ce0ec8e56b19e6b3a683d8099cb786

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: > the comment needs to be updated? this doesn't have to be current committee

resolve with b6ca941f04ce0ec8e56b19e6b3a683d8099cb786

consensus/tendermint/accountability/contracts.go
Jason-Zhangxin-Chen: > can be removed

resolve with b6ca941f04ce0ec8e56b19e6b3a683d8099cb786

consensus/tendermint/accountability/contracts.go
Jason-Zhangxin-Chen: > Probably it's not directly related to this change, But i have noticed that the height here is a uint64 type but in solidity it's uint256, Shouldn't we keep it same both the places. Also i believe instead of change it here. We should probably use uint64 in solidity, That's a big enough number and should save us some space

Let's discuss it at standup, we may need a new issue to make this change.

consensus/tendermint/accountability/contracts.go
```diff
@@ -730,11 +738,45 @@ func maxEvidenceMessages(committee *types.Committee) int {
...
+	var committeeSet []types.CommitteeMember
+	err := acCall(evm, "getCommitteeOfHeight", &committeeSet, new(big.Int).SetUint64(height))
+	if err != nil {
+		return nil, err
+	}
+	committee := &types.Committee{}
+	if len(committeeSet) != 0 {
+		committee.Members = make([]*types.CommitteeMember, len(committeeSet))
+		for i, m := range committeeSet {
+			committee.Members[i] = &types.CommitteeMember{
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > Is there a need of copy here? shouldn't we direcly assign it to members?



> Is there a need of copy here? shouldn't we direcly assign it to members?

Directly assignment does not work in this case, the unpacked data returned from contract is in committee member object array, it does not fit into the pointer array. If we change the typing of the member array, then it could be possible. I will think about this change and create an issue for it if we need such change.

consensus/tendermint/accountability/innocence_proof_handler.go
```diff
@@ -222,7 +223,13 @@ func (fd *FaultDetector) handleOffChainAccusation(accusation *Proof, sender comm
...
 
 	// last param represent the current height for which we are doing consensus (lastBlock + 1)
-	if err := preVerifyAccusation(fd.blockchain, accusation.Message, fd.blockchain.CurrentBlock().NumberU64()+1); err != nil {
+	if err := preVerifyAccusation(accusation.Message, fd.blockchain.CurrentBlock().NumberU64()+1); err != nil {
+		return nil
+	}
+
+	// if the suspicious message is for a value that got committed in the same height --> reject accusation
+	if fd.blockchain.GetBlock(accusation.Message.Value(), accusation.Message.H()) != nil {
+		// return nil just skip to drop the peer connection.
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > this comment is confusing, we don't want to drop peer connection here right?

resolve with b6ca941f04ce0ec8e56b19e6b3a683d8099cb786

core/types/bft.go
Jason-Zhangxin-Chen: > We are taking up lock too early, this should be taken just before the member's access, to keep the critical section short

resolve with b6ca941f04ce0ec8e56b19e6b3a683d8099cb786

consensus/tendermint/accountability/contracts.go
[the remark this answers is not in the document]
vpiyush: I think you misunderstood my comment i meant that do we really need to iterate and copy each member. shouldn't this be possible:  	
```
var committeeSet []*types.CommitteeMember
	committee := &types.Committee{}
	err := acCall(evm, "getCommitteeByHeight", &committeeSet, new(big.Int).SetUint64(height))
	if err != nil {
		return nil, err
	}
	committee.Members = committeeSet
```

Or even could we try directly passing committee.Members in acCall?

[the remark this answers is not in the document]
vpiyush: Also the the committeeOfHeight should also be adjust to committeeByHeight since we are calling getCommitteeByHeight now.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -726,6 +732,26 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
+     */
+    function getCommitteeByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory) {
+        require(_height <= block.number, "cannot get committee for a future height");
+        if (_height == 0) {
+            return committee;
+        }
+        uint256 eID = blockEpochMap[_height];
+
+        // if current epoch haven't been finalized, then return current committee.
+        CommitteeMember[] memory members = epochCommittees[eID];
```
vpiyush, line 747: Just curious, Why not also keep the current committee in the epochCommittes so that we don't have to do this. also we could always access the current committee from the map

consensus/tendermint/backend/engine.go
```diff
@@ -338,13 +338,10 @@ func (sb *Backend) Finalize(chain consensus.ChainReader, header *types.Header, s
...
-			committee.Members[i] = &types.CommitteeMember{
-				Address:      m.Address,
-				VotingPower:  new(big.Int).Set(m.VotingPower),
-				ConsensusKey: m.ConsensusKey,
-			}
+			c := m
+			committee.Members[i] = &c
 		}
-		committee.Sort()
+		// As committee as sorted by contract already, thus we don't sort it again.
```
vpiyush, line 344: typo: "As committee is sorted"

autonity/solidity/contracts/Autonity.sol
```diff
@@ -726,6 +732,26 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
     }
 
+    /**
+     * @notice Returns the committee of a specific height.
+     * @param _height the input block number.h
+     */
+    function getCommitteeByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory) {
+        require(_height <= block.number, "cannot get committee for a future height");
+        if (_height == 0) {
+            return committee;
```
Jason-Zhangxin-Chen, line 742: Here, it should return epochCommittees[0];

Jason-Zhangxin-Chen:
Replace this PR with https://github.com/autonity/autonity/pull/1022.

PR #979, opened 2024-05-09
NTN Inflation mechanism

yazzaoui:


tbssajal (changes requested):

params/protocol_contracts.go
```diff
@@ -47,6 +52,14 @@ var (
 		InitialAllocation: (*math.HexOrDecimal256)(new(big.Int).Sub(new(big.Int).Exp(big.NewInt(2), big.NewInt(256), nil), common.Big1)),
 	}
 
+	DefaultInflationControllerGenesis = &InflationControllerGenesis{
+		IInit:  (*math.HexOrDecimal256)(new(big.Int).Div(new(big.Int).Mul(big.NewInt(75), DecimalFactor), big.NewInt(1000*SecondsInYear))),        // 7.5% AR
+		ITrans: (*math.HexOrDecimal256)(new(big.Int).Div(new(big.Int).Mul(big.NewInt(55), DecimalFactor), big.NewInt(1000*SecondsInYear))),        // 5.5% AR
+		IPerm:  (*math.HexOrDecimal256)(new(big.Int).Div(new(big.Int).Mul(big.NewInt(17_328), DecimalFactor), big.NewInt(100_000*SecondsInYear))), // 17.328% AR
```
tbssajal, line 58: `inflationReserve` is already set to 40 million NTN, so `IPerm` should be just 17.328% / year, why are we multiplying it with `DecimalFactor` ?

autonity/solidity/contracts/InflationController.sol
```diff
@@ -0,0 +1,125 @@
...
+        uint256 _lastEpochTime,
+        uint256 _currentEpochTime
+    )
+        external
+        view
+        returns (uint256)
+    {
+        SD59x18 _lastTime = convert(int256(_lastEpochTime));
+        SD59x18 _currentTime = convert(int256(_currentEpochTime));
+        if (_currentTime <= params.T) {
```
tbssajal, line 44: Not my concern but according to the [discussion](https://github.com/clearmatics/autonity-protocol/discussions/213), it looks like we are in transition period if `_lastEpochTime <= params.T` is true, otherwise it is the permanent regime.

params/protocol_contracts.go
```diff
@@ -47,6 +52,14 @@ var (
 		InitialAllocation: (*math.HexOrDecimal256)(new(big.Int).Sub(new(big.Int).Exp(big.NewInt(2), big.NewInt(256), nil), common.Big1)),
 	}
 
+	DefaultInflationControllerGenesis = &InflationControllerGenesis{
+		IInit:  (*math.HexOrDecimal256)(new(big.Int).Div(new(big.Int).Mul(big.NewInt(75), DecimalFactor), big.NewInt(1000*SecondsInYear))),        // 7.5% AR
```
tbssajal, line 56: I guess we should not have `DecimalFactor` in any of the following following parametes, or am I missing something? If it is multiplied to keep it as ineger after division, then we should handle it in `calculateTransitionRegime` function in `InflationController.sol`, right?

tbssajal (changes requested):

params/protocol_contracts.go
yazzaoui: We multiply by 1e18 to conform to the `SD59x18` Solidity type used for fixed-point arithmetic by those initial parameters.

autonity/solidity/contracts/InflationController.sol
yazzaoui: From my understanding we could be in a situation where the transition is happening mid-epoch so we have to account for it by splitting the calculation in two parts - maybe @fjribi can provide some feedback here ?

params/protocol_contracts.go
yazzaoui: See comment above about `SD59x18` type

autonity/solidity/contracts/Autonity.sol
```diff
@@ -887,59 +921,80 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
...
-            uint256 _reward = (committee[i].votingPower * _amount) / epochTotalBondedStake;
-            if (_reward > 0) {
+            uint256 _atnReward = (committee[i].votingPower * _atn) / epochTotalBondedStake;
+            uint256 _ntnReward = (committee[i].votingPower * _ntn) / epochTotalBondedStake;
+            if (_atnReward > 0 || _ntnReward > 0) {
                 // committee members in the jailed state were just found guilty in the current epoch.
                 // committee members in jailbound state are permanently jailed
                 if (_val.state == ValidatorState.jailed || _val.state == ValidatorState.jailbound) {
-                    config.contracts.accountabilityContract.distributeRewards{value: _reward}(committee[i].addr);
+                    config.contracts.accountabilityContract.distributeRewards{value: _atnReward}(committee[i].addr);
```
tbssajal, line 957: Should we distribute `_ntnRewards` as well or is it left out on purpose?

autonity/solidity/contracts/InflationController.sol
tbssajal: Another point, from my understanding, in $t < T$, here $t$ is the time passed from genesis, so it should be `_currentTime - genesisTime`, right?

autonity/solidity/contracts/InflationController.sol
```diff
@@ -0,0 +1,126 @@
...
+        SD59x18 _lastEpochTime,
+        SD59x18 _currentEpochTime
+    )
+        internal
+        view
+        returns (uint256)
+    {
+        SD59x18 _rate;
+        if (params.aE == convert(0)){
+            _rate =  params.iInit + (params.iTrans - params.iInit) * _lastEpochTime / params.T;
```
tbssajal, line 99: From my understanding, the time $t$ used is this formula is the time passed after genesis, so it should be `_lastEpochTime - genesisTime`.

autonity/solidity/contracts/InflationController.sol
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            SD59x18 _lExp = (params.aE * _lastEpochTime)/params.T;
```
tbssajal, line 101: From my understanding, the time $t$ used is this formula is the time passed after genesis, so it should be `_lastEpochTime - genesisTime`.

tbssajal (changes requested):

params/protocol_contracts.go
tbssajal: Got it.

params/protocol_contracts.go
```diff
@@ -19,6 +19,11 @@ import (
 )
 
 var (
+	DecimalPrecision = int64(18)
+	SecondsInYear    = int64(365 * 24 * 60 * 60)
```
Jason-Zhangxin-Chen, line 23: To double check if a leap year with 366 days in not consider in our case.

Jason-Zhangxin-Chen (commented):
LGTM, just left a question on the board.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -139,10 +142,11 @@ contract Autonity is IAutonity, IERC20, Upgradeable {
     uint256 public epochID;
     mapping(uint256 => uint256) internal blockEpochMap;
     uint256 public lastEpochBlock;
+    uint256 public lastEpochTime;
```
Jason-Zhangxin-Chen, line 145: Just a QQ: Why we use physical time rather than the number of blocks to compute the inflation delta?

autonity/solidity/contracts/InflationController.sol
fjribi: > rom my understanding we could be in a situation where the transition is happening mid-epoch so we have to account for it by splitting the calculation in two parts - maybe @fjribi can provide some feedback here ?

Yes, indeed, see [comment](https://github.com/clearmatics/autonity-protocol/discussions/213#discussioncomment-9262617) 
>  * Besides, the discrete-time implementation of the inflation mechanism could violate the conservation law $$\forall t, \, N_t + R_t + L_t = \Gamma $$, particularly at the end dates of unlocking contracts and at the end of the transition period, potentially causing insufficient balance-like errors. A close examination of the dynamics of the Inflation Reserve $R_t$ confirms that there is no risk of the cumulative released amount exceeding the initial Inflation Reserve $\gamma$, and therefore no breach of the conservation law should occur from the dynamics of the Inflation Reserve. However, there is a risk that the amounts unlocked from the unlocking contracts could exceed the initially locked amounts $L^{(j)}_0$ at the end of the unlocking contracts ${T'}^{(j)}$. To prevent this, the last amount to be unlocked ($- \delta L(t) \cdot dt$) at $t+dt$ when $t < {T'}^{(j)} \leq t+dt$ can be replaced by the remaining locked amount $L^{(j)}(t)$.

fjribi: > Another point, from my understanding, in , here  is the time passed from genesis, so it should be `_currentTime - genesisTime`, right?

Yes, $t$ = `_currentTime - genesisTime` (using the same time unit)

autonity/solidity/contracts/Autonity.sol
fjribi: In [sepecifications](https://github.com/clearmatics/autonity-protocol/discussions/213#discussioncomment-9206024) 
> * The inflation reward allocation follows the same schema as the transaction fee (BaseFee), except that inflation rewards are not subject to treasury fee taxation.

autonity/solidity/contracts/InflationController.sol
fjribi: see https://github.com/autonity/autonity/pull/979#discussion_r1596677280

autonity/solidity/contracts/Autonity.sol
tbssajal: I guess it is not treasury fee taxation, it goes to the reporter who reported the validator to jail.

params/protocol_contracts.go
fjribi: Actually, we are using the following convention 

> * Time $t$ is measured in year fractions: $$t = \frac{\text{number of seconds}}{365 \times 24 \times 60 \times 60}$$ with $t=0$ at genesis. For each block, it corresponds to the difference between the current timestamp and the genesis timestamp.

It is a day count convention, meaning that `1 year = 365 * 24 * 60 * 60`, (we can use `365.25*24*60*60` instead). The same convention is used to convert annualized rates into second-based rates. Leap-year treatment is included in the $T$ parameter calculation (the same rule applies to $T^{(j)}$ and $T'^{(j)}$ as in [specifications](https://github.com/clearmatics/autonity-protocol/discussions/213#discussioncomment-9206024)). For instance, if we are targeting a particular start date-time and transition date-time, that includes a leap-year additional day, the time difference in seconds will include additional that leap-year day. Example, T = (365 +365 +365 + 366) * 24*60*60 / (  * 24*60*60) => T = 4.002739726 years or T = (365 +365 +365 + 366) * 24*60*60 in seconds (actual parameter to set in the contract code)

This avoids any particular treatment of leap years in the code it self.

Does this answer your question?

autonity/solidity/contracts/Autonity.sol
fjribi: Thanks for the remark. 

First of all, note that the same convention used for inflation rate time $t$ must be applied to all related unlocking schedules; otherwise, if, for some technical reasons, we start suffering a structural block time stretching (an increase of the block time in average), we could suffer some mismatch between the inflation release schedule and unlocking schedules, such a mismatch could lead to a breach in the conservation law and discontinuity of inflation rate function.

Although it is possible and easier to use block heights for all schedules, we prefer using the timestamp-based approach, mainly for the following reasons:
- If we use absolute dates for "contractual" unlocking schedules and transition dates, and we start suffering a notable increase in block time we can get a large mismatch between the contractual dates and the block-height-based timing (ex if the block time realized average is 1.05 seconds instead of the theoretical target of 1 second, could leads to 18 days mismatch over one year).
- Interest rate convention standardization across the protocol. For instance, CDP Debt interest rate compounding is timestamp-based.

_Note that there are another related remarks and arguments in this issue [Interest compounding discrete or continuous](https://github.com/clearmatics/research-stabilization/issues/299)_

autonity/solidity/contracts/Autonity.sol
fjribi: My understanding is that the "rough rule" is the following: any minted but not distributed/allocated NTNs (slashed NTNs for locked stakable NTNs, suspended vesting contract NTNs, etc.) should go to the foundation treasury account. 

> I guess it is not treasury fee taxation, it goes to the reporter who reported the validator to jail.

IMO, this case could be treated as part of the reward allocation logic (neither a "non-distributed" NTN case nor a network activity taxation treasury fee); ie. apply the same existing schema for ATN reporting reward. This could offer a higher incentive for reporting faults, @rmsams, @yazzaoui what do you think?

params/protocol_contracts.go
fjribi: Note that before the MainNet launch, we are supposed to have the final dates (launch date, transition date, and all unlocking schedules dates) and inflation parameters ($\tau_{\frac{1}{2}}$, $i_{\text {trans }}$, and  $i_{\text {init }}$). Therfore we can determine the $T$, $T^{(j)}$, and $T'^{(j)}$ values according to the 365 convention. We re-run the [curve analysis notebook](https://github.com/clearmatics/autonity-protocol/blob/arn-xx-inflation-mechanism/research/12/Inflation_Mechanism_Analysis-Supply_Curves.ipynb) to solve for $a_e$ and update the controller contract parameters (subject to "seconds conversion").

autonity/solidity/contracts/InflationController.sol
yazzaoui: Thanks that's a good point I overlooked when I changed for the timestamp based calculation.

autonity/solidity/contracts/Autonity.sol
yazzaoui: I don't foresee any side effects, I'll add it.

autonity/solidity/contracts/InflationController.sol
```diff
@@ -0,0 +1,129 @@
...
+        return uint256(convert(_result));
+    }
+
+    /**
+    * @notice Calculate inflation after transition.
+    */
+    function calculatePermanentRegime(
+        uint256 _inflationReserve,
+        SD59x18 _lastEpochBlock,
+        SD59x18 _currentEpochBlock
```
tbssajal, line 118: 'Time' instead of 'Block'

autonity/tests/base.go
```diff
@@ -149,6 +157,7 @@ func (r *runner) waitNBlocks(n int) { //nolint
 		// consider monitoring gas cost here and fail if it's too much
 		require.NoError(r.t, err, "finalize function error in waitNblocks", i)
 		r.evm.Context.BlockNumber = new(big.Int).Add(big.NewInt(int64(i+1)), start)
+		r.evm.Context.Time = new(big.Int).Add(r.evm.Context.Time, common.Big1)
```
tbssajal, line 160: I guess, modification of Context should be done before calling 'Finalize'. Otherwise the contract has access to previous block number and time, right?

tbssajal (changes requested):

autonity/tests/base.go
yazzaoui: This is fine? This is just reflecting that all previous operations on the state are accounted for the yet to be finalized block. It's the same thing we're doing when the chain is running for real.

tbssajal: I am not sure what is the issue then, but there is something wrong. Like if you run `r.waitNextEpoch()` no matter how many times, you will be still at block 30 (epoch period = 30).

tbssajal: My point is that contract function should get the information of the block beign mined when they access `block.number` or `block.timestamp`, so context should be updated before function is called. And I tested with truffle and looks like the claim is true. If you update context before calling Finalize, then the problem I just mentioned about `r.waitNextEpoch()` gets fixed.

tbssajal: Or I am missing the problem here I guess.

tbssajal (approved):
LGTM Except these comments: https://github.com/autonity/autonity/pull/979#discussion_r1600255648

yazzaoui:
Merging @tbssajal, I managed to fix a bunch of tests but there is still a few ones broken in the truffle probably due to the accounting for rewards. Thanks for taking care of that :+1:

autonity/solidity/contracts/InflationController.sol
fjribi: @yazzaoui, @tbssajal, there's no way to avoid testing params.aE == 0 each time, using a conditional compilation (pre-processing) as we know at inception the value once for all ?

yazzaoui: Not possible. I can make the adjustment on the contract logic once we know the value of params.aE.

Issue #978, opened 2024-05-08
A bug in the implementation of tendermint line36 upon-condition, we need to fix it and add extra tests to cover it.


Jason-Zhangxin-Chen:
There is an interesting mismatch between our tendermint implementation and the tendermint protocol. As line 36 of page 6 of https://arxiv.org/pdf/1807.04938 specified: for the 1st time on that upon-condition, we'd need to do the locking of value, round, and setting of valid value and valid round. However in our implementation, we introduced a flag: https://github.com/autonity/autonity/blob/develop/consensus/tendermint/core/core.go#L352 with which to be reset on each round start, I think this resetting is not need at all, otherwise we might overwrite the locked value, locked round, valid value and valid round. 
We need to fix it and add extra tests to cover this scenario.

vpiyush:
I think this makes sense, we should probably set the flag only when round=0

lorenzo-dev1:
I think that the current implementation is correct. It is allowed for a node to change `lockedValue` and `lockedRound` if it receives a quorum of prevotes for another value in a later round. This behaviour is called [flip-flopping](https://github.com/tendermint/tendermint/blob/v0.34.x/spec/light-client/accountability/README.md#flip-flopping).

Moreover the definition of `for the first time` from the tendermint paper is:

> with ”for the first time” constraint to denote that it is triggered only the first time a corresponding condition evaluates to true.

They do not clearly specify if they mean "first time in a round" or "first time in an height". However from condition on lines 34 and 47 we can infer that they mean "first time in a round", since timeouts can be scheduled once every round.

Jason-Zhangxin-Chen:
> I think that the current implementation is correct. It is allowed for a node to change `lockedValue` and `lockedRound` if it receives a quorum of prevotes for another value in a later round. This behaviour is called [flip-flopping](https://github.com/tendermint/tendermint/blob/v0.34.x/spec/light-client/accountability/README.md#flip-flopping).
> 
> Moreover the definition of `for the first time` from the tendermint paper is:
> 
> > with ”for the first time” constraint to denote that it is triggered only the first time a corresponding condition evaluates to true.
> 
> They do not clearly specify if they mean "first time in a round" or "first time in an height". However from condition on lines 34 and 47 we can infer that they mean "first time in a round", since timeouts can be scheduled once every round.
> 
> Additionally that is also why we have no rules in the FD to detect this kind of behaviour (commit for V, then later on commit for V').
> 
[...]

PR #981, opened 2024-05-09
[ReadyForReview] State transition line36

Jason-Zhangxin-Chen:
To close issue https://github.com/autonity/autonity/issues/978, we don't overwrite the locked value/round and the valid value/round of upon condition line36 of tendermint protocol.


lorenzo-dev1 (changes requested):
I believe the current `develop` logic is the correct one. For the reason, see my comment [here](https://github.com/autonity/autonity/issues/978#issuecomment-2102178536)

PR #982, opened 2024-05-15
allow ip/port update for validator enode

vpiyush:
a validator who chooses to migrate to a different ip needs to ensure that it is out of the committee while doing so.

To gracefully migrate it should follow the following guidelines if it is the part of committee currently:

- User needs to pause the validator
- Wait for an epoch to go out of committee
- Invoke updateEnode with modified enode
    - internal checks are done if the caller is the registered treasury of the validator
    - Validator is registered and not part of the current committee  
- Bring the node back up on new machine having the ip as provided in the modified enode
- Activate validator again


yazzaoui (approved):
LGTM

tbssajal (approved):
LGTM

yazzaoui:
> Bring the node back up on new machine having the ip same as modified validator

Typo ? The same public key not the same ip right?

cmjc:
I read it as a typo...

vpiyush:
> > Bring the node back up on new machine having the ip same as modified validator
> 
> Typo ? The same public key not the same ip right?

I meant it as "new machine having the same ip as provided in the modified enode".  do you  think it is still confusing?

cmjc:
Thanks @vpiyush - that's clear.

How to  migrate a client is currently in [Run Autonity, sub-section "Migrating an Autonity Go Client"](https://docs.autonity.org/node-operators/run-aut/#migrating-an-autonity-go-client).

So, this can be documented in there. (It might be better to put it into a separate page, tbh, to make it more visible. I'll look at that in the docs update 👍

PR #984, opened 2024-05-20
enable suspension for spammers

vpiyush:
Redefining p2p jailing to peer suspenstion. If a peer spams a node it would get suspended (will not be allowed to make connection ) for upto 60 blocks. This is the initial implementation.


p2p/server.go
```diff
@@ -1061,9 +1067,11 @@ func (srv *Server) checkInboundConn(remoteIP net.IP) error {
 	return nil
 }
 
-func (srv *Server) AddJail(id enode.ID, jailTime time.Duration) {
-	now := srv.clock.Now()
-	srv.jailed.add(id.String(), now.Add(jailTime))
+func (srv *Server) needsSuspension(pd peerDrop) {
```
yazzaoui, line 1070: The current name suggests it checks if a suspension is needed, but from the code, it actually performs the suspension by adding a peer to a suspension list. I'd suggest to rename it to `processPeerSuspension`, `checkAndSuspendPeer` or something else.

yazzaoui (approved):
LGTM, would have preferred though using the more widespread word `blacklist`.

tbssajal (approved):
LGTM

vpiyush:
> LGTM with naming comment, would have preferred though using the more widespread word `blacklist`.

The term "suspension" implies a temporary restriction, whereas "blacklist" conveys a more permanent prohibition, which might not be acurrate in this context. What do you think? I am open to rename it to blacklist though.

yazzaoui:
>The term "suspension" implies a temporary restriction, whereas "blacklist" conveys a more permanent prohibition, which might not be acurrate in this context. What do you think? I am open to rename it to blacklist though.

Ok it's fine let's not quibble over it unless @cmjc you have a strong view on that ?

cmjc:
No quibbling! lgtm Agree with @vpiyush .

PR #989, opened 2024-05-28
Afd more tests

Jason-Zhangxin-Chen:
This PR contains full tests for AFD precompiled contracts base on the msg aggregation work. It also include a few enhancement to make the contract more robust for some edge cases.


Jason-Zhangxin-Chen:
To be closed with new one.

PR #992, opened 2024-05-29
fix, we don't suspend peer with network infra errors.

Jason-Zhangxin-Chen:
To get the network recovery from partitioning disaters, we cannot suspend peer with network infra errors.

### DIFF
--- a/p2p/peer_error.go
+++ b/p2p/peer_error.go
@@ -19,6 +19,8 @@ package p2p
 import (
 	"errors"
 	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


vpiyush:
@Jason-Zhangxin-Chen I have used a different approach to fix this problem. Please check this PR. https://github.com/autonity/autonity/pull/994

Jason-Zhangxin-Chen:
To be replaced by https://github.com/autonity/autonity/pull/994

PR #996, opened 2024-06-03
Start mining at genesis time

yazzaoui:
A small PR to ensure that mining starts to happen at genesis time, avoiding potential round changes.
Also, I added a banner when launching autonity with the `--piccadilly` flag

### DIFF
--- a/cmd/autonity/main.go
+++ b/cmd/autonity/main.go
@@ -252,7 +252,33 @@ func prepare(ctx *cli.Context) {
 	// If we're running a known preset, log it for convenience.
 	switch {
 	case ctx.GlobalIsSet(utils.PiccadillyFlag.Name):
-		log.Info("Starting Autonity on Piccadilly testnet")
+		log.Info(`Starting Autonity on Piccadilly Testnet
+
+ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα		
+ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα
+
+                        888                     d8b 888             
+                        888                     Y8P 888             
+                        888                         888             
+       8888b.  888  888 888888 .d88b.  88888b.  888 888888 888  888 
+          "88b 888  888 888   d88""88b 888 "88b 888 888    888  888 
+      .d888888 888  888 888   888  888 888  888 888 888    888  888 
+      888  888 Y88b 888 Y88b. Y88..88P 888  888 888 Y88b.  Y88b 888 
+      "Y888888  "Y88888  "Y888 "Y88P"  888  888 888  "Y888  "Y88888 
+                                                                888 
+                         autonity.org                      Y8b d88P 
+                                                            "Y88P"  
+
+ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα
+ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα
+
+Take part in Piccadilly Circus Games Round 6! More infos @ https://game.autonity.org
+Discord:  https://discord.com/invite/autonity 
+Telegram: https://t.me/autonity
+X:        https://twitter.com/autonity_
+
+
+`)
 	case ctx.GlobalIsSet(utils.BakerlooFlag.Name):
 		log.Info("Starting Autonity on Bakerloo testnet")
 	case ctx.IsSet(utils.DeveloperFlag.Name):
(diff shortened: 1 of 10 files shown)


eth/backend.go
```diff
@@ -666,6 +673,65 @@ func (s *Ethereum) Stop() error {
...
+		lastSecond = 0
+	)
+	for {
+		now := time.Now()
+		duration := genesisTime.Sub(now)
+
+		if duration <= 0 {
+			s.log.Warn("Launch!")
+			go func() {
+				time.Sleep(3 * time.Second)
```
vpiyush, line 698: This could create inconsistent behaviour depending on the environment. Using a periodic check in a loop could be more sound

eth/backend.go
```diff
@@ -666,6 +673,65 @@ func (s *Ethereum) Stop() error {
...
+		}
+		days := int(duration.Hours() / 24)
+		hours := int(duration.Hours()) % 24
+		minutes := int(duration.Minutes()) % 60
+		seconds := int(duration.Seconds()) % 60
+
+		if days > 0 {
+			if days != lastDays {
+				lastDays = days
+				s.log.Info(fmt.Sprintf("%d days remaining before genesis", days))
```
vpiyush, line 713: when the valus is 1 for days, hours and minutes, It would be an incorrect statment something like "1 days remaining". The alternative here would be to use "1 day(s) remaining" . What do you think @jay

lorenzo-dev1 (commented):
Code LGTM but I think we should use `time.Now().UTC()`, otherwise validators in different timezones might start at different times. 

Specifically because of this part of the code: 
```
now := time.Now()
duration := genesisTime.Sub(now)
```

cmd/autonity/main.go
```diff
@@ -252,7 +252,28 @@ func prepare(ctx *cli.Context) {
...
+     "88b 888  888 888   d88""88b 888 "88b 888 888    888  888 
+ .d888888 888  888 888   888  888 888  888 888 888    888  888 
+ 888  888 Y88b 888 Y88b. Y88..88P 888  888 888 Y88b.  Y88b 888 
+ "Y888888  "Y88888  "Y888 "Y88P"  888  888 888  "Y888  "Y88888 
+                                                           888 
+                    autonity.org                      Y8b d88P 
+                                                       "Y88P"  
+
+ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα
+Take part of Piccadilly Circus Games Round 6! More infos : https://game.autonity.org/
```
cmjc, line 271:
```suggestion
Take part in Piccadilly Circus Games Round 6! More infos : https://game.autonity.org/
```

vpiyush (approved):
LGTM, except minor comments

tbssajal (approved):
LGTM

tbssajal:
> Code LGTM but I think we should use `time.Now().UTC()`, otherwise validators in different timezones might start at different times.
> 
> Specifically because of this part of the code:
> 
> ```
> now := time.Now()
> duration := genesisTime.Sub(now)
> ```

I guess it is fine because `genesisTime` is also converted to local time in the following line.
```
genesisTime := time.Unix(int64(s.blockchain.Genesis().Time()), 0)
```
However I was just wondering, if `time.Sub(anotherTime)` converts both location into same location before doing subtraction.

Jason-Zhangxin-Chen (approved):
LGTM

lorenzo-dev1 (approved):
LGTM

yazzaoui:
> Code LGTM but I think we should use time.Now().UTC(), otherwise validators in different timezones might start at different times.

A `time` object represents an instant in time and is location independent, the location is only used for presentation. See the docs:
```
// The methods Local, UTC, and In return a Time with a specific location.
// Changing the location in this way changes only the presentation; it does not
// change the instant in time being denoted and therefore does not affect the
// computations described in earlier paragraphs.
```
Internally it keeps track of the second elapsed since a certain starting date in UTC timezone, it's like a more advanced unix timestamp.
> However I was just wondering, if time.Sub(anotherTime) converts both location into same location before doing subtraction.

As per above it doesn't need to but you can check it directly in the implementation of Sub in the standard library.

cmd/autonity/main.go
yazzaoui: Thanks, it there is more things you think we can add let me know

eth/backend.go
vpiyush: Not a big deal but, because of some weird reason if we take more than 3 seconds. Than, the "launch success" would not be printed. if that's an expectation that would be inconsistent.

yazzaoui: It's fine for the time being if we are taking more than 3 seconds it means probably something went wrong, we should ideally mint block 1 in one second.

PR #998, opened 2024-06-03
Bls integration tests

lorenzo-dev1:


tbssajal (approved):
LGTM

vpiyush (approved):
LGTM

consensus/tendermint/core/message/message.go
```diff
@@ -578,10 +577,9 @@ func AggregateVotes[E Prevote | Precommit](votes []Vote) *E {
 	// signers of the aggregate
 	signers := types.NewSigners(representative.Signers().CommitteeSize())
 
-	// order votes by decreasing number of distinct signers.
-	// This ensures that we reduce as much as possible the number of duplicated signatures for the same validator
+	// we want to privilege vote with higher voting power, since we are creating a complex aggregate carrying quorum
 	sort.Slice(votes, func(i, j int) bool {
-		return votes[i].Signers().Len() > votes[j].Signers().Len()
+		return votes[i].Signers().Power().Cmp(votes[j].Signers().Power()) > 0
```
tbssajal, line 582: Not sure if it is relevant here though:
```
The sort is not guaranteed to be stable: equal elements may be reversed from their original order. For a stable sort, use [SliceStable](https://pkg.go.dev/sort#SliceStable).
```

core/types/signers.go
```diff
@@ -216,7 +216,7 @@ func (s *Signers) RespectsBoundaries(other *Signers) bool {
 
 		secondCoefficient = int(other.Bits.Get(i))
 		if secondCoefficient == int(multipleSignatures) {
-			secondCoefficient = int(other.Coefficients[count])
+			secondCoefficient = int(other.Coefficients[count2])
```
tbssajal, line 219: I guess it is better to use long names that would explain the usage instead of using name1, name2. I also encountered similar bug in my code.

tbssajal (commented):
Just some comments.

consensus/tendermint/core/message/message.go
lorenzo-dev1: Yeah here it is fine to not have it stable

Jason-Zhangxin-Chen (approved):
LGTM

yazzaoui (approved):
LGTM

PR #1001, opened 2024-06-04
Vesting Contract Documentation

tbssajal:
Contains documentation for both user and developer for vesting contracts only.


cmjc:
Looking good @tbssajal 👍  Summary of comments from the Slack just now: README on NatSpec conventions, adding @return, and @custom for constraint checks.

PR #1003, opened 2024-06-06
Test vesting contract

tbssajal:
- Bug fix in Autonity.sol, in case failing to notify the delegator contract.
- More tests


yazzaoui:
https://github.com/autonity/autonity/blob/71fac1a9f64f4c4c04a111ff63479d55b3c55187/autonity/solidity/contracts/Autonity.sol#L781-L784

I would not recommend creating dedicated events for debug purposes, this is just cluttering the ABI. Use instead a general `debug(string)` event if you want. `block.timestamp` can be retrieved easily.

tbssajal:
Closing as there is major rework, not just tests update. New PR for the rework: https://github.com/autonity/autonity/pull/1007

PR #1006, opened 2024-06-10
aggregator tests & refinements

lorenzo-dev1:
Refinements:
- Refactor of the `Merge` function in `signers.go`. This version is more CPU efficient.
- Removes the `alreadyProcessed` check in the `aggregator` as it is redundant. We shouldn't process two times the same message thanks to the backend caches. Moreover, even if it happens sparingly it is not a big deal (Core and FD are resilient to it).
- Modifies bypass for not suspending jailed validators. Returning an error from the backend handler when the signer is jailed makes it easier to test for it.
- remove `ErrUnprovableRule`. It can never get raised.
- performance improvements in the handling of deserialization of public keys
- detection of splitting zero attack in the aggregator

Tests:
- Adds tests for `aggregator.go`
- Adds tests for other part of the code modified by https://github.com/autonity/autonity/pull/957


consensus/tendermint/backend/aggregator.go
```diff
@@ -649,24 +663,66 @@ func (a *aggregator) handleVote(voteEvent events.UnverifiedMessageEvent, committ
...
+		return
+	}
+	if msg.H() > coreHeight {
+		// future messages are dealt with at backend peer handler level
+		a.logger.Crit("future message in aggregator", "msgHeight", msg.H(), "coreHeight", coreHeight)
+	}
+
+	header := a.backend.BlockChain().GetHeaderByNumber(msg.H() - 1)
+	if header == nil { //nolint
+		a.logger.Crit("cannot fetch header for non-future message", "headerHeight", msg.H()-1, "coreHeight", coreHeight)
```
vpiyush, line 694: Question; why does this log say non-future message. aren't we dealing with future messages here

crypto/blst/signature.go
```diff
@@ -83,19 +83,37 @@ func (s *BlsSignature) Verify(pubKey PublicKey, msg []byte) bool {
 	return s.s.Verify(false, pubKey.(*BlsPublicKey).p, false, msg, generalDST)
 }
 
+// Same as AggregateVerify, but enforces the distinct messages constraint.
+// This constraint should always be enforced, as not doing it can lead to inconsistencies in how signatures are validated across nodes
+// see TestBlsAttacks for more information
+func (s *BlsSignature) AggregateVerifyStrict(pubKeys []PublicKey, msgs [][32]byte) bool {
```
vpiyush, line 89: Naming suggestion: AggregateVerifyDistinct seems more appropriate to me.

vpiyush (approved):
LGTM!! Nice improvements overall and testing coverage.

crypto/blst/signature.go
lorenzo-dev1: I thought about it but I did not choose that name because: both `AggregateVerify` and `AggregateVerifyStrict` always needs to be used on distinct messages, however only `AggregateVerifyStrict` actually does the check!

I feel like naming one `AggregateVerifyDistinct` might hint at the fact that `AggregateVerify` can be used on non-distinct messages, which is not true.

vpiyush: Ah. I see. Your reasoning makes sense.

PR #1009, opened 2024-06-13
feature, save and query committee with epoch header.

Jason-Zhangxin-Chen:
The PR contains below improvements for the protocol:
- Saving committee only in epoch header
- Adding a unified committee query interface in the blockchain
- Providing an epoch head event subscription function, and refining the committee watcher routine.


Jason-Zhangxin-Chen:
To be replaced by https://github.com/autonity/autonity/pull/1022

PR #1016, opened 2024-06-20
add FastAggregateVerifyBatch

lorenzo-dev1:
Adds FastAggregateVerifyBatch. 

This new function ensures that when verifying an aggregate signature:
- if the aggregate is valid, then also all the individual signatures are valid.
- if the aggregated is invalid, then at least one signature is invalid.

The standard `FastAggregateVerify` function doesn't ensures these properties when dealing with edge cases (e.g. splitting zero attack and consensus attack. See `TestBlsAttacks` in `crypto/bls/signature_test.go` for more information).

This function however is not defined by the [BLS standard](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html), therefore **it should be checked for correctness by a cryptographer**. 

Its implementation is inspired by the `VerifyMultipleSignatures` function (which is also not present in the standard), originally proposed in https://ethresear.ch/t/fast-verification-of-multiple-bls-signatures/5407/14 and currently implemented in [prysm](https://github.com/prysmaticlabs/prysm/blob/develop/crypto/bls/blst/signature.go#L235)

Performance overhead is ~1% CPU more used.

### DIFF
--- a/consensus/tendermint/backend/aggregator.go
+++ b/consensus/tendermint/backend/aggregator.go
@@ -531,9 +531,8 @@ func (a *aggregator) processBatches(batches [][]events.UnverifiedMessageEvent, e
 			continue
 		}
 
-		aggregateSignature := blst.Aggregate(signatures)
 		hash := batch[0].Message.SignatureInput()
-		valid := aggregateSignature.FastAggregateVerify(publicKeys, hash)
+		valid := blst.FastAggregateVerifyBatch(signatures, publicKeys, hash)
 
 		var validVotes []message.Vote
 		var invalids []uint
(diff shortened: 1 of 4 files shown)


ScottRMalley (commented):
Looks pretty good to me, other than one small question.

consensus/tendermint/backend/aggregator.go
```diff
@@ -541,30 +540,15 @@ func (a *aggregator) processBatches(batches [][]events.UnverifiedMessageEvent, e
...
-				a.logger.Error("Please report the following data to the Autonity team!")
-				for i, msg := range messages {
-					a.logger.Error("Message", "i", i, "hash", msg.Hash(), "signers", msg.Signers().String())
-				}
-
-				// consider all messages as valid since single signatures are valid.
-				// other nodes might already have considered them as valid depending on how they were aggregated.
-				// so this choice maximizes coherence across the network.
-				validVotes = messages
-			} else {
```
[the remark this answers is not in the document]
ScottRMalley: Discussed offline: the answer is that it would be impossible for colluding validators to know that sequence of random points we are generating, therefore making the splitting zero attack no longer feasible.

ScottRMalley (approved):

crypto/blst/signature.go
```diff
@@ -152,6 +152,59 @@ func (s *BlsSignature) FastAggregateVerify(pubKeys []PublicKey, msg [32]byte) bo
...
+		// convert from affine coordinates
+		key := new(blst.P1)
+		key.FromAffine(pubkeys[i].(*BlsPublicKey).p)
+		sig := new(blst.P2)
+		sig.FromAffine(sigs[i].(*BlsSignature).s)
+
+		// multiply key and sig with the random scalar
+		// TODO: to verify with cryptographer, can the scalar multiplication make the sigs or the pubkeys exit the group?
+		scalar := randomScalar()
+		modifiedKey := key.Mult(scalar)
```
tbssajal, line 195: Do we know the mathematical relation between signature and public key? If the relation is not linear then multiplying should not work. For example if the relation is in second order then one of them should be multiplied by $scaler$ and the other should be multiplied by $scaler^2$.

tbssajal: Resolving the comment because the tests cover the function properly.

crypto/blst/signature.go
```diff
@@ -152,6 +152,59 @@ func (s *BlsSignature) FastAggregateVerify(pubKeys []PublicKey, msg [32]byte) bo
...
+		modifiedSig := sig.Mult(scalar)
+
+		modifiedKeys[i] = &BlsPublicKey{p: modifiedKey.ToAffine()}
+		modifiedSigs[i] = &BlsSignature{s: modifiedSig.ToAffine()}
+	}
```
tbssajal, line 200: For further optimization that can be done in separate PR: the function `FastAggregateVerifyBatch` can be refactored in two separate functions. This part can prepare the modifed kyes and the rest does the verification. Then in `findInvalid` function we can just prepare the keys once and use them to verify recursively.

crypto/blst/signature.go
```diff
@@ -152,6 +152,59 @@ func (s *BlsSignature) FastAggregateVerify(pubKeys []PublicKey, msg [32]byte) bo
...
+	modifiedSigs := make([]Signature, n)
+	for i := 0; i < n; i++ {
+		// convert from affine coordinates
+		key := new(blst.P1)
+		key.FromAffine(pubkeys[i].(*BlsPublicKey).p)
+		sig := new(blst.P2)
+		sig.FromAffine(sigs[i].(*BlsSignature).s)
+
+		// multiply key and sig with the random scalar
+		// TODO: to verify with cryptographer, can the scalar multiplication make the sigs or the pubkeys exit the group?
```
tbssajal, line 193: Can this be verified by multiplying with a big number in a test?

tbssajal (approved):
LGTM! Just some comments.

PR #1017, opened 2024-06-26
add private gossiper api

lorenzo-dev1:


consensus/tendermint/backend/gossiper.go
```diff
@@ -77,7 +78,7 @@ func (g *Gossiper) Gossip(committee types.Committee, message message.Msg) {
 				defer func() {
 					<-g.concurrencyLimiter
 				}()
-				p.SendRaw(code, payload) //nolint
+				p.SendRaw(NetworkCodes[code], payload) //nolint
```
vpiyush, line 81: this causes a muliple map access, which is not needed, previous one was better to fetch the code before spawning these goroutines. Also, this is concurrent on a normal map, however since it's only initialized once, doesn't cause a problem

lorenzo-dev1: fixed with https://github.com/autonity/autonity/pull/1017/commits/7d5e7a521c1ac444ca76ea6119a73a9bc0ba3e34

Jason-Zhangxin-Chen (approved):
LGTM

PR #1019, opened 2024-06-27
[WIP] Improve spam resilience on the client

lorenzo-dev1:
Improvements:
- if a message is so old that it is not useful even for accountability, discard it right away without verifying it.

###  future height message spam 
 A future height message spam can cause two problems:
1. slight increase in CPU usage (however not really significant from the test I ran)
2. eviction of future messages received from honest nodes. I.e. the spammer spams so many messages that the cache fills and we start removing useful messages.

Solution: 
1. Discard messages that are so far in the future that they belong to the next epoch. If we receive messages for the next epoch it means that either we are very out of sync, or that someone is spamming. In any case it is safe to discard those messages (but removing them from the knownMessage cache, so that we can re-receive them). Depends on https://github.com/autonity/autonity/pull/1009 to easily determine epoch end.
2. Partition the future height messages buffer based on the sender (`p2pAddress`). This way a spammer will not cause eviction of messages received by other peers. 

With these two mitigations in place we will ensure that honest peers messages will not be evicted from the buffer + the spamming node will likely be slashed and he cannot unbond before that (because we do not accept messages that do not belong to the current epoch).

### spam of current/old messages from jailed validator

We will make the signature group check and the PreValidate before discarding the message because it comes from a jailed validators. These are wasted resources.

Solution:
- P2p suspend jailed validators until the end of the epoch. The in the following epoch they will be evicted from the committee, so we will not be connected to them anymore. Depends on https://github.com/autonity/autonity/pull/1009 to easily determine epoch end.

However note that messages from a jailed validator can still arrive to us from another colluding peer which gossips them even if the signer is jailed. 

### additional metrics 

Add metrics for jailings and p2p suspends.

### DIFF
--- a/consensus/tendermint/accountability/fault_detector.go
+++ b/consensus/tendermint/accountability/fault_detector.go
@@ -168,7 +168,7 @@ func (fd *FaultDetector) Start() {
 	go fd.consensusMsgHandlerLoop()
 }
 
-func (fd *FaultDetector) isHeightExpired(headHeight uint64, height uint64) bool {
+func IsHeightExpired(headHeight uint64, height uint64) bool {
 	return headHeight > HeightRange && height < headHeight-HeightRange
 }
 
@@ -190,7 +190,7 @@ tendermintMsgLoop:
 			// handle consensus message or innocence proof messages
 			switch e := ev.Data.(type) {
 			case events.MessageEvent:
-				if fd.isHeightExpired(currentHeight, e.Message.H()) {
+				if IsHeightExpired(currentHeight, e.Message.H()) {
 					fd.logger.Debug("Fault detector: discarding old message")
 					continue tendermintMsgLoop
 				}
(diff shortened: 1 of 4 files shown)


tbssajal:
> Discard messages that are so far in the future that they belong to the next epoch. If we receive messages for the next epoch it means that either we are very out of sync, or that someone is spamming.

I guess we should not discard future messages from next epoch, as next epoch can be just one block away. But we can discard future messages from $\delta$ height ahead in future, where $\delta \ge epochPeriod$. But I guess the safest option would be to pick $\delta \ge 2*epochPeriod$, because this ensures that if the future message is valid, then the receiver end has missed at least one full epoch. For example, if the receiver is at epoch $e_1 = 5$ and the future message is from epoch $e_2 \ge 7$, then the receiver missed epoch 6. Which means if the receiver validator was chosen to be a committee member at epoch 6, and then he missed the whole epoch, then the protocol must punish the validator and he will not be a member for epoch 7 or higher. So receiving a valid message from epoch $e_2$ while the current epoch is $e_1$ where $e_2 > e_1 + 1$ means that the receiver is too far behind to actually be a committee member for the actual current committee. So he can safely discard the future message even if it is valid and just import blocks to catch up with the network.

lorenzo-dev1:
I will leave this PR open for the moment, as I did not have time yet to implement all the improvements listed. However the old messages discarding improvement has been incorporated in https://github.com/autonity/autonity/pull/1069 so that we can merge it earlier

PR #1021, opened 2024-07-05
[WIP] omission accountability on-chain contract

lorenzo-dev1:
Still very WIP, do not spend time looking at it yet.


autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -2,82 +2,310 @@
...
+    }
+
+    function distributeProposerRewards(uint256 _ntnReward) external payable virtual onlyAutonity {
+        uint256 atnReward = msg.value;
+
+        for(uint256 i=0; i < committee.length; i++) {
+           if(proposerEffort[committee[i]] > 0){
+               // TODO(lorenzo) is it possible that numerator is always too small and therefore rewards = 0 all the time?
+                    // write a test for it
+               uint256 atnProposerReward = (proposerEffort[committee[i]] * atnReward) / totalEffort;
```
Jason-Zhangxin-Chen, line 273: On an edge case, such as there are F nodes being off for the entire epoch, thus the net proposer effort will always be zero, making the totalEffort being zero, it could cause the finalize() of block being reverted.

lorenzo-dev1: That will not happen because in that case also all the entries in the `proposerEffort` map will be zero, therefore we will never actually compute this line. But I will add a more explicit check to make it clearer, doesn't hurt

Jason-Zhangxin-Chen: > That will not happen because in that case also all the entries in the `proposerEffort` map will be zero, therefore we will never actually compute this line. But I will add a more explicit check to make it clearer, doesn't hurt

No worries, I created an e2e test case to cover this scenario. It doesn't happen by tthe running result.

.github/workflows/go.yml
```diff
@@ -307,13 +307,10 @@ jobs:
       - name: run test
         run: cd ./autonity/solidity && ape --verbosity WARNING test --network ::hardhat ./test/asm/${{ matrix.tests }}
 
-  # docker-e2e-test runs the docker e2e test, it only runs on merges to the
-  # develop branch because it takes hours to complete and so is not feasible to
-  # run during normal development.
-  docker-e2e-tests:
+  # break docker-e2e-test into 6 tasks to speed up the tests.
+  docker-e2e-tests-part1:
```
vpiyush, line 311: We don't need to manually break it down into 6 parts. We can utilize the matrix strategy and partition the tests into 6 different parts. we could use more if we like.  e.g.: 
matrix:
  partition: [0, 1, 2, 3, 4, 5]
for the run part:
make docker-e2e-tests START=${{ matrix.partition * 4 }} END=${{ (matrix.partition + 1) * 4 }}

autonity/autonity.go
```diff
@@ -233,7 +234,7 @@ func (c *AutonityContract) GetCommitteeByHeight(header *types.Header, db vm.Stat
 
 // EpochInfo get the committee and the corresponding epoch boundary base on the input header's state.
 // it returns the committee, previousEpochBlock, curEpochBlock, and the nextEpochBlock.
-func (c *AutonityContract) EpochInfo(header *types.Header, db vm.StateDB) (*types.Committee, uint64, uint64, uint64, error) {
+func (c *AutonityContract) EpochInfo(header *types.Header, db vm.StateDB) (*types.Committee, uint64, uint64, uint64, uint64, error) {
```
vpiyush, line 237: the comments also needs an update of return params

consensus/tendermint/backend/engine.go
```diff
@@ -82,19 +87,29 @@ func (sb *Backend) VerifyHeader(chain consensus.ChainHeaderReader, header *types
 		return consensus.ErrUnknownAncestor
 	}
 
-	// get latest epoch info for signature checks and epoch boundary checks latter on.
-	committee, _, curEpochHead, nextEpochHead, err := chain.LatestEpoch()
+	// get latest epoch info for signature checks and epoch boundary checks later on.
```
Jason-Zhangxin-Chen, line 90: I am thinking if we can replace this LatestEpoch() with EpochOfHeight().

consensus/tendermint/backend/engine.go
```diff
@@ -323,11 +355,66 @@ func (sb *Backend) Prepare(chain consensus.ChainHeaderReader, header *types.Head
...
+
+	// for the 1st delta blocks of the epoch, the proposer does not have to provide an activity proof
+	if h <= epochBlock+delta {
+		sb.logger.Debug("Skip to assemble activity proof at the start of epoch", "height", h, "epochBlock", epochBlock)
+		return nil, 0, nil
+	}
+
+	// after delta blocks, get quorum certificates from height h-delta.
+	targetHeight := h - delta
+	targetHeader := sb.BlockChain().GetHeaderByNumber(targetHeight)
```
Jason-Zhangxin-Chen, line 386: GetHeaderByNumber() might return nil in some edge cases, it is worth to check it.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,435 @@
...
+
+    mapping(address => uint256) public proposerEffort;
+
+    // epoch inactivity score for each committee member. Updated at every epoch.
+    mapping(address => uint256) public inactivityScores;
+
+    mapping(address => uint256) public probationPeriods; // in epochs
+    mapping(address => uint256) public repeatedOffences; // reset as soon as an entire probation period is completed without offences.
+
+    uint256[] public epochCollusionDegree; // maps epoch number to the collusion degree
```
Jason-Zhangxin-Chen, line 47: From the comment left, it seems a mapping from epoch id to the collusion degress, however it saves the epoch collusion degree one by one by pushing it back to this array. So I guess we save per epoch collusion degree at the end of epoch by appending it to the array.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,435 @@
...
+
+        uint256[1] memory _committeeSlot; // declare it as array to easily access from assembly
+        assembly{
+            mstore(_committeeSlot,committee.slot)
+        }
+
+        (bool _isProposerOmissionFaulty, uint256 _proposerEffort, address[] memory _absentees) = Precompiled.computeAbsentees(_mustBeEmpty, config.delta, _committeeSlot[0]);
+
+        // short-circuit function if the proof has to be empty
+        if(_mustBeEmpty){
```
Jason-Zhangxin-Chen, line 85: Just cirious, why wouldn't we put this short-circuit checker before the compute of absentees?

autonity/solidity/contracts/Autonity.sol
```diff
@@ -15,14 +15,15 @@ import "./asm/IACU.sol";
...
+import "./interfaces/IOmissionAccountability.sol";
 import "./interfaces/IOracle.sol";
 import "./interfaces/IAutonity.sol";
 import "./interfaces/IInflationController.sol";
 import "./interfaces/ILiquidLogic.sol";
 import "./ReentrancyGuard.sol";
 
 /** @title Proof-of-Stake Autonity Contract */
-enum ValidatorState {active, paused, jailed, jailbound}
+enum ValidatorState {active, paused, jailed, jailbound, jailedForInactivity, jailboundForInactivity}
```
Jason-Zhangxin-Chen, line 26: Regarding the state transition of a validator in its lifecycle, can we have a diagram to  describe it? It would be very helpful for Jay and the end user if we can present this into the official document.

autonity/solidity/contracts/OmissionAccountability.sol
Jason-Zhangxin-Chen: For the time being, this `epochCollusionDegree` is not used yet at anywhere.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,435 @@
...
+    * @notice called by the Autonity contract at epoch finalization, to redistribute the proposer rewards based on the effort
+    * @param _ntnRewards, amount of NTN reserved for proposer rewards
+    */
+    function distributeProposerRewards(uint256 _ntnReward) external payable virtual onlyAutonity {
+        uint256 atnReward = msg.value;
+
+        for(uint256 i=0; i < committee.length; i++) {
+            address nodeAddress = committee[i].addr;
+            if(proposerEffort[nodeAddress] > 0){
+               uint256 atnProposerReward = (proposerEffort[nodeAddress] * atnReward) / totalEffort;
```
Jason-Zhangxin-Chen, line 254: As we comput just net effort, thus the totalEffort could be zero in some valid edge case, this will revert the finalize() of the epoch block which is a potential liveness issue?

autonity/solidity/contracts/Autonity.sol
cmjc: Hi @Jason-Zhangxin-Chen - there is an issue to add diagrams to the docs here - https://github.com/autonity/docs.autonity.org/issues/104. I will add this to the listing of diagrams to add under Validator. 👍

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,435 @@
...
+        operator = _operator;
+    }
+
+    // config update methods
+
+    /* @notice sets the inactivity threshold
+    * @dev restricted to the operator
+    * @param _inactivityThreshold, the new value for inactivity threshold
+    */
+    function setInactivityThreshold(uint256 _inactivityThreshold) external virtual onlyOperator {
```
Jason-Zhangxin-Chen, line 353: In the setter() method below, should assigning a value of `zero` be considered a critical/sucide operation? Is it necessary to implement a check for this scenario?

Jason-Zhangxin-Chen (commented):
LGTM, just left few comments and quesions there, good work!

.github/workflows/go.yml
Jason-Zhangxin-Chen: Thank you piyush, I will take this action. I will try it in this way in another PR.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1162,8 +1208,27 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
+            _transfer(address(this), address(config.contracts.omissionAccountabilityContract), ntnProposerRewards);
+            config.contracts.omissionAccountabilityContract.distributeProposerRewards{value: atnProposerRewards}(ntnProposerRewards);
+            _atn -= atnProposerRewards;
+            _ntn -= ntnProposerRewards;
+        }
+
+        uint256 omissionScaleFactor = config.contracts.omissionAccountabilityContract.getScaleFactor();
+
         // Redistribute fees through the Liquid Newton contract
         atnTotalRedistributed += _atn;
```
tbssajal, line 1229: I think this line should be above `if(config.contracts.omissionAccountabilityContract.getTotalEffort() > 0)` because those atn distributed to proposer are still distributed. I am not sure but I guess only the atn going to protocol treasury are excluded from `atnTotalRedistributed` variable. This makes me question if we need any state variable for distributed ntn.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1203,6 +1291,19 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
                 emit Rewarded(_val.nodeAddress, _atnReward, _ntnReward);
             }
         }
+
+        // send withheld funds to the appropriate pool
+        if (atnTotalWithheld > 0) {
+            // Using "call" to let the treasury contract do any kind of computation on receive.
+            (bool _sent, bytes memory _returnData) = config.policy.withheldRewardsPool.call{value: atnTotalWithheld}("");
```
tbssajal, line 1298: Although this address `config.policy.withheldRewardsPool` might be set by us, maybe it would be safe to put a gas value in this call?

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -13,6 +14,33 @@ library Precompiled {
...
+        uint256[1] memory _outputLength; // need to declare it as array to be able to easily access it from assembly
+
+        assembly {
+            // the call doesn't modify omission state, but it reads the committee from it
+            // however we could actually use a static-call, it doesn't change anything from the precompile point of view
+            //delegatecall(gasLimit, to, inputOffset, inputSize, outputOffset, outputSize)
+            if iszero(delegatecall(gas(), to, add(_input,32), mload(_input), 0, 0)) {
+                revert(0, 0)
+            }
+            mstore(_outputLength,returndatasize())
```
tbssajal, line 31: I guess `_outputLength` stores the size of the array. I guess you need `mstore(add(_outputLength, 32),returndatasize())` but you can check with tests to be sure.

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -13,6 +14,33 @@ library Precompiled {
...
+        }
+
+        bytes memory _output = new bytes(_outputLength[0]);
+
+        assembly{
+            returndatacopy(add(_output,32), 0, mload(_outputLength))
```
tbssajal, line 37: Same here. Maybe you need `mload(add(_outputLength,32))`.

tbssajal (changes requested):

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,435 @@
...
+    * @param _ntnRewards, amount of NTN reserved for proposer rewards
+    */
+    function distributeProposerRewards(uint256 _ntnReward) external payable virtual onlyAutonity {
+        uint256 atnReward = msg.value;
+
+        for(uint256 i=0; i < committee.length; i++) {
+            address nodeAddress = committee[i].addr;
+            if(proposerEffort[nodeAddress] > 0){
+               uint256 atnProposerReward = (proposerEffort[nodeAddress] * atnReward) / totalEffort;
+               uint256 ntnProposerReward = (proposerEffort[nodeAddress] * _ntnReward) / totalEffort;
```
tbssajal, line 255: You can handle the dust by doing
```
totalEffort -= proposerEffort[nodeAddress];
atnReward -= atnProposerReward;
_ntnReward -= ntnProposerReward;
```

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: Tbh I am not sure what is the purpose of this variable, since it is not used in the logic and seems just a "stats" variable. Do you know if we use it for something outside the client?

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: I am not sure, as we might want to do some computations when the fund arrive. We use the same approach for the Autonity treasury:
```
// Using "call" to let the treasury contract do any kind of computation on receive.
(bool sent,) = config.policy.treasuryAccount.call{value: _atnTreasuryReward}("");
```

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: I currently handled it by using the contract balances as a base for the computation instead of the sent values. It is simpler it code terms, but I wonder if your solution is more fair?

autonity/autonity.go
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/e4c6eaa28c0fdec38ee198504cec5d980f8b98b5

consensus/tendermint/backend/engine.go
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/e4c6eaa28c0fdec38ee198504cec5d980f8b98b5

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: This variable is just for facilitating validators in re-computing and verifying their slashing rate when and if slashing occurs. It is not used by the protocol.

The data structure is an array but it maps the epoch number to the collusion degree, meaning that:
- collusion degree of epoch 0 is in epochCollusionDegree[0]
- collusion degree of epoch 1 is in epochCollusionDegree[1]

etc. etc.

Does this clarify your doubts?

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: Because if the proof needs to be empty, once we have verified that it is we do not have anything more to do

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: I think a value of zero is a valid value for the inactivity threshold. It would mean no tolerance, which is allowed and handled by the protocol.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -583,13 +593,29 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
+        config.policy.proposerRewardRate = _proposerRewardRate;
+    }
+
+    function setWithholdingThreshold(uint256 _withholdingThreshold) public virtual onlyOperator {
+        require(_withholdingThreshold <= WITHHOLDING_THRESHOLD_PRECISION,"Cannot exceed 100%");
+        config.policy.withholdingThreshold = _withholdingThreshold;
+    }
+
+    function setWithheldRewardsPool(address payable pool) public virtual onlyOperator {
+        config.policy.withheldRewardsPool = pool;
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/e4c6eaa28c0fdec38ee198504cec5d980f8b98b5

autonity/solidity/contracts/Accountability.sol
```diff
@@ -504,7 +432,7 @@ contract Accountability is IAccountability {
     }
 
     function _baseSlashingRate(uint256 _severity) internal view returns (uint256) {
-        //
+        // TODO: complete
```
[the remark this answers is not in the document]
lorenzo-dev1: It is to remember that we have to fix this function, it currently returns the same slashing rate for all severities

cmd/gengen/gengen/new_genesis.go
```diff
@@ -56,6 +57,26 @@ type Validator struct {
...
+	err = json.Unmarshal(jsonBytes, genesisCopy)
+	if err != nil {
+		panic("cannot unmarshal genesis config: " + err.Error())
+	}
+	// deep copying through json marshaling kinda messes up the ABIs (we have `{}` instead of `nil` for empty elements)
+	// so let's just copy them from the original, we do not modify them in tests anyways for now
+	// TODO: find a better solution to deep copy also the ABIs.
+	// I suspect it is not straightforward due to how the generated ABIs are computed.
+	// See gen-contract target in Makefile and `TestAPIGetContractABI`
+	genesisCopy.AutonityContractConfig.ABI = original.AutonityContractConfig.ABI
```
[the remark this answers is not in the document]
lorenzo-dev1: I don't understand, wouldn't the pointer always point to the same memory area anyways?

consensus/tendermint/backend/engine.go
```diff
@@ -181,20 +196,64 @@ func (sb *Backend) verifyHeaderAgainstLastView(header, parent *types.Header, com
...
+	if header.ActivityProof != nil {
+		targetHeight := header.Number.Uint64() - delta
+		targetHash, err := hash(targetHeight)
+		if err != nil {
+			return err
+		}
+		headerSeal := message.PrepareCommittedSeal(targetHash, int64(header.ActivityProofRound), new(big.Int).SetUint64(targetHeight))
+		_, _, err = header.ActivityProof.Validate(headerSeal, committee, true)
+		if err != nil {
+			return errors.Join(errInvalidActivityProof, err)
```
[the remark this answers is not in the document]
lorenzo-dev1: Re: this comment and https://github.com/autonity/autonity/pull/1021#discussion_r1808508967, I am not sure these diagnostic information would be useful. This error will pop up when we are verifying an invalid header, so even adding number information wouldn't help us much (since the invalid header will just be discarded). I think the only useful thing would be to dump the entire header.

core/types/block.go
```diff
@@ -105,12 +110,53 @@ type AggregateSignature struct {
...
+func (a *AggregateSignature) Validate(message common.Hash, committee *Committee, checkQuorum bool) (map[common.Address]struct{}, *big.Int, error) {
+	// validate signers information first
+	if _, err := a.Signers.validate(committee.Len()); err != nil {
+		return nil, nil, fmt.Errorf("invalid signers information: %w", err)
+	}
+
+	// verify signature
+	var keys []blst.PublicKey //nolint
+	for _, index := range a.Signers.flatten(committee.Len()) {
+		keys = append(keys, committee.Members[index].ConsensusKey)
```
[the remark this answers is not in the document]
lorenzo-dev1: I can't, I don't know how many keys will be there in advance

autonity/tests/omission_accountability_test.go
```diff
@@ -0,0 +1,1124 @@
...
+	"github.com/stretchr/testify/require"
+	"math"
+	"math/big"
+	"math/rand"
+	"testing"
+)
+
+var omissionEpochPeriod = 130
+
+const ScaleFactor = 10_000           // must match the scale factor used in OmissionAccountability.sol
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/064ba8e150712627cc918d67fc6f33c701966541

consensus/tendermint/backend/engine.go
```diff
@@ -323,11 +355,66 @@ func (sb *Backend) Prepare(chain consensus.ChainHeaderReader, header *types.Head
...
+	targetHeader := sb.BlockChain().GetHeaderByNumber(targetHeight)
+	targetRound := targetHeader.Round
+
+	precommits := sb.MsgStore.GetPrecommits(targetHeight, func(m *message.Precommit) bool {
+		return m.R() == int64(targetRound) && m.Value() == targetHeader.Hash()
+	})
+
+	// we should have provided an activity proof, but we do not have past messages
+	if len(precommits) == 0 {
+		sb.logger.Warn("Failed to provide activity valid activity proof as proposer", "height", h, "targetHeight", targetHeight)
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/064ba8e150712627cc918d67fc6f33c701966541

consensus/tendermint/backend/engine.go
```diff
@@ -323,11 +355,66 @@ func (sb *Backend) Prepare(chain consensus.ChainHeaderReader, header *types.Head
...
+	for i, p := range precommits {
+		votes[i] = p
+	}
+
+	aggregatePrecommit := message.AggregatePrecommits(votes)
+
+	// if we do not have enough voting power, leave the proof empty
+	quorum := bft.Quorum(committee.TotalVotingPower())
+	if aggregatePrecommit.Power().Cmp(quorum) < 0 {
+		sb.logger.Warn("Failed to provide activity valid activity proof as proposer, not enough voting power", "height", h, "targetHeight", targetHeight, "power", aggregatePrecommit.Power(), "quorum", quorum)
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/064ba8e150712627cc918d67fc6f33c701966541

core/types/block.go
```diff
@@ -105,12 +110,53 @@ type AggregateSignature struct {
...
+		panic("Failed to aggregate keys from committee members: " + err.Error())
+	}
+	valid := a.Signature.Verify(aggregatedKey, message[:])
+	if !valid {
+		return nil, nil, errInvalidSignature
+	}
+
+	// Total assembled voting power for the activity proof
+	power := new(big.Int)
+	signers := make(map[common.Address]struct{})
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/064ba8e150712627cc918d67fc6f33c701966541

core/types/signers.go
```diff
@@ -407,9 +417,12 @@ func (s *Signers) AssignPower(powers map[int]*big.Int, power *big.Int) {
 }
 
 func (s *Signers) Copy() *Signers {
-	powers := make(map[int]*big.Int)
-	for index, power := range s.powers {
-		powers[index] = new(big.Int).Set(power)
+	var powers map[int]*big.Int
+	if s.powers != nil {
+		powers = make(map[int]*big.Int)
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/064ba8e150712627cc918d67fc6f33c701966541

consensus/tendermint/backend/engine.go
lorenzo-dev1: fixed

consensus/tendermint/backend/engine.go
```diff
@@ -162,13 +177,13 @@ func (sb *Backend) verifyHeader(chain consensus.ChainHeaderReader, header, paren
 	if parent == nil {
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed

autonity/solidity/contracts/OmissionAccountability.sol
tbssajal: > but I wonder if your solution is more fair?

It doesn't bother me because it would be a couple of wei. At best maybe 1000 wei if we have 1000 proposers.

> I currently handled it by using the contract balances as a base for the computation instead of the sent values.

In that case I guess we just need to be careful when adding any other payable functions, so that the amount of ATN transferred is handled immediately. Although we cannot control or monitor the transfer of NTN.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -330,14 +339,14 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
         }
     }
 
-    function finalizeInitialization() onlyProtocol nonReentrant public {
+    function finalizeInitialization(uint256 delta) onlyProtocol nonReentrant public {
```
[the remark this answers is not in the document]
lorenzo-dev1: as we discussed I added a safety check https://github.com/autonity/autonity/pull/1021/commits/863a086249187d7e7c87d3594760b732dbe08602

autonity/solidity/contracts/Autonity.sol
tbssajal: I am not aware of its usage. But it is a public variable so it could have some usage to others like tracking how much atn tokens have been distributed since genesis. So I think the variable should reflect the definition.

Jason-Zhangxin-Chen (approved):
LGTM now! Thank you Lorenzo.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1162,8 +1208,27 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
                 _atn -= _atnTreasuryReward;
             }
         }
+
+        // proposer fees redistribution based on effort put into activity proofs
+        // if the total effort is 0, just redistribute the proposer rewards based on stake
+        // NOTE: reward forfeiting and withholding based on accountability and omission accountability are not applied to proposer rewards
+        // e.g. a validator punished for equivocation will still receive his share of proposer rewards
+        if(config.contracts.omissionAccountabilityContract.getTotalEffort() > 0){
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: No, it is okay.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,435 @@
...
+                if(!inactiveValidators[h][_absentees[i]]){
+                    confirmedAbsent = false;
+                    break;
+                }
+            }
+            // if the absentee was absent for the entirety of the lookback period, increment his inactivity counter
+            if (confirmedAbsent) {
+                inactivityCounter[_absentees[i]]++;
+            }
+        }
```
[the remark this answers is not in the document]
lorenzo-dev1: this logic has been implemented as discussed

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -13,6 +14,33 @@ library Precompiled {
...
+            // the call doesn't modify omission state, but it reads the committee from it
+            // however we could actually use a static-call, it doesn't change anything from the precompile point of view
+            //delegatecall(gasLimit, to, inputOffset, inputSize, outputOffset, outputSize)
+            if iszero(delegatecall(gas(), to, add(_input,32), mload(_input), 0, 0)) {
+                revert(0, 0)
+            }
+            mstore(_outputLength,returndatasize())
+        }
+
+        bytes memory _output = new bytes(_outputLength[0]);
```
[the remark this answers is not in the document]
lorenzo-dev1: Could you provide an example on how to allocate a dynamic array in assembly? I am not sure on how to do this.

autonity/solidity/contracts/Precompiled.sol
```diff
@@ -13,6 +14,33 @@ library Precompiled {
     address constant public MISBEHAVIOUR_CONTRACT = address(0xfe);
     address constant public ENODE_VERIFIER_CONTRACT = address(0xff);
 
+
+    function computeAbsentees(bool _mustBeEmpty, uint256 _delta, uint256 _committeeSlot) internal returns (bool,uint256,address[] memory){
+        address to = ACTIVITY_CONTRACT;
+
+        bytes memory _input = abi.encodePacked(_mustBeEmpty, _delta, _committeeSlot);
+        uint256[1] memory _outputLength; // need to declare it as array to be able to easily access it from assembly
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed https://github.com/autonity/autonity/pull/1021/commits/2b5cefba6d690cb60af8c3013a55545a84ac6428

.github/workflows/go.yml
lorenzo-dev1: created issue https://github.com/autonity/autonity/issues/1058

core/types/block.go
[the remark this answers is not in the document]
vpiyush: I  meant doing something like this: 
```
	s := a.Signers.flatten(committee.Len())
	keys = make([]blst.PublicKey, 0, len(s))
	for _, index := range s {
		keys = append(keys, committee.Members[index].ConsensusKey)
	}
```

cmd/gengen/gengen/new_genesis.go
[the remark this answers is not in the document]
vpiyush: When you de-reference it will copy the content not the pointer itself. Though there are some maps for them only referneces would be copied. So just the de-reference is not sufficeint. I think you can leave this if not really needed.

autonity/solidity/contracts/Precompiled.sol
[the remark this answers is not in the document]
vpiyush: can you try this:  
```
bytes memory _output;
        assembly {
            _output := mload(0x40)
            mstore(_output, outputLength)
            let outputData := add(_output, 32)
            returndatacopy(outputData, 0, outputLength)
            mstore(0x40, add(outputData, outputLength))
        }
```

core/types/block.go
```diff
@@ -105,12 +110,53 @@ type AggregateSignature struct {
...
+	if _, err := a.Signers.validate(committee.Len()); err != nil {
+		return nil, nil, fmt.Errorf("invalid signers information: %w", err)
+	}
+
+	// verify signature
+	var keys []blst.PublicKey //nolint
+	for _, index := range a.Signers.flatten(committee.Len()) {
+		keys = append(keys, committee.Members[index].ConsensusKey)
+	}
+	aggregatedKey, err := blst.AggregatePublicKeys(keys)
```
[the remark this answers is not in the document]
lorenzo-dev1: The problem is what to use to index the cache. Because using indexes is not safe, as committee member can changes across epochs. The only other option is to use the concatenation of all the keys? but I think this way we would end up having very big keys.

Another option could be to hash the keys bytes and use the hash as key, but then we are adding the overhead of hashing and I am not sure whether we will get a net benefit

autonity/solidity/contracts/Precompiled.sol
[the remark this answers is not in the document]
lorenzo-dev1: Tbh I don't really see the benefit of doing that. I believe the gas cost should be the same, and I think it worsen the readability of the code. It is harder to understand what is going on and more error prone

cmd/gengen/gengen/new_genesis.go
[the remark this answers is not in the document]
lorenzo-dev1: Yeah from the point of view of the code execution it wouldn't change much, and this is already a quick-and-dirty workaround which should be substituted later on

core/types/block.go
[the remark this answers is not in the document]
lorenzo-dev1: aaah I see, fixed in https://github.com/autonity/autonity/pull/1021/commits/9a89dd3664fe19a9423f1e19465292fed1bb043c

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: Yeah... however I don't see any problem caused by someone sending NTN to the omission contract. Basically it will just get treated as additional rewards. Do you think that can cause issues?

core/types/block.go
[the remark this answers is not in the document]
vpiyush: I agree there is no straight-forward way to create a key to index this cache. we can do this exercise later on.

autonity/solidity/contracts/Precompiled.sol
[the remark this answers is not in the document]
vpiyush: Hmm, I think the gas cost differs because Solidity’s new initializes memory to zero, which doesn’t happen when we handle the free memory pointer ourselves. Although the difference may not be significant, I do see your point about code readability and somewhat agree with it. I’ll leave it to your discretion to decide if you’d like to keep it as is and resolve this conversation.

vpiyush (approved):
LGTM!!

autonity/solidity/contracts/Precompiled.sol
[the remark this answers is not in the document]
lorenzo-dev1: Yeah, since we are close to release I would keep it like this to avoid introducing problems. It can then be refactored later on if we want to optimize for gas consumption.

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.19;
+
+import "./Autonity.sol";
+
+contract Slasher {
+    Autonity private autonity;
```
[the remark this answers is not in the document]
lorenzo-dev1: The slasher contract is now a separate protocol contract, so I believe this comment is out of date

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    uint256 private constant SLASHING_RATE_PRECISION = 10_000;
+
+    constructor(address payable _autonity){
+        autonity = Autonity(_autonity);
+    }
+
+    /**
+    * @notice generic slashing function
```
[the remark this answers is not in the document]
lorenzo-dev1: Agree, fixed

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.19;
+
+import "./Autonity.sol";
+
+contract Slasher {
+    Autonity private autonity;
+
+    uint256 private constant SLASHING_RATE_PRECISION = 10_000;
```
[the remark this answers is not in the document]
lorenzo-dev1: Agree, made it public

autonity/tests/omission_accountability_test.go
```diff
@@ -0,0 +1,1124 @@
...
+const SlashingRatePrecision = 10_000 // must match the precision used in Slasher.sol
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/bdb6366558b744399f35705790a5b6c1ce758b0e

lorenzo-dev1:
@Jason-Zhangxin-Chen @yazzaoui @vpiyush @tbssajal @ScottRMalley 

There are substantial new changes which came out of review discussion. The most important are:
1. relegation of the slashing functionality in a separate protocol contract
2. major refactoring of the `recordAbsentee` in the omission contract
3. fixing the GetHashFn function to work with optimistic block building
4. adaptation of activity proof assembling to optimistic block building

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,195 @@
...
+    address private autonity;
+
+    uint256 public constant SLASHING_RATE_PRECISION = 10_000;
+
+    constructor(address _autonity){
+        autonity = _autonity; // we could use msg.sender but it would make deploying a new upgraded slasher contract harder
+    }
+
+    /**
+    * @dev modifies the passed validator structure to enable jailing
```
tbssajal, line 16: I guess for public function we should use @notice instead of @dev and also mention that the function is restricted to autonity only. We can add @dev for some comments that are only needed for developers.

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,195 @@
...
+        Autonity.Validator memory _val,
+        uint256 _jailtime,
+        ValidatorState _newJailedState
+    ) internal virtual {
+        _val.jailReleaseBlock = block.number + _jailtime;
+        _val.state = _newJailedState;
+    }
+
+    /**
+    * @dev modifies the passed validator structure to enable jailbounding (perpetual jail)
```
tbssajal, line 43: Same here, maybe notice would fit.

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,195 @@
...
+        Autonity.Validator memory _val,
+        ValidatorState _newJailboundState
+    ) internal virtual {
+        _val.jailReleaseBlock = 0;
+        _val.state = _newJailboundState;
+    }
+
+
+    /**
+    * @dev applies slashing to the passed validator struct
```
tbssajal, line 68: Same here

autonity/solidity/contracts/Slasher.sol
```diff
@@ -0,0 +1,195 @@
...
+            _val.bondedStake = 0;
+            _val.selfBondedStake = 0;
+            _val.selfUnbondingStake = 0;
+            _val.unbondingStake = 0;
+            _val.totalSlashed += availableFunds;
+            _jailbound(_val, _newJailboundState);
+            return (_val, availableFunds, true);
+        }
+
+        uint256 slashingAmount = _slash(_val, _slashingRate);
```
tbssajal, line 184: I am not entirely sure but I think the `_val` is being copied because it is in memory. Then the changes done in `_slash` will not reflect here. Did you test it?

vpiyush:
Reviewed new changes as well. LGTM!

autonity/solidity/contracts/Slasher.sol
vpiyush: I think it's not being reflected.  Also Since it's a big struct may be passing using memory is not the most efficient.

tbssajal: Also it seems to me that slasher doesn't need to be a separate contract because only autonity calls it, there is no external usage. It can be an abstract contract and part of the Autonity contract. Then we can just pass the storage pointer.

autonity/solidity/contracts/OmissionAccountability.sol
tbssajal: Nope.

autonity/solidity/contracts/Autonity.sol
tbssajal: I think `atnTotalRedistributed` accounts for the redistribution of atn fees from transactions, except the amount that goes to the treasury account. But @yazzaoui can cofirm.

autonity/solidity/contracts/Slasher.sol
tbssajal: LOL! Just noticed my previous comment to use `@dev` instead of `@notice`, I guess the function was internal at that time.

lorenzo-dev1: mmh, I choose @dev because reading from here https://docs.soliditylang.org/en/latest/natspec-format.html#tags it says "@notice Explain to an end user what this does". But these function are all "protocol-only" so users will never need to call them. I think this info is more relevant to developers? but tbh I do not have a strong opinion on it

autonity/solidity/contracts/Slasher.sol
lorenzo-dev1: The memory is not copied if the function is in the same contract of the caller. It is copied only in calls across contracts. You can verify the behaviour in Remix.

For what concerns a separate slasher contract, it is just for making it easier to upgrade the slashing logic if we want to. We can just deploy a new slasher contract and point autonity to use that one

vpiyush: > The memory is not copied if the function is in the same contract of the caller. It is copied only in calls across contracts. You can verify the behaviour in Remix.

Ah, I wasn't aware of this. Good to know. Thanks :)

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1133,6 +1298,11 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
     ============================================================
     */
 
+    function _getValidator(address _addr) internal virtual view returns (Validator memory){
```
tbssajal, line 1301: I think for zero address it will not revert. Although it doesn't make sense to query withe zero address.

core/vm/contracts.go
```diff
@@ -1358,6 +1364,136 @@ func (c checkEnode) Run(input []byte, _ uint64, _ *EVM, _ common.Address) ([]byt
 	return out, nil
 }
 
+func readCommittee(db StateDB, caller common.Address, committeeSlot common.Hash) *types.Committee {
```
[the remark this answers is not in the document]
lorenzo-dev1: yeah that's true, I will add a comment for it. I think also the `committeeSelector` relies on it right?

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    // config update methods
+
+    /* @notice sets the inactivity threshold
+    * @dev restricted to the operator
+    * @param _inactivityThreshold, the new value for inactivity threshold
+    */
+    function setInactivityThreshold(uint256 _inactivityThreshold) external virtual onlyOperator {
+        require(_inactivityThreshold <= SCALE_FACTOR, "cannot exceed scale factor");
```
tbssajal, line 413: I think similar check can be done in constructor as well.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+        config.inactivityThreshold = _inactivityThreshold;
+    }
+
+    /* @notice sets the past performance weight
+    * @dev restricted to the operator
+    * @param _pastPerformanceWeight, the new value for the past performance weight
+    */
+    function setPastPerformanceWeight(uint256 _pastPerformanceWeight) external virtual onlyOperator {
+        require(_pastPerformanceWeight <= SCALE_FACTOR, "cannot exceed scale factor");
+        require(_pastPerformanceWeight <= config.inactivityThreshold, "pastPerformanceWeight cannot be greater than inactivityThreshold");
```
tbssajal, line 423: I think similar check can be done in constructor as well.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+    /* @notice sets delta. It will get updated at epoch end
+    * @dev restricted to the operator
+    * @param _delta, the new value for delta
+    */
+    function setDelta(uint256 _delta) external virtual onlyOperator {
+        require(_delta >= 2, "delta needs to be at least 2"); // cannot be 1 due to optimistic block building
+        uint256 epochPeriod = autonity.getEpochPeriod();
+
+        // utilize newLookbackWindow for comparison, so that if delta is also being changed in this epoch we take the new value
+        require(epochPeriod > _delta + newLookbackWindow - 1, "epoch period needs to be greater than delta+lookbackWindow-1");
```
tbssajal, line 474: A similar check for `delta` and `lookbackWindow` can be done in constructor as well.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+    function setInitialProbationPeriod(uint256 _initialProbationPeriod) external virtual onlyOperator {
+        config.initialProbationPeriod = _initialProbationPeriod;
+    }
+
+    /* @notice sets the initial slashing rate
+    * @dev restricted to the operator
+    * @param _initialSlashingRate, the new value for the initial slashing rate
+    */
+    function setInitialSlashingRate(uint256 _initialSlashingRate) external virtual onlyOperator {
+        require(_initialSlashingRate <= SCALE_FACTOR, "cannot exceed scale factor");
```
tbssajal, line 448: I am not sure about the relation between `config.initialSlashingRate` and `SCALE_FACTOR`. Although `SLASHING_RATE_PRECISION` from `Slasher` can be used here I guess.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+                (
+                    inactivityCounter[nodeAddress] * SCALE_FACTOR * currentPerformanceWeight
+                    + inactivityScores[nodeAddress] * config.pastPerformanceWeight * qualifiedBlocks
+                )
+                / (SCALE_FACTOR * qualifiedBlocks);
+
+            if (aggregatedInactivityScore > config.inactivityThreshold) {
+                collusionDegree++;
+            }
+            inactivityScores[nodeAddress] = aggregatedInactivityScore;
```
tbssajal, line 235: `inactivityScores` is never reduced, is that on purpose? So `inactivityScores` always impacts new score and if someone's `aggregatedInactivityScore` exceeds `config.inactivityThreshold` and gets jailed, technically we can jail him forever if we want by setting `config.pastPerformanceWeight = SCALE_FACTOR`.

core/vm/contracts.go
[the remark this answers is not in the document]
tbssajal: No I think you added this function. `committeeSelector` does not read committee.

tbssajal (changes requested):

core/vm/contracts.go
```diff
@@ -1358,6 +1364,136 @@ func (c checkEnode) Run(input []byte, _ uint64, _ *EVM, _ common.Address) ([]byt
...
+
+	out := make([]byte, length)
+	if isProposerFaulty {
+		copy(out[0:DataLen], common.LeftPadBytes([]byte{1}, DataLen))
+	} else {
+		copy(out[0:DataLen], common.LeftPadBytes([]byte{0}, DataLen))
+	}
+	copy(out[DataLen:DataLen*2], common.LeftPadBytes(proposerEffort.Bytes(), DataLen))
+
+	copy(out[DataLen*2:DataLen*3], common.LeftPadBytes(new(big.Int).SetUint64(DataLen*3).Bytes(), DataLen))              // offset
```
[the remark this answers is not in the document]
lorenzo-dev1: By reading solidity docs I figured out that dynamic length arrays have two "metadata" bytes at the end, the offset and the length of the array.

The offset points at where the array starts (at the byte where the length is stored). So for example in the case of this return data we will have:
isProposerFaulty | proposerEffort | offset | length | array elements ...

and offset will always be 96 because the length slot of the array always starts at the 96th byte in the stream.

I am not sure why they introduced it but that's what it is needed for `abi.decode` to work on the solidity side

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: As discussed in standup, I removed this variable in https://github.com/autonity/autonity/pull/1021/commits/afb276d64a47709e107b38120f834f5a8b5aa703

core/vm/contracts.go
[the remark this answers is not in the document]
lorenzo-dev1: Added comments in https://github.com/autonity/autonity/pull/1021/commits/afb276d64a47709e107b38120f834f5a8b5aa703

core/vm/contracts.go
```diff
@@ -1358,6 +1364,136 @@ func (c checkEnode) Run(input []byte, _ uint64, _ *EVM, _ common.Address) ([]byt
...
+	for i := 0; i < committeeSize; i++ {
+		nodeAddress := common.BytesToAddress(db.GetState(caller, committeeOffset).Bytes())
+		committeeOffset = common.BytesToHash(new(big.Int).Add(committeeOffset.Big(), common.Big1).Bytes())
+		votingPower := db.GetState(caller, committeeOffset).Big()
+		committeeOffset = common.BytesToHash(new(big.Int).Add(committeeOffset.Big(), common.Big1).Bytes())
+		consensusKeySlot := crypto.Keccak256Hash(committeeOffset.Bytes())
+		consensusKeyBytes1 := db.GetState(caller, consensusKeySlot).Bytes()
+		consensusKeySlot = common.BytesToHash(new(big.Int).Add(consensusKeySlot.Big(), common.Big1).Bytes())
+		consensusKeyBytes2 := db.GetState(caller, consensusKeySlot).Bytes()
+		consensusKeyBytes := append(consensusKeyBytes1, consensusKeyBytes2[:blst.BLSPubkeyLength-DataLen]...)
```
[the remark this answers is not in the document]
lorenzo-dev1: added comment in https://github.com/autonity/autonity/pull/1021/commits/afb276d64a47709e107b38120f834f5a8b5aa703

core/vm/contracts.go
```diff
@@ -1358,6 +1364,136 @@ func (c checkEnode) Run(input []byte, _ uint64, _ *EVM, _ common.Address) ([]byt
...
+		return nil, errBadInput
+	}
+	if !(input[0] == byte(0) || input[0] == byte(1)) {
+		return nil, errBadInput
+	}
+	mustBeEmpty := input[0] == byte(1)
+	delta := new(big.Int).SetBytes(input[1:33])
+	committeeSlot := common.BytesToHash(input[33:])
+
+	committee := readCommittee(evm.StateDB, caller, committeeSlot)
```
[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/afb276d64a47709e107b38120f834f5a8b5aa703

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+        uint256 targetHeight = block.number - config.delta;
+
+        // decrease the faulty proposer count if the one dropping out of the window is faulty
+        // window: (h - delta - lookback - faultyProposers, h - delta]
+        if (
+            (targetHeight > epochBlock + config.lookbackWindow + faultyProposersInWindow) &&
+            (faultyProposers[targetHeight - config.lookbackWindow - faultyProposersInWindow])
+        ) {
+            faultyProposersInWindow--;
+        }
```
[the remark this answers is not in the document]
lorenzo-dev1: yeah I think you are right... I am rethinking about it and now I am not even sure that we need to count the previous `faultyProposersInWindow` in the if... I will write some tests and verify

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+    function _computeInactivityScoresAndCollusionDegree() internal virtual returns (uint256) {
+        uint256 epochPeriod = autonity.getCurrentEpochPeriod();
+        uint256 collusionDegree = 0;
+
+        // first config.lookbackWindow-1 blocks of the epoch are accountable, but we do not have enough info to determine if a validator was offline/online
+        // last delta blocks of the epoch are not accountable due to committee change
+        uint256 qualifiedBlocks = epochPeriod - config.lookbackWindow + 1 - config.delta;
+
+        // weight of current epoch performance
+        uint256 currentPerformanceWeight = SCALE_FACTOR - config.pastPerformanceWeight;
```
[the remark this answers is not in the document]
lorenzo-dev1: If you check `config.go` Prepare() function you will see that there is this check:
```
	if c.OmissionAccountabilityConfig.PastPerformanceWeight > c.OmissionAccountabilityConfig.InactivityThreshold {
		return fmt.Errorf("PastPerformanceWeight is too high. PastPerformanceWeight: %d, InactivityThreshold: %d", c.OmissionAccountabilityConfig.PastPerformanceWeight, c.OmissionAccountabilityConfig.InactivityThreshold)
	}
```

so we have this guarantee. For the fact that SCALE_FACTOR needs to be greater than the values in the config, I think in other parts of the codebase we assume that without any checks (e.g. the treasury fee). But I think it doesn't hurt to make additional checks.

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+            *  A more intuitive but equivalent construction:
+            *  aggregatedInactivityScore = (currentInactivityScore * currentPerformanceWeight + pastInactivityScore * pastPerformanceWeight) / SCALE_FACTOR
+            *  with currentInactivityScore = (currentInactivityCounter * SCALE_FACTOR) / qualifiedBlocks
+            */
+            uint256 aggregatedInactivityScore =
+                (
+                    inactivityCounter[nodeAddress] * SCALE_FACTOR * currentPerformanceWeight
+                    + inactivityScores[nodeAddress] * config.pastPerformanceWeight * qualifiedBlocks
+                )
+                / (SCALE_FACTOR * qualifiedBlocks);
```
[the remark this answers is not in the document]
lorenzo-dev1: Yeah I wrote this way because I think it is better for precision to do multiplcations first and then divisions. Otherwise you lose more precision

autonity/solidity/contracts/OmissionAccountability.sol
```diff
@@ -0,0 +1,493 @@
...
+            // validator is having a streak of offline blocks
+            if (uint256(lastActive[_absentees[i]]) <= _targetHeight - (config.lookbackWindow + faultyProposersInWindow)) {
+                // the validator was not active at some point in the lookback window
+                inactivityCounter[_absentees[i]]++;
+            }
+        }
+    }
+
+    // returns collusion degree
+    function _computeInactivityScoresAndCollusionDegree() internal virtual returns (uint256) {
```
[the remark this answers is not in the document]
lorenzo-dev1: I researched that, and it seems to me that the convention is that only function parameters should have the `_` prefix. I might be wrong though

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: It is already done in the Config.Prepare() function, before deploying contracts

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: theoretically yes but I set the constraint that `pastPerformanceWeight < InactivityThreshold`. see the checks in config.go, Prepare function

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: It is already done in the Config.Prepare() function, before deploying contracts

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: Yeah that is a good point, I will change it as it should respect the precision defined by the slashing contract. I think this should be changed also in `Accountability.sol`, but that's for another PR.

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: added in https://github.com/autonity/autonity/pull/1021/commits/cc3713258b6cb12048040bb797976003b156db66

autonity/solidity/contracts/OmissionAccountability.sol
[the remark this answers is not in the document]
lorenzo-dev1: added in https://github.com/autonity/autonity/pull/1021/commits/cc3713258b6cb12048040bb797976003b156db66

autonity/solidity/contracts/OmissionAccountability.sol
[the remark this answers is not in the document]
ScottRMalley: Yeah there are actually a couple conventions for this. In general, if a function parameter has the possibility of colliding with a state variable, you can use either a `_singleUnderscorePrefix` or a `singleUnderscoreSuffix_`. You will often see function parameters with these, even in the absence of a collision, just to ensure that no future state variables would collide. (see [Solidity style guide](https://docs.soliditylang.org/en/latest/style-guide.html#avoiding-naming-collisions))

In the absence of a collision though, I'd say we don't need them for locally declared variables, and in general I haven't seen that used. That being said, currently in our code, we are often using locally declared variables with a `_singleUnderscorePrefix`, so I'd say that for code consistency, we should stick with what we currently have. Might be something later to define our own style guide for Solidity code.

autonity/solidity/contracts/OmissionAccountability.sol
tbssajal: So in genesis config we have `pastPerformanceWeight < InactivityThreshold`, but later it is possible for the operator to set `config.pastPerformanceWeight = SCALE_FACTOR`. Maybe we can avoid that by design by allowing max value of `config.pastPerformanceWeight` to be `SCALE_FACTOR - 1`, but it is not super important.

autonity/solidity/contracts/OmissionAccountability.sol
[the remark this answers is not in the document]
tbssajal: Yeah, they should not overflow.

autonity/solidity/contracts/Autonity.sol
tbssajal: >Do you think it can happen there to have a call to query the 0 validator?

It shouldn't happen.

lorenzo-dev1: added check in 9c6657b53271aaca004a1c6d820fbc9b6e462852

autonity/solidity/contracts/OmissionAccountability.sol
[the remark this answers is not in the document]
lorenzo-dev1: applied our temporary convention in 9c6657b53271aaca004a1c6d820fbc9b6e462852

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: Not really, we do the same check in the setter function for `pastPerformanceWeight`:
```
    function setPastPerformanceWeight(uint256 _pastPerformanceWeight) external virtual onlyOperator {
        require(_pastPerformanceWeight <= SCALE_FACTOR, "cannot exceed scale factor");
        require(_pastPerformanceWeight <= config.inactivityThreshold, "pastPerformanceWeight cannot be greater than inactivityThreshold");
        config.pastPerformanceWeight = _pastPerformanceWeight;
    }
```

lorenzo-dev1: check added in b747e4bcb1068e0f4c774dcb1201718d26f83228

autonity/solidity/contracts/OmissionAccountability.sol
lorenzo-dev1: I did the fix in 8a69311dee1866c07164147857f3dea3d367c84d

autonity/solidity/contracts/OmissionAccountability.sol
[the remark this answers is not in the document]
tbssajal: A scenario where two or more adjacent blocks have faulty proposers could trigger a bug.

Consider an example where `lookbackWindow = 10` and `faultyProposersInWindow = 0` and we have faulty proposers for both block 100 and 101. From block 101 we have `faultyProposersInWindow = 2` so we will look 12 blocks back from target height in the `_recordAbsentees` function. When we are at target height = 112, we will check if block 100 had a faulty proposer, and we will reduce `faultyProposersInWindow` by 1, then we move on. But at target height = 113, we cannot reduce `faultyProposersInWindow` anymore because we will look 11 blocks back at 113 - 11 = 102 instead of 101. So what we should have done is when we were at target height = 112, we should have also looked at block 101 and reduced `faultyProposersInWindow` by 1 again.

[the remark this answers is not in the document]
lorenzo-dev1: fixed in https://github.com/autonity/autonity/pull/1021/commits/0126b4b222b2a1b4f8464c08b8a510175f0afed5

tbssajal (approved):

core/types/bft.go
```diff
@@ -81,33 +67,3 @@ func ECRecover(header *Header) (common.Address, error) {
...
-
-// TODO: All these Write* functions do useless checks as we always create the input ourselves. Remove them?
-
-// WriteSeal writes the extra-data field of the given header with the given seals.
-func WriteSeal(h *Header, seal []byte) error {
-	if len(seal) != common.SealLength {
-		return ErrInvalidSignature
-	}
-	h.ProposerSeal = make([]byte, len(seal))
-	copy(h.ProposerSeal, seal)
```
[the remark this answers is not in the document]
lorenzo-dev1: yes exactly, if you check the `addSeal` function you will see the the seal is just a local variable created inside that same function. So there is no need for a deep copy.

PR #1029, opened 2024-07-24
don't crash the node if repeated sync failures + malicious syncers suspension

lorenzo-dev1:
Removes node panic in case of repeated sync failures. The node operator will still be notified via verbose warnings if this scenario is happening.

Moreover this PR also brings back suspension based on time (not blocks) and adds a 30 seconds suspension period for malicious syncers (peers that we try to sync with but provide invalid data / time out).

We need time-based suspension because adding a block-based suspension for a peer while syncing, basically means blacklisting that peer until we successfully sync (without chances to retry with the same peer after a while). This would be too harsh.


miner/miner.go
```diff
@@ -142,13 +142,13 @@ func (miner *Miner) update() {
 				miner.eth.Logger().Info("Chain syncing failed", "#failures", syncFailures, "shouldStart", miner.shouldStart, "canStart", miner.canStart)
 				// if we fail more than maxSyncFailures times consequently, assume we are under attack
 				if syncFailures >= maxSyncFailures {
-					miner.eth.Logger().Warn("************************** SYNC ATTACK DETECTED **************************")
+					miner.eth.Logger().Warn("************************** PROBLEM DETECTED ******************************")
```
vpiyush, line 145: Is the idea to keep printing these messages on every sync failure after the max failures until operator takes an action OR problem resolves by itself?

vpiyush (approved):
Overall LGTM. just a question and suggestion.

ScottRMalley (approved):
LTGM. My only thought was that we could stand to have some more sophisticated testing of the peer suspension behavior, but I think that is out of the scope of this PR.

miner/miner.go
lorenzo-dev1: The problem needs operator action to be resolved

lorenzo-dev1:
successfull CI run: https://github.com/autonity/autonity/actions/runs/11597356448

PR #1032, opened 2024-08-08
Refactor autonity contract tests

ScottRMalley:
This pull request represents a start to moving the Truffle tests into our Golang solidity test framework, as well as some upgrades to the testing framework to allow event emission testing.

This PR contains:
* reworked genesis sequence from autonity-package-refactor
* updates to the Golang smart contract testing framework to support event emissions
* implementation of the tests from autonity/solidity/test/autonity.js in Golang
* Simplification of the `autonity` package to remove calls that were not used


ScottRMalley:
This PR is being separated out into PR's for the updates to the genesis sequence, and updates to the golang testing framework. Genesis sequence PR is already available in #1106

PR #1033, opened 2024-08-21
Consensus pipeline optimizations

vpiyush:
Various optimizations:
1. Trigger block creation as soon as proposal is verified.
2. Remove pending snapshot creation
3. no update for non-validator nodes on new transaction
4. Cache processed state on verify proposal reuse the same later on processing blocks

Minor refactoring is still expected on this but logic is complete


Jason-Zhangxin-Chen:
> d on this but logic is complete

@yazzaoui @lorenzo-dev1 @cmjc @ScottRMalley @tbssajal : When reviewing this PR, another improvement comes to mind: initiating the proposal propagation earlier by allowing participants to send the current height precommit message to the proposer of the next height (consensus instance) with the highest priority. This means sending the precommit to the proposer of the next height first, allowing the proposer of the next height to commit the block earlier and thus finishing the new proposal propagation earlier as well. I believe that the proposal propagation is the most heavy OP in our tendermint communication channel. What is your opinion on this?

vpiyush:
> > d on this but logic is complete
> 
> @yazzaoui @lorenzo-dev1 @cmjc @ScottRMalley @tbssajal : When reviewing this PR, another improvement comes to mind: initiating the proposal propagation earlier by allowing participants to send the current height precommit message to the proposer of the next height (consensus instance) with the highest priority. This means sending the precommit to the proposer of the next height first, allowing the proposer of the next height to commit the block earlier and thus finishing the new proposal propagation earlier as well. I believe that the proposal propagation is the most heavy OP in our tendermint communication channel. What is your opinion on this?

I don’t believe there’s a practical way to prioritize data over TCP. One option might be to adjust the sequence of send calls to peers. However, since the current implementation already sends messages asynchronously, and goroutine scheduling is random, making that change likely wouldn’t have a significant impact. Also, In our tests, sending precommits are not a bottleneck, they reach in milliseconds. It's the proposal propogation which takes a lot of time

consensus/tendermint/core/propose.go
```diff
@@ -137,6 +144,8 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
 		return err
 	}
 
+	// notify miner
+	go c.Backend().ProposalVerified(proposal.Hash())
```
Jason-Zhangxin-Chen, line 148: QQ: On the quorumPrecommitsChecks of rules.go Line139, after we verified an old round proposal, should we need to notify the miner too?

vpiyush: I think that's redundant since it's an optimization and here  we either already have block Or ready to commit the block. So we don't optimize anything

miner/worker.go
```diff
@@ -462,11 +451,24 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+				continue
+			}
+			timestamp = time.Now().Unix()
+			if h, ok := w.engine.(consensus.Handler); ok {
+				h.NewChainHead()
+			}
+			lastBlock = head.Block.Hash()
+			commit(false, commitInterruptNewHead)
+
+		case blockHash := <-w.proposalVerifiedEventCh:
```
Jason-Zhangxin-Chen, line 465: May be it is too early to issue the NewChainHead event to the consensus engine from here?Because in this context, we just know that a proposal is already verified by the engine, it is still under the consensus settlement.

eth/state_accessor.go
```diff
@@ -131,7 +131,7 @@ func (eth *Ethereum) StateAtBlock(block *types.Block, reexec uint64, base *state
 		if current = eth.blockchain.GetBlockByNumber(next); current == nil {
 			return nil, fmt.Errorf("block #%d not found", next)
 		}
-		_, _, _, err := eth.blockchain.Processor().Process(current, statedb, vm.Config{})
+		_, _, statedb, _, err = eth.blockchain.Processor().ProcessFromCache(current, statedb, vm.Config{})
```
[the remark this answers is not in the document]
vpiyush: I agree with this.

miner/worker.go
vpiyush: Yes, Will correct this.

Jason-Zhangxin-Chen:
> Various optimizations:
> 
> 1. Trigger block creation as soon as proposal is verified.
> 2. Remove pending snapshot creation
> 3. no update for non-validator nodes on new transaction
> 4. Cache processed state on verify proposal reuse the same later on processing blocks
> 
> Minor refactoring is still expected on this but logic is complete

QQ: Regarding `1. Trigger block creation as soon as proposal is verified.`, is it going to mine new block on top of the verified proposal?

core/blockchain.go
```diff
@@ -225,6 +232,7 @@ type BlockChain struct {
 	// senderCacher is a concurrent transaction sender recoverer and cacher
 	senderCacher *TxSenderCacher
 	log          log.Logger
+	cachedState  atomic.Pointer[blockStateCache]
```
Jason-Zhangxin-Chen, line 235: As different round can have individual proposals, we don't know which one will be taken as the decision until we have quorum certificates at the end, shall we need to cache multiple states of different proposal?

vpiyush:
> > Various optimizations:
> > 
> > 1. Trigger block creation as soon as proposal is verified.
> > 2. Remove pending snapshot creation
> > 3. no update for non-validator nodes on new transaction
> > 4. Cache processed state on verify proposal reuse the same later on processing blocks
> > 
> > Minor refactoring is still expected on this but logic is complete
> 
> QQ: Regarding `1. Trigger block creation as soon as proposal is verified.`, is it going to mine new block on top of the verified proposal? I guess, in most of the case the verified proposal will become the final decision of the height. But it is not safe since a verified proposal can not be the final decision at sometime, mining new block on top of it unless we can deal with it properly.

Yes, We do mine the new block on top of it. and i have done few updates in the code to handle this specific scenario. Please have a look on the latest commit

core/blockchain.go
vpiyush: yes, that's possible. IMO In most cases(99%) the latest verified block would be the one which gets finalized. If not we would fall back to normal proposal verification which should be alright. Do you agree?

Jason-Zhangxin-Chen: > yes, that's possible. IMO In most cases(99%) the latest verified block would be the one which gets finalized. If not we would fall back to normal proposal verification which should be alright. Do you agree?

Yes, that is true. It will work without the state caching. Would like to ask the others' opinion on this: Shall we need to cache state of different proposal of different round?

consensus/tendermint/core/propose.go
```diff
@@ -25,6 +25,15 @@ func (c *Proposer) SendProposal(_ context.Context, block *types.Block) {
 	if c.Height().Cmp(block.Number()) != 0 {
 		panic("proposal block height incorrect")
 	}
+
+	// We start preparing block as soon as proposal is verified, but there are situation
```
Jason-Zhangxin-Chen, line 29: We can schedule a call to discuss this. I will write some concerns on this for potential liveness issue.

consensus/tendermint/core/propose.go
```diff
@@ -25,6 +25,15 @@ func (c *Proposer) SendProposal(_ context.Context, block *types.Block) {
 	if c.Height().Cmp(block.Number()) != 0 {
 		panic("proposal block height incorrect")
 	}
+
```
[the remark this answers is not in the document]
vpiyush: with the current state of the PR, your initial thinking is absolutely right. we are building the block on top of the verified proposal.

miner/worker.go
```diff
@@ -710,42 +630,27 @@ func (w *worker) resultLoop() {
...
-			// Different block could share same sealhash, deep copy here to prevent write-write conflict.
-			var (
-				receipts = make([]*types.Receipt, len(task.receipts))
-				logs     []*types.Log
-			)
-			copyStart := time.Now()
-			for i, taskReceipt := range task.receipts {
-				receipt := new(types.Receipt)
-				receipts[i] = receipt
-				*receipt = *taskReceipt
```
lorenzo-dev1, line 722: Why we do not need anymore this deep copy?

miner/worker.go
```diff
@@ -710,42 +630,27 @@ func (w *worker) resultLoop() {
...
+			for _, log := range task.env.state.Logs() {
+				log.BlockHash = block.Hash()
 			}
 
 			// Commit block and state to database.
 			persistStart := time.Now()
-			_, err := w.chain.WriteBlockAndSetHead(block, receipts, logs, task.state, true)
+			// protect state access, Practically this lock will always be available, by this time
+			// a new pending task will be created(since task creation starts just after the proposal verification)
+			// and this will not be the point of contention, but in theory this can happen, hence the lock
```
lorenzo-dev1, line 652: which lock?

miner/worker.go
```diff
@@ -710,42 +630,27 @@ func (w *worker) resultLoop() {
...
+			_, err := w.chain.WriteBlockAndSetHead(block, task.env.receipts, task.env.state.Logs(), task.env.state, true)
```
lorenzo-dev1, line 653: Do we have the guarantee that the Logs() of statedb are the same ones as the ones inside the receipts?

lorenzo-dev1:
Also, what happens if after verifying round 0 proposal, we actually do not commit it and move to round 1. Will the nodes be able to propose another block for the same height? Or since they will all have already prepared the block for the next height, they will not have another block ready?

consensus/consensus.go
```diff
@@ -92,7 +92,7 @@ type Engine interface {
 
 	// Prepare initializes the consensus fields of a block header according to the
 	// rules of a particular engine. The changes are executed inline.
-	Prepare(chain ChainHeaderReader, header *types.Header) error
+	Prepare(consensus ChainHeaderReader, parentHeader, header *types.Header) error
```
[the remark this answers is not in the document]
vpiyush: unintentional fixed

consensus/ethash/sealer.go
```diff
@@ -48,9 +47,11 @@ var (
 
 func (ethash *Ethash) SetResultChan(_ chan<- *types.Block) {}
 
+func (ethash *Ethash) SetProposalVerifiedEventChan(_ chan<- *types.Block) {}
+
 // Seal implements consensus.Engine, attempting to find a nonce that satisfies
 // the block's difficulty requirements.
-func (ethash *Ethash) Seal(chain consensus.ChainReader, block *types.Block, results chan<- *types.Block, stop <-chan struct{}) error {
+func (ethash *Ethash) Seal(parentHeder *types.Header, block *types.Block, results chan<- *types.Block, stop <-chan struct{}) error {
```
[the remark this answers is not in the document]
vpiyush: fixed

miner/worker.go
```diff
@@ -547,80 +526,15 @@ func (w *worker) mainLoop() {
...
-				txset := types.NewTransactionsByPriceAndNonce(w.current.signer, txs, w.current.header.BaseFee)
-				tcount := w.current.tcount
-				w.commitTransactions(w.current, txset, nil)
-
-				// Only update the snapshot if any new transactions were added
-				// to the pending block
-				if tcount != w.current.tcount {
-					w.updateSnapshot(w.current)
-				}
-			}
```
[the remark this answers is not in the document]
vpiyush: could be to reduce load sharing and avoid dependency on validator nodes for the pending state. But another way to do this would be directly monitoring the transaction pool

miner/worker.go
```diff
@@ -987,28 +872,32 @@ type generateParams struct {
...
+	var (
+		optimisticCandidate bool
+		parentHeader        *types.Header
+	)
+
 	// Find the parent block for sealing task
-	parent := w.chain.CurrentBlock()
+	if parent == nil {
+		parent = w.chain.CurrentBlock()
+		parentHeader = w.chain.GetHeader(parent.Hash(), parent.NumberU64())
```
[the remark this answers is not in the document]
vpiyush: on line 885: yes it can be, depends on which flow we are coming, only non-nil flow is optimistic block preparation. in rest of the flows it would be nil.

on line 887: It can't be in my understanding.

miner/worker.go
```diff
@@ -335,29 +315,13 @@ func (w *worker) enablePreseal() {
...
 }
 
 // pendingBlock returns pending block.
 func (w *worker) pendingBlock() *types.Block {
-	// return a snapshot to avoid contention on currentMu mutex
-	w.snapshotMu.RLock()
-	defer w.snapshotMu.RUnlock()
-	return w.snapshotBlock
-}
-
```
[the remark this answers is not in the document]
vpiyush: I agree with this. We will discuss this in standup

miner/worker.go
```diff
@@ -1146,21 +1019,21 @@ func (w *worker) commitWork(interrupt *int32, noempty bool, timestamp int64) {
 
 	// Create an empty block based on temporary copied state for
 	// sealing in advance without waiting block execution finished.
-	if !noempty && w.chainConfig.Ethash != nil && atomic.LoadUint32(&w.noempty) == 0 {
-		w.commit(work.copy(), nil, false, start)
```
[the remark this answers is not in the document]
vpiyush: Yeah, because we don't create any snapshot in parallel now

miner/worker.go
```diff
@@ -1146,21 +1019,21 @@ func (w *worker) commitWork(interrupt *int32, noempty bool, timestamp int64) {
...
-	w.fillTransactions(interrupt, work)
+	w.fillTransactions(req.interrupt, work)
 	if metrics.Enabled {
 		now := time.Now()
 		FillWorkTimer.Update(now.Sub(fillTxStart))
 		FillWorkBg.Add(now.Sub(fillTxStart).Nanoseconds())
 	}
 
 	commitWorkStart := time.Now()
-	w.commit(work.copy(), w.fullTaskHook, true, start)
```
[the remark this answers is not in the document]
vpiyush: yes

miner/worker.go
```diff
@@ -652,31 +565,38 @@ func (w *worker) taskLoop() {
...
 			w.pendingMu.Unlock()
 
-			sealStart := time.Now()
-			if err := w.engine.Seal(w.chain, task.block, w.resultCh, stopCh); err != nil {
-				w.eth.Logger().Warn("Block sealing failed", "err", err)
-				w.pendingMu.Lock()
-				delete(w.pendingTasks, sealHash)
-				w.pendingMu.Unlock()
-			}
+			// go routine is needed to be able to interrupt the task, in case a more recent block arrives
```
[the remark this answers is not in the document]
vpiyush: yes, It was

miner/worker.go
vpiyush: obsolete comment, removed now

core/types.go
```diff
@@ -44,8 +44,14 @@ type Prefetcher interface {
 
 // Processor is an interface for processing blocks using a given initial state.
 type Processor interface {
+
 	// Process processes the state changes according to the Ethereum rules by running
 	// the transaction messages using the statedb and applying any rewards to both
-	// the processor (coinbase) and any included uncles.
-	Process(block *types.Block, statedb *state.StateDB, cfg vm.Config) (types.Receipts, []*types.Log, uint64, error)
+	// the processor (coinbase) and any included uncles. It checks the availability of state cache
```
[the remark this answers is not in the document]
vpiyush: Right

miner/worker.go
```diff
@@ -452,21 +416,40 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
 			}
-			commit(false, commitInterruptNewHead)
+			lastBlock = head.Block.Hash()
+			commit(false, commitInterruptNewHead, nil)
+
+		case block := <-w.proposalVerifiedEventCh:
+			if block.Hash() == lastBlock {
+				log.Debug("block already prepared")
+				continue
+			}
```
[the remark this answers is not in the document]
vpiyush: Yeah, OR when you receive a proposal in parallel to the new chain event Because these 2 are separate flows you could have a new chain head event being fired before the verified proposal event.

miner/worker.go
```diff
@@ -452,21 +416,40 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+			lastBlock = block.Hash()
+			clearPending(w.chain.CurrentBlock().NumberU64())
+			timestamp = time.Now().Unix()
+			commit(false, commitInterruptNewHead, block)
```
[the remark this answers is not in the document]
vpiyush: To me commitInterrupNewHead is the correct Usage. A new verified proposal is nothing but a new head from the miner's point of view. also the implication is same.

miner/worker.go
```diff
@@ -769,83 +674,63 @@ func (w *worker) resultLoop() {
...
-	}
-	if err != nil {
-		return nil, err
+	var (
+		state *state.StateDB
+		err   error
+		hash  common.Hash
+	)
+	if optimisticCandidate {
+		if parent.Header().Coinbase == w.coinbase { // we were the proposer for the parent
```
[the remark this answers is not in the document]
vpiyush: In general, Yes, we would because we were the one who initially created the block and we do not clear the pending Tasks untill the task is older than 7 blocks. But if the node went down in between proposing and came back up again we won't have the cache. In such case the node can't prepare the block but will wait for others to prepare. This is quite rare though and we have the recovery. But we have means to totally elimitate this problem. I will think about it

miner/worker.go
vpiyush: As far as i understand recipts logs are assigned from the stateDb logs which are the same array of pointers. any change in either of these reflects on each other. Do you think of any flow where it can defer?

vpiyush:
> Also, what happens if after verifying round 0 proposal, we actually do not commit it and move to round 1. Will the nodes be able to propose another block for the same height? Or since they will all have already prepared the block for the next height, they will not have another block ready?

Yes, they can still prepare a new block if there is no chain event for the next 3 seconds through the worker timer.

core/state_processor.go
```diff
@@ -51,58 +50,40 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+		receipts  types.Receipts
+		usedGas   uint64
+		blockHash = block.Hash()
+		allLogs   []*types.Log
+		err       error
 	)
 	// Mutate the block and state according to any hard-fork specs
 	if p.config.DAOForkSupport && p.config.DAOForkBlock != nil && p.config.DAOForkBlock.Cmp(block.Number()) == 0 {
 		misc.ApplyDAOHardFork(statedb)
 	}
```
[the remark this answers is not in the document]
vpiyush: I agree we can remove it

e2e_test/e2e_test.go
```diff
@@ -261,7 +261,7 @@ func TestNodeAlreadyHasProposedBlock(t *testing.T) {
 
 	// handle the proposal
 	err = proposer.HandleProposal(context.TODO(), proposal)
-	require.True(t, errors.Is(err, constants.ErrAlreadyHaveBlock))
+	require.Error(t, constants.ErrAlreadyHaveProposal)
```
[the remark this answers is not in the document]
lorenzo-dev1: However why it now checks for `ErrAlreadyHaveProposal`? It has a different meaning than `ErrAlreadyHaveBlock`. The latter is raised when a proposal is received for an already finalized block, while the former when we already have a proposal for the consensus instance.

consensus/tendermint/core/propose.go
```diff
@@ -26,25 +25,25 @@ func (c *Proposer) SendProposal(_ context.Context, block *types.Block) {
...
+		panic("not proposer")
+	}
 
 	// We start preparing block as soon as proposal is verified, but there are situation
 	// that verified proposal is not finalized in the particular round hence this safety
 	// check to ensure that the block parent hash is same as last hash in core
-	if c.LastHeader().Hash() != block.ParentHash() {
-		log.Info("verified proposal was not finalized in the last round", "aborting send proposal", "last header hash", c.lastHeader.Hash(), "block parent hash", block.ParentHash())
+	if c.Backend().HeadBlock().Hash() != block.ParentHash() {
+		log.Info("verified proposal was not finalized in the last round", "aborting send proposal", "last header hash", c.Backend().HeadBlock().Hash(), "block parent hash", block.ParentHash())
```
lorenzo-dev1, line 36: Is there any difference between `c.LastHeader()` and `c.Backend().HeadBlock()`?

consensus/tendermint/backend/engine.go
```diff
@@ -379,20 +379,14 @@ func (sb *Backend) Seal(parent *types.Header, block *types.Block, _ chan<- *type
...
-	committee, _, _, _, err := chain.LatestEpoch()
-	if err != nil {
-		sb.logger.Error("misssing epoch header", "err", err)
-		return err
-	}
 
-	nodeAddress := sb.Address()
-	if committee.MemberByAddress(nodeAddress) == nil {
-		sb.logger.Error("error validator errUnauthorized", "addr", sb.address)
-		return errUnauthorized
```
lorenzo-dev1, line 400: I am not sure that this check and the one in core are equivalent. In core we could still have the old committee I think.

consensus/tendermint/core/propose.go
```diff
@@ -102,9 +110,16 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
...
-	// Verify the proposal we received
-	start := time.Now()
-	duration, err := c.backend.VerifyProposal(proposal.Block()) // youssef: can we skip the verification for our own proposal?
+	var (
+		duration time.Duration
+		err      error
+		start    = time.Now()
+	)
+
+	// skip verification for our own proposal
```
Jason-Zhangxin-Chen, line 119: As we cached the verified proposal states, can we check it by using those cached states? Not only local mined block should be skipped, but also those valid value of a previous round should be skipped as well, if we already verified and cached a previous round's proposal, then we can skip it.

consensus/tendermint/backend/backend.go
```diff
@@ -289,33 +286,14 @@ func (sb *Backend) VerifyProposal(proposal *types.Block) (time.Duration, error)
...
-				return 0, receiptErr
-			}
-			receipts = append(receipts, receipt)
-		}
-
-		state.Prepare(common.ACHash(proposal.Number()), len(proposal.Transactions()))
-		receipt, epochInfo, err := sb.Finalize(sb.blockchain, header, state, proposal.Transactions(), nil, receipts)
+		receipts, _, usedGas, epochInfo, err := sb.blockchain.Processor().Process(proposal, state, *sb.vmConfig)
 		if err != nil {
+			sb.logger.Error("state processing failed", "error", err)
```
Jason-Zhangxin-Chen, line 291: Please add the height number in the logging.

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: I am thinking even the verified proposal was not committed yet, checking a new block with the authority is still okay on top of the verified proposal, because we have the state cached where it contain the latest committee.

core/blockchain.go
```diff
@@ -460,6 +468,15 @@ func NewBlockChain(db ethdb.Database,
 	return bc, nil
 }
 
+func (bc *BlockChain) CacheProposalState(hash common.Hash, receipts types.Receipts, usedGas uint64, db *state.StateDB) {
+	bc.cachedState.Store(&blockStateCache{hash: hash, receipts: receipts, usedGas: usedGas, stateDb: db})
+}
+
+func (bc *BlockChain) LoadProposalState() (*state.StateDB, common.Hash) {
+	st := bc.cachedState.Load()
```
Jason-Zhangxin-Chen, line 476: Is it possible for st to be a nil?

eth/api_backend.go
```diff
@@ -135,8 +135,9 @@ func (b *EthAPIBackend) BlockByNumberOrHash(ctx context.Context, blockNrOrHash r
 	return nil, errors.New("invalid arguments; neither block nor hash specified")
 }
 
+// not supported anymore
 func (b *EthAPIBackend) PendingBlockAndReceipts() (*types.Block, types.Receipts) {
-	return b.eth.miner.PendingBlockAndReceipts()
+	return nil, nil
```
Jason-Zhangxin-Chen, line 140: I'm curious, why is this no longer supported?

miner/worker.go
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Recall back to the cach of verified proposal's state, shall we re-think to cache verified proposal of multiple rounds? Now the cach might overwriten by different verified proposal of different round, I am just concern about the atomic operation of it and the race of it.

miner/worker.go
```diff
@@ -452,21 +416,40 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
 	for {
 		select {
 		case <-w.startCh:
 			clearPending(w.chain.CurrentBlock().NumberU64())
 			timestamp = time.Now().Unix()
-			commit(false, commitInterruptNewHead)
+			commit(false, commitInterruptNewHead, nil)
 
 		case head := <-w.chainHeadCh:
+			if head.Block.Hash() == lastBlock {
```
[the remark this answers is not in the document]
lorenzo-dev1: I see.  Do you think we could reset the `lastBlock` in case there is an error in the flows after `commit`? I think that would mitigate the issue.

miner/worker.go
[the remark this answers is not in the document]
lorenzo-dev1: Yeah I see but in my opinion a new head is a finalized block that is forever written in the chain. But tbh I don't think this is a big deal, let's see what other think nonetheless

miner/worker.go
lorenzo-dev1: Yeah I agree, however then why do we need to loop through both the `env.receipts` ant the `state.Logs()` when assigning the `BlockHash`? Is it because we could have performed a deep copy somewhere before?

eth/api_backend.go
vpiyush: yeah, we have removed that with the understanding that these APIs are not worth the added complexity. People aren't using these anymore. We instead use the latest block now.

core/blockchain.go
vpiyush: No, because we don't invoke load before cache. and once it's loaded we always have the older reference.

consensus/tendermint/core/propose.go
vpiyush: the last header got removed in Jason's PR. I am fetching the same info though.

consensus/tendermint/backend/engine.go
vpiyush: @lorenzo-dev1 @Jason-Zhangxin-Chen the checks i am talking about are in sendProposal. I don't believe we can have old committee [here](https://github.com/autonity/autonity/blob/0ed8eafd468d72e7c43343c11fe44c33f74e5eaf/consensus/tendermint/core/propose.go#L43).

miner/worker.go
vpiyush: I couldn't find any reason why it was needed in the first place. this seems to be very old code. At least in our current flow I don't see any reason to copy it.

miner/worker.go
vpiyush: Actually we don't have to. I retained this as previous logic. but I am going to remove that now.

consensus/tendermint/core/propose.go
```diff
@@ -117,7 +117,9 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
 	)
 
 	// skip verification for our own proposal
-	if c.backend.ProposedBlockHash() != proposal.Hash() {
+	// skip if our own OR cached
+	// verify if not in our own and not state cached
+	if c.backend.ProposedBlockHash() != proposal.Hash() && !c.backend.IsProposalStateCached(proposal.Hash()) {
```
lorenzo-dev1, line 122: I don't think this will work as hinted here: https://github.com/autonity/autonity/pull/1033/files/0ed8eafd468d72e7c43343c11fe44c33f74e5eaf..6df84d2f13edb55729bb51ab20e3b08252fff7c7#r1798831954

We use the proposal hash and not the block hash to check if the proposal is verified. So even if a proposer proposes an old valid value, the round is going to be increase wrt the old proposal and therefore the proposal hash will be different.

I.E.
```
proposal.Hash() = hash( rlp.EncodeToBytes([]any{ProposalCode, ext.Round, ext.Height, ext.ValidRound, ext.IsValidRoundNil, p.block.Hash()}))
```

while the block.Hash() is the hash of only the block.

lorenzo-dev1: Also, I think there is another call to `VerifyProposal` in rules.go that is not cached.

lorenzo-dev1: Oh I actually saw that proposedBlockHash and the proposal cache actually use the hash of the block, so this part of the code needs to be changed to use that as well. This would allow to use caching to validate old proposals.

vpiyush: Thanks @lorenzo-dev1 , I realized that it was overlooked

lorenzo-dev1:
@vpiyush I was reflecting on the implications of using the block.hash to index the cache. I think it works fine, however I was thinking about the `round` and `quorumCertificate` fields, which are not included into the hash. This means that if we cache proposal x, the a node can re-send it and put garbage in `round` and `quorumCertificate`, but we would not detect it (because we would skip verification). 

This is not a big deal I think, because both these fields are overwritten when the proposal is committed. However I was thinking, how about enforcing that un-finalized proposal have empty quorum certificate and round=0? That would simplify things and rule out any unforeseen scenarios.

miner/worker.go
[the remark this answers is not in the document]
lorenzo-dev1: Discussed in meeting: best solution is probably to make the miner flow synchronous. But that is a big refactor so it doesn't belong to this PR. Right now the analyzed scenario should not cause any issues.

core/blockchain.go
lorenzo-dev1: I think caching for multiple rounds will not improve performance of much. As long as the fallback mechanism works, I think we are fine.

miner/worker.go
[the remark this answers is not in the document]
lorenzo-dev1: I am leaving this open @vpiyush, as we should probably capture this into an issue. I can write it up unless you want to do it yourself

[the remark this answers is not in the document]
vpiyush: Sure, I have created an issue [here](https://github.com/autonity/autonity/issues/1051)

miner/worker.go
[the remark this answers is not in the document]
lorenzo-dev1: FYI @cmjc

consensus/tendermint/core/propose.go
```diff
@@ -89,6 +89,14 @@ func (c *Proposer) HandleProposal(ctx context.Context, proposal *message.Propose
 		return constants.ErrNotFromProposer
 	}
 
+	// these checks ensure that nodes don't exploit the cached state, cache is indexed by block hash which
+	// excludes quorum certificate and round that allows nodes to send garbage for these values for pre-verified
+	// proposal
+	qc := proposal.Block().Header().QuorumCertificate
+	if qc.Signature != nil || qc.Signers != nil || proposal.Block().Header().Round != 0 {
+		return constants.ErrInvalidMessage
+	}
```
lorenzo-dev1, line 98: I think the check is good, but I am wondering if we could actually do it a RLP decoding time. I cannot think of any scenario where a proposal can have a non empty quorum certificate.

I think that once the proposal is committed, we will just broadcast the block directly through announcements?

vpiyush: I agree, since we are not expecting these to be set on the proposal message, making these checks at decoding would be better.

miner/worker.go
```diff
@@ -335,29 +309,13 @@ func (w *worker) enablePreseal() {
...
 func (w *worker) pending() (*types.Block, *state.StateDB) {
-	// return a snapshot to avoid contention on currentMu mutex
-	w.snapshotMu.RLock()
-	defer w.snapshotMu.RUnlock()
-	if w.snapshotState == nil {
-		return nil, nil
-	}
-	return w.snapshotBlock, w.snapshotState.Copy()
+	st, _ := w.chain.State()
+	return w.chain.CurrentBlock(), st
```
lorenzo-dev1, line 313: It would be better to use StateAt to avoid that the currentblock changes right after we fetched the state

lorenzo-dev1 (approved):
LGTM! 

Just left a minor comment + I have a question: while having a final look at the changes in `blockchain.go` I noticed that [here](https://github.com/autonity/autonity/blob/develop/core/blockchain.go#L1666) we copy the statedb pointer. We are still doing that in this PR but now the pointer could be changed to the cached state db copy. Can this cause issue with this `activeState` prefetcher logic?

vpiyush:
> LGTM!
> 
> Just left a minor comment + I have a question: while having a final look at the changes in `blockchain.go` I noticed that [here](https://github.com/autonity/autonity/blob/develop/core/blockchain.go#L1666) we copy the statedb pointer. We are still doing that in this PR but now the pointer could be changed to the cached state db copy. Can this cause issue with this `activeState` prefetcher logic?

The prefetcher would still be working on the older state while we use the cache state to write blocks so there is no issue as such. but the prefetcher effort would be  wasted cpu cycles when we use the cached state. so this could be an opportunity to optimize.

Jason-Zhangxin-Chen (approved):
LGTM.

yazzaoui (approved):
LGTM let's merge it

PR #1035, opened 2024-08-28
Generate NatSpec data together with contract ABIs

szemate:
Currently, the protocol contract documentation is maintained manually in the [docs.autonity.org repository](https://github.com/autonity/docs.autonity.org).

As part of the documentation effort for mainnet launch, we are planning to [migrate protocol contract documentation to the source code](https://github.com/autonity/docs.autonity.org/issues/217) and [automatically generate documentation from NatSpec data](https://github.com/autonity/docs.autonity.org/issues/216).

Using the existing Autonity build process for [generating NatSpec files](https://docs.soliditylang.org/en/latest/natspec-format.html#documentation-output) would reduce the maintenance burden on the documentation generator, compared to maintaining a similar Makefile in the docs repository. With the `--userdoc` and `--devdoc` Solc compiler options the NatSpec JSONs will be available next to the ABI files with `.docdev` and `.docuser` file extensions that can be used as inputs to the generator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,8 @@ __pycache__/
 /package.json
 /params/generated/**/*.abi
 /params/generated/**/*.bin
+/params/generated/**/*.docdev
+/params/generated/**/*.docuser
 nohup.out
 
 # travis
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ autonity-docker:
 	@echo "Run \"$(BINDIR)/autonity\" to launch autonity."
 
 define gen-contract
-	$(SOLC_BINARY) --overwrite --optimize --optimize-runs 10000 --evm-version london --abi --bin -o $(GENERATED_CONTRACT_DIR) $(CONTRACTS_DIR)/$(1)$(2).sol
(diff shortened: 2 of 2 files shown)


Dedenwrg (approved):

.gitignore
```diff
@@ -36,6 +36,8 @@ __pycache__/
 /package.json
 /params/generated/**/*.abi
 /params/generated/**/*.bin
+/params/generated/**/*.docdev
```
tbssajal, line 39: I guess `.docdev` and `.docuser` files should not be in gitignore. Otherwise if we make some changes in the contract, docs will not be updated.

tbssajal: Unless we don't want the docs available in this repo.

szemate: These are interim data files, the docs generated with them will be in the [docs.autonity.org](https://github.com/autonity/docs.autonity.org) repo. Similarly, the `.abi` files are in gitignore and only the generated Go bindings are committed to the repo.

tbssajal (approved):
LGTM

yazzaoui (approved):

PR #1037, opened 2024-09-09
fix aggregator height/round map cleanup

vpiyush:
Piccadilly issue - Number of goroutines kept growing due to aggregator being busy in iterating over height Map.
Changed few things:
1. remove goroutine in backend and increase the size of buffered message ch
2. increase fixSize cache capacity
3. Fix clean up of message map in aggregator

### DIFF
--- a/consensus/tendermint/backend/aggregator.go
+++ b/consensus/tendermint/backend/aggregator.go
@@ -905,7 +905,6 @@ loop:
 							signatureInput := sameValueVotes[0].Message.SignatureInput() // all votes have same (h,r,c,v)
 							a.staleMessages[signatureInput] = append(a.staleMessages[signatureInput], sameValueVotes...)
 						}
-
 						delete(a.messages[h], r)
 						roundMapSize.Dec(1)
 						continue
--- a/consensus/tendermint/backend/aggregator.go
+++ b/consensus/tendermint/backend/aggregator.go
@@ -42,6 +42,8 @@ var (
 	BatchesBg                  = metrics.NewRegisteredBufferedGauge("aggregator/batches", nil, metrics.GetIntPointer(100))          // size of batches (aggregated together with a single fastAggregateVerify)
 	InvalidBg                  = metrics.NewRegisteredBufferedGauge("aggregator/invalid", nil, metrics.GetIntPointer(100))          // number of invalid sigs
 	BackendAggregatorTransitBg = metrics.NewRegisteredBufferedGauge("aggregator/backend/transit", nil, metrics.GetIntPointer(1000)) // measures time for message passing from backend to aggregator
+	roundMapSize               = metrics.NewRegisteredGauge("aggregator/rounds/all", nil)
+	heightMapSize              = metrics.NewRegisteredGauge("aggregator/height/all", nil)
 )
 
 func recordMessageProcessingTime(code uint8, start time.Time) {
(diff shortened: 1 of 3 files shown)


consensus/tendermint/backend/aggregator.go
```diff
@@ -899,11 +906,16 @@ loop:
...
 						}
 						delete(a.messages[h], r)
+						roundMapSize.Dec(1)
 						continue
 					}
 					// if current height, process them
 					a.processRound(h, r)
 				}
+				if h < coreHeight {
+					delete(a.messages, h)
```
tbssajal, line 916: Is this the fix? Then don't we want it on `develop`? Or we want to test it before?

consensus/tendermint/backend/aggregator.go
```diff
@@ -899,11 +906,16 @@ loop:
...
+					heightMapSize.Dec(1)
```
tbssajal, line 917: Should we also decrease `roundMapSize` by `len(roundMap)`?

consensus/tendermint/backend/handler.go
```diff
@@ -215,7 +215,7 @@ func (sb *Backend) handleDecodedMsg(msg message.Msg, errCh chan<- error, sender
 		sb.logger.Crit("Tendermint backend processing unknown message")
 	}
 
-	go sb.Post(events.UnverifiedMessageEvent{
+	sb.Post(events.UnverifiedMessageEvent{
```
tbssajal, line 218: Shall we make this change in `develop` as well? If not then why make the change here?

consensus/tendermint/backend/aggregator.go
tbssajal: Saw slack messages.

consensus/tendermint/backend/handler.go
vpiyush: yeah, we will push this into develop for next release

consensus/tendermint/backend/aggregator.go
vpiyush: We are decrementing roundMapSize when we delete roundMap from the height. line 909

tbssajal (approved):

consensus/tendermint/backend/aggregator.go
```diff
@@ -324,8 +328,11 @@ func (a *aggregator) processRound(h uint64, r int64) {
 
 	a.processBatches(batches, currentHeightEventBuilder)
 
-	//clean up
-	delete(a.messages[h], r)
+	if _, ok := a.messages[h][r]; ok {
+		//clean up
+		delete(a.messages[h], r)
+		roundMapSize.Dec(1)
+	}
```
lorenzo-dev1, line 335: Just to understand, this shouldn't change performance right? Is it just to correctly decrease the gauge?

consensus/tendermint/backend/handler.go
lorenzo-dev1, line 218: Is this because the messageCh is buffered so we shouldn't block anyways?

lorenzo-dev1 (approved):
A couple of questions but LGTM! thanks for debugging!

consensus/tendermint/backend/aggregator.go
vpiyush: Yes, Exactly

consensus/tendermint/backend/handler.go
vpiyush: Yes

Jason-Zhangxin-Chen (approved):
LGTM.

lorenzo-dev1:
I think we should merge this into develop to include it in the next release

vpiyush:
> I think we should merge this into develop to include it in the next release

Yeah,  We also need to merge the concurrency limiter related change. This PR would actually not get merged as it is. We don't need those metrics permanently. But I didn't want to close this untill we have a separte PR to merge both of those changes. I will do that though after i am done with the work in hand.

vpiyush:
Closing this PR, because these changes were already taken in relase v14 and in there is new PR [here](https://github.com/autonity/autonity/pull/1055) to pick these.

Issue #1025, opened 2024-07-16
Dump debug info during abnormal cpu/mem usage - CPU and Mem profile


vpiyush:
### Description

It would be a good idea to dump debug information for eg. cpu/mem profile, stack, trace etc. when system params are abnormal for e.g. cpu and memory usage being abnormally high.

This can aid in debugging later and could save some cycles of reproducing the issue and then debugging the same

vpiyush:
**Design Overview:**

- Implement a new monitoring service to track system performance.
- Gather diagnostic data when the following metrics exceed acceptable threshold values:
  - CPU usage percentage
  - Heap memory utilization
  - Number of goroutines

**Diagnostics to Collect:**
- CPU profile
- Memory profile
- Goroutine stack trace
- Trace profile

**Additional Considerations:**
- To prevent excess or redundant diagnostics, we propose the following:
  - Increase the threshold by 20% after each diagnostic collection to adjust for further diagnostics.
  - Limit the number of diagnostics collected to a maximum of 5 per day.

lorenzo-dev1:
@vpiyush I think we can close this as completed right?

PR #1041, opened 2024-09-18
Dump profiles

vpiyush:
See [here](https://github.com/autonity/autonity/issues/1025) for issue and design consideration.


internal/monitor/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	time.Sleep(ms.config.traceDuration)
+	trace.Stop()
+	f.Close()
+}
+
+func (ms *monitorService) checkSystemState() {
+	currentDate := time.Now().Format("2006-01-02")
+
+	if currentDate != ms.lastProfileDay {
```
lorenzo-dev1, line 190: Maybe it would be better `lastProfileDate` since it contains the full date

internal/monitor/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+func (ms *monitorService) updateThresholds() {
+	// update threshold by 10%
+	ms.config.cpuThreshold = ms.config.cpuThreshold * 1.1
+	ms.config.memThreshold = uint64(float64(ms.config.memThreshold) * 1.1)
+	ms.config.numGoroutines = int(float64(ms.config.numGoroutines) * 1.1)
+}
+
```
lorenzo-dev1, line 112: Maybe we should also have some code that resets the threshold to their initial values. For example after an entire day without any profile collected?

internal/monitor/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+func (ms *monitorService) collectDiagnostics(currentDate string) {
+	profileDir := filepath.Join(ms.config.profileDir, currentDate)
+	err := os.MkdirAll(profileDir, os.ModePerm)
```
lorenzo-dev1, line 115: I would give permissions 770, so no permission for `others`. Just in case the profiling info contains any sensitive information (e.g. keys).

internal/monitor/monitor.go
```diff
@@ -0,0 +1,213 @@
...
+		log.Error("Couldn't write mem profile", "error", err)
+		return
+	}
+	f.Close()
+
+	// goroutines stack trace
+	goroutines := filepath.Join(profileDir, goroutineDumpFile+postfix)
+	f, err = os.Create(goroutines)
+	if err != nil {
+		log.Error("Couldn't create file to write goroutines", "error", err)
```
lorenzo-dev1, line 158: file to write stack trace?

lorenzo-dev1 (commented):
Overall LGTM, just a few comments

internal/monitor/monitor.go
vpiyush: I had thought about it, but i don't feel strongly about it may be it's not worth it. Here is what can happen, a process can stay on increased resource usage due to other factors and we would keep creating the profiles which we have already got besides we already have profiles for an specifc resource spikes, Most cases we would just get the duplicate info. It would be  filling the disk and creating additional resource load.

internal/monitor/monitor.go
vpiyush: the kind of profiles we are creating doesn't contain the memory data information. It's just stats about memory, cpu usage. This could be true in case of a core dump. where you get the snapshot of the memory. I can still update it to 0664. so not allow others to corrupt it

internal/monitor/monitor.go
lorenzo-dev1: but what if let's say there is a validator which is experiencing resource spikes, and send us the profiles. Then we do our analysis and work on a fix for next release. Now that validator needs to restart his client if he wants to detect again spikes (which could be due to other issues potentially, and maybe could happen after weeks). But anyways I do not have  a strong opinion as well. Let's see what other think

Jason-Zhangxin-Chen:
Good work, LGTM with little comments on the configuration side as we discussed in the pair-work session.

vpiyush:
> Good work, LGTM with little comments on the configuration side as we discussed in the pair-work session.

Updated the PR with the functionality we talked about. Please approve if it's alright.

internal/monitor/monitor.go
vpiyush: As long as these are spikes and resource usage go back to normal In most cases (99% of the time) that's not a issue. Though once in a while there could be an optimization oporrtunity there. While the other way around where the resource usage stays up, it is most likely an issue needs to be addressed. let's hear what others think.

internal/monitor/monitor.go
```diff
@@ -0,0 +1,231 @@
...
+	memThreshold         uint64
+	profilePerDay        int
+	monitoringInterval   time.Duration
+	cpuProfilingDuration time.Duration
+	traceDuration        time.Duration
+	profileDir           string
+}
+
+var DefaultMonitorConfig = Config{
+	cpuThreshold:         80,
```
lorenzo-dev1, line 44: If that's for a single core, I would even start from a lower value. 80% is already quite high I think

internal/monitor/monitor.go
```diff
@@ -0,0 +1,231 @@
...
+	numGoroutines:        6000,
+	memThreshold:         6 * 1024 * 1024,
```
lorenzo-dev1, line 46: If that's 6GB, also here I would maybe start from a bit lower. Also Ideally I think we could make all these cli parameters, so that every operator can customize them. But it can be for later.

lorenzo-dev1 (approved):
LGTM

internal/monitor/monitor.go
vpiyush: It's a combined value, I have intentionally kept it high. As long as we are below 80% we can function normal. Above 80% it start getting into red zone and a potential invoction of oom killer. that's when we need to get involved otherwise it should be ok  i believe. Idea is to only write these profiles when things to wrong.

internal/monitor/monitor.go
vpiyush: Same for memory i kept it high. We do tend to use approximately 4-5 GB in normal scenario. I was thinking it keeping even higher but i settled for 6GB. The reason is - if there is a memory related issue it would keep growing and eventually go beyond 6 GB even in normal traffic scenario. before that we don't have much to do if it dumps.

Jason-Zhangxin-Chen (approved):
LGTM.

PR #1044, opened 2024-09-23
test, create individual chain lifecycle for docker e2e testcases.

Jason-Zhangxin-Chen:
This PR presents the fix of docker e2e test to co-exist with the omission jailing protocol of D4.
It also break down the playbook into multiple ones, with each has its own CI job to speedup the test process.


Jason-Zhangxin-Chen:
To be replaced by https://github.com/autonity/autonity/pull/1049

PR #1045, opened 2024-09-23
fix, header verification with epoch header.

Jason-Zhangxin-Chen:
- check if header is already known before checking it.
An edge case was addressed by the docker e2e test, that was, the epoch block was inserted by the enqueuer (a local virtual peer to insert the decision of consensus) consensus engine, however block sync routine of the execution layer can sync the same epoch block from peers, thus it cause the epoch boundary check failed due to a wrong epoch resolved. So we skip the known header verification as the VerifyHeader() interface did.


consensus/tendermint/backend/engine.go
```diff
@@ -199,8 +199,16 @@ func (sb *Backend) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*t
...
 			} else {
-				// check header against parent and parent epoch head.
-				err = sb.verifyHeader(chain, header, parent, committee)
+				// check not imported header against parent, skip the header that was inserted.
+				// The edge case was addressed by the docker e2e test, that was, the epoch block was inserted by
+				// the enqueuer (a local virtual peer to insert the decision of consensus) consensus engine, however
+				// block sync routine of the execution layer can sync the same epoch block from peers, thus it cause
+				// the epoch boundary check failed due to a wrong epoch resolved. So we skip the known header verification
+				// as the VerifyHeader() interface did.
+				number := header.Number.Uint64()
```
vpiyush, line 208: The header number could be nil here.

consensus/tendermint/backend/engine.go
```diff
@@ -199,8 +199,16 @@ func (sb *Backend) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*t
...
+				if chain.GetHeader(header.Hash(), number) == nil {
```
vpiyush, line 209: this introduces a database call in most cases. which isn't good. Can we think a different way to get around this problem without incurring performance penalty.

consensus/tendermint/backend/engine.go
```diff
@@ -209,14 +199,20 @@ func (sb *Backend) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*t
...
+					err = sb.verifyHeader(chain, header, parent, committee)
+				}
 			}
 			// if the processing header is a new epoch header, update the epoch info for un-processed headers .
 			if header.IsEpochHeader() {
```
vpiyush, line 214: Also what does this assignemt do. It is not being used. and assignments can also cause panics if the verifyHeader above failed

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: I thought there was a sanity check in RLP decoding and the block sanity checker, however there was no. Thus I add it in 45b00ddbf5cd9c231b8d24ec0c745e299d6d4bab.

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: Refine this in 45b00ddbf5cd9c231b8d24ec0c745e299d6d4bab

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: In the rlp decoding and block sanity checker, if a header is an epoch header, then we checked the sanity of epoch, thus in this context, it wouldn't be nil.

vpiyush: Still, If the verifyHeader fails above do you want to still do these assignments?

Jason-Zhangxin-Chen: We still need to do these, since the caller of this VerifyHeaders() function depends on every single header's verification result to proceed the block importing logic, for example, for those already known blocks will be allowed in the batch, they will be skipped for importing, only those unknown blocks in the batch will be imported.

consensus/tendermint/backend/engine.go
lorenzo-dev1: We can leave the check as it doesn't hurt but rlp decoding will always instantiate pointers unless using the `rlp:nil` tag. Therefore header.Number can never be nil unless we are creating the block locally

consensus/tendermint/backend/engine.go
lorenzo-dev1: I think we should still do the check for every header if it is needed, and not just for epoch headers. Also if we remove the epoch boundary check, do we still need this additional `knowBlock` check?

lorenzo-dev1:
For what concerns the `ErrOutOfEpochRange` I think it was good to have, as it ensure that we are never using information from a wrong epoch to validate an header. This shouldn't happen normally, but since this is a new feature I think that having an additional safeguard in place was good. Was this check causing issues?

For what concerns the `ErrInvalidEpochBoundary` instead, I agree that a wrong epoch boundary value should trigger quorum certificate invalidation, however I think as well that it was good to have this additional safeguard already at header verification to filter out blatantly invalid headers. 

My point is, are these checks hurtful? If not, I think it is good to have them. At the end of the day, theoretically we could skip the entire header verification and just verify the quorum certificate, but I am not sure this is what we want.

consensus/tendermint/backend/backend.go
```diff
@@ -318,7 +318,11 @@ func (sb *Backend) VerifyProposal(proposal *types.Block) (time.Duration, error)
 			sb.logger.Error("proposal proposed, bad root state", err)
 			return 0, err
 		}
-
+		// As the epoch infos(committee, lastEpochBlock, nextEpochBlock) are saving in the contract state, thus the
+		// consistency of them are checked by the state validation, moreover that since the epoch info in block header
+		// is a factor to compute the block hash, thus they are immutable in the hash chain once there are quorum
+		// certificates to finalize the block. Thus, the epoch boundary checking is not required anymore, however
+		// the epoch info in the proposal's header should be equal to the epoch info dumped from the contract state.
```
lorenzo-dev1, line 325: I think this comment is out of date since we are still doing the epoch boundary check in header verification (and it is correct to do it imo).

consensus/tendermint/backend/engine.go
```diff
@@ -188,14 +197,22 @@ func (sb *Backend) verifyHeaderAgainstLastView(header, parent *types.Header, com
...
-	committee, _, curEpochHead, nextEpochHead, err := chain.LatestEpoch()
-	if err != nil {
-		panic(err)
-	}
+
 	go func() {
+		epoch, err := chain.EpochOfHeight(headers[0].Number.Uint64())
+		// short circuit, if we cannot find the correct epoch for the 1st header, we quit this batch of verification.
+		if err != nil {
+			sb.logger.Error("VerifyHeaders", "cannot find epoch for the 1st header of the batch: ", err.Error())
```
lorenzo-dev1, line 208: Maybe also print the header number

consensus/tendermint/backend/engine.go
```diff
@@ -188,14 +197,22 @@ func (sb *Backend) verifyHeaderAgainstLastView(header, parent *types.Header, com
...
+			results <- err
+			return
```
lorenzo-dev1, line 210: Is it possible that this early termination makes the iteration in `insertChain` block? 

I didn't dig deep into the code but at first look it seems like it expects a result for every header in the chain

consensus/tendermint/backend/engine.go
```diff
@@ -388,14 +406,14 @@ func (sb *Backend) Seal(chain consensus.ChainReader, block *types.Block, _ chan<
 		return ErrStoppedEngine
 	}
 	// update the block header and signature and propose the block to core engine
-	committee, _, _, _, err := chain.LatestEpoch()
+	epoch, err := chain.EpochOfHeight(block.NumberU64())
```
lorenzo-dev1, line 409: I think that this might fail if for some reason we end up fetching the epoch information from state. Due to:
```
require(_height <= block.number, "cannot get committee for a future height");
```

In this case since we are sealing we will request height = currentChainHead + 1, but the block context of the EVM will be set based on the current chain head, and the call will fail.

Can this happen?

lorenzo-dev1 (commented):
Overall LGTM, just a few comments

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: > Is it possible that this early termination makes the iteration in `insertChain` block?
> 
> I didn't dig deep into the code but at first look it seems like it expects a result for every header in the chain

No, the `insertChain` checks if the 1st head verification of the batch failed then it break the entire process. Thus it would not block at all.

consensus/tendermint/backend/engine.go
```diff
@@ -144,24 +145,32 @@ func (sb *Backend) verifyHeader(chain consensus.ChainHeaderReader, header, paren
...
+	// We expect the parent to be non nil when header is not the genesis header.
+	if parent == nil {
+		return errUnknownBlock
+	}
+
+	// Due to the synchronisation issue, the verifying header might be already known, thus we return the ErrKnownBlock
+	// to skip the following epoch boundary checks. As recent headers are always buffered in LRU cache, thus it does
+	// not introduce too much performance dropping.
+	if chain.GetHeader(header.Hash(), header.Number.Uint64()) != nil {
+		return core.ErrKnownBlock
```
lorenzo-dev1, line 157: Looking more at the code of `insertChain`, I am not sure that it is the best to use `ErrKnownBlock`. By reading this comment: 
```
	// We can only ever bypass processing if the only error returned by the validator
	// is ErrKnownBlock, which means all checks passed, but we already have the block
	// and state.
```
It seems that returning `ErrKnownBlock` means that the block is valid, but we already have it. However we are returning it when we do not known if the block is valid or not.

It seems like `insertChain` still does some operations on knownBlocks related to snapshots.

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: resolved in 8e1527ea7dd96aaa48400294fe9c75fe9ff00ff1

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: In the mining context, calling EpochOfHeight() returns at the latest epoch checker, it wouldn't go to the epoch querying base on the state db, however, I think it would be more logical to replace EpochOfHeight() with LatestEpoch() in this Seal() function. As it is as now resolved in 8e1527ea7dd96aaa48400294fe9c75fe9ff00ff1

lorenzo-dev1: I am not sure this is a good idea, because it can happen that the chain moves while we are sealing a block (e.g. we realize that we are not at the top of the chain). I think using `EpochOfHeight` solves this problem at the root, we just need to make sure that we can always fetch the epoch information for the next block going to be produced.

consensus/tendermint/core/core.go
```diff
@@ -371,10 +373,14 @@ func (c *Core) setInitialState(r int64) {
...
-			log.Debug("on epoch rotation, update committee!", "number", lastBlockMined.Number())
-			c.committee.SetCommittee(lastBlockMined.Header().Epoch.Committee)
+
+		// if epoch rotated, update committee.
+		if epoch, err := c.Backend().LatestEpoch(); err == nil {
+			if c.epoch.EpochBlock.Cmp(epoch.EpochBlock) != 0 {
+				log.Debug("on epoch rotation, update committee!", "number", lastBlockMined.Number())
+				c.epoch = epoch
+				c.committee.SetCommittee(epoch.Committee)
+			}
```
lorenzo-dev1, line 383: What if `err != nil`? 

Also what improvement does this change brings? Aren't we making more calls to the db wrt the previous code?

Jason-Zhangxin-Chen: > What if `err != nil`?
> 
> Also what improvement does this change brings? Aren't we making more calls to the db wrt the previous code?

I added some more comments for this change in 5b62ac26131aef5d913bd7bfc80892d2082f5f6e. I think check epoch rotation from consensus engine should not rely on checking if the head of the chain is an epoch header, because at the start of a client, chain head changes due to the peer synchronization. In case of error, I will panic it now as it should not happen.

lorenzo-dev1: I agree on the fact that the chain head can change fast while syncing, but then with the current code we will have a mismatch between height (which is still based on the chain head, and epoch information). At this point it might be better to call `EpochOfHeight` with the chain head height as parameter?

consensus/tendermint/backend/engine.go
lorenzo-dev1: @Jason-Zhangxin-Chen I think we have a similar issue also in the `getEpochFromBlock` function

consensus/tendermint/core/core.go
Jason-Zhangxin-Chen: > I agree on the fact that the chain head can change fast while syncing, but then with the current code we will have a mismatch between height (which is still based on the chain head, and epoch information). At this point it might be better to call `EpochOfHeight` with the chain head height as parameter? That would solve both problems imo

Yeah, I am also thinking EpochOfHeight will be more logical.

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: > Looking more at the code of `insertChain`, I am not sure that it is the best to use `ErrKnownBlock`. By reading this comment:
> 
> ```
> 	// We can only ever bypass processing if the only error returned by the validator
> 	// is ErrKnownBlock, which means all checks passed, but we already have the block
> 	// and state.
> ```
> 
> It seems that returning `ErrKnownBlock` means that the block is valid, but we already have it. However we are returning it when we do not known if the block is valid or not.
> 
> It seems like `insertChain` still does some operations on knownBlocks related to snapshots.
> 
> Any thoughts?

Now we return nil if the header is already known by the blockchain, that mean it was checked and valid. See d34fa035782089d1a9db944bf4a17159ff95629b

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1069,7 +1069,12 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
     */
     function getEpochByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory, uint256, uint256, uint256) {
         // the lastFinalizedBlock+1 height's committee is always available.
-        require(_height <= lastFinalizedBlock+1, "cannot get committee for a future height");
+        // resolve the cap as lastFinalizedBlock might be 0 in the contract testing framework.
+        uint256 cap = lastFinalizedBlock+1;
+        if (block.number > lastFinalizedBlock) {
+            cap = block.number;
+        }
+        require(_height <= cap, "cannot get epoch for a future block");
```
lorenzo-dev1, line 1077: I see the reason for this change but I think that if this is needed just for Tests it would be better to either fix the tests or make these changes in `AutonityTest.sol`, so that we do not pollute production code with test logic.

consensus/tendermint/backend/engine.go
lorenzo-dev1: I think returning nil is correct. However I am thinking, if we have already an header in the chain, we know it is valid. However we do not have any guarantee on the quorum certificate of this new header that we are verifying.

More in details, when we receive header $H$ for header validation, we realize that we already have it in the chain ($H'$). However we have no guarantees that $H.QuorumCertificate$ = $H'.QuorumCertificate$. Do you think this could be a problem?

Jason-Zhangxin-Chen: Good catch! I think it is possible, although it does not change the state of the chain, however it would potentially break the synchronization protocol, for example such client provides overwriten headers as a source of blockchain sync, then the remote requesting client would break its sync process from this source. Thus I resolved it with double checking the header against its parent and the corresponding quorum certificates in 067fbc6c2b52f73873ed72509af93aa70c2a9b9d

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: I find that it would cost a lots of time to fix the truffle test, I will try to repair it in that way on time, otherwise I will keep current cap for the range as it for now.

lorenzo-dev1: What about putting this code in `AutonityTest` and deploying that in the truffle tests? should be pretty fast to do

consensus/tendermint/backend/engine.go
lorenzo-dev1: I agree with @vpiyush that this might be a problem. We are basically using an untrusted committee to validate following headers. 

This could cause a crash later on when verifying quorum certificates using this untrusted committee, since we panic if we cannot aggregate the keys of the committee (because we assume that it is trusted, but we are now breaking this assumption). Therefore we should either lower that log level OR we just abort the header verification unless the error is one that `insertChain` deals with and still allows the processing of subsequent blocks.

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: Resolved in c8daffb4e39871bf76c40ac3d69280c98fa8ab80, and tests were repaired in [11bc7d2](https://github.com/autonity/autonity/pull/1045/commits/11bc7d2139c3ab69ec1a7e485404a18ec3750c5a). CI is failed due to the ASM tests which requires 3rd party dependency to be upgraded.

lorenzo-dev1:
Hi @Jason-Zhangxin-Chen , the PR now LGTM except for this last point: https://github.com/autonity/autonity/pull/1045#discussion_r1792071553 which could cause a crash when we are verifying a chain of headers and the epoch header is faulty.

I outline possible solutions in the comment, but we can discuss at standup if needed.

consensus/tendermint/backend/engine.go
Jason-Zhangxin-Chen: > I agree with @vpiyush that this might be a problem. We are basically using an untrusted committee to validate following headers.
> 
> This could cause a crash later on when verifying quorum certificates using this untrusted committee, since we panic if we cannot aggregate the keys of the committee (because we assume that it is trusted, but we are now breaking this assumption). Therefore we should either lower that log level OR we just abort the header verification unless the error is one that `insertChain` deals with and still allows the processing of subsequent blocks.

resolved in 5c031ebe783bcbb893ddccf4d6d960c387a25068.

Jason-Zhangxin-Chen:
> Hi @Jason-Zhangxin-Chen , the PR now LGTM except for this last point: [#1045 (comment)](https://github.com/autonity/autonity/pull/1045#discussion_r1792071553) which could cause a crash when we are verifying a chain of headers and the epoch header is faulty.
> 
> I outline possible solutions in the comment, but we can discuss at standup if needed.

Resolved in 5c031ebe783bcbb893ddccf4d6d960c387a25068 by replacing the critical logging with error logging and return that err to tell the verification result.

core/headerchain.go
```diff
@@ -550,14 +550,24 @@ func (hc *HeaderChain) GetHeaderByNumber(number uint64) *types.Header {
...
+	// as header chain does not have the state db, thus we cannot query it from state db.
+	// In this case, we just return this error which means the caller should discard the batch of headers,
+	// and the protocol will start another batch of sync with the correct head.
+	if height <= epochHead.Number.Uint64() || height > epochHead.Epoch.NextEpochBlock.Uint64() {
+		return nil, consensus.ErrOutOfEpochRange
 	}
 
-	return head.Epoch.Committee, head.Epoch.PreviousEpochBlock.Uint64(), head.Number.Uint64(), head.Epoch.NextEpochBlock.Uint64(), nil
+	epoch := &types.EpochInfo{
+		EpochBlock: epochHead.Number,
```
lorenzo-dev1, line 567: Since in `EpochOfHeight` in blockchain_reader.go we are returning a copy of number, I would return a copy also here for consistency.

consensus/tendermint/core/handler.go
```diff
@@ -20,15 +20,17 @@ const syncTimeOut = 30 * time.Second
 
 // Start implements core.Tendermint.Start
 func (c *Core) Start(ctx context.Context, contract *autonity.ProtocolContracts) {
-	committee, _, _, _, err := c.backend.BlockChain().LatestEpoch()
+
+	chainHead := c.backend.BlockChain().CurrentBlock().Header()
```
lorenzo-dev1, line 24: Can use `c.backend.HeadBlock()` instead of `c.backend.BlockChain().CurrentBlock()`

lorenzo-dev1 (approved):
LGTM! left two more low priority comments.

Also I see the ASM tests are failing, but I am not sure it is related to the changes in this PR.

core/headerchain.go
Jason-Zhangxin-Chen: Resolved in 624ccb4977446957c399aa204ab1433824e95589

consensus/tendermint/core/handler.go
Jason-Zhangxin-Chen: Resolved in 624ccb4977446957c399aa204ab1433824e95589

Jason-Zhangxin-Chen:
> LGTM! left two more low priority comments.
> 
> Also I see the ASM tests are failing, but I am not sure it is related to the changes in this PR.

Those ASM tests were failed due to 3rd dependency failures in CI, it does not relate to the changes in this PR.

core/blockchain_reader.go
```diff
@@ -68,22 +78,27 @@ func (bc *BlockChain) CommitteeOfHeight(height uint64) (*types.Committee, error)
 	if err != nil {
 		return nil, err
 	}
-	committee, err = bc.protocolContracts.GetCommitteeByHeight(currentHeader, stateDB, new(big.Int).SetUint64(height))
+
+	epoch, err = bc.protocolContracts.EpochByHeight(currentHeader, stateDB, new(big.Int).SetUint64(height))
```
yazzaoui, line 82: Can you add a DEBUG log statement here to better understand when this is called with 
requested height and current local height

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1065,30 +1065,31 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
     /**
-     * @notice Returns the committee of a specific height.
-     * @param _height the input block number
-     * @return committee The next epoch's consensus committee, if there is no epoch rotation, an empty set is returned.
-     */
-    function getCommitteeByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory) {
-        require(_height <= block.number, "cannot get committee for a future height");
+    * @notice Returns the epoch info of the height.
+    */
+    function getEpochByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory, uint256, uint256, uint256) {
```
yazzaoui, line 1070: Please add some tests for it using golang testing framework thanks

yazzaoui (approved):
LGTM let's get it merged, we can add the tests later but let's not forget

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1065,30 +1065,31 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
         if (_height <= lastFinalizedBlock) {
-            uint256 blockEpochID = blockEpochMap[_height];
-            CommitteeMember[] memory members = epochInfos[blockEpochID].committee;
-            return members;
+            blockEpochID = blockEpochMap[_height];
         }
 
-        // otherwise, this _height is the latest consensus instance, return current committee.
-        return committee;
+        CommitteeMember[] memory members = epochInfos[blockEpochID].committee;
```
vpiyush, line 1079: It would be better to load the epochInfo first into the memory and then access the fields to avoid multiple sload.

vpiyush: It's present in one other place too.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1065,30 +1065,31 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
-     */
-    function getCommitteeByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory) {
-        require(_height <= block.number, "cannot get committee for a future height");
+    * @notice Returns the epoch info of the height.
+    */
+    function getEpochByHeight(uint256 _height) public view virtual returns (CommitteeMember[] memory, uint256, uint256, uint256) {
+        require(_height <= lastFinalizedBlock+1, "cannot get epoch for a future block");
 
-        // if the block was already finalized, get committee by its corresponding epoch id.
+        uint256 blockEpochID = epochID;
```
vpiyush, line 1073: This causes an overwrite which redundant, blockEpochID is assgined twice. May be you could an if-else or a ternary operator assignement.

core/blockchain_reader.go
Jason-Zhangxin-Chen: Resolved in 90d0d03458bbf06e6d9d8f1282b20029e2133461

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: Resolved in 90d0d03458bbf06e6d9d8f1282b20029e2133461

autonity/solidity/contracts/Autonity.sol
Jason-Zhangxin-Chen: It is on purpose to take the current epoch id as the default value, if the height was finalized, then it get the epoch id from historic storage. I will refine it in other PR.

vpiyush: I meant you could do something like this to acheive the same thing but without double write
 uint256 blockEpochID = (_height <= lastFinalizedBlock) ? blockEpochMap[_height] : epochID;

PR #1056, opened 2024-10-25
Update getEpochInfo to return struct

ScottRMalley:
This is a small update to change the way that `getEpochInfo` on the Autonity contract returns. It now returns a struct, which is decoded in `autonity/calls.go`.

There is also a small change to the Makefile in here, as echo behaves differently on Zsh vs Bash, and this causes issues when using `make contracts`.


autonity/solidity/contracts/Autonity.sol
```diff
@@ -961,13 +961,8 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
     * @notice Returns the current epoch info of the chain.
     */
-    function getEpochInfo() external view virtual returns (CommitteeMember[] memory, uint256, uint256, uint256) {
-        EpochInfo memory epochInfo = epochInfos[epochID];
-        CommitteeMember[] memory members = epochInfo.committee;
-        uint256 previous = epochInfo.previousEpochBlock;
-        uint256 current = epochInfo.epochBlock;
-        uint256 next = epochInfo.nextEpochBlock;
-        return (members, previous, current, next);
+    function getEpochInfo() external view virtual returns (EpochInfo memory) {
```
tbssajal, line 964: I think the function `getEpochByHeight` should also be updated.

ScottRMalley: Good point. Updated in https://github.com/autonity/autonity/pull/1056/commits/3e125e80ad358569bd43510ecdb817d47d0a8f50

autonity/calls.go
```diff
@@ -562,23 +563,37 @@ func (c *AutonityContract) callGetCommitteeEnodes(state vm.StateDB, header *type
...
+		return nil, err
+	}
 
-	if err := c.AutonityContractCall(state, header, "getEpochByHeight", &[]any{&committeeMembers, &previousEpochBlock, &curEpochBlock, &nextEpochBlock}, height); err != nil {
+	data, err := c.contractABI.Unpack("getEpochByHeight", output)
+	if err != nil {
 		return nil, err
 	}
+
+	info := *abi.ConvertType(data[0], new(AutonityEpochInfo)).(*AutonityEpochInfo)
```
vpiyush, line 576: just curious, why couldn't we direct use this struct during contract call itself, avoiding the need to unpack into raw and convert

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1069,7 +1064,7 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
     * @notice Returns the epoch info of the height.
     */
     // todo: Jason, add some test for this function by using the new contract test framework.
```
vpiyush, line 1066: I think this todo can be removed now

vpiyush (approved):
LGTM, with minor comments

autonity/calls.go
ScottRMalley: From what I could tell by chasing errors, the `contractABI.UnpackToInterface` function doesn't seem to be able to handle nested structs (this is necessary because `EpochInfo` has a slice of `CommitteeMember` structs within it). 

Eventually, we may want to investigate whether `UnpackToInterface` is the correct function to be using withing `AutonityContractCall`, particularly if we want to return more structs from Solidity code, but for now I thought it was best to handle the casting for the individual cases where we need it.

autonity/solidity/contracts/Autonity.sol
ScottRMalley: I'm not sure whether the tests I created will cover all logic of this function. I'd leave it in for now, as the test I wrote is specifically to test code in the `calls.go` of the autonity package, and not the solidity code. (eg. I'd expect a test here to make sure this reverts on a future block.)

lorenzo-dev1 (approved):
LGTM

yazzaoui (approved):
LGTM

PR #1059, opened 2024-10-28
auto-bond ntn and test fix

tbssajal:
- [x] Implemented auto-bonding by increasing LNTN conversion ratio
- [ ] Tests for auto-bonding


autonity/solidity/contracts/Autonity.sol
```diff
@@ -1192,15 +1194,17 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
...
                 }
+                // auto-bonding `_ntnReward`
+                _val.bondedStake += _ntnReward;
                 uint256 _ntnSelfReward = (_val.selfBondedStake * _ntnReward) / _val.bondedStake;
-                if (_ntnSelfReward > 0) {
-                    _transfer(address(this), _val.treasury, _ntnSelfReward);
+                uint256 _commission = ILiquidLogic(_val.liquidStateContract).calculateValidatorCommission(_ntnReward - _ntnSelfReward);
+                if (_ntnSelfReward + _commission > 0) {
+                    // auto-bonding to `selfBondedStake`
+                    _val.selfBondedStake += _ntnSelfReward + _commission;
```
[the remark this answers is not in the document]
tbssajal: I guess reading from state is more costly than the if block but I could be wrong.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1192,15 +1194,17 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, Upgradeable {
                         emit CallFailed(_val.treasury, "", _returnData);
                     }
                 }
+                // auto-bonding `_ntnReward`
+                _val.bondedStake += _ntnReward;
```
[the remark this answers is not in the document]

autonity/solidity/contracts/liquid/LiquidLogic.sol
```diff
@@ -341,16 +313,20 @@ contract LiquidLogic is ILiquid, LiquidStorage {
      ============================================================
      */
 
+    function calculateValidatorCommission(uint256 _reward) public virtual view returns (uint256) {
+        uint256 _commission = (_reward * commissionRate) / COMMISSION_RATE_PRECISION;
+        require(_commission <= _reward, "commission is greater than reward");
```
vpiyush, line 318: Is this even possible ? The failure of this depends commissionRate being more than commission_rate_precision. and none of that is the parameter of this function and being set elsewhere. So this check Ideally belongs while setting those two values. Second: if this failure does occur. This would cause finalize failure halting chain on epoch block. If we decide to keep this check this should be invoked in try-catch.

tbssajal: Yeah I think we don't need this check because we have a check when we update `commissionRate`.

autonity/solidity/contracts/Autonity.sol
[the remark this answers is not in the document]
ScottRMalley: I believe we are already paying these gas fees when we load the struct like this:
```
Validator storage _val = validators[committee[i].addr];
```

You can only benefit from gas fee optimizations if you don't load the whole struct when accessing the field, so if we had something like:
```
 validators[committee[i].addr].selfBondedStake += _ntnSelfReward + _commission;
```
there might be a difference. In this cas there is no difference and we are only increasing the lines of code.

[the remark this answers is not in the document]
ScottRMalley: Yeah you are right. I checked the gas fees for both methods, and the 
```
Validator storage _val = validators[committee[i].addr];
```
only adds gas fees from variable assignment and not from storage access. In addition, 
```
if (_ntnSelfReward + _commission > 0) {
                    // auto-bonding to `selfBondedStake`
                    _val.selfBondedStake += _ntnSelfReward + _commission;
}
```
does actually make the gas fees slightly less when the reward is zero, but slightly more when the reward is nonzero.

The gas fee difference is very negligible. Really it only depends how often we expect this to be zero, which I would argue is almost never (the validator would have to have no self bonded stake I believe).

vpiyush (approved):
LGTM now.

autonity/solidity/contracts/Autonity.sol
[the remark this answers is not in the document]
tbssajal: Right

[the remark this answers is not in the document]
tbssajal: Although I think it is very probable to have `selfBondedStake = 0` because it absorbs slashing penalties, commission should never be zero.

tbssajal:
I had to do a bit of refactoring due to the fact that there are multiple protocol contracts that distribute rewards. Please review again if you already reviewed @vpiyush @ScottRMalley @Jason-Zhangxin-Chen @lorenzo-dev1 @yazzaoui .

autonity/solidity/contracts/Autonity.sol
```diff
@@ -869,6 +882,15 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
         return (_oracleVoters, _afdReporters, _treasuries);
     }
 
+    function autobond(address _address, uint256 _selfBond, uint256 _delegated) external virtual onlyAtEpochEnd onlyRewardDistributer(_address) {
+        require(accounts[msg.sender] >= _selfBond + _delegated, "not enough balance");
+        require(_address != address(0), "validator address cannot be zero");
+        require(validators[_address].nodeAddress == _address, "validator not registered");
+
+        accounts[msg.sender] -= _selfBond + _delegated;
```
lorenzo-dev1, line 890: Just thinking, does it make sense to use `burn` and `mint` for the `autobonding` logic? Or does it just overcomplicate things?

Also should the `autobond` procedure generated a net change in `stakeSupply` and `stakeCirculating`? or should they remain the same?

autonity/tests/omission_accountability_test.go
```diff
@@ -690,8 +677,8 @@ func TestProposerRewardDistribution(t *testing.T) {
 		proposer := r.Committee.Validators[0].NodeAddress
 		proposerTreasury := r.Committee.Validators[0].Treasury
 		atnBalanceBefore := newFloat(r.GetBalanceOf(proposerTreasury))
-		ntnBalanceBefore := newFloat(ntnBalance(r, proposerTreasury))
-		t.Logf("atn balance before: %s, ntn balance before %s", toString(atnBalanceBefore), toString(ntnBalanceBefore))
+		stakesBefore := newFloat(r.Committee.Validators[0].BondedStake)
```
lorenzo-dev1, line 680: Shouldn't it be the `SelfBonded` one?

lorenzo-dev1 (approved):
Just a couple of questions, but LGTM!

autonity/solidity/contracts/Autonity.sol
tbssajal: > Also should the autobond procedure generated a net change in stakeSupply and stakeCirculating? or should they remain the same?

They remain the same. That's why burn and mint will make it a bit complicated because you will then need to do the following 
```
accounts[treasury] -= _selfbond + _delegated
```

autonity/tests/omission_accountability_test.go
tbssajal: It's a bit complicated I guess. The proposer reward will go to self-bonded stake. But the validator reward will go to both the self-bonded and delegated stake. So I just wanted to make it simple. Anyways, I will add tests that will focus on auto-bonding.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1231,6 +1253,22 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
+            || address(config.contracts.omissionAccountabilityContract) == msg.sender
+            || address(validators[_validator].liquidStateContract) == msg.sender
+            , "caller is not a reward distributer"
+        );
+        _;
+    }
+
+    modifier onlyAtEpochEnd() {
+        // require(true, "it passes");
+        require(block.number >= epochInfos[epochID].nextEpochBlock, "epoch is not ended");
```
vpiyush, line 1268: The value of epochID should be passed to this function. Otherwise it's error prone. an order change of epochID increment in finalize function which would seem harmless but ultimately impact here. It's also better for code readbility

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1231,6 +1253,22 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
+            address(config.contracts.accountabilityContract) == msg.sender
+            || address(config.contracts.omissionAccountabilityContract) == msg.sender
+            || address(validators[_validator].liquidStateContract) == msg.sender
+            , "caller is not a reward distributer"
+        );
+        _;
+    }
+
+    modifier onlyAtEpochEnd() {
+        // require(true, "it passes");
```
vpiyush, line 1267: remove this.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1524,6 +1561,28 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
         }
     }
 
+    function _autobond(address _address, uint256 _selfBond, uint256 _delegated) internal virtual {
+        if (_selfBond == 0 && _delegated == 0) {
+            return;
+        }
+        Autobond storage _request = autobondingMap[_address];
+        if (_request.totalBond == 0) {
+            validatorsUnderAutobonding.push(_address);
```
vpiyush, line 1570: validatorsUnderAutobonding need not be a state variable i believe. It seems to be only used temporarily

vpiyush: same for autobondingMap as well

ScottRMalley: `_autobond` is being called both internally, and externally from the Accountability contracts, so even though it's occurring in a single finalize call, we need to have a way of maintaining state between function calls if we want to apply the total autobonding amounts at the end of all redistribution actions.

 In addition, there is no way to have `memory` storage allocation to a `mapping` in solidity, so if we want to use mappings, they will have to be in `storage`.

That being said, why are we gathering all the autobondings and applying them at once? I can imagine there might be a gas benefit, but I think it's unlikely as we are still reading and writing to storage on every `_autobond` call (just writing to `autobindingMap` instead of `_validator`. @tbssajal Why not just apply autobondings as they occur? I don't think it's the end of the world if a validators stake gets incremented a few times over the course of an epoch end.

tbssajal: > That being said, why are we gathering all the autobondings and applying them at once?

If you see the flow in the `finalize()` function, all new staking operations are to be applied at last because they increase the stakes and might affect other operations. For example, if omission rewards are applied as autobonding immediately then the validator rewards distribution will be wrong.

tbssajal: I think it is better to apply all the autobonding together in staking operation because it keeps the flow clean.

tbssajal: I see you changed the flow quite a bit to remove the new struct and state variables. It works as well.

ScottRMalley: Unfortunately I just pushed some changes that remove the mapping. We discussed it a bit in standup and decided to go without it. I wasn't expecting you to be back online today, or I would have waited for some input.

In general, I think both methods have some trade-offs. The order of rewards distribution in the `_performRedistribution` now matters because external contracts will be increasing the bonded stake through autobonding. However, I do think that there is more that can be done to clean this up. It doesn't make so much sense to me in the first place that these external contracts need to be called to do the rebonding. I'm going to think about it some more.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1304,22 +1297,20 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
-        if(config.contracts.omissionAccountabilityContract.getTotalEffort() > 0){
-            uint256 _atnProposerRewards = (_atn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
-            uint256 _ntnProposerRewards = (_ntn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
-            address _omission = address(config.contracts.omissionAccountabilityContract);
-            _transfer(address(this), _omission, _ntnProposerRewards);
-            config.contracts.omissionAccountabilityContract.distributeProposerRewards{value: _atnProposerRewards}(accounts[_omission]);
+        // Calculate initial proposer rewards (actual distribution is done after regular rewards)
+        uint256 _atnProposerRewards = (_atn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
+        uint256 _ntnProposerRewards = (_ntn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
```
lorenzo-dev1, line 1302: This works but for code clarity I think it would be better to assign them only if totalEffort > 0. Or maybe copy the initial `_atn` and `_ntn` values and use it later when redistributing rewards for omission.

With the current code we might have proposer rewards variables != 0 but actually no reward getting re-distributed, which I think is more error prone.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1304,22 +1297,20 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
+
+        if (config.contracts.omissionAccountabilityContract.getTotalEffort() > 0) {
             _atn -= _atnProposerRewards;
             _ntn -= _ntnProposerRewards;
         }
 
         uint256 _omissionScaleFactor = config.contracts.omissionAccountabilityContract.getScaleFactor();
 
+        uint256[] memory _jailedValidators = new uint256[](committee.length);
```
lorenzo-dev1, line 1311: I feel like it would be easier to read if we would store directly a pointer to the committee struct. But I am not sure if it would increase gas consumption.
 
 The double indrection later on makes it a bit hard to read, but it is not a big deal really.

autonity/solidity/contracts/Autonity.sol
ScottRMalley: I'm not sure how you are imagining this. This modifier is called on an external function, are you suggesting that `epochId` be passed into `autobond`? That would require tracking of epoch id in other rewards distributing contracts.

I don't think it makes sense to change this. Having epoch id incremented at the very end of the finalize function is probably best practice in either case.

autonity/solidity/contracts/Autonity.sol
ScottRMalley: removed

autonity/solidity/contracts/Autonity.sol
vpiyush: Yeah, That's how i was thinking about it. To be passing epochId in autobond. But if that makes things complex then there is no advantage of doing it.  Although, we should have a note in the finalize to avoid epochId increment order getting messed up.

ScottRMalley: The external contracts would either need to fetch the `epochID` from the Autonity contract, or have a `setNewEpochId` function that is set from the finalize function. I don't see how this would solve the issue of ordering, as now there would be an order constraint on when the external contracts fetch the ID, or when the ID is updated by the Autonity contract.

autonity/solidity/contracts/Autonity.sol
vpiyush: Yes It does. @ScottRMalley

autonity/solidity/contracts/Autonity.sol
ScottRMalley: Added a note to not move the epoch ID increment.

PR #1061, opened 2024-11-06
Epoch fix

vpiyush:
Fix epoch info fetch failure after and unclean shutdown.

### DIFF
--- a/consensus/tendermint/backend/engine.go
+++ b/consensus/tendermint/backend/engine.go
@@ -266,6 +266,10 @@ func (sb *Backend) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*t
 				err = sb.verifyHeader(chainConfig, header, parent, epoch, hash)
 			}
 
+			if err != nil {
+				sb.logger.Error("VerifyHeaders", "Error verifying header", "error", err)
+			}
+
 			// cross epoch header check, update the committee and epoch boundary if current header is an epoch head.
 			// the verification behind this header will be continued with the updated epoch info.
 			if header.IsEpochHeader() {
--- a/consensus/tendermint/backend/engine.go
+++ b/consensus/tendermint/backend/engine.go
@@ -88,7 +88,7 @@ func (sb *Backend) VerifyHeader(chain consensus.ChainHeaderReader, header *types
 	}
 
 	// get the epoch information for the header we are verifying
-	epoch, err := chain.EpochOfHeight(header.Number.Uint64(), nil)
+	epoch, err := chain.EpochByHeight(header.Number.Uint64(), nil)
 	if err != nil {
 		return fmt.Errorf("cannot fetch epoch information for height %d: %w", header.Number.Uint64(), err)
 	}
@@ -240,7 +240,7 @@ func (sb *Backend) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*t
 
 	go func() {
 		firstHeight := headers[0].Number.Uint64()
-		epoch, err := chain.EpochOfHeight(firstHeight, nil)
+		epoch, err := chain.EpochByHeight(firstHeight, nil)
 		// short circuit, if we cannot find the correct epoch for the 1st header, we quit this batch of verification.
 		if err != nil {
 			sb.logger.Error("VerifyHeaders", "cannot find epoch for the 1st header of the batch: ", err.Error(), "height", firstHeight)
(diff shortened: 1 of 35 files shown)


core/blockchain.go
```diff
@@ -564,6 +555,7 @@ func (bc *BlockChain) loadLastState() error {
 	bc.log.Info("Loaded most recent local header", "number", currentHeader.Number, "hash", currentHeader.Hash(), "td", headerTd, "age", common.PrettyAge(time.Unix(int64(currentHeader.Time), 0)))
 	bc.log.Info("Loaded most recent local full block", "number", currentBlock.Number(), "hash", currentBlock.Hash(), "td", blockTd, "age", common.PrettyAge(time.Unix(int64(currentBlock.Time()), 0)))
 	bc.log.Info("Loaded most recent local fast block", "number", currentFastBlock.Number(), "hash", currentFastBlock.Hash(), "td", fastTd, "age", common.PrettyAge(time.Unix(int64(currentFastBlock.Time()), 0)))
+	bc.log.Info("Loaded most recent epochBlock", "number", epochBlock.Number(), "hash", epochBlock.Hash(), "td", fastTd, "age", common.PrettyAge(time.Unix(int64(epochBlock.Time()), 0)))
```
lorenzo-dev1, line 558: It seems like we just load this epoch block for logging purposes, is it even useful to load it?

Or is it just to check that the blockchain is not corrupted?

core/blockchain_reader.go
```diff
@@ -67,63 +66,28 @@ func (bc *BlockChain) EpochOfHeight(height uint64, customFetcher consensus.Heade
...
-	header, stateDB, err := fetcher()
-	if err != nil {
-		return nil, err
+		// when we are building an optimistic block we will use the cached state via customFetcher
+		header, stateDB, err := customFetcher()
+		if err != nil {
+			return nil, err
+		}
+		// we intentionally do not add in the epochCache because this flow is only for optimistic blocks
+		return bc.protocolContracts.EpochByHeight(header, stateDB, new(big.Int).SetUint64(height))
```
lorenzo-dev1, line 76: Since we utilize the state to fetch epoch information only when building optimistic blocks, I guess the default fetcher can be removed.

lorenzo-dev1 (approved):
LGTM with a couple of comments. moreover some questions:
1. It seems that the PR changes calls to `StateAt` to use the `CurrentBlock()` instead of the `CurrentHeader()`. I assume that's because the state for `CurrentHeader()` could be missing. However there are still some calls to `StateAt` which rely on `CurrentHeader` (such as `EpochByHeight` in engine.go)
2. the `customFetcher` logic was initially introduced because we were potentially fetching from state as a part of the normal flow. So I needed something to play well with that. However rn we don't fetch from the state as a part of the normal flow anymore, so I believe this part could be refactored. But it is not blocking to merge the PR, just maybe capture it into an issue to not forget.

Jason-Zhangxin-Chen (approved):
LGTM.

core/blockchain_reader.go
vpiyush: yes, removing that

core/blockchain.go
vpiyush: removed it

Jason-Zhangxin-Chen (approved):
LGTM.

vpiyush:
> LGTM with a couple of comments. moreover some questions:
> 
> 1. It seems that the PR changes calls to `StateAt` to use the `CurrentBlock()` instead of the `CurrentHeader()`. I assume that's because the state for `CurrentHeader()` could be missing. However there are still some calls to `StateAt` which rely on `CurrentHeader` (such as `EpochByHeight` in engine.go)
> 2. the `customFetcher` logic was initially introduced because we were potentially fetching from state as a part of the normal flow. So I needed something to play well with that. However rn we don't fetch from the state as a part of the normal flow anymore, so I believe this part could be refactored. But it is not blocking to merge the PR, just maybe capture it into an issue to not forget.

1. Changed the behaviour for epochByHeight in engine.go to use the current block. I think there are no other places where we use currentHeader now to fetch epoch block
2. The customerFetcher logic still goes well with the current flow. We don't use state anymore in normal flow. but for optimistic flow we do need it. So we could either have a separate function Or the similar logic which we already have. I would say it's fine as it is.

consensus/tendermint/backend/engine.go
```diff
@@ -347,9 +350,7 @@ func (sb *Backend) Prepare(_ consensus.ChainHeaderReader, parentHeader, header *
 
 	height := header.Number.Uint64()
 	// use a custom fetcher to be able to fetch epoch information in the case we are building an optimistic block at the epoch boundary
```
ScottRMalley, line 352: I think this comment doesn't make much sense anymore.

ScottRMalley (approved):
LGTM, only one small non-essential comment.

consensus/tendermint/backend/engine.go
vpiyush: yes, It has been removed

PR #1065, opened 2024-11-06
Restrict opening of CDPS

ScottRMalley:
This PR is in reference to the [ADR-014](https://github.com/clearmatics/autonity-protocol/pull/300) regarding restriction of opening CDPs to an ADF controlled account.

As per the discussion [here](https://github.com/clearmatics/autonity-protocol/discussions/298#discussioncomment-11165840), the modifications include:

- add an `onlyOperator` function `setATNSupplyOperatorAccount(address atnSupplyOperator)`
- leave `BorrowInterestRate` in the `DefaultStabilizationGenesis` at its 5% setting, but initialize the contract to a borrow interest rate of 0.
- restrict the following contract functions to be only available to the `_atnSupplyOperator`
    - `deposit`
    - `withdraw`
    - `borrow`
    - `repay`
- completely disable `liquidate`
- add an `onlyOperator` toggle method `removeCDPRestrictions()` with no params, which: enables `liquidate`, removes the supply operator constraint on `borrow`, and resets `config.borrowInterestRate` in the contract to the `DefaultStabilizationGenesis` value at 5%.


autonity/tests/omission_accountability_test.go
```diff
@@ -255,136 +257,140 @@ func TestProposerLogic(t *testing.T) {
 
 // checks that the inactivity counters are correctly updated according to the lookback window (_recordAbsentees function)
 func TestInactivityCounter(t *testing.T) {
-	r := Setup(t, configOverrideIncreasedStake)
+	setup := func() *Runner {
+		return Setup(t, configOverrideIncreasedStake)
+	}
+	RunWithSetup("Inactivity counter increases correctly", setup, func(r *Runner) {
```
lorenzo-dev1, line 263: What's the difference with previous code here?

autonity/solidity/contracts/asm/Stabilization.sol
```diff
@@ -149,6 +154,18 @@ contract Stabilization is IStabilization {
         _;
     }
 
+    // Restricted to the atnSupplyOperator during the initial CDP restrictions
+    modifier restrictedSupplyOperator() {
+        if (_restricted && msg.sender != _atnSupplyOperator) revert Unauthorized();
```
lorenzo-dev1, line 159: Just wondering, could it be a problem that the `_atnSupplyOperator` right after genesis will be the zero address?

lorenzo-dev1 (approved):
LGTM just a couple questions

autonity/tests/omission_accountability_test.go
ScottRMalley: Honestly, I can't be 100% sure. This test was failing for me, even though it should be unrelated to the changes I made. That being said, it wasn't failing when run alone, only when it was being run together with other tests. @tbssajal made the `RunWithSetup` function, which better isolates test environments from each other. I decided to try and see whether using it fixed my problem, and it did.

I think we should probably migrate most of the tests to use `RunWithSetup` in either case.

autonity/solidity/contracts/asm/Stabilization.sol
cmjc: This is OK @lorenzo-dev1 - the sequence would be to set the address, then open the CDP. It's temporary, so no need (and cleaner not to have) for an address set in the genesis config.

autonity/tests/omission_accountability_test.go
yazzaoui: Hmm. A `run` block should be running in isolation and we shouldn't see those problems. I would prefer to investigate what's wrong when we can. Perhaps that's because we're not saving the block/transaction context in the state snapshot.

yazzaoui (approved):
LGTM

autonity/tests/omission_accountability_test.go
lorenzo-dev1: I think it is just that the test itself is flaky. Running it 100 times will yield some failures. I am looking into it.

@ScottRMalley I think the change to the test can be reverted, it doensn't really solve the issue.

ScottRMalley: Ok, the change is reverted!

autonity/solidity/contracts/asm/Stabilization.sol
ScottRMalley: Yeah I've been thinking about this as well. The only scenario that this would be a problem is when `msg.sender` is the zero address, and we have not set the `_atnSupplyOperator`. Considering we are planning on immediately setting the supply operator after genesis, and `msg.sender` should never be the zero address outside of core contract deployment.

autonity/tests/omission_accountability_test.go
ScottRMalley: Just to update this here @lorenzo-dev1 investigated the issue, and the test seems to occasionally fail due to accumulated errors between the off-chain calculation and the on-chain one, which results in a very slight difference between expected and actual inactivity score. As a fix, I updated the test to check within a tolerance of 1. This doesn't significantly change the test expectation but allows more stable test outcomes.

Issue #1034, opened 2024-08-21
Revise the `aut` JSON-RPC Namespace


aiman:
### Description

We'd like to remove all JSON-RPC methods from the `aut` Namespace that can be replicated through `eth_call`. So in the end, we only keep in the `aut` Namespace:

- `aut_address`: The `address` of the Autonity Contract itself
- `aut_config`: The current `config` of the Autonity Contract (justifiable exception to principle)
- `aut_acnPeers`: Information about each consensus network peer

PR #1066, opened 2024-11-08
remove json-rpc api's from aut namespace

vpiyush:
See details here https://github.com/autonity/autonity/issues/1034


eth/api.go
```diff
@@ -611,7 +611,25 @@ func (api *PrivateDebugAPI) GetAccessibleState(from, to rpc.BlockNumber) (uint64
...
+	ac     *autonity.ProtocolContracts
+	server *p2p.Server
+}
+
+func (a *AutonityContractAPI) Config() (*autonity.AutonityConfig, error) {
+	st, err := a.bc.State()
+	if err != nil {
+		return nil, err
+	}
+	return a.ac.CallConfig(st, a.bc.CurrentHeader())
```
lorenzo-dev1, line 624: I think it would be better to use `StateAt` and `CurrentBlock`

lorenzo-dev1 (approved):
LGTM just a minor comment

Jason-Zhangxin-Chen (approved):
LGTM, please double check if those e2e/docker test frameworks enabled `aut` namespace, should be disabled from the config if they were enabled.

autonity/autonity.go
```diff
@@ -125,10 +125,15 @@ func newCache(ac *AutonityContract, head *types.Header, state vm.StateDB) (*Cach
...
 	*AutonityFilterer // allows to watch for Autonity Contract events
+	address           common.Address
 
 	proposers map[uint64]map[int64]common.Address // map[height][round] --> proposer
 }
 
+func (ac *AutonityContract) Address() common.Address {
+	return ac.address
+}
+
```
ScottRMalley, line 136: Is there a reason we need to add this? Won't the AC address always be as listed in `params.AutonityContractAddress`?

vpiyush: you are right. this is redundant. Removed it.

vpiyush:
> LGTM, please double check if those e2e/docker test frameworks enabled `aut` namespace, should be disabled from the config if they were enabled.

I don't see a problem with aut namespace being enabled though, because we still support it.Is there any other reason we should disable it?

PR #1067, opened 2024-11-14
Oracle accountability

yazzaoui:
Implement oracle accountability

WIP:
- Autobonding integration
- Tests missing


autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,200 +2,295 @@
...
+            // punish outliers if found
+            for(uint256 i = 0; i < _totalOutliers; i++) {
+                _penalize(_outliers[i], _priceMedian, reports[_symbol][_outliers[i]]);
+                emit Penalized(_outliers[i], _symbol, _priceMedian, reports[_symbol][_outliers[i]].price);
+            }
+            _success = true;
+        } else {
+            // use past value for price if unsuccesful
+            _price = prices[round-1][_symbol].price;
         }
```
lorenzo-dev1, line 285: I think that the `_price` does not get updated in case that `_count` > 0. I guess it should be the median of the filtered reports?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -341,8 +518,81 @@ contract Oracle is IOracle {
...
+            if (_ratio <= config.outlierDetectionThreshold && -1 * _ratio <= config.outlierDetectionThreshold) {
+                _filteredReports[_totalReports++] = reports[_symbol][_voter];
+                // take advantage of this iteration to include performance calculation
+                voterInfo[_voter].performance += reports[_symbol][_voter].confidence;
+                aggregatedScore += reports[_symbol][_voter].confidence;
+            } else {
+                _outliers[i - _totalReports] = _voter;
+            }
+        }
+        _totalOutliers = voters.length - _totalReports;
```
lorenzo-dev1, line 567: I think it would be better to keep it simple and increase the `_totalOutliers` return variable, it is already declared as a part of the return parameters anyways

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1270,16 +1278,24 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
+        uint256 _ntnProposerRewards = 0;
         if(config.contracts.omissionAccountabilityContract.getTotalEffort() > 0){
-            uint256 _atnProposerRewards = (_atn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
-            uint256 _ntnProposerRewards = (_ntn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
+            _atnProposerRewards = (_atn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
+            _ntnProposerRewards = (_ntn * config.policy.proposerRewardRate * committee.length) / (PROPOSER_REWARD_RATE_PRECISION * config.protocol.committeeSize);
             address _omission = address(config.contracts.omissionAccountabilityContract);
             _transfer(address(this), _omission, _ntnProposerRewards);
             config.contracts.omissionAccountabilityContract.distributeProposerRewards{value: _atnProposerRewards}(accounts[_omission]);
-            _atn -= _atnProposerRewards;
```
tbssajal, line 1279: Removing this enforces some additional constraints on `config.policy.proposerRewardRate` and `config.policy.oracleRewardRat`. What we need now is the following:
```
config.policy.oracleRewardRat / ORACLE_REWARD_RATE_PRECISION + config.policy.proposerRewardRat / PROPOSER_REWARD_RATE_PRECISION <= 1
```

If we consider `PROPOSER_REWARD_RATE_PRECISION = ORACLE_REWARD_RATE_PRECISION`, then we have the condition
```
config.policy.oracleRewardRat + config.policy.proposerRewardRat <= REWARD_RATE_PRECISION
```

Otherwise when doing the following
```
_atn -= _atnProposerRewards + _atnOracleRewards
```
we can get overflow error.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1270,16 +1278,24 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
-            _ntn -= _ntnProposerRewards;
         }
 
+        uint256 _atnOracleRewards = _atn * config.policy.oracleRewardRate / ORACLE_REWARD_RATE_PRECISION;
+        uint256 _ntnOracleRewards = _ntn * config.policy.oracleRewardRate / ORACLE_REWARD_RATE_PRECISION;
+        _transfer(address(this), address(config.contracts.oracleContract), _ntnOracleRewards);
+        config.contracts.oracleContract.receiveRewards{value:_atnOracleRewards}();
```
tbssajal, line 1294: `_ntnOracleRewards` is not passed. Is it intentional?

tbssajal (changes requested):

autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,200 +2,295 @@
...
+            uint256 _atn = (_totalATN * voterInfo[_voter].performance) / aggregatedScore;
+            uint256 _ntn = (_totalNTN * voterInfo[_voter].performance) / aggregatedScore;
+
+            // Transfer ATN rewards
+            // 2300 gas fowarded with send()
+            // funds for failed transfers will be redistributed for the next round
+            payable(_voter).send(_atn);
+
+            // Transfer NTN rewards
+            config.autonity.transfer(_voter, _ntn);
```
tbssajal, line 241: Shouldn't we transfer the rewards to treasury address?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -341,8 +522,81 @@ contract Oracle is IOracle {
...
+        )
+    {
+        _filteredReports = new Report[](voters.length);
+        _outliers = new address[](voters.length);
+        for(uint256 i = 0; i < voters.length; i++) {
+            address _voter = voters[i];
+            if(!voterInfo[_voter].reportAvailable) {
+                continue;
+            }
+            // median here is assumed to be non-0.
```
vpiyush, line 559: a safety check for median != 0 could be added in the begining of the function

autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,200 +2,299 @@
...
+
+        if (_count > 0){
+            int256 _priceMedian = int256(uint256(_getMedian(_totalReports, _count)));
+            // exclude and detect outliers
+            (address[] memory _outliers, uint256 _totalOutliers, Report[] memory _filteredReports, uint256 _totalConfidence)
+                = _findOutliers(_priceMedian, _symbol);
+            // There is an extreme edge-case where everyone is detected outlier. This is left todo.
+            // punish outliers if found
+            for(uint256 i = 0; i < _totalOutliers; i++) {
+                _penalize(_outliers[i], _priceMedian, reports[_symbol][_outliers[i]]);
```
vpiyush, line 282: len( _outliers) won't always be same as _totalOutliers. The way we are calculating _totalOutliers.  we could potentially invoke this with address 0 and report 0.

autonity/solidity/contracts/Oracle.sol
yazzaoui: Yes, calling the price calculation function was missing here

autonity/solidity/contracts/Oracle.sol
yazzaoui: Depreciated with autobond integration

autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,284 +2,510 @@
...
-        int256 _priceMedian = prices[round-1][_symbol].price;
-        bool success = false;
+
+        uint256 _price = 0;
+        bool _success = false;
+
         if (_count > 0) {
-            _priceMedian = _getMedian(_totalReports, _count);
-            success = true;
+            int256 _priceMedian = int256(uint256(_getMedian(_totalReports, _count)));
```
ScottRMalley, line 299: I think there is an issue, here. We never set `_price` here, so assuming that `_count > 0` we will never have a price that is non-zero. In addition, weren't we planning on including the confidence interval when calculating the price?

yazzaoui: Mentionned here https://github.com/autonity/autonity/pull/1067#discussion_r1852481511 that indeed the price calculation invocation was missing.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,200 +2,299 @@
...
+        voterInfo[msg.sender].round = round;
         // new voter/first round
         if (_lastVotedRound == 0 ) {
             return;
         }
 
         // if data is not supplied and voter is not a new voter
         // report must contain the correct price
-        if(_reports.length != symbols.length)  {
+        if(_reports.length != symbols.length)  { // todo: why not require?
```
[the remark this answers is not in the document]
yazzaoui: Good points, I'll remove this comment.

autonity/solidity/contracts/Autonity.sol
yazzaoui: Agree but this constraint enforcement can be done offline as this being a governance operation

tbssajal: In that case we should add some checks in Prepare() function that checks the genesis config. And we should have `@notice` that explain the enforcement in both `setOracleRewardRate` and `setProposerRewardRate` function.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -335,14 +563,85 @@ contract Oracle is IOracle {
...
+            uint256 _totalOutliers,
+            Report[] memory _filteredReports,
+            uint256 _totalReports
+        )
+    {
+        _filteredReports = new Report[](voters.length);
+        _outliers = new address[](voters.length);
+        for (uint256 i = 0; i < voters.length; i++) {
+            address _voter = voters[i];
+            if (!voterInfo[_voter].reportAvailable) {
```
tbssajal, line 603: Here the voter who did not vote will not be considered an outlier and will not be penalized. If voting is mendatory, then this voter should be an outlier or at least be penalized. Or maybe we could consider his last reported price for this symbol to determine if the voter is an outlier or not.

tbssajal (changes requested):

autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,200 +2,299 @@
...
             return;
         }
-        // Voter has to vote on all the symbols
-        // uint256 MAX_INT = uint256(-1) is a special value
+
+        // Voter voted on every symbols
         for (uint256 i = 0; i < _reports.length; i++) {
-             reports[symbols[i]][msg.sender] = _reports[i];
+            require(_reports[i].confidence <= 100, "invalid confidence score");
+            require(
```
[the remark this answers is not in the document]
vpiyush: if a voter calculated a 0 price and sent a commit for it in round r in the round r+1 it can not register a commit for the price report which it is going to send in r+2. because r+1 will always revert.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -2,284 +2,511 @@
...
+        }
+        emit NewRound(round, block.number, block.timestamp, config.votePeriod);
+        return true;
+    }
+
+    function _finalizeRewards() internal {
+        for (uint256 i = 0; i < voters.length; i++) {
+            address _voter = voters[i];
+            if (voterInfo[_voter].performance > 0) {
+                rewardReceivers.add(_voter);
```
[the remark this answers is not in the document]
tbssajal: #1085 contains a test for this which is currently failing. After the fix the test should pass.

autonity/solidity/contracts/Oracle.sol
vpiyush: This got fixed [here](https://github.com/autonity/autonity/pull/1067/commits/ddede31011b01c9b27adea04ed1d4786b35b77a1), Resolving this comment now.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -335,14 +564,92 @@ contract Oracle is IOracle {
...
+    }
+
+    function _penalize(address _outlier, int256 _median, Report memory _report) internal {
+        // Stop considering this reporter for any future calculation.
+        // This is symbol independant.
+        voterInfo[_outlier].reportAvailable = false;
+        int256 _diffRatio = (int256(uint256(_report.price)) - _median) * 100 / _median;
+        //price is 120 bits max so _diffratio squared is at most 240 bits
+        _diffRatio = _diffRatio * _diffRatio;
+        if (_diffRatio < config.outlierSlashingThreshold) {
```
tbssajal, line 638: Consider writing `if (_diffRatio <= config.outlierSlashingThreshold)` instead. Otherwise we are not realling slashing when `_diffRatio == config.outlierSlashingThreshold)` but still doing some unnecessary calls.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: In this case, the voter is omission faulty, its loss of the performance score causes the decreasing of its fractions of the ATN & NTN reward, and there is no slashing penalty for omission faulty node. 

We don't reset `reportAvailable` after a voting round, so in general we are already using the previously submitted price in the case that a user does not submit a new price. The only scenario where `reportAvailable` will be false is if a voter has never submitted a report, or the voter set just changed (ie. epoch end).

tbssajal: I see. But in that case someone who never submits a vote will be in profit compared to an outlier, right? Because the outlier is getting slashed and not getting any rewards?

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Fair point, done.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Yes, this is correct. Outlier behavior is counted as malicious, rather than just omission faulty.

tbssajal: Alright makes sense.

tbssajal (approved):
LGTM

PR #1071, opened 2024-11-18
fix, do not trigger the optimistic parent block's logs on child blocks.

Jason-Zhangxin-Chen:
This PR fix the duplicated TXN log notification on the client when making environment for the child block of an optimistic parent block for consensus pipeline optimization.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -625,6 +625,9 @@ func (w *worker) resultLoop() {
 				continue
 			}
 
+			// The logs inside the task.env.state may contain parent block's logs as we copied the parent state when
+			// making environment for the child of an optimistic parent block for consensus pipeline optimization.
+			// Thus, we construct logs from the task.env.receipts that is unique for block.
 			var logs []*types.Log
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -625,20 +625,28 @@ func (w *worker) resultLoop() {
...
+			var logs []*types.Log
 			// Update the block hash in all logs since it is now available and not when the
 			// receipt/log of individual transactions were created.
 			for i, r := range task.env.receipts {
 				r.BlockHash = block.Hash()
 				r.BlockNumber = block.Number()
 				r.TransactionIndex = uint(i)
-				for _, l := range r.Logs {
-					l.BlockHash = block.Hash()
+				for i, taskLog := range r.Logs {
```
vpiyush, line 638: What does this loop do? isn't this just copying the same thing over again.

vpiyush (approved):
LGTM now

lorenzo-dev1 (approved):
LGTM

PR #1072, opened 2024-11-18
Cap size before block gossiping

Jason-Zhangxin-Chen:
This PR port an improvement from https://github.com/clearmatics/autonity-internal/pull/272 to this public repo, it aims at limiting the fault scope. Cap the block size to align with the limits of dev-p2p protocol.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -779,8 +779,8 @@ func (w *worker) commitTransactions(env *environment, txs *types.TransactionsByP
 			break
 		}
 
-		// cap the maximum txs size of a block.
-		// The current ENV is not always exist in legacy tests, thus to check it nil to prevent panic.
+		// Cap the maximum size of packed TXNs in a block before we run into the unexpected errors due to
+		// the limits of msg size in dev-p2p layer. As current ENV is not always exist in legacy tests, check nil to prevent panic.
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -771,6 +778,14 @@ func (w *worker) commitTransactions(env *environment, txs *types.TransactionsByP
 		if tx == nil {
 			break
 		}
+
+		// Cap the maximum size of packed TXNs in a block before we run into the unexpected errors due to
+		// the limits of msg size in dev-p2p layer. As current ENV is not always exist in legacy tests, check nil to prevent panic.
+		if w.current != nil && w.current.txsSize+tx.Size() > capBlockSize {
```
lorenzo-dev1, line 784: I think we should check the `env` parameter, as that is what is passed to `commitTransaction` for adding the txs. Not sure if it always coincide with the `w.current`. @vpiyush

yazzaoui:
To be honest I'm not so sure about this being needed.. we can always cap the block gas limit and that would achieve indirectly the same goal..

lorenzo-dev1:
I agree, I had not thought about controlling it this way.

Jason-Zhangxin-Chen:
Close it now as the team decide to control in in gas limit.

PR #1073, opened 2024-11-18
Oracle accountability finalize and autobond updates

ScottRMalley:
This PR includes:

- incorporation of Autobond into the Oracle rewards
- Refactored finalize() in Oracle.sol to separate voters' info update and rewards distribution from the function and call finalize() at the beginning of finalize() in Autonity.sol like accountability contracts.
- Bug fix in _findOutliers function when calculating ratio.


autonity/solidity/contracts/Autonity.sol
```diff
@@ -1408,6 +1414,9 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
             config.contracts.omissionAccountabilityContract.distributeProposerRewards{value: _atnProposerRewards}(accounts[_omission]);
         }
 
+        _transfer(address(this), address(config.contracts.oracleContract), _ntnOracleRewards);
+        config.contracts.oracleContract.distributeRewards{value: _atnOracleRewards}(_ntnOracleRewards);
```
tbssajal, line 1418: Maybe use `accounts[address(config.contracts.oracleContract)]` instead of `_ntnOracleRewards`?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -50,14 +52,21 @@ contract Oracle is IOracle {
...
     address[] private voters;
     address[] private newVoters;
+    bool private newVotersSet;
+    bool private newVotersAccessUpdated;
     uint256 private round;
     mapping(string => Price)[] internal prices;
+
     //rewards accounting
-    bool private rewardsReady;
     uint256 private aggregatedScore;
```
tbssajal, line 64: Are we using this variable anymore?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -170,84 +184,91 @@ contract Oracle is IOracle {
...
-    function receiveRewards() onlyAutonity external payable returns (bool) {
-        rewardsReady = true;
+    function _finalizeRewards() onlyAutonity internal {
+        for (uint256 i = 0; i < voters.length; i++) {
+            address _voter = voters[i];
+            if (voterInfo[_voter].performance > 0) {
+                rewardReceivers.add(_voter);
+                rewardPeriodPerformance[_voter] += voterInfo[_voter].performance;
+                rewardPeriodAggregatedScore += voterInfo[_voter].performance;
+                voterInfo[_voter].performance = 0;
```
tbssajal, line 222: I am not sure if it falls under validator mismanagement but what if multiple validator nodes have single treasury address and single voter address? In that case we will count its performance only once where I am not sure if we should count it more than once. If we should count it more than once, we can set this `voterInfo[_voter].performance = 0` in a 2nd for loop.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -551,15 +579,15 @@ contract Oracle is IOracle {
...
+        for (uint256 i = 0; i < voters.length; i++) {
             address _voter = voters[i];
-            if(!voterInfo[_voter].reportAvailable) {
+            if (!voterInfo[_voter].reportAvailable) {
                 continue;
             }
             // median here is assumed to be non-0.
-            // we don't want the following to underflow 
+            // we don't want the following to underflow
             int256 _ratio = (_median - int256(uint256(reports[_symbol][_voter].price)) * 100) / _median;
```
tbssajal, line 589:
```suggestion
            int256 _ratio = (_median - int256(uint256(reports[_symbol][_voter].price))) * 100 / _median;
```
I think the current formula is multiplying 100 with the price and then subtracting from median and then dividing by median. I think it is wrong. I guess it should be to subtract the price from the median first and then multiply with 100 and then divide by median. Also the other formula in `_penalize()` function does it this way.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -170,84 +184,91 @@ contract Oracle is IOracle {
...
+            voterInfo[_voter].treasury.call{value:_atn, gas: 2300}("");
 
             // Transfer NTN rewards
-            config.autonity.transfer(_voter, _ntn);
+            config.autonity.autobond(voterInfo[_voter].validator, _ntn, 0);
 
-            // Reset performance score
-            voterInfo[_voter].performance = 0;
+            rewardPeriodPerformance[_voter] = 0;
+            rewardReceivers.remove(_voter);
```
tbssajal, line 250: I think removing should be done in another loop. Right now, remove function will change the order of the set. To be exact, the position at `i` will have a new address. So another way to handle it, if you want to do it in a single for loop, is doing `rewardReceivers.at(0)` instead of `rewardReceivers.at(i)` until set becomes empty.

tbssajal (changes requested):

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Single treasury address shouldn't be a problem because we only use it when sending ATN rewards. Single voter address would indeed be an issue. This would go beyond just how we are zeroing performance in this for loop, it would also be an issue in how we store `rewardsReceivers` as an `EnumerableSet.AddressSet` because we assume voter addresses are unique.

Is there any chance voter addresses would not be unique? Should that be allowed in the protocol? I was operating under the assumption that the same voter address would imply the same node address.

autonity/solidity/contracts/Autonity.sol
ScottRMalley: Yes great point. Had meant to do that. Updated now.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: No, it was used, but only to write to and never to read. I've removed it now.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Yeah that is a really good point. I missed that.

I've updated it now to loop through `rewardReceivers.values()` instead of using `at(i)`. I'm still thinking whether that is the best way to do this, as it could become a problem if we have a large set of voters.

autonity/solidity/contracts/Oracle.sol
tbssajal: I don't think we have a check for voter address. So technically it is possible for different validator nodes to have same voter address. I am not sure if it is a mismanagement from the validator side or if we should handle this scenario. If we want to handle it, I think storing them in `rewardsReceivers` should not be an issue, as by the definition, it's a set. I guess only the map `rewardPeriodPerformance` should be updated properly. That's why I was suggesting to set `voterInfo[_voter].performance = 0` in a 2nd loop if we want to handle such scenario.

autonity/solidity/contracts/Oracle.sol
tbssajal: I guess using `rewardReceivers.at(0)` until the set becomes empty would avoid creating big list in memory. But current code is also fine for me. We create such big lists here and there very often.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -271,21 +291,21 @@ contract Oracle is IOracle {
...
-                = _findOutliers(_priceMedian, _symbol);
+            = _findOutliers(_priceMedian, _symbol);
             // There is an extreme edge-case where everyone is detected outlier. This is left todo.
             // punish outliers if found
-            for(uint256 i = 0; i < _totalOutliers; i++) {
+            for (uint256 i = 0; i < _totalOutliers; i++) {
                 _penalize(_outliers[i], _priceMedian, reports[_symbol][_outliers[i]]);
                 emit Penalized(_outliers[i], _symbol, _priceMedian, reports[_symbol][_outliers[i]].price);
             }
             _success = true;
```
Jason-Zhangxin-Chen, line 305: Shoun't we set the `_price` after the aggregation is success?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -271,21 +291,21 @@ contract Oracle is IOracle {
         uint256 _price = 0;
         bool _success = false;
 
-        if (_count > 0){
+        if (_count > 0) {
```
Jason-Zhangxin-Chen, line 294: Shoun't we check the num of reports are at least 2/3 of the num of participants of consensus committee?

autonity/calls.go
```diff
@@ -351,12 +351,19 @@ func DeployAutonityContract(genesisConfig *params.AutonityContractGenesis, genes
 
 func DeployOracleContract(genesisConfig *params.ChainConfig, evmContracts *GenesisEVMContracts) error {
 	voters := make([]common.Address, len(genesisConfig.AutonityContractConfig.Validators))
+	treasuries := make([]common.Address, len(genesisConfig.AutonityContractConfig.Validators))
+	nodeAddresses := make([]common.Address, len(genesisConfig.AutonityContractConfig.Validators))
+
 	for _, val := range genesisConfig.AutonityContractConfig.Validators {
 		voters = append(voters, val.OracleAddress)
+		treasuries = append(treasuries, val.Treasury)
+		nodeAddresses = append(nodeAddresses, *val.NodeAddress)
```
lorenzo-dev1, line 360: There is a bug here, the slices are allocated with a length!=0 and then `append` is used, this will lead to having a bunch of empty spots at the start of the slice

autonity/solidity/contracts/Autonity.sol
```diff
@@ -561,7 +562,11 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
         uint256 _lookbackWindow = config.contracts.omissionAccountabilityContract.getLookbackWindow();
         uint256 _delta = config.contracts.omissionAccountabilityContract.getDelta();
         require(_period > 0, "epoch period cannot be 0");
-        require(_period > _delta + _lookbackWindow -1, "epoch period needs to be greater than delta+lookbackWindow-1");
+        require(_period > _delta + _lookbackWindow - 1, "epoch period needs to be greater than delta+lookbackWindow-1");
+
+        // we need this check to update new voters at the end of voting round
+        uint256 _votePeriod = config.contracts.oracleContract.getVotePeriod();
+        require(_votePeriod * 2 <= _period, "epoch period is too small");
```
lorenzo-dev1, line 569: Consider adding this check also for the genesis parameters (for example in the `Prepare` function)

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1408,6 +1414,11 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
             config.contracts.omissionAccountabilityContract.distributeProposerRewards{value: _atnProposerRewards}(accounts[_omission]);
         }
 
+        _transfer(address(this), address(config.contracts.oracleContract), _ntnOracleRewards);
+        config.contracts.oracleContract.distributeRewards{value: _atnOracleRewards}(
+            accounts[address(config.contracts.oracleContract)]
+        );
+
```
lorenzo-dev1, line 1421: Just to understand, is there an economical reason for which the transfer of funds happens here and not anymore before? Or is it just for code cleanliness

tbssajal (approved):
LGTM

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: nevermind, I saw it is done at the oracle contract deployment

autonity/solidity/contracts/Oracle.sol
```diff
@@ -170,84 +182,92 @@ contract Oracle is IOracle {
...
         return true;
     }
 
-   /**
-    * @dev this function is used to receive ATN rewards and to signal availability of NTN rewards.
-    * This is decoupled from {finalize} to keep the reward flow clean on the autonity contract side.
-   */
-    function receiveRewards() onlyAutonity external payable returns (bool) {
-        rewardsReady = true;
+    function _finalizeRewards() onlyAutonity internal {
```
lorenzo-dev1, line 213: I think it shouldn't be `onlyAutonity` since the function is internal and called by the oracle cotnract

autonity/solidity/contracts/Oracle.sol
```diff
@@ -50,14 +52,19 @@ contract Oracle is IOracle {
...
     mapping(string => Price)[] internal prices;
-    //rewards accounting
-    bool private rewardsReady;
-    uint256 private aggregatedScore;
 
+    // voting periods are not necessarily in sync with epoch rewards, so we need to keep track of
+    // rewards for each over the epoch, and distribute for all voting periods that concluded this epoch
+    EnumerableSet.AddressSet private rewardReceivers;
+    mapping(address => uint256) private rewardPeriodPerformance;
+    uint256 private rewardPeriodAggregatedScore;
```
lorenzo-dev1, line 68: Am I right to assume that there could be voting period that is half in epoch x and half in epoch x+1, whose rewards will be distributed at the end of epoch x+1?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -170,84 +182,92 @@ contract Oracle is IOracle {
...
-    function receiveRewards() onlyAutonity external payable returns (bool) {
-        rewardsReady = true;
+    function _finalizeRewards() onlyAutonity internal {
+        for (uint256 i = 0; i < voters.length; i++) {
+            address _voter = voters[i];
+            if (voterInfo[_voter].performance > 0) {
+                rewardReceivers.add(_voter);
+                rewardPeriodPerformance[_voter] += voterInfo[_voter].performance;
+                rewardPeriodAggregatedScore += voterInfo[_voter].performance;
+                voterInfo[_voter].performance = 0;
```
lorenzo-dev1, line 220: I understand how this works but why do we need the `rewardPeriodPerformance` mapping? Why not keeping to accumulate into the `voterInfo.performance` and reset it only at epoch end?

lorenzo-dev1: Is it to facilitate tracking per-round performance for the oracle nodes?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -409,6 +430,7 @@ contract Oracle is IOracle {
     * @dev IOracle interface method implementation..
     */
     function setVotePeriod(uint _votePeriod) external onlyOperator {
+        _checkVotePeriod(_votePeriod);
         config.votePeriod = _votePeriod;
```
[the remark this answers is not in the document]
ScottRMalley: No, I don't think so. Basically, we are persisting the rewards of every ended round and only distributing them at the end of the epoch. The only thing that would change when updating the voting period would be that fewer/more rounds would occur before the next reward distribution, but the accumulated rewards from previous rounds should not change.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Yes that is correct. Rewards are distributed based on the rounds that ended during the epoch that is finalized, so when the round was started is not considered in reward distribution.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Yeah the idea was to have a variable that represented this round's performance, and another one to represent the aggregated epoch performance so that they could be tracked. That being said, on a second look, the voter performance will be incremented and set to zero in the same finalize transaction, so there's never any time to read round performance.

I think the best thing to do is to get rid of the mapping. Will update.

ScottRMalley: Oh, wait, no I remember now, the voter info is deleted when a new voter set is added, so we need the mapping in order to keep track of rewards between epochs.

autonity/solidity/contracts/Oracle.sol
ScottRMalley: Good point. Fixed.

lorenzo-dev1 (approved):
LGTM

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
lorenzo-dev1: But what about the `NewRound` event? In case of period change it will not be accurate anymore. And I think this is tracked by the oracle, right @Jason-Zhangxin-Chen

autonity/solidity/contracts/Oracle.sol
lorenzo-dev1: thanks

PR #1083, opened 2024-11-21
Duplicate oracle address check

### LINKED ISSUE #1081: Same Oracle Address for Different Validator Nodes
Right now, it is possible to register multiple validator nodes using a single Oracle address. It could be a feature except that this leads to a reduced number of data points for price reporting in a single voting round. Also, in our current implementation, a single Oracle address for multiple validator nodes messes up accounting performance scores and reward distribution. So, this should be considered a bug instead of a feature, and we should enforce different Oracle addresses for different validator nodes in the validator registration process.

tbssajal:
Closes #1081


params/protocol_contracts.go
```diff
@@ -495,7 +495,7 @@ func (s *Schedule) Validate() error {
 	if s.Amount == nil {
 		return errors.New("amount must be specified")
 	}
-	if s.VaultAddress == common.HexToAddress("0") {
+	if deep.Equal(s.VaultAddress, common.Address{}) {
```
lorenzo-dev1, line 498: I think using `bytes.Equal` would be better for performance because it does not use reflection. But not a big deal really if `deep.Equal` also gives the correct result.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -182,6 +182,7 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
     /* Newton ERC-20. */
     mapping(address => uint256) internal accounts;
     mapping(address => Validator) internal validators;
+    mapping(address => uint256) internal oraclesExist; // 0 for false, true otherwise
```
lorenzo-dev1, line 185: Might as well use a boolean if we allow only 0 and 1

lorenzo-dev1 (approved):
LGTM with two minor comments

autonity/solidity/contracts/Autonity.sol
tbssajal: Yeah, boolean would be more readable. Used uint256 to save gas instead. Not sure how much I am saving though.

params/protocol_contracts.go
tbssajal: Just copied it from another part of the code.

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: ah ok didn't know it saved gas. That's ok then

Jason-Zhangxin-Chen (approved):
LGTM.

PR #1085, opened 2024-11-22
More Tests for Oracle

tbssajal:
- tests for edges cases while updating voters
- tests to trigger a bug when old voters participate partially in an epoch and earn rewards


ScottRMalley (approved):

yazzaoui (approved):
LGTM

Issue #1082, opened 2024-11-21
Chain Halt Issue due to P2P Jailing Mechanism


tbssajal:
Our p2p jailing mechanism jails a validator immediately when we have an `NewFaultProof` event. This mechanism does not follow consensus protocol or our own protocols and it is a bit hard to get immediate consensus on p2p jailing. So it raises some issues that could potentially leat to chain halt described below
- **ISSUE_1:** Consensus algorithm does not require us to jail any validator and Autonity protocol contracts jail some validator at epoch end and immediately change the committee set by excluding the jailed validators. But the immediate jailing of p2p jailing mechanism does not follow these protocols, so theoritically it is possible that p2p jailing mechanism jails more than 1/3rd validators of the committee without breaking any rules of the tendermint consensus algorithm. For example, at any given height, there are less than 1/3rd validators behaving maliciously, but when multiple height are taken into account, more than 1/3rd validators are found behaving maliciously. If we did not have p2p jailing mechanism, this scenario will not cause chain halt, but at the end of the epoch, the Autonity protocol contracts can jail the misbehaving validators and kick them out of the committee.
- **ISSUE_1:** This one is brought up by @lorenzo-dev1, I am writing here just to address the solutions in one place. Because it is a bit hard to get consensus on p2p jailing, it is possible that some validator node (`nodeA`) have some aggregated prevotes or precommits which includes a jailed validator from some other node's (`nodeB`) point of view because `nodeA` was a bit late to jail the validator in question. This will lead that `nodeB` to reject the aggregated message but `nodeA` will keep it even after jailing the validator in question, because the aggregated message was prepared before the validator was jailed in `nodeA`. This can lead to not reach quorum.
Lets discuss the possible solutions in this thread.

tbssajal:
@lorenzo-dev1 please feel free to update the **ISSUE_2** or describle it again in your language.

tbssajal:
For **ISSUE_1**, the solution I propose is to get consensus on p2p jailing, which is possible because the jailing is based on events emitted form cotracts. The idea would be to track the voting power of the validators that are jailed in p2p jailing mechanism in the client and not jail anymore if the voting power of the remaining validators does not reach quorum. There should be a particular order on how to jail. It could be based on the events or the voting power of the misbehaving validators.

tbssajal:
[...]

PR #1086, opened 2024-11-22
jailing rotation draft

lorenzo-dev1:


consensus/tendermint/backend/engine.go
```diff
@@ -662,36 +733,78 @@ func (sb *Backend) faultyValidatorsWatcher(ctx context.Context) {
...
+								belowF = true // still unjail all other validators for this height to ensure fairness, but then stop unjailing
+							}
+						}
+						delete(atHeight, unjailingHeight)
+						if belowF {
+							break
+						}
+					}
+					// i always <= len(heights-1), since we will unjail everyone at the last height and go below F for sure
+					heights = heights[i+1:]
```
tbssajal, line 803: Consider adding a function for this whole process. That would help repopulate things from database at initialization.

consensus/tendermint/backend/engine.go
```diff
@@ -662,36 +733,78 @@ func (sb *Backend) faultyValidatorsWatcher(ctx context.Context) {
...
+				}
 				sb.jailed.Unlock()
 
+				// TODO: store also the additional data struct (heights, atHeight and power) in the db
```
tbssajal, line 807: I don't think we need this info in db. What we need is to store a struct{validator, height} for some epochID. It is like storing the necessary information of the event. Then when we jail validator by processing these information at initialization using a single function as mentioned in above comment, it will reproduce these data structs. It would be a lot less work I guess.

tbssajal: Also, the db only contains validators for which the voting power is less than F. So we don't need to modify any logic or think about ang edge cases or think about the order when we initialize from db by processing {validator, height}.

tbssajal:
Closses #1082.

PR #1087, opened 2024-11-24
fix, oracle accountability slashing is reverted.

Jason-Zhangxin-Chen:
This PR contain a few corrections for oracle accountability slashing logics.

- [x] grant the autonity.slash() access to oracle contract
- [x] process the over 100% slashing rate for oracle accountability.

**TODO:** The slashing rate computing can result a >= SLASHING_RATE_PRECISION (10,000) rate, which means a >= 100% slashing, we need to evaluate the correctness of the computing formula.

@lorenzo-dev1 please help to double check the changes in the Slasher.sol contract. I just added an processing of an edge case for it.

### DIFF
--- a/autonity/solidity/contracts/ProtocolConstants.sol
+++ b/autonity/solidity/contracts/ProtocolConstants.sol
@@ -2,4 +2,5 @@
 pragma solidity ^0.8.19;
 
 uint256 constant SLASHING_RATE_PRECISION = 10_000; // 100% slashing rate.
-uint256 constant ORACLE_SLASHING_RATE_CAP = 1_000; // Capped with 10% slashing rate.
+// Todo: resolve a better cap for the protocol.
+uint256 constant ORACLE_SLASHING_RATE_CAP = 1_000; // Capped with 10% slashing rate for oracle accountability.
--- a/autonity/solidity/contracts/ProtocolConstants.sol
+++ b/autonity/solidity/contracts/ProtocolConstants.sol
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 pragma solidity ^0.8.19;
 
-uint256 constant SLASHING_RATE_PRECISION = 10_000;
+uint256 constant SLASHING_RATE_PRECISION = 10_000; // 100% slashing rate.
+uint256 constant ORACLE_SLASHING_RATE_CAP = 1_000; // Capped with 10% slashing rate.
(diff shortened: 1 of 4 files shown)


autonity/solidity/contracts/Slasher.sol
```diff
@@ -77,6 +77,23 @@ contract Slasher {
...
+            }
+            _val.bondedStake = 0;
+            _val.selfBondedStake = 0;
+            _val.selfUnbondingStake = 0;
+            _val.unbondingStake = 0;
+            _val.totalSlashed += _availableFunds;
+            //_jailbound(_val, _newJailboundState); // shall we do this jailBound?
+            return (_val, _availableFunds);
+        }
+
```
lorenzo-dev1, line 96: I don't think this is the right solution, because when slashing 100% we should jailbound, otherwise it can lead to other problems down the line. 

So either:
1. we cap the slashing in the oracle contract
2. we create a new function like `slashUnbounded` or something like that, which allows 100% slash and jailbounds the validator if 100% slash happens

lorenzo-dev1 (changes requested):

autonity/solidity/contracts/Slasher.sol
Jason-Zhangxin-Chen: > I don't think this is the right solution, because when slashing 100% we should jailbound, otherwise it can lead to other problems down the line.
> 
> So either:
> 
> 1. we cap the slashing in the oracle contract
> 2. we create a new function like `slashUnbounded` or something like that, which allows 100% slash and jailbounds the validator if 100% slash happens

Thank you, Lorenzo. I think capping the slashing rate would be the best option.

ScottRMalley (approved):
This LGTM with the latest slashing rate cap changes.

autonity/solidity/contracts/ProtocolConstants.sol
```diff
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 pragma solidity ^0.8.19;
 
-uint256 constant SLASHING_RATE_PRECISION = 10_000;
+uint256 constant SLASHING_RATE_PRECISION = 10_000; // 100% slashing rate.
+// Todo: resolve a better cap for the protocol.
+uint256 constant ORACLE_SLASHING_RATE_CAP = 1_000; // Capped with 10% slashing rate for oracle accountability.
```
tbssajal, line 6: Can we write `ORACLE_SLASHING_RATE_CAP = SLASHING_RATE_PRECISION / 10` instead or is it not allowed due to being constant?

tbssajal (approved):
LGTM except some minor comments.

lorenzo-dev1 (approved):
LGTM as well

yazzaoui (approved):
LGTM but we'll need later for the max slashing rate to be a governance parameter

PR #1091, opened 2024-11-29
Discard Message Too Far in The Future

### LINKED ISSUE #1090: Exploit on Future Message Handling Mechanism
Discussed here: https://github.com/autonity/autonity/pull/1069#discussion_r1860247887, the message handling mechanism for height greater than the current height at node is exploitable by any validator which can cause some potential issues. The most serious concern is that if any other validator node can make some honest node crash by using some exploit then the honest is actually behaving like a malicious node and the amount of voting power in the malicious validator group is increasing. Possible solution is discussed in the thread.

tbssajal:
Closes #1090.
Implemented the soln described here: https://github.com/autonity/autonity/issues/1090#issuecomment-2507189924.

### DIFF
--- a/consensus/tendermint/backend/handler.go
+++ b/consensus/tendermint/backend/handler.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"time"
 
@@ -36,6 +37,9 @@ var (
 	// errJailed is returned when a consensus message is discarded because the signer is jailed
 	ErrJailed    = errors.New("signer is jailed")
 	ErrNotFuture = errors.New("message is not future height anymore")
+	// ErrTooFarInFuture returned when msg is so far in the future that it is not relevant for this node
+	ErrTooFarInFuture = errors.New("message is too far in the future")
+
 	NetworkCodes = map[uint8]uint64{
 		message.ProposalCode:  ProposeNetworkMsg,
 		message.PrevoteCode:   PrevoteNetworkMsg,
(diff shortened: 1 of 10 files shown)


consensus/tendermint/backend/handler.go
```diff
@@ -295,6 +302,38 @@ func (sb *Backend) saveFutureMsg(msg message.Msg, errCh chan<- error, sender com
 	return nil
 }
 
+func (sb *Backend) isFutureMsgTooFar(myHeight, msgHeight uint64) bool {
```
vpiyush, line 305: Why shouldn't we just reject message if it is further than 2 epoch periods. Is it really necessary to be accurate here?  an estimation wouldn't be enough.

vpiyush: IMO buffer size is not going to have any performance penalty as long as it's capped to certain size, whereas this logic would have some. So i am still of the opinion that this is not really needed a simpler check should be enough. Although We can check with others in the team and take a decision

tbssajal: > IMO buffer size is not going to have any performance penalty as long as it's capped to certain size, whereas this logic would have some.
Because we are fetching state info? We will still have to fetch state info if we want to check with only epoch period. But in that case we will do one call and here we are doing two calls. In both case we will do these calls at most once per epoch.

vpiyush: We would use the cached epoch period value and update the cache on epoch period change. No state queries

tbssajal: > update the cache on epoch period change

By listening to events? Yeah that is better I guess.

params/node_params.go
```diff
@@ -0,0 +1,9 @@
+package params
+
+import "time"
+
+const (
+	// approximate time the node might need to start all services after shutting down
+	// in case of a crash or restart
+	BootingTime = time.Minute * 2
```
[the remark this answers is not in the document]
tbssajal: So my thinking was that in case of a restart, we would fall behind some blocks, let's say $x$ blocks. So we should accept any messages that can come from those future $x$ blocks. And to make things simpler, we do this all the time, even if the node did not restart. That's why I wanted to calculate this threshold based on the `BootingTime` instead. It makes more sense to me if we want to change the value, instead of randomly deciding the thresold on height, we can make assumption on the time needed for the node to restart.

> Is there any empirical backing to why we are using the choice of two minutes here?

Not really.

PR #1096, opened 2024-12-04
Allow partial oracle reports

ScottRMalley:
Currently we require all oracle votes to include a price for every symbol. This means that when one price source is unavailable to an oracle, it will ultimately influence the price accuracy of every other symbol, as they can no longer submit prices for any symbol.

This is a draft PR for one way we could solve this issue, by having the oracle interface allow specifying a subset of symbols when reporting on prices.


autonity/solidity/contracts/Oracle.sol
```diff
@@ -183,7 +184,8 @@ contract Oracle is IOracle {
                 (_reports[i].price == 0 && _reports[i].confidence == 0),
```
vpiyush, line 184: FYI, We have removed the second part of this if condition in the patch, meaning we don't allow 0 price and/or confidence anymore. This would affect your approach.

ScottRMalley:
After further consideration, we have decided that this is not a priority change. For the moment, partial voting, and re-submission of stale prices should be handled by the Autonity Oracle.

PR #1097, opened 2024-12-04
Update supply control to burn genesis ATN supply

ScottRMalley:
Under normal operation, ATN repaid on a CDP is returned to the SupplyControl contract, where it can be minted on a subsequent `borrow` call. This change updates the SupplyControl contract to send an initial amount `burnableGenesisSupply` to the zero address, rather than keeping it in circulation.

This ensures that over time, all ATN on the network will correspond exactly to debt on a CDP position.

In order to achieve this:
* `SupplyControl` constructor is initialized with a new variable `burnableGenesisSupply`
* `burn` on `SupplyControl` is updated to send to the zero address while `burnableGenesisSupply > 0`
* contract deployment is updated to deploy `SupplyControl` with a `burnableGenesisSupply` value derived from the GenesisAlloc


autonity/solidity/contracts/asm/SupplyControl.sol
```diff
@@ -90,6 +95,16 @@ contract SupplyControl is ISupplyControl {
...
+            uint256 _toBurn = msg.value;
+            if (msg.value > burnableGenesisSupply) {
+                _toBurn = burnableGenesisSupply;
+                burnableGenesisSupply = 0;
+            } else {
+                burnableGenesisSupply -= msg.value;
+            }
+            payable(address(0)).transfer(_toBurn);
+        }
         emit Burn(msg.value);
```
Jason-Zhangxin-Chen, line 108: This legacy code looks wired, it does not burn anything at all by emitting a message telling value is burnt. Shall we need to move this `emit Burn(msg.value)` into the added logic block where the burning happens?

Jason-Zhangxin-Chen (commented):
LGTM, just leave one comment.

tbssajal (approved):
LGTM!

autonity/solidity/contracts/asm/SupplyControl.sol
vpiyush, line 108: Shouldn't we emit _toBurn in Burn event instead of msg.value?

PR #1098, opened 2024-12-05
Stabilization Bug Fix

tbssajal:
A harmless bug that works without fixing, but from user point of view, does not give proper revet reason. Fixed to give proper revert reason.


lorenzo-dev1 (approved):
LGTM

vpiyush (approved):
LGTM!

PR #1107, opened 2024-12-09
Oracle Outlier Incentive Removal

tbssajal:
There are scenarios where an outlier can get slashed but also get incentives. This PR fixes the bug.


autonity/solidity/contracts/Oracle.sol
```diff
@@ -636,6 +636,7 @@ contract Oracle is IOracle {
         // Stop considering this reporter for any future calculation.
         // This is symbol independant.
         voterInfo[_outlier].reportAvailable = false;
+        voterInfo[_outlier].performance = 0;
```
ScottRMalley, line 639: This doesn't make much sense to me. We are already withholding their performance increment for this report on line 614.

This change would mean that when a report is an outlier for a single symbol, we remove their rewards for all symbols, which doesn't seem fair.

tbssajal: I kinda agree but I think it also relates to both fairness and consistency issue #1110. Right now it is not consistent when the outlier is rewarded but also getting slashed.

ScottRMalley: I agree that the current implementation is dependent on symbol order, and this would at least set rewards to zero regardless of the symbol order, however, I do think that the eventual correct implementation will be to only withhold rewards for symbols where the oracle was an outlier, rather than withhold rewards for all reports in the round where an oracle posted at least one outlier report.

I don't think we are moving forward with merging this PR at the moment, so I'm not requesting changes at the necessarily, but I do think it's interesting to keep these options in the discussion.

tbssajal: > I do think that the eventual correct implementation will be to only withhold rewards for symbols where the oracle was an outlier, rather than withhold rewards for all reports in the round where an oracle posted at least one outlier report.

Yeah I kinda agree.

tbssajal: Removed this line. Now outliers can get rewards.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -668,7 +729,65 @@ contract Oracle is IOracle {
...
+        }
+    }
+
+    function _removePenalizedVoters() internal {
+        uint256 _length = penalizedVoters.length();
+        while (_length > 0) {
+            address _voter = penalizedVoters.at(0);
+            // don't consider past reports from the penalized voters
+            voterInfo[_voter].reportAvailable = false;
+            require(penalizedVoters.remove(_voter), "voter not removed");
```
Jason-Zhangxin-Chen, line 779: Just curious, if this remove() returns false, the finalize of the block will be failed, leading a chain on-holding. When will the remove() be failed?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -668,7 +729,65 @@ contract Oracle is IOracle {
...
+            _length--;
+        }
+
+        _length = outliers.length();
+        while (_length > 0) {
+            address _voter = outliers.at(0);
+            // don't consider past reports from the outliers
+            voterInfo[_voter].reportAvailable = false;
+            require(outliers.remove(_voter), "voter not removed");
```
Jason-Zhangxin-Chen, line 788: Same question here.

Jason-Zhangxin-Chen (commented):
The overall looks good to me, just left two questions on the failure of the remove() function.

PR #1108, opened 2024-12-10
fix api to fetch acn peer infos

vpiyush:
The wrong type was returned which didn't have any json tags either. so result was not displayed properly. Fixed the return type add test to make sure Json marshalling/unmarshalling is successful.

### DIFF
--- a/e2e_test/api_test.go
+++ b/e2e_test/api_test.go
@@ -34,7 +34,7 @@ func TestAPI_AcnPeers(t *testing.T) {
 	require.NoError(t, err)
 
 	// test json un-marshaling
-	var result  []*p2p.PeerInfo
+	var result []*p2p.PeerInfo
 	err = json.Unmarshal(jsonPayload, &result)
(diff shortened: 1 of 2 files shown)


ScottRMalley (approved):
LGTM

params/generated/Oracle.go
```diff
@@ -1,11 +1,8 @@
 package generated
 
-import (
-	"strings"
-
-	"github.com/autonity/autonity/accounts/abi"
-	"github.com/autonity/autonity/common"
-)
+import "strings"
```
yazzaoui, line 3: was that because of your local linting configuration?

yazzaoui (approved):
LGTM but please remove the oracle.go update from the commit before merging

params/generated/Oracle.go
vpiyush: not because of that, but i couldn't figure why exactly that happened

PR #1109, opened 2024-12-10
Refactor core event

vpiyush:
Changes the interface to send events from core to aggregator to a buffered chanel. the space in buffered chanel is shared by all three event types.


vpiyush:
> LGTM, my only comment is whether you think it could be useful to add a metric measuring how long it takes for the event to reach Core.
> 
> Just to make sure that 100 of buffer size is enough, and we do not end up blocking Core due to the aggregator not reading fast enough.
 
Since we already have a lot of metrics enabled by default, I'd rather not add more unless we start experiencing performance issues. However, we could consider increasing the default buffer size as an extra precaution. thoughts?

lorenzo-dev1:
> Since we already have a lot of metrics enabled by default, I'd rather not add more unless we start experiencing performance issues.

ok makes sense

>  However, we could consider increasing the default buffer size as an extra precaution. thoughts?

Intuitively I think the channel shouldn't really fill up much... I would say it is fine to leave it at 100 and see how it goes.

Jason-Zhangxin-Chen (approved):
LGTM.

Issue #1111, opened 2024-12-11
Prometheus metrics have incorrect type


Szymongib:
### Description

With new version 1.0.2 Prometheus is not able to scrape metrics, it fails with error:
```
invalid metric type "header gauge"
```

Metrics are enabled and queryable directly from the endpoint.

### System Information

Autonity Version: `1.0.2`

### How to Reproduce

Start the node with Prometheus metrics enabled, and setup Prometheus instance to scrape it.

### Expected Behaviour

Prometheus is able to scrape and collect Autonity metrics.

### Actual Behaviour

Prometheus scrape fails.

Szymongib:
Proposed a fix in #1112. This fixes issues on our side, let me know if it is any good from your side :)

PR #1112, opened 2024-12-11
Fix Prometheus metrics format

Szymongib:
This solves https://github.com/autonity/autonity/issues/1111

This fixes incorrect type of this metric
```
# TYPE chain_head_epoch-header gauge
```
to
```
# TYPE chain_head_epoch_header gauge
```

I deployed this fix to one of our nodes and our Prometheus instance can now scrape metrics from this node.

As per https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels regex for allowed names does not contain `-` - `[a-zA-Z_:][a-zA-Z0-9_:]*`


vpiyush (approved):
LGTM!

lorenzo-dev1 (approved):
LGTM

PR #1116, opened 2024-12-13
feature, trace the protocol contract gas at the block finalization.

Jason-Zhangxin-Chen:
This PR contains a feature which trace the gas consumption of different protocol contracts at the block finalization phase.


core/vm/gas_tracer.go
```diff
@@ -0,0 +1,59 @@
...
+	acuUpdateGas                      = metrics.NewRegisteredBufferedGauge("autonity/acu/update", nil, nil)
+)
+
+type Tracer map[string]struct{}
+
+// TracerRegistry save the tracked functions' signature in 4 bytes as the method ID.
+var TracerRegistry = map[common.Address]Tracer{
+	// params.AutonityContractAddress:               {"4bb278f3": struct{}{}}, // finalize(), duplicated with legacy metrics.
+	params.AccountabilityContractAddress:         {"6c9789b0": struct{}{}}, // finalize(bool)
+	params.OmissionAccountabilityContractAddress: {"6c9789b0": struct{}{}}, // finalize(bool)
```
vpiyush, line 24: omission and accountability address is same here?

Jason-Zhangxin-Chen: The two contracts have the same signature of the finalize(bool) function, thus the 4bytes method ID is the same.

PR #1117, opened 2024-12-13
Oracle Bug Fix and Penalized Event Improvement

### LINKED ISSUE #1115: Changing Oracle Symbols Trigger Zero Median Bug
In oracle, we require voters to submit positive price and rely on the fact that the median of the prices will be also positive. This fact is true when we have reported prices from voters. However there is a bug which can be triggered by changing the oracle symbols which can lead to scenario where the median of the prices can be zero.

### Steps To Reproduce
- Submit valid vote in two consecutive rounds, then vote no more
- Change the symbol so we have new symbol in oracle
- Progress three more rounds
- Third round will trigger the bug which will lead to `median = 0` and will cause revert the finalize call

### Possible Soln
A simple solution that comes to mind is when symbol change is in effect, make every voters' report from `available = true` to `available = false`.

tbssajal:
- Fixed symbol update bug described in #1115
- Push price once per voting round
- Update `Penalized` event to include the slashing amount

Closes #1115

**Consensus breaking with piccadilly testnet**


autonity/solidity/contracts/Oracle.sol
```diff
@@ -302,21 +305,22 @@ contract Oracle is IOracle {
...
             // use past value for price if unsuccesful
-            _price = prices[round - 1][_symbol].price;
+            prices[round][_symbol] = Price(
+                prices[round - 1][_symbol].price,
+                block.timestamp,
+                false
+            );
         }
-
-        prices.push();
```
yazzaoui, line 315: This was required to give room for the `round`  element to be saved in the `prices` array otherwise the underlying length value would not get updated. Are you sure that is working or I'm missing something?

tbssajal: I removed it here but added it before entering aggregate function in finalize. This way you only push once per voting round. But currently you push once for each symbol per voting round.

yazzaoui: Ah ok I see now - good spot 👍

Jason-Zhangxin-Chen:
@tbssajal , this interface update of the slashing event would need a upgrade on the oracle protocol contract, so, just wondering how will we do the upgrade in current Piccadilly? A chain fork might be requried.

tbssajal:
I was suggesting to do the upgrade via transaction. But for that we will have to include the fix that @yazzaoui did for the scenario where everyone is outlier.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -291,32 +296,41 @@ contract Oracle is IOracle {
...
+                    block.timestamp,
+                    true
+                );
+            } else {
+                // use past value for price if unsuccesful
+                prices[round][_symbol] = Price(
+                    prices[round - 1][_symbol].price,
+                    block.timestamp,
+                    false
+                );
```
lorenzo-dev1, line 324: Maybe add a comment specifying that this branch handles the case where all reports are detected as outliers.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -272,9 +267,19 @@ contract Oracle is IOracle {
...
+
+        // symbol update should happen in the symbolUpdatedRound+2 since we expect
+        // oracles to send commit for newSymbols in symbolUpdatedRound+1 and reports
+        // for the new symbols in symbolUpdatedRound+2
+        if (int256(round) == symbolUpdatedRound + 2) {
+            symbols = newSymbols;
+            for (uint i = 0; i < voters.length; i++) {
+                voterInfo[voters[i]].reportAvailable = false;
+            }
+        }
```
lorenzo-dev1, line 279: I am thinking about the scenario when `SymbolUpdatedRound + 2` is the same round where `newVotersSet` == true.

If I understand correctly, we do not need to set `reportAvailable` to false for the `newVoters`, because even if by coincidence the `SymbolUpdatedRound + 2` is the same round where `newVotersSet` == true, the `newVoters` will have by default `reportAvailable` set to false?

lorenzo-dev1: Also might be worth to leave a comment explaining why it is needed to set `reportAvailable` to false when there are new symbols.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -291,32 +296,41 @@ contract Oracle is IOracle {
...
-            _price = _calculateWeightedPrice(_filteredReports, _reportsCount);
-            _success = true;
         } else {
             // use past value for price if unsuccesful
-            _price = prices[round - 1][_symbol].price;
+            prices[round][_symbol] = Price(
+                prices[round - 1][_symbol].price,
+                block.timestamp,
+                false
+            );
```
lorenzo-dev1, line 332: So let's say that there is a symbol change and no ones votes after, so everyone has `reportAvailable` == false. 

We will try to use the past round price for the new symbols, however no previous price exists. Will the contract return a 0 price?

lorenzo-dev1 (approved):
I have a couple questions, but the code LGTM

autonity/solidity/contracts/Oracle.sol
tbssajal: Yes

autonity/solidity/contracts/Oracle.sol
tbssajal: > If I understand correctly, we do not need to set reportAvailable to false for the newVoters, because even if by coincidence the SymbolUpdatedRound + 2 is the same round where newVotersSet == true, the newVoters will have by default reportAvailable set to false?

Yes you are right. But do you think we need to include this logic in the code. I did this way to just keep it simple.

autonity/solidity/contracts/interfaces/IOracle.sol
```diff
@@ -144,6 +134,7 @@ interface IOracle {
      * @param _symbol Outlier symbol.
      * @param _median Median price calculate for this symbol.
      * @param _reported Reported outlier price.
+     * @param _slashingAmount Slashing amount of the validator stakes.
```
Jason-Zhangxin-Chen, line 137: _slashingAmount can be zero if it does not exceed the protocol's threshold. We can add this comment to this interface file.

autonity/solidity/contracts/Oracle.sol
lorenzo-dev1: No I think it is fine like this, just wanted to confirm whether my understanding was fine.

PR #1118, opened 2024-12-17
POC: Network clustering k-means optimization

ScottRMalley:
This draft PR is intended to be a POC for implementing a network clustering model for block propogation.

As has been discussed and investigated, block proposals can be quickly disseminated through the network using a cluser-based graph model based on the relative latency of nodes in the committee set. In this POC, the flow works as follows:
* At each committee change (epoch), a node measures it's latency with respect to it's peers by pinging them over TCP
* The relative latency is submitted to an on-chain contract where an event is emitted on latency reporting
* When an on-chain latency reporting event is received, a node fetches the full latency matrix and performs a deterministic clustering of the nodes
* When a node proposes a block, instead of broadcasting to everyone, they select `ClusterRedundancyParameter` nodes in each cluster and sends them the proposal
* When a node receives a proposal, if it's from outside their cluster, they forward it to everyone in their cluster

For the purposes of this POC, the k-means clustering algorithm has been borrowed from [here](https://github.com/parallelo-ai/kmeans). We may need our own implementation of kmeans in a production-level implementation, particularly if the performance of the algorithm becomes a bottleneck.


ScottRMalley:
We have decided to go a different direction with network clustering. Closing this POC PR.

PR #1119, opened 2024-12-20
Migrate ASM test to golang testing framework

ScottRMalley:
This PR migrates the ASM tests from python to our golang testing framework.

It includes:
* Upstream commit to allow error decoding
* Updated abigen template to generate custom errors from abi spec
* All current tests in /autonity/solidity/test/asm migrated

The primary ToDo, outside the scope of this PR will be to include testing for emitted events, once we have implemented support in the Golang framework.


accounts/abi/bind/source.java.tpl
yazzaoui, line 1: Do we really need that?

accounts/abi/abi.go
```diff
@@ -261,6 +278,27 @@ func (abi *ABI) HasReceive() bool {
...
 
+// panicSelector is a special function selector for panic reason unpacking.
+var panicSelector = crypto.Keccak256([]byte("Panic(uint256)"))[:4]
+
+// panicReasons map is for readable panic codes
+// see this linkage for the details
+// https://docs.soliditylang.org/en/v0.8.21/control-structures.html#panic-via-assert-and-error-via-require
+// the reason string list is copied from ether.js
+// https://github.com/ethers-io/ethers.js/blob/fa3a883ff7c88611ce766f58bdd4b8ac90814470/src.ts/abi/interface.ts#L207-L218
+var panicReasons = map[uint64]string{
```
yazzaoui, line 289: Good idea👍

yazzaoui (approved):
LGTM, the new error testing mechanism is neat

accounts/abi/abi.go
ScottRMalley: Wish I could say this one was me, but I pulled it from upstream

vpiyush (approved):
Looks Good to me!

PR #1123, opened 2024-12-31
improvement, apply new vote period at the end of round.

Jason-Zhangxin-Chen:
This PR contains an improvement for the update of vote period in oracle contract, now the new vote period is applied at the end of a round to get oracle nodes synced with it.

### DIFF
--- a/autonity/solidity/contracts/Oracle.sol
+++ b/autonity/solidity/contracts/Oracle.sol
@@ -51,6 +51,7 @@ contract Oracle is IOracle {
     uint8 private constant DECIMALS = 18;
     string[] private symbols;
     string[] private newSymbols;
+    uint private newVotePeriod;
 
     address[] private voters;
     address[] private newVoters;
@@ -97,6 +98,7 @@ contract Oracle is IOracle {
         _votersSort(_voters, int(0), int(_voters.length - 1));
         voters = _voters;
         newVoters = _voters;
+        newVotePeriod = config.votePeriod;
         round = 1;
         // create the space for first index in prices array
         prices.push();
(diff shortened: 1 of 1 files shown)


tbssajal (changes requested):
The `setEpochPeriod` function in Autonity also needs to be updated to check if the new vote period passes the requirements.

Jason-Zhangxin-Chen:
> The function `setEpochPeriod` in Autonity also needs to be updated to check if the new epoch period passes the requirements with the new vote period.

Resolved in 15a2cdfb2060e2f080e1f997da4fe1dd84f4aacc

autonity/solidity/contracts/Oracle.sol
```diff
@@ -216,7 +218,11 @@ contract Oracle is IOracle, IConfigEvents {
 
         lastRoundBlock = block.number;
         round += 1;
-        emit NewRound(round,  block.timestamp, config.votePeriod);
+        // apply the new vote period at the end of the round to get the oracle network synced with it.
+        if (config.votePeriod != newVotePeriod) {
+            config.votePeriod = newVotePeriod;
+        }
```
lorenzo-dev1, line 224: There is no need for an `if` statement here, you can just do the assigment `config.votePeriod = newVotePeriod;`. If it is equal the assigment will just have no effect as it will be re-assigning the same value. However the code will be easier to read.

lorenzo-dev1 (approved):
Left a comment for an improvement but LGTM!

yazzaoui (approved):
LGTM

autonity/solidity/contracts/Oracle.sol
lorenzo-dev1: By trying on [evm.codes](https://www.evm.codes/) I see that a SSTORE for the same value present in state costs 2200 if the slot is cold and 100 if the slot is hot. So I think it wouldn't be much different wrt doing the if statement (at least with the new instruction set)

PR #1127, opened 2025-01-13
Allow Transfer Failure in case of Operation Call

tbssajal:
- Use maximum 2300 gas to send rewards to beneficiary.
- In `changeContractBeneficiary` call, allow transfer to fail, in which case the rewards go to treasury account.

The reason for above changes is that `changeContractBeneficiary` should not fail if the transfer fails.

**_Consensus breaking with piccadilly testnet_**


tbssajal:
Closing this as https://github.com/autonity/autonity/pull/1136 covers everything

PR #1130, opened 2025-01-15
improve, don't do clustering for small scale network.

Jason-Zhangxin-Chen:
By according to the simulation and testing, a small network with <=32 nodes are not clustered anymore since clustering does not benefit anymore. Thus this PR skip the clustering for small network.


consensus/tendermint/latency/routing.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/autonity/autonity/log"
 )
 
+// ScaleThresholdForClustering is the number of nodes to split the network into multiple clusters.
```
ScottRMalley, line 25: edit: 
```
// ScaleThresholdForClustering is the minimum number of validators required to do network clustering
```

consensus/tendermint/latency/routing.go
```diff
@@ -139,14 +141,18 @@ func (r *Router) Start(ctx context.Context, chain *core.BlockChain) error {
 			r.measurementWindow = epochPeriod.Uint64() / uint64(r.curEpochInfo.Committee.Len())
 			r.measured = false
 
+			// on new epoch, we have a small scale of network, clustering does not benefit anymore.
+			if r.curEpochInfo.Committee.Len() <= ScaleThresholdForClustering {
+				r.resetClusters()
+			}
```
ScottRMalley, line 147: If we are resetting clusers to nil, I think we need to have an early return in `Route` that returns the full committee when `r.clusters == nil`, or did I miss something here?

ScottRMalley (changes requested):

consensus/tendermint/latency/routing.go
Jason-Zhangxin-Chen: In the Route function, if there was no clusters formed or there is no clusters(nil) at all, isn't it going to broadcast the message to the full committee by returning the full commitee? https://github.com/autonity/autonity/blob/no-clustering-small-net/consensus/tendermint/latency/routing.go#L79-L83

ScottRMalley: Ah right, we already have that. Looks good to me then.

ScottRMalley (approved):

PR #1133, opened 2025-01-22
fix, clean shutdown msg router.

Jason-Zhangxin-Chen:
This PR contains:
1. a fix on the clean shutdown of router, it will fix those failed tests under worker_test.go and all of those timeout tests in consensus package.
2. a data race in router introduced by a correction I made before.


Jason-Zhangxin-Chen:
The correction was merged via https://github.com/autonity/autonity/pull/1132, close this one

PR #1134, opened 2025-01-22
reduce penalty for starters

tbssajal:
Reduces penalty for outliers who are off from median by minimal margin.

**Consensus breaking with piccadilly testnet**


lorenzo-dev1 (approved):
LGTM

autonity/solidity/contracts/Oracle.sol
```diff
@@ -644,9 +644,10 @@ contract Oracle is IOracle {
         }
 
         // TODO: to formal evaluate the correctness of this formula.
-        uint256 _slashingRate = uint256(_diffRatio - config.outlierSlashingThreshold) *
+        // `_diffRatio` is a percentage squared, so dividing it by 10_000
```
Jason-Zhangxin-Chen, line 647: I think now we can remove the TODO statement at line 646.

Jason-Zhangxin-Chen (approved):
LGTM with a single comment to remove legacy TODO around the correction.

PR #1139, opened 2025-02-10
Contract metrics

vpiyush:
Added new events, discussed over chain state metric meetings.


Jason-Zhangxin-Chen (approved):
LGTM, just remember to remove those legacy todos comments.

autonity/solidity/contracts/UpgradeManager.sol
```diff
@@ -20,12 +22,26 @@ contract UpgradeManager {
...
-            let result := delegatecall(gas(), precompile, add(_input,32), mload(_input), 0, 0)
-            returndatacopy(0, 0, returndatasize())
-            switch result
-            case 0 { revert(0, returndatasize()) }
-            default { return(0, returndatasize()) }
+            if iszero(success) {
+                revert(add(returnData, 32), returnSize)
+            }
+            return(add(returnData, 32), returnSize)
         }
```
lorenzo-dev1, line 45: Would be useful to have some comments for the assembly operations as they are not straightforward. My understanding is that (correct me if I am wrong):
1. you modify the free memory pointer to allocate memory for the return data. You allocate len(returnData) + 32 to prefix the byte array with its length.
2. you copy the return data in the allocated memory
3. revert or return based on the success variable

I also wonder if it is necessary to prefix the return data with its length, since we are always skipping it when we return. Is it to be able to correctly assign it to the non-assembly `returnData` variable?

lorenzo-dev1 (commented):
Overall LGTM but I have some minor questions and doubts.

autonity/solidity/contracts/Accountability.sol
```diff
@@ -143,6 +153,7 @@ contract Accountability is IAccountability, AccessAutonity {
         if(!ok) {
             autonity.getTreasuryAccount().call{value:msg.value}("");
         }
+        emit ReporterRewarded(_reporter.nodeAddress, _offender, _ntnReward);
```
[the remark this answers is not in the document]
vpiyush: agree, I will add the atn rewards as well

autonity/solidity/contracts/UpgradeManager.sol
vpiyush: your understanding is fine. the question about prefixing the return data with it's length, since i had to split these 2 assembly blocks, now there is a solidity code between these blocks which has access to returnData variable. if we were to use the returnData in the solidity code without copying the length prefix that would have been wrong. But i have still kept the behaviour same and still returning only the raw data to the caller.

lorenzo-dev1 (approved):
LGTM

PR #1144, opened 2025-02-13
Staking Pool

tbssajal:


autonity/solidity/contracts/StakingPool.sol
```diff
@@ -0,0 +1,830 @@
...
+
+    /**
+     * @notice Claims and tracks rewards for each committee member. Restricted to autonity contract.
+     * @dev Must be called before doing any staking operation.
+     * @param _validators committee members
+     * @param _liquidContracts liquid state contract addresses of the respective committee member
+     */
+    function collectRewards(address[] memory _validators, ILiquid[] memory _liquidContracts) external onlyAutonity {
+        uint256 _count = _validators.length;
+        require(_count == _liquidContracts.length, "invalid inputs");
```
Jason-Zhangxin-Chen, line 115: This error message should be refined to add more specific info.

Jason-Zhangxin-Chen (changes requested):
I have a few comments and questions left, thank you, Tariq. The structure looks better. However, if we can add a README.md for this feature in our contract directory would be better. In the README, you can high light the state machine transitions for the staking tasks and the asset steaming in those pools. For the other reviews, please start with the data models in IStakingPool.sol first, that would help a lots.

PR #1145, opened 2025-02-14
enable peer consensus endpoint check on receiving inbound connection

vpiyush:
During piccadilly network testing it was observed that some of the acn peers didn't open their consensus port correctly. due to which the peer count on consensus not was not consistent across cluster. Find more detailed description [here](https://github.com/clearmatics/web-priv-team-platform/issues/529).

This change would eliminate scenarios like this from occurring and the validators who don't have consensus port open will not be able to connect at all to any of the peers

### DIFF
--- a/p2p/peer_error.go
+++ b/p2p/peer_error.go
@@ -78,9 +78,9 @@ const (
 	DiscPeerNotInCommittee
 	DiscPeerOutsideTopology
 	DiscSyncFailed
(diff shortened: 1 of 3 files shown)


Jason-Zhangxin-Chen (approved):
LGTM.

lorenzo-dev1 (approved):
LGTM

PR #1147, opened 2025-02-21
fix revert due to arithetic underflow/overflow in autonity finalize

vpiyush:
during proposer reward calculation we use the current configured committee size. however this can be changed mid epoch, if the new committee size is signicately lower than the current committee the proposerRewards end up being more than the rewards itself. which ultimately causes solidity exception of  arithmetic underflow/overflow.


lorenzo-dev1 (approved):
LGTM nice catch

PR #1148, opened 2025-02-21
imporve acn network connection setup time

vpiyush:
Currently retry dialTimer and inboundThrottle timer configuration for both networks (acn and execution) are same. This PR reduces these timers for consensus network. My theory is that we want to retry to consensus nodes as fast as possible. This also helps us in scenario where multiple nodes starts at the same time. In such cases first dial attempt for some of the nodes would not succeed. Now with a smaller retry timer they would retry the setup faster.

This is still open for discussion though, because this comes with a increased DOS risk due to the reduced inboundThrottle timer. IMO the new 10 second duration is not too low, it should still be sufficient to avoid DOS attacks.

Another Fix in this PR is to start core before aggregator, because aggregator seem to access the core height which is only set when core starts. Else it panics with the nil pointer exception.


lorenzo-dev1 (approved):
LGTM 

> This is still open for discussion though, because this comes with a increased DOS risk due to the reduced inboundThrottle timer. IMO the new 10 second duration is not too low, it should still be sufficient to avoid DOS attacks.

I think 10 seconds is plenty of time to avoid DOS, especially on the "permissioned" consensus network.

Jason-Zhangxin-Chen (approved):
LGTM.

PR #1150, opened 2025-03-04
Add GitHub Actions workflow to trigger downstream autonity.py build

szemate:
In order to keep the latency between new Autonity releases and new releases of the Python tooling around it minimal, we'd like to keep the Python contract bindings in sync with the `develop` branch of Autonity. As a result we will be able work on autonity-cli updates continuously, not all changes at once after a new Autonity release comes out.

This commit adds a new GitHub workflow that builds the AGC binary and the protocol contracts, uploads them to GitHub artifacts, and triggers another workflow in the autonity.py repository that updates the Python bindings.

Requires a GitHub access token with "workflow" scope for the autonity/autonity.py repository, exposed in the `ACCESS_TOKEN` repository secret.


lorenzo-dev1 (approved):
LGTM

PR #1151, opened 2025-03-04
upgrade docker e2e test to run on ubuntu 24.04

lorenzo-dev1:
ubuntu 20.04 on the github ci runner is being deprecated, and since we are in brownout time the CI is faling. See https://github.com/actions/runner-images/issues/11101


Jason-Zhangxin-Chen:
We might need to update the docker file of the test engine and the Autonity Client in the e2e test framework.

docker_e2e_test/client/client.py
```diff
@@ -375,37 +351,43 @@ def get_balance(self):
...
     def get_chain_height(self):
         try:
             if self.rpc_client is None:
-                self.rpc_client = RpcClient(host=self.host, port=self.rpc_port)
-                self.rpc_client.session.headers.update({"Content-type": "application/json"})
-            height = self.rpc_client.get_block_number()
+                self.rpc_client = Web3(Web3.HTTPProvider(f"http://{self.host}:{self.rpc_port}"))
+                if not self.rpc_client.isConnected():
+                    raise Exception("Cannot connect to L1 node")
+            height = self.rpc_client.eth.blockNumber
```
lorenzo-dev1, line 390: Does this func gets executed even without () ?

lorenzo-dev1:
small comment + I pushed a commit that removes some debug prints I had initially added.

I cannot approve since I originally opened the PR, but LGTM thanks @Jason-Zhangxin-Chen 

Feel free to approve to count my review towards the 2 approvals threshold for merging

lorenzo-dev1:
Actually the latest run failed with `2025-05-19T16:46:57.8287402Z cannot build image:  toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit`

I think we might need to create an account on docker hub to avoid getting rate limited

docker_e2e_test/client/client.py
Jason-Zhangxin-Chen: Yes, blockNumber is tagged as a property of the object of eth, however it will be called.

Jason-Zhangxin-Chen (approved):
LGTM

Jason-Zhangxin-Chen:
> Actually the latest run failed with `2025-05-19T16:46:57.8287402Z cannot build image: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit`
> 
> I think we might need to create an account on docker hub to avoid getting rate limited

Let's experiment the free quote first, I think I might be abused it during the debugging yesterday. As 100 docker hub pulls per IP per 6 hours should be sufficient unless we have a lots of pull requests in CI. Anyway, we can still merge those tests as we did in the history, but it would take more time to finish the test.

lorenzo-dev1:
@Jason-Zhangxin-Chen seems like there are a lot of warnings about being unable to send transactions in some docker tests run, is this normal?

yazzaoui (approved):
Good to go if that works

PR #1156, opened 2025-03-17
Penalize Oracle Voters for No Reveal

tbssajal:


autonity/solidity/contracts/Oracle.sol
```diff
@@ -670,6 +723,55 @@ contract Oracle is IOracle {
...
+                _voterInfo.revealsMissed = 0;
+                // penalize with highest
+                config.autonity.slash(voters[i], ORACLE_SLASHING_RATE_CAP);
+            }
+            else if (resetCounter == 0) {
+                // counter has been reset
+                _voterInfo.revealsMissed = 0;
+            }
+        }
+    }
```
lorenzo-dev1, line 763: I see the logic behind this function, however I wonder if it would be simpler to just increment the missed reveal counter where you set `voterInfo.commitRevealed` to false. Moreover we should also increment it when there is a gap in the voting rounds (the condition `_lastVotedRound != round - 1`).

Am i missing something?

tbssajal: In the letter case, we increment the counter later than when we should. We might never increment if he never votes and the reset window is passed.

lorenzo-dev1: I think that if we make the reset window last the entire epoch, it wouldn't be too much of an issue if we do not penalize a node that commits and then stop voting for the entire epoch.

Effectively it is like we have an implicit +1 on the tolerance.

autonity/solidity/contracts/interfaces/IOracle.sol
```diff
@@ -134,4 +149,12 @@ interface IOracle {
...
      */
     event Penalized(address indexed _participant, uint256 _slashingAmount, string _symbol, int256 _median, uint120 _reported);
+
+    /**
+     * @notice Emitted when a participant gets penalized for missing too many reveals in a certain window
+     * @param _voter Voter address
+     * @param _round Round where penalized
+     * @param _missedReveal Count of missed reveal
+     */
+    event NoRevealPenalty(address indexed _voter, uint256 _round, uint256 _missedReveal);
```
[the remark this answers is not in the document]
tbssajal: We always slash with the highest offence so I omitted the slashing amount.

autonity/solidity/contracts/Oracle.sol
tbssajal: Yeah. But we will need to itereate through all the voters anyway to slash them for missed reveal at each oracle round. Also I am not sure if we should make the reset window last the entire epoch.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -670,6 +723,55 @@ contract Oracle is IOracle {
...
+    
+    function _penalizeForNoReveal() internal {
+        // reset missed reveal counter
+        resetCounter++;
+        if (resetCounter >= config.missedRevealWindow) {
+            resetCounter = 0;
+        }
+
+        // penalize for commit without reveal
+        for (uint i = 0; i < voters.length; i++) {
```
vpiyush, line 735: this penalilze could be done in the agregate loop itself, we already handle outlier penalize their. Saving some gas cost. Also if a voter is penalized for non - reveal we could right away not use their reports and probably we won't need the penalizedVoter variable as well. currently this is helping in maintaining some metaData across function scopes

autonity/solidity/contracts/Oracle.sol
```diff
@@ -71,6 +75,10 @@ contract Oracle is IOracle {
     mapping(address => uint256) private rewardPeriodPerformance;
     uint256 private rewardPeriodAggregatedScore;
 
+    uint256 private resetCounter;
+    // set of voters penalized in the current round
+    EnumerableSet.AddressSet private penalizedVoters;
```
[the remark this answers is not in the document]
tbssajal: Thanks for the suggestions! I just want to say that I used `penalizedVoters` because it can have more usecases. Currently the idea here is that we don't want to slash one voter more than once in a voting round. In case we want to add some more layers for penalizing the voters, we can use this set to slash at most once.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -147,33 +159,36 @@ contract Oracle is IOracle {
...
             return;
         }
 
-        if (_lastVotedRound != round - 1 ||
-            _pastCommit != uint256(keccak256(abi.encode(_reports, _salt, msg.sender)))
-        ) {
+        if (_pastCommit != uint256(keccak256(abi.encode(_reports, _salt, msg.sender)))) {
             // we return the tx fee in all cases, because in both cases voter is slashed during aggregation
             // phase, because the reports contain invalid prices
+            _voterInfo.commitRevealed = false; // did not reveal commit, cannot anymore
```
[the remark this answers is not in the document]
tbssajal: > current variabled commitRevealed=false doesn't really mean that commit was not revealed, It is more of either reveal was not needed Or it was not revealed.

I guess it means commit was not revealed unless the voter did not vote in the current round. If the voter does not vote in current round, then both variable, `commitRevealed` and `missedReveal` lose their meaning, doesn't matter what is stored in them.

I guess whatever variable we use, we will need to update it every time a new vote comes. Even if we use `missedReveal`, we need to set `false` or `true` explicitly because it could be that we already have it set to `true` or `false`, whichever is not appropriate. Unless if we reset the variable to its default value at the end of the voting round.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -670,6 +723,55 @@ contract Oracle is IOracle {
 
         return config.autonity.slash(voterValidators[_outlier], _slashingRate);
     }
+    
+    function _penalizeForNoReveal() internal {
+        // reset missed reveal counter
+        resetCounter++;
```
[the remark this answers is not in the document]
tbssajal: Good point

autonity/solidity/contracts/Oracle.sol
```diff
@@ -670,6 +723,55 @@ contract Oracle is IOracle {
...
+                    1. Voter voted in this round but did not reveal his commit from
+                        the last round (it's not the first round for the voter).
+                    2. Voter provided commit in the last round but did not vote in this round.
+                In both cases, the voter submitted commit in the last round but did not reveal
+            */
+            if (
+                (!_voterInfo.commitRevealed && _voterInfo.round == round) ||
+                (_voterInfo.round == round-1 && _voterInfo.commit > 0)
+            ) {
+                _voterInfo.revealsMissed++;
```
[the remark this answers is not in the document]
tbssajal: An event is a good idea. But it could emit too many events. I guess I can change the flow a bit to support event.

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
tbssajal: I think your suggestion is more intuitive.

autonity/solidity/contracts/Oracle.sol
tbssajal: > however I wonder if it would be simpler to just increment the missed reveal counter where you set `voterInfo.commitRevealed to false`.

I ened up doing it this way to support event emission each time we increase the counter. However the logic is a bit different. I am increasing the counter in both vote() function and finalize() function. But in finalize(), it's only increased if a voter did not submit any vote in the current round but voted in the previous round.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -84,6 +92,10 @@ contract Oracle is IOracle {
         string[] memory _symbols,
         Config memory _config
     ) {
+        require(
+            _config.missedRevealTolerance < _config.noRevealPenaltyResetRound && _config.missedRevealTolerance > 0,
+            "invalid config"
```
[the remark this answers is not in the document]
tbssajal: updated

autonity/solidity/contracts/Oracle.sol
```diff
@@ -474,6 +501,28 @@ contract Oracle is IOracle {
         config.votePeriod = _votePeriod;
     }
 
+    /**
+     * @notice Setter for the tolerance of missed reveal count before the voter gets punished.
+     */
+    function setMissedRevealTolerance(uint256 _tolerance) external onlyOperator {
+        require(
```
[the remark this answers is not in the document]
tbssajal: updated

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
tbssajal: updated

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
tbssajal: updated

autonity/solidity/contracts/Oracle.sol
```diff
@@ -147,33 +159,36 @@ contract Oracle is IOracle {
...
-        if (_lastVotedRound == 0) {
+        if (_lastVotedRound == 0 || _lastVotedRound != round - 1) {
+            _voterInfo.commitRevealed = true; // because we don't have past commit
             return;
         }
 
         // if data is not supplied and voter is not a new voter
         // report must contain the correct price
         if (_reports.length != symbols.length) {
+            _voterInfo.commitRevealed = false; // did not reveal commit, cannot anymore
```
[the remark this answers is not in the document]
tbssajal: updated

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
tbssajal: updated

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
tbssajal: Removed this variable.

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
vpiyush: I see your point, however a general guideline is to not base the current decision on a future requirement. In most cases, future requirements don't materialize in the same way we currently think about them. And if they we can always make the change accordingly.

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
vpiyush: I believe you would still benefit from the "missedReveal" naming, because the default is false. So you don't really have to set it and only set it to true in one case. So much less code. However not a strong opinion, May be i am missing something.

autonity/solidity/contracts/Oracle.sol
vpiyush: My initial thought was to avoid iterating the same loop twice to save gas cost, Removal of PenalizedVoter is a bonus. you can choose not to remove it .

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
vpiyush: I see your point, however a general guideline is to not base the current decision on a future requirement. In most cases, future requirements don't materialize in the same way we currently think about them. And if they we can always make the change accordingly.

Jason-Zhangxin-Chen (approved):
LGTM.

autonity/solidity/contracts/Oracle.sol
tbssajal: Actually we iterate through `voters` multiple times, `symbols.length` to be exact. With the current implementation, it's `symbols.length+1` times. I guess the improvement here is not much compared to current implementation. In case we run into issues, I guess we can address this separately, like maybe aggregating reports as they come in `vote` function.

autonity/solidity/contracts/Oracle.sol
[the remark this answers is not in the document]
tbssajal: updated

vpiyush (approved):
LGTM!

tbssajal:
@ScottRMalley @lorenzo-dev1 one of your approval would be good to have after the latest changes.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -75,6 +77,8 @@ contract Oracle is IOracle, IConfigEvents {
     mapping(address => uint256) private rewardPeriodPerformance;
     uint256 private rewardPeriodAggregatedScore;
 
+    EnumerableSet.AddressSet private nonRevealSlashedVoters;
```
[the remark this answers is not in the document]
tbssajal: I like this approach. Using storage variable is gas costly compared to this approach. Can you please push this commit in my branch? Please create two commits, one for *.sol and another for auto generated files. @ScottRMalley

autonity/solidity/contracts/Oracle.sol
```diff
@@ -706,6 +758,55 @@ contract Oracle is IOracle, IConfigEvents {
...
+        for (uint i = 0; i < voters.length; i++) {
+            address _voter = voters[i];
+            VoterInfo storage _voterInfo = voterInfo[_voter];
+            /*
+                Following two scenarios can happen:
+                    1. Voter voted in this round but did not reveal his commit from
+                        the last round (it's not the first round for the voter).
+                    2. Voter provided commit in the last round but did not vote in this round.
+                In both cases, the voter submitted commit in the last round but did not reveal
+            */
```
lorenzo-dev1, line 774: If I understand correctly point 1 is not addressed here in this func right? 

here we only address if an oracle didn't reveal and didn't commit 0 on his last committment due to lack of data.

autonity/solidity/contracts/Oracle.sol
```diff
@@ -706,6 +758,55 @@ contract Oracle is IOracle, IConfigEvents {
...
+            voterInfo[voters[i]].nonRevealCount = 0;
+        }
+    }
+
+    function _removePenalizedVoters() internal {
+        uint256 _length = nonRevealSlashedVoters.length();
+        while (_length > 0) {
+            address _voter = nonRevealSlashedVoters.at(0);
+            // don't consider past reports from the penalized voters
+            voterInfo[_voter].reportAvailable = false;
```
lorenzo-dev1, line 800: I am not sure about this, I think we could keep using their last valid report if it does not get detected as outlier.  Or am I missing something?

autonity/solidity/contracts/Oracle.sol
```diff
@@ -507,6 +531,30 @@ contract Oracle is IOracle, IConfigEvents {
...
+     */
+    function setRevealResetInterval(uint256 _resetInterval) external onlyOperator {
+        require(
+            config.nonRevealThreshold < _resetInterval && _resetInterval > 0,
+            "invalid config"
+        );
+        emit ConfigUpdateUint("revealResetInterval", config.revealResetInterval, _resetInterval);
+        config.revealResetInterval = _resetInterval;
+    }
+
```
lorenzo-dev1, line 557: I think that to avoid confusion we could apply the new threshold and the new period at the end of the current reset period. I think the two will always be correlated anyways... Maybe an easy way to do it is to merge the two setters in one

lorenzo-dev1 (commented):
Some more comments but the approach LGTM!

autonity/solidity/contracts/Oracle.sol
tbssajal: yes

autonity/solidity/contracts/Oracle.sol
tbssajal: I don't have any take on this except that in our current model, if someone is detected as outlier for a single symbol, all his remaining reports also get discarded. This is a separate topic to discuss but for this case I have no opinion really.

autonity/solidity/contracts/Oracle.sol
tbssajal: > Maybe an easy way to do it is to merge the two setters in one

That's a good idea.

> I think that to avoid confusion we could apply the new threshold and the new period at the end of the current reset period.

It requires additional state variables and probably getters and also modifying tests. Is it necessary?

autonity/solidity/contracts/Oracle.sol
lorenzo-dev1: I just raised it because his last reports could still be valid (inside the outlier threshold band) even though he is missing reveals. However I do not have a strong opinion on this either.

autonity/solidity/contracts/Oracle.sol
tbssajal: So whether we change it immeidately or at the boundary, it's the same because we are not using any counter, instead we are doing `round % interval == 0` check. So it's the same thing. The downside is that the voters' missed reveal count will be reset sooner, but it's due to `round % interval == 0`, whenever we set the interval.

autonity/solidity/contracts/Oracle.sol
tbssajal: Anyone else have any opinion on this? @yazzaoui @vpiyush @ScottRMalley @Jason-Zhangxin-Chen

autonity/solidity/contracts/Oracle.sol
lorenzo-dev1: right, however if we change the threshold before resetting everything, we could end up with:
- people who did not get slashed which now do get slashed at the following round even if they did not missed any more reveals (but I guess that's fine)
- people who did get slashed which are now under threshold, which could be confusing for the validator.

However I do agree that it seems like it doesn't pose any critical security risks for the protocol, so I am fine to leave it like this if you think it is not necessary.

autonity/solidity/contracts/Oracle.sol
lorenzo-dev1: Tbh I don't want to held hostage this PR any longer, so I'll approve now, feel free to merge in if you are satisfied

lorenzo-dev1 (approved):
LGTM!

autonity/solidity/contracts/Oracle.sol
ScottRMalley: I agree with @lorenzo-dev1 that there isn't an obvious connection here between non-reveal slashing and the validity of past reports. Previously we were invalidating past reports when they were penalized, but that was because penalizing was only a result of outlier detection, which could have an impact on price.

That being said, I think it is fine as is. We are only marking the reports as invalid when they have exceeded the threshold. If we are confident in slashing them at the max rate, we should be comfortable with excluding their past reports. Past reports shouldn't be frequently relied upon in either case as we do expect voters to submit every round.

lorenzo-dev1: yeah my concern was also due to the gas consumption of looping through the entire voters for a non-essential operation. But we can always remove it later if we see that round change takes too much gas

PR #1164, opened 2025-05-02
Improvement, a view exchanging based AskSync for consensus SMR recovery.

Jason-Zhangxin-Chen:
This PR contains below improvements to make the consensus engine be more scalable with large scale network.
1. Decouple the processing of AskSync from core to release the locking of round messages.
2. The sync is based on view exchanging to reduce the messaging overhead.
3. Future messages are also be sent to the asker client.


consensus/tendermint/accountability/lost_sync_handler.go
```diff
@@ -0,0 +1,238 @@
...
+	timeDiff := now - lastTS
+
+	return timeDiff < int64(constants.AskSyncInterval)
+}
+
+func (r *AskSyncRateLimiter) resetRateLimiter() {
+	for k := range r.lastRequestTSs {
+		delete(r.lastRequestTSs, k)
+	}
+}
```
lorenzo-dev1, line 48: I see why we need to delete entries from the map from time to time (to avoid having a memory leak), however I think this way kind of changes the rate limiting that we are doing. Maybe we can avoid deleting recent entries? That would ensure that spammers will still be detected as overrated

consensus/tendermint/core/message/round_messages.go
```diff
@@ -217,3 +223,51 @@ func (s *RoundMessages) AllMessages() []Msg {
...
+	}
+	return view
+}
+
+type RoundMsgView struct {
+	Round uint64
+
+	// the 1st received proposal that the client prevoted, normally only one, could be multiple if proposer equivocated.
+	// different proposals will be exchanged if one find there is an equivocated one.
+	Proposal common.Hash
```
lorenzo-dev1, line 256: I think that for the proposal really we just need a flag that says whether the remote node has a proposal for the round or not. In any case if he already has a proposal, we will not send other ones for the same round.

lorenzo-dev1 (changes requested):
I think that the approach generally looks good. I left some improvements comments and I also opened a PR that refactors the `lost_sync_handler` code to make it more readable and more performance efficient. I felt like it would be easier to just do the refactor and let you see if you like it rather than posting dozen of comments. The refactoring PR is https://github.com/autonity/autonity/pull/1169

consensus/tendermint/backend/gossiper.go
```diff
@@ -50,17 +51,19 @@ func (g *Gossiper) UpdateStopChannel(stopCh chan struct{}) {
...
 		return
 	}
-	code := NetworkCodes[message.Code()]
-	payload := message.Payload()
-	for _, val := range committee.Members {
+
+	// forward future epoch proposal to all the committee members, as most of them are still in the committee.
+	recipients := committee.Members
+	code := message.NetworkCodes[msg.Code()]
+	for _, val := range recipients {
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: > Why are these changes to the Gossip function needed?

That was because the msg code mapping now have to be moved to a common package, thus it reduce the code duplication when network clustering be merged. In this PR, I picked it when I cherry pick this improvement from the network clustering feature.

consensus/tendermint/core/constants/constant.go
```diff
@@ -3,4 +3,5 @@ package constants
 const (
 	MaxRound                      = 999 // consequence of backlog priority
 	ProposerElectionCycleConstant = 99  // seed for controlling proposer election repetition
+	AskSyncInterval               = 5   // the interval in seconds to check the liveness and rise AskSync request.
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Node can send a sync msg from an every single restart, so in some devop operation case, it can send ask sync more than once per 30s, thus I would leave it 5s here, if it abuse this 5s, we will ban it from p2p network layer.

consensus/tendermint/core/sync_msg.go
```diff
@@ -0,0 +1,87 @@
...
+				roundView.Proposal = m.Value()
+			}
+
+			if m.Code() == message.PrevoteCode {
+				value := m.Value()
+				_, ok := preVoteSigners[value]
+				if !ok {
+					preVoteSigners[value] = message.NewAggregatedPower()
+				}
+				for index, power := range m.(message.Vote).Signers().Powers() {
```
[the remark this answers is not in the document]
lorenzo-dev1: Moreover I am a bit concerned that the future round message snapshot could be abused since it has a lot of nested iterations (for every round, iterate through all the messages, and for all messages iterate through all signers). I am thinking about a scenario where a malicious node sends us a lot of future round nil prevotes and nil precommits.

However there is not a straightforward way for the remote attacker to cause the local node to askSync, so I think it is not easy to exploit.

I would say that if we have time we could refactor the futureRound messages data structures to allow easy access to the vote signers for each value. However it is not a priority.

consensus/tendermint/core/sync_msg.go
```diff
@@ -0,0 +1,87 @@
...
+	// and future rounds
+	future := c.futureRoundSnapshot()
+	if len(future) > 0 {
+		msg.RoundsViews = append(msg.RoundsViews, future...)
+	}
+
+	return msg
+}
+
+func (c *Core) futureRoundSnapshot() []*message.RoundMsgView {
```
lorenzo-dev1, line 23: I am a bit concerned that the `futureRoundSnapshot` function could be abused since it has a lot of nested iterations (for every round, iterate through all the messages, and for all messages iterate through all signers). I am thinking about a scenario where a malicious node sends the victim a lot of future round nil prevotes and nil precommits.

However there is not a straightforward way for the remote attacker to cause the local node to askSync, so I think it is not easy to exploit.

I would say that if we have time we could refactor the futureRound messages data structures to allow easy access to the vote signers for each value. However it is not a priority.

consensus/tendermint/core/message/round_messages.go
Jason-Zhangxin-Chen: I thought it would miss a pontential equivocation of a malicious proposer then.

consensus/tendermint/core/handler.go
```diff
@@ -292,7 +289,8 @@ func (c *Core) syncLoop(ctx context.Context) {
 	height := c.Height()
 
 	// Ask for sync when the engine starts
-	c.backend.AskSync(c.committee.Committee())
+	syncMsg := c.snapshotLostSyncMsg()
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resovled in f73b61fb04455de4fda20b4bc04bcffa2ae42fd3

consensus/tendermint/core/handler.go
```diff
@@ -301,25 +299,19 @@ eventLoop:
...
 			currentHeight := c.Height()
 
-			// we only ask for sync if the current view stayed the same for the past 10 seconds
+			// we only ask for sync if the current view stayed the same for the past 30 seconds
 			if currentHeight.Cmp(height) == 0 && currentRound == round {
-				c.logger.Warn("⚠️ Consensus liveliness lost")
-				c.logger.Warn("Broadcasting sync request..")
-				c.backend.AskSync(c.committee.Committee())
+				c.logger.Warn("⚠️ Consensus liveliness lost", "node", c.Address(), "height", height, "round", currentRound, "step", c.Step())
+				c.logger.Warn("Broadcasting sync request..", "node", c.Address(), "height", height, "round", currentRound, "step", c.Step())
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resovled in https://github.com/autonity/autonity/commit/f73b61fb04455de4fda20b4bc04bcffa2ae42fd3

consensus/tendermint/events/events.go
```diff
@@ -144,8 +144,10 @@ func (f FuturePowerChangeEvent) Value() common.Hash {
 	panic("not implemented")
 }
 
-type SyncEvent struct {
-	Addr common.Address
+type LostSyncEvent struct {
+	Sender  common.Address
+	Payload []byte
+	ErrCh   chan<- error
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resovled in https://github.com/autonity/autonity/commit/f73b61fb04455de4fda20b4bc04bcffa2ae42fd3

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -233,6 +238,19 @@ tendermintMsgLoop:
 					}
 					continue tendermintMsgLoop
 				}
+			case events.LostSyncEvent:
+				// process liveness fault from msg store context to release the consensus core locking in large scale network
+				// in which the processing of huge num of AskSyncs would lock the core for a long period.
+				err := fd.handleLostSyncEvent(e.Payload, e.Sender)
+				if err != nil {
+					fd.logger.Error("Accountability: lost sync recovery", "error", err)
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resovled in https://github.com/autonity/autonity/commit/f73b61fb04455de4fda20b4bc04bcffa2ae42fd3

consensus/tendermint/core/message/round_messages.go
```diff
@@ -36,10 +39,16 @@ func (s *Map) GetOrCreate(round int64) *RoundMessages {
...
 
-// TODO: this function has a mutex that can be taken by:
-// 1. the core routine
-// 2. the routine that syncs other peers
-// can this be exploited by a malicious peer to slow Core down (by requesting ask sync lots of times)
+func (s *Map) Snapshot() []*RoundMsgView {
+	s.RLock()         // nolint
+	defer s.RUnlock() // nolint
+	var views []*RoundMsgView
+	for r, v := range s.internal {
```
lorenzo-dev1, line 46: what is defined here as `w` is generally called `state` in the other functions. Would be good to keep that code consistency.

consensus/tendermint/core/sync_msg.go
Jason-Zhangxin-Chen: There is a truth that only a node who is out of sync that will call this function locally to dump itself's view of round messages, it would not DoS itself unless someone let this node enter into an out-of-sync state. In our current messaging topology, and the strategy of relaying, it can hardly happen.

consensus/tendermint/core/constants/constant.go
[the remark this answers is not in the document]
lorenzo-dev1: I see. I think that ideally we would want something that actually limits the rate (e.g. max x requests per minute). With the current approach, if a node just sent a askSync requests, then it crashes and sends another askSync requests right away, it would get rate limited.

consensus/tendermint/core/sync_msg.go
lorenzo-dev1: I agree it is an edge case, I would suggest to leave it for later.

consensus/tendermint/core/message/round_messages.go
Jason-Zhangxin-Chen: resolved in 1fc9db52ae1a9b68cc1723fe6658ef61e7962972

consensus/tendermint/core/constants/constant.go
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Refined the rate limiter in 50556bc4d8a6acbf2e4ccbf4e69928014ed1f729

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -70,9 +72,10 @@ var (
 // read state db on each new height to get the latest challenges from autonity contract's view,
 // and to prove its innocent if there were any challenges on the suspicious node.
 type FaultDetector struct {
-	innocenceProofBuff *InnocenceProofBuffer
-	protocolContracts  *autonity.ProtocolContracts
-	rateLimiter        *AccusationRateLimiter
+	innocenceProofBuff    *InnocenceProofBuffer
+	protocolContracts     *autonity.ProtocolContracts
+	accusationRateLimiter *AccusationRateLimiter
+	askSyncRateLimiter    *AskSyncRateLimiter
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: refined the rate limiter in 50556bc4d8a6acbf2e4ccbf4e69928014ed1f729

consensus/tendermint/core/sync_msg.go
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: resolved in 7d72c1c1008124bf5c0834fd720e6f97cc92d3e8

consensus/tendermint/core/message/round_messages.go
Jason-Zhangxin-Chen: resolved in 287c5c63f330c3ca9df0c495e277f8f659bb7301

consensus/tendermint/core/message/round_messages.go
```diff
@@ -311,7 +309,7 @@ func (lsm *AskSyncMsg) Validate() error {
 		}
 
 		// if the remote peer does not have a proposal for this round, mark it
-		if v.Proposal == (common.Hash{}) {
+		if v.HaveProposal {
```
lorenzo-dev1, line 312: Shouldn't it be `if !v.HaveProposal` (if the remote peer doesn't have the proposal, then we send it to him

consensus/tendermint/core/sync_msg.go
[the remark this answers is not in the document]
lorenzo-dev1: I understand your change, however AggregatedPower semantically has nothing to do here. We are just storing the signers bitmap, so I think it is better to just use a *big.Int instead of misusing the AggregatedPower object to store only that. I pushed a commit that implements the change I had in mind. feel free to cherry pick it if you like it 96eb5e23385abb16b740309bc7e814ec488843fc

consensus/tendermint/core/message/round_messages.go
Jason-Zhangxin-Chen: Yeah, this is a bug. Resolved in 36e678c559282a81dfe39ecf7e9d3b38e0cb4666

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,202 @@
...
+	for addr, record := range l.limits {
+		if now.After(record.expiration) {
+			delete(l.limits, addr)
+		}
+	}
+}
+
+type HeightBasedLimiter struct {
+	rwMutex      sync.RWMutex
+	accusations  map[common.Address]map[uint64]int
```
lorenzo-dev1, line 78: I guess since the rate limiter is generic now, it doesn't make sense to name this field accusations

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,202 @@
...
+var (
+	ErrRateLimitExceeded    = errors.New("rate limit exceeded")
+	ErrHeightQuotaExhausted = errors.New("block height quota exhausted")
+	ErrDuplicateMessage     = errors.New("duplicate message detected")
+)
+
+type TimeWindowLimiter struct {
+	rwMutex    sync.RWMutex
+	limits     map[common.Address]*rateRecord
+	timeWindow time.Duration
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Removed the useless field in 98d43df462e3f570ddfa73c71ed1030e511d9115, I just realized that the TTL is not required for the time window limiter.

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		return nil
+	}
+
+	if record.count >= l.burst {
+		return ErrRateLimitExceeded
+	}
+
+	record.count++
+	record.last = now
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: In the very beginning, `last` is used to implement a rolling time window, but I think that is over designed. I just removed it in https://github.com/autonity/autonity/commit/98d43df462e3f570ddfa73c71ed1030e511d9115

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		if len(heights) == 0 {
+			delete(l.accusations, addr)
+		}
+	}
+}
+
+type DuplicateTracker struct {
+	rwMutex  sync.RWMutex
+	messages map[common.Address]map[common.Hash]time.Time
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resolved in 4b7877571713e6478ecbbf33eddbfba1560acb4e

consensus/tendermint/accountability/rate_limiter.go
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/4b7877571713e6478ecbbf33eddbfba1560acb4e

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -130,8 +130,9 @@ func NewFaultDetector(
 	fd := &FaultDetector{
 		innocenceProofBuff:    NewInnocenceProofBuffer(),
 		protocolContracts:     protocolContracts,
-		accusationRateLimiter: NewAccusationRateLimiter(),
-		askSyncRateLimiter:    NewAskSyncRateLimiter(),
+		accusationRateLimiter: NewAFDRateLimiter(),
+		// 1 ask sync per 5s, rate limit reset per 5s.
+		askSyncRateLimiter:    NewTimeWindowLimiter(AskSyncInterval*time.Second, 1, AskSyncInterval*time.Second),
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/4b7877571713e6478ecbbf33eddbfba1560acb4e

consensus/tendermint/core/sync_msg.go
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Thank you, lorenzo. I just cherry picked it.

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+		if now.After(record.expiration) {
+			delete(l.limits, addr)
+		}
+	}
+}
+
+type HeightBasedLimiter struct {
+	rwMutex      sync.RWMutex
+	records      map[common.Address]map[uint64]int
+	maxPerHeight int
```
lorenzo-dev1, line 74: Is there a reason to not use uint64 in place of int?

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+func NewAFDRateLimiter() *AFDRateLimiter {
+	limiter := &AFDRateLimiter{
+		// since communication channel is asynchronous, those pending write of off chain accusation msgs from a sender
+		// could potentially be received once the peer connection get established from a disaster recovery, thus it
+		// could exceed the number of accusation that could be produced by rule engine over a height, so we set higher
+		// rate limit during 1 second to be tolerant for such case.
+		// 8 accusations per 1s window for per client, rate limit reset per 1s.
+		timeLimiter: NewTimeWindowLimiter(time.Second, maxAccusationPerHeight*2),
+		// 4 accusations per height for per client.
+		heightLimiter: NewHeightBasedLimiter(maxAccusationPerHeight, msgGCInterval),
```
lorenzo-dev1, line 185: I think that it would make sense to set the block to live parameter to `HeightRange`, as that's what we use to determine if an accusation is too old in the precompiled contracts ( see https://github.com/autonity/autonity/blob/develop/consensus/tendermint/accountability/contracts.go#L82-L87). 

Anything older than that is invalid, so I think we don't have to keep rate limiting info for that. What do you think?

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+		// duplicated accusation checker, reset per 10 minutes.
+		duplicateCheck: NewDuplicateTracker(time.Minute * 10),
```
lorenzo-dev1, line 187: 10 minutes seems a lot to me. I would suggest to cut that back to 5.

lorenzo-dev1 (commented):
Some small comments but overall LGTM

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+	ErrRateLimitExceeded    = errors.New("rate limit exceeded")
+	ErrHeightQuotaExhausted = errors.New("block height quota exhausted")
+	ErrDuplicateMessage     = errors.New("duplicate message detected")
+)
+
+type TimeWindowLimiter struct {
+	rwMutex    sync.RWMutex
+	limits     map[common.Address]*rateRecord
+	timeWindow time.Duration
+	burst      int
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resolved in 711f52c741248ba71ed8d4d520c11583685e95b1

consensus/tendermint/accountability/rate_limiter.go
Jason-Zhangxin-Chen: Resolved in 711f52c741248ba71ed8d4d520c11583685e95b1

consensus/tendermint/accountability/rate_limiter.go
Jason-Zhangxin-Chen: Resolved in 711f52c741248ba71ed8d4d520c11583685e95b1

lorenzo-dev1 (approved):
LGTM

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -233,6 +233,19 @@ tendermintMsgLoop:
 					}
 					continue tendermintMsgLoop
 				}
+			case events.SyncRequestEvent:
```
vpiyush, line 236: I like the fact that we have decoupled the synchandler from the core, But I am not coninved with the implantation in accountability module, I think it deserves it's separate module. By coupling it with accountability we overload the accoutability message handler loop and there is no direct dependency between these two. if it's separate it would be much easier to maintain. We can discuss this further on call if you like.

eth/backend.go
```diff
@@ -291,7 +291,7 @@ func New(stack *node.Node, config *Config) (*Ethereum, error) {
 	eth.accountability = accountability.NewFaultDetector(
 		eth.blockchain,
 		eth.address,
-		evMux.Subscribe(events.MessageEvent{}, events.AccountabilityEvent{}, events.OldMessageEvent{}),
+		evMux.Subscribe(events.MessageEvent{}, events.AccountabilityEvent{}, events.OldMessageEvent{}, events.SyncRequestEvent{}),
```
vpiyush, line 294: the current multiplexer is already overladed and we have done few iterations to separate out events from it. the sync request event don't need to go into multiplexer becuase it's target towards a single handler. Although this would be automatically corrected if we move the sync handling outside the accountability module.

consensus/tendermint/accountability/lost_sync_handler.go
```diff
@@ -0,0 +1,144 @@
...
+
+// handleAskSyncEvent handles the ask sync request from a lost sync validator or from a rebooting validator.
+// Any error return from this function will drop the remote peer.
+func (fd *FaultDetector) handleAskSyncEvent(payload []byte, sender common.Address) error {
+
+	if err := fd.askSyncRateLimiter.Allow(sender); err != nil {
+		return err
+	}
+
+	lostSync := new(message.AskSyncMsg)
```
vpiyush, line 25: minor: askSync would have been better name here. lostSync breaks the flow of someone reading the code. sticking to askSync retains the consistency throughout the code flow.

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+		if len(hashes) == 0 {
+			delete(t.records, addr)
+		}
+	}
+}
+
+type AFDRateLimiter struct {
+	timeLimiter    *TimeWindowLimiter
+	heightLimiter  *HeightBasedLimiter
+	duplicateCheck *DuplicateTracker
```
vpiyush, line 173: naming: duplicateCheck ==> duplicateLimiter for consistency

consensus/tendermint/accountability/lost_sync_handler.go
```diff
@@ -0,0 +1,144 @@
...
+		return fmt.Errorf("ask sync msg sanity check failed: %w", err)
+	}
+
+	// fetch remote's peer missing messages
+	proposals := fd.missingProposals(lostSync)
+	prevotes := fd.missingPrevotes(lostSync)
+	precommits := fd.missingPrecommits(lostSync)
+
+	// broadcast them to the missing peer
+	if fd.broadcaster == nil {
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resolved in 7154fc1e6a6cdacb59800ea6865493496eb542b7

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+	t.rwMutex.RLock()
+	if hashes, exists := t.records[sender]; exists {
+		if _, duplicate := hashes[hash]; duplicate {
+			t.rwMutex.RUnlock()
+			return ErrDuplicateMessage
+		}
+	}
+	t.rwMutex.RUnlock()
+
+	t.rwMutex.Lock()
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/7154fc1e6a6cdacb59800ea6865493496eb542b7

consensus/tendermint/accountability/rate_limiter.go
```diff
@@ -0,0 +1,197 @@
...
+	now := time.Now()
+	for addr, record := range l.limits {
+		if now.After(record.expiration) {
+			delete(l.limits, addr)
+		}
+	}
+}
+
+type HeightBasedLimiter struct {
+	rwMutex      sync.RWMutex
```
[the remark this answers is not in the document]
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/7154fc1e6a6cdacb59800ea6865493496eb542b7

consensus/tendermint/accountability/lost_sync_handler.go
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/7154fc1e6a6cdacb59800ea6865493496eb542b7

consensus/tendermint/accountability/rate_limiter.go
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/7154fc1e6a6cdacb59800ea6865493496eb542b7

eth/backend.go
Jason-Zhangxin-Chen: Creating a separate module for it might be too heavy, I am thinking to put it back to the core, binding with an individual go routine, is this idea satisfied everyone?

consensus/tendermint/accountability/fault_detector.go
vpiyush: do you want to discuss this on call?  do you mean too havey because the change required at this point? I think it doesn't matter where we put it as long as it doesn't interfere the core and accountability loop.

eth/backend.go
Jason-Zhangxin-Chen: Resolved in https://github.com/autonity/autonity/commit/27cfd446b9eda18a620849d2ba5d73cdbff3454a where I moved the askSyncEvent.

consensus/tendermint/backend/peer_synchronizer.go
```diff
@@ -0,0 +1,158 @@
...
+
+import (
+	"fmt"
+	"github.com/autonity/autonity/common"
+	"github.com/autonity/autonity/consensus/tendermint/core/message"
+	"github.com/autonity/autonity/rlp"
+)
+
+const AskSyncInterval = 5          // the interval in seconds to check the liveness and rise AskSync request.
+const RateLimiterGCThreshold = 256 // try to GC out of updated records if the total record of rate limiter is over 256.
```
lorenzo-dev1, line 11: I am not sure whether 256 is enough... as it means we do the cleanup when we have more than 256 distinct addresses in the ask sync limiter state. However if the network is temporary stuck and we have say 500 nodes in the committee we will most likely reach this threshold at every message. This will trigger a lot of iterations on the ask sync rate limiter... However yeah it is hard to settle on a number because it really depends on the committee size.

I think it was better to do it based on height like it was before, but I think that might be harder to do here. What do you think @Jason-Zhangxin-Chen @vpiyush

Jason-Zhangxin-Chen: Let's finalize it with dd43e2e30d6eeef492d363f79104706a3273759a which places the msg handling context into the core's syncEventLoop, and the clean-up if base on a timer of that loop as well.

vpiyush: cleanup is fine now, But why move the sync handling back to core and event into the eventMux, there is no need to send an event to eventmux.

consensus/tendermint/core/handler.go
```diff
@@ -301,25 +302,44 @@ eventLoop:
 			currentRound := c.Round()
 			currentHeight := c.Height()
 
-			// we only ask for sync if the current view stayed the same for the past 10 seconds
+			// we only ask for sync if the current view stayed the same for the past 30 seconds
```
vpiyush, line 305: I realised that my changes wrt askSync are not ported here completely. I had more sophisticated mechanism to detect when we are out of sync based on the time when we received last valid message, Was there a reason to not port those?
Because, In some cases when the network is slow chain can naturally take more than 30 sec and every node would go into sync mode increasing overall traffic in the network while the network is already slow making the situation even worse.

Jason-Zhangxin-Chen: Ported the previous work in 0e8f4c4c557b8db3c8cf86b2b5babb18d6d4c86a

consensus/tendermint/core/handler.go
```diff
@@ -278,12 +282,11 @@ eventLoop:
 	c.stopped <- struct{}{}
 }
 
-func (c *Core) livenessTrackerLoop(ctx context.Context) {
```
lorenzo-dev1: I think liveness tracker loop was actually a good name

lorenzo-dev1 (commented):
Some comments

consensus/tendermint/helpers/rate_limiter.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+func (t *DuplicateLimiter) Allow(sender common.Address, hash common.Hash) error {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+
+	if hashes, exists := t.records[sender]; exists {
+		if _, duplicate := hashes[hash]; duplicate {
+			return ErrDuplicateMessage
```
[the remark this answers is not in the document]
lorenzo-dev1: mmh no in general we should never re-process duplicates, the `ttl` just determines how long after we are going to delete them. But this is because we cannot store every hash forever in memory. But yeah theoretically we should never re-process a message once it is inserted in the map

consensus/tendermint/core/message/round_messages.go
```diff
@@ -217,3 +226,162 @@ func (s *RoundMessages) AllMessages() []Msg {
...
+type AskSyncMsg struct {
+	Height        uint64
+	KnownMessages []*RoundMsgView
+	// following fields are ignored when rlp/json encoding/decoding.
+	// They are built locally when validating the message
+	validated        bool                                `rlp:"-"`
+	rounds           map[uint64]struct{}                 `rlp:"-"`
+	prevoteSigners   map[uint64]map[common.Hash]*big.Int `rlp:"-"` // maps point to the signers bitmap for that specific value
+	precommitSigners map[uint64]map[common.Hash]*big.Int `rlp:"-"` // maps point to the signers bitmap for that specific value
+	nilProposal      map[uint64]struct{}                 `rlp:"-"` // marks round where remote node doesn't have a proposal
```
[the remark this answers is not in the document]
lorenzo-dev1: Ok maybe I see the point, I am going to try to refactor it and see if it can work with just 1 map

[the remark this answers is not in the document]
lorenzo-dev1: was this what you had in mind? https://github.com/autonity/autonity/pull/1164/commits/8f2b0b476a7fe36538b10f9822fcc35da7b414a9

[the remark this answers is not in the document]
tbssajal: Yes

tbssajal (approved):
LGTM

PR #1166, opened 2025-05-16
Cluster based message routing in consensus network

vpiyush:
Introducing clustering for message routing in consensus netowrk. In this model while we maintain the full mesh connection with all nodes in consensus network. We create clusters of node for efficient routing of the message. Every node relies on the elected committee and forms the identical clusters. While routing a message a node selects the routing nodes based on the assigned role for the message. Below are the roles:

- Originator
- First relayer origin cluster
- First relayer remote cluster
- Local relayer origin cluster
- Local relayer remote cluster


eth/backend.go
```diff
@@ -630,12 +631,40 @@ func (s *Ethereum) validatorController() {
 	if err != nil {
 		panic(err)
 	}
+
+	startMiningWhenReady := func(committee *types.Committee) {
+		go func() {
+			ticker := time.NewTicker(1 * time.Second)
+			timeout := time.After(1 * time.Minute) // max wait for 1 minute
```
Jason-Zhangxin-Chen, line 638: Is 1 minute timeout optimal for different scale of network?

eth/handler.go
```diff
@@ -563,9 +564,9 @@ func (h *handler) BroadcastTransactions(txs types.Transactions) {
 		annoCount += len(hashes)
 		peer.AsyncSendPooledTransactionHashes(hashes)
 	}
-	log.Debug("Transaction broadcast", "txs", len(txs),
-		"announce packs", annoPeers, "announced hashes", annoCount,
-		"tx packs", directPeers, "broadcast txs", directCount)
+	//log.Debug("Transaction broadcast", "txs", len(txs),
```
Jason-Zhangxin-Chen, line 567: Is this logging not required anymore?

params/config.go
```diff
@@ -99,7 +99,7 @@ var (
 		TerminalTotalDifficulty: nil,
 		Ethash:                  nil,
 		AutonityContractConfig: &AutonityContractGenesis{
-			MinBaseFee:              500_000_000,
+			MinBaseFee:              200_000_000_000,
```
Jason-Zhangxin-Chen, line 102: Let's leave this EIP1559 configs at the final config PR for mainnet, we will resolve all of those protocol configurations once we finalized the decision.

params/protocol_params.go
```diff
@@ -119,9 +119,9 @@ const (
 	// Introduced in Tangerine Whistle (Eip 150)
 	CreateBySelfdestructGas uint64 = 25000
 
-	BaseFeeChangeDenominator = 8          // Bounds the amount the base fee can change between blocks.
-	ElasticityMultiplier     = 2          // Bounds the maximum gas limit an EIP-1559 block may have.
-	InitialBaseFee           = 1000000000 // Initial base fee for EIP-1559 blocks.
+	BaseFeeChangeDenominator = 88              // Bounds the amount the base fee can change between blocks.
```
Jason-Zhangxin-Chen, line 122: Same for EIP1559 params.

core/types/epoch.go
```diff
@@ -292,45 +293,38 @@ func (c *Committee) MemberByAddress(address common.Address) *CommitteeMember {
 }*/
 
 func (c *Committee) TotalVotingPower() *big.Int {
-	c.lock.RLock() // Acquire read lock
-	if c.totalVotingPower != nil {
-		defer c.lock.RUnlock()                      // Release read lock
-		return new(big.Int).Set(c.totalVotingPower) // Return a copy of the cached value
+	total := new(big.Int)
```
Jason-Zhangxin-Chen, line 296: Any idea on why we change this function in this way? Don't we need a buffered power and some lock for race condition?

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,325 @@
...
+		if lat, ok := latencyMap[addr]; ok {
+			latency = lat
+		}
+		if latency < c.minLatency {
+			c.minLatency = latency
+		}
+		if latency > c.maxLatency {
+			c.maxLatency = latency
+		}
+		c.base[clusterID] = append(c.base[clusterID], Node{Addr: addr, Lat: uint(latency), ClusterID: clusterID})
```
Jason-Zhangxin-Chen, line 63: No need a uint() convertion.

Jason-Zhangxin-Chen (commented):
The overall structure look very nice, just a few comments left. I will add few more but they wouldn't be a big changes.

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,336 @@
...
+			c.base[clusterID] = []Node{}
+			continue
+		}
+		sort.Slice(peers, func(i, j int) bool { return peers[i].Lat < peers[j].Lat })
+		c.base[clusterID] = peers
+	}
+}
+
+// IDByAddress returns the ID of the cluster which containing the given address
+func (c *Clusters) IDByAddress(address common.Address) int {
```
Jason-Zhangxin-Chen, line 289: Those internal field of clusters can be accessed from different `go sb.router.Forward(committee, msg, sender)` routines, but if they are all read-read race, that would be fine. Just need to double check if we have read-write race.

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+			for _, bucketIdx := range bucketIndices {
+				if !filledBuckets[bucketIdx] {
+					clusterCandidatesForBucket[bucketIdx] = append(clusterCandidatesForBucket[bucketIdx], clusterID)
+				}
+			}
+		}
+
+		// Find the bucket with the fewest candidates (but at least one)
+		// we want to prioritize the buckets with lowest candidates
```
Jason-Zhangxin-Chen, line 149: Just curious, is this results a balanced bucket allocation for nodes?

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,344 @@
...
+		c.maxLatency = c.minLatency
+	}
+	return c, nil
+}
+
+// Base returns the base cluster views
+func (c *Clusters) Base() [][]Node {
+	return c.base
+}
+
```
Jason-Zhangxin-Chen, line 87: Add more comments for this function to help the reader to understand the purpose, as it group nodes into different buckets by according to their latency distribution, mapping the clusters to buckets for better routing performance, etc...

consensus/tendermint/core/handler.go
```diff
@@ -177,54 +177,71 @@ eventLoop:
...
 							break // do not gossip single message, only complex aggregate
 						}
 					}
 
-					// gossip message. We should arrive here only if we did not already gossip a complex aggregate
-					go c.backend.Gossip(c.CommitteeSet().Committee(), msg)
-					recordMessageProcessingTime(msg.Code(), start)
+					if err != nil && errors.Is(err, constants.ErrOldRoundMessage) {
+						go func() {
+							time.Sleep(5 * time.Millisecond) // minor sleep for old round messages
```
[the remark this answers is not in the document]
vpiyush: Idea was to prioritize current round and height messages by doing that. neverhless this has been removed now.

eth/handler.go
vpiyush: i will revert this change

params/protocol_params.go
vpiyush: removed them from there

consensus/tendermint/router/network/cluster.go
vpiyush: fixed

vpiyush:
> The unit tests of cluster.go and the network.go are missing, if you really trust on it, I can trust you.

I have added the unit tests for these as well

consensus/tendermint/router/routing.go
```diff
@@ -0,0 +1,347 @@
...
+			epoch := epochEv.Header.Epoch
+			m.inCommittee = epoch.Committee.MemberByAddress(m.self) != nil
+			if !m.inCommittee || len(m.committee) < ScaleThresholdForClustering {
+				log.Info("Router: clustering not needed, skipping measurement")
+				continue
+			}
+			m.updateCommittee(epoch)
+			nw := network.New(m.committee, m.latestLatencies, m.self)
+			m.updateNetwork(nw)
+			if err := m.measureLatency(); err != nil {
```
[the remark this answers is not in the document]
vpiyush: I agree but we also have retry every 30 seconds. which would measure for those failed nodes. so it should be fine. moreover in normal scenario we don't expect many new connection on a epoch change

eth/backend.go
vpiyush: yes should be fine. i updated it so that it would just proceed with starting minor, which should not have any side effects.

consensus/tendermint/router/network/cluster.go
vpiyush: not necessarily, It depends on the latency spread of the cluster nodes, It's just tries to find best suitable.

lorenzo-dev1 (commented):
I am still going through all the logic under `tendermint/router`, but it looks very good and well organized.

Concerning the interactions with the other parts of the clients, I raised  a bunch of smaller comments, but my biggest doubts that I would like to discuss are:
- addition of originator in the message struct and the logic around it
- early forwarding logic in the backend handler
- change in the start of mining process according to number of connected peers
- gossiping changes in `core/handler.go`

node/defaults.go
```diff
@@ -59,7 +59,7 @@ var DefaultConfig = Config{
 	ConsensusP2P: p2p.Config{
 		ListenAddr:      DefaultATCPort,
 		MaxPeers:        math.MaxInt,
-		MaxPendingPeers: 100,
+		MaxPendingPeers: 1000,
```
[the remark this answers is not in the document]
vpiyush: this allows us to connect to large number of peers simultaneously. only needed fro simultaneous connections though

p2p/dial.go
```diff
@@ -466,10 +468,11 @@ func (d *dialScheduler) removeFromStaticPool(idx int) {
 func (d *dialScheduler) startDial(task *dialTask) {
 	d.log.Trace("Starting p2p dial", "id", task.dest.ID(), "ip", task.dest.IP(), "flag", task.flags, "server", d.net.String())
 	hkey := string(task.dest.ID().Bytes())
+	randomDelay := time.Duration(rand.Intn(1000)) * time.Millisecond
 	if d.net == Consensus {
-		d.history.add(hkey, d.clock.Now().Add(acnDialHistoryExpiration))
+		d.history.add(hkey, d.clock.Now().Add(acnDialHistoryExpiration+randomDelay))
 	} else {
-		d.history.add(hkey, d.clock.Now().Add(dialHistoryExpiration))
+		d.history.add(hkey, d.clock.Now().Add(dialHistoryExpiration+randomDelay))
```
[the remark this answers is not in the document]
vpiyush: yes, I will get rid of it from here.

core/blockchain_reader.go
```diff
@@ -65,8 +65,18 @@ func (bc *BlockChain) EpochByHeight(height uint64) (*types.EpochInfo, error) {
...
+	head := bc.currentBlock.Load().(*types.Block)
+	if head.IsEpochHead() {
+		return &types.EpochInfo{
+			EpochBlock: new(big.Int).Set(head.Number()),
+			Epoch:      *head.Header().Epoch.Copy(),
+		}, nil
+	}
+
+	// otherwise, query the latest epoch for current head.
+	return bc.hc.EpochByHeight(head.NumberU64())
```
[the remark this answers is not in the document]
vpiyush: @Jason-Zhangxin-Chen I agree with this,  let me know if you think otherwise. this change had come from one of your PR merges in clustering.

core/types/epoch.go
```diff
@@ -277,60 +279,22 @@ func (c *Committee) MemberByAddress(address common.Address) *CommitteeMember {
...
-		c.membersMap = make(map[common.Address]*CommitteeMember)
-		for _, member := range c.Members {
-			m := member
-			c.membersMap[member.Address] = &m
-		}
-	}
-	return c.membersMap[address]
-}*/
-
 func (c *Committee) TotalVotingPower() *big.Int {
```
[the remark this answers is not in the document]
vpiyush: i have removed the readlock from the totalVotingPower, it is replaced by TotalVotingPower() invocation idea is to not used the totalVotingPower directly outside the TotalVotingPower().

eth/handler.go
```diff
@@ -545,6 +545,7 @@ func (h *handler) BroadcastTransactions(txs types.Transactions) {
 		peers := h.peers.peersWithoutTransaction(tx.Hash())
 		// Send the tx unconditionally to a subset of our peers
 		numDirect := int(math.Sqrt(float64(len(peers))))
+		//numDirect := len(peers)
```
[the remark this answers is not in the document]
vpiyush: removed

p2p/server.go
```diff
@@ -1266,6 +1266,10 @@ func (srv *Server) runPeer(p *Peer) {
 	})
 }
 
+func (srv *Server) Committee() []*enode.Node {
+	return srv.committee
+}
+
```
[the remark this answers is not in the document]
vpiyush: yes, it does and a copy would be definitely better.

consensus/tendermint/backend/backend.go
```diff
@@ -394,15 +418,16 @@ func (sb *Backend) SyncPeer(address common.Address) {
 		return
 	}
 	sb.logger.Debug("Syncing", "peer", address)
+	//todo: only to selected peers
```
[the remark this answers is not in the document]
vpiyush: removed this todo, out of the scope of this PR

consensus/tendermint/backend/aggregator.go
```diff
@@ -903,6 +907,59 @@ loop:
...
+						case message.PrecommitCode:
+							counts[2]++
+						default:
+							counts[3]++
+						}
+						stats[height][round] = counts
+					}
+				}
+
+				sb := strings.Builder{}
```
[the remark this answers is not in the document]
vpiyush: Go automatically takes the address to satisfy the pointer receiver requirement as long as the object is addressable.

consensus/tendermint/backend/gossiper.go
```diff
@@ -50,29 +67,58 @@ func (g *Gossiper) UpdateStopChannel(stopCh chan struct{}) {
...
 			p.Cache().Add(hash, true)
 			go p.SendRaw(code, payload) //nolint
+		} else {
+			// todo: Jason, shall we select other backups for liveness?
+			lostPeers = append(lostPeers, addr)
 		}
 	}
+	if len(lostPeers) > 0 {
+		g.logger.Debug("peers not found", "len", len(lostPeers), "peers", lostPeers)
+	}
```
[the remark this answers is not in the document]
vpiyush: yes, disconnection on consensus network

consensus/tendermint/core/handler.go
```diff
@@ -286,7 +295,8 @@ func (c *Core) syncLoop(ctx context.Context) {
 		this method is responsible for asking the network to send us the current consensus state
 		and to process sync queries events.
 	*/
-	timer := time.NewTimer(syncTimeOut)
+	// syncTime out should be dynamic based on the current round timer
+	// TODO: think about sending the bitmap of messages you currently have in sync request
```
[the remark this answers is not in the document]
vpiyush: yes, this needs to be handled in ask sync PR

consensus/tendermint/router/cache/cache.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func New() Recipients {
+	return &peerCache{
+		recipients:   make(map[string]Entry),
+		cacheVersion: 1,
+	}
+}
+
+func (c *peerCache) Get(key string) (Entry, bool) {
```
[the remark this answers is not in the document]
vpiyush: the last used i updated only if the entry is usable so ti's updated manually using updatedLastUsed.

consensus/tendermint/router/cache/cache.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (c *peerCache) Invalidate() {
+	c.Lock()
+	defer c.Unlock()
+	c.cacheVersion++
+}
+
+func (c *peerCache) Cleanup() {
```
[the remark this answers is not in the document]
vpiyush: yes, will do that.

consensus/tendermint/backend/engine.go
```diff
@@ -585,6 +585,11 @@ func (sb *Backend) Start(ctx context.Context) error {
 
 	sb.wg.Add(1)
 	go sb.faultyValidatorsWatcher(ctx)
+	if sb.router != nil {
+		sb.router.Start(ctx, sb.BlockChain())
+	} else {
+		sb.logger.Warn("Latency msgRouter is not set, cannot start")
+	}
```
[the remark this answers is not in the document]
vpiyush: removed this check

consensus/tendermint/backend/engine.go
```diff
@@ -606,6 +611,9 @@ func (sb *Backend) Close() error {
 	// Stop Tendermint
 	sb.aggregator.stop()
 	sb.core.Stop()
+	if sb.router != nil {
+		sb.router.Stop()
+	}
```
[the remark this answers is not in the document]
vpiyush: this as well

p2p/dial.go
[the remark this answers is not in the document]
lorenzo-dev1: Ok I see, then I will comment it here since I think that #1167 will get merged after this one. What about using `d.rand` as a source of randomness? It uses `math/rand` but it seeds it with some random data from `crand`. Maybe that is going to help avoiding that different nodes generate same random delays?

ScottRMalley (changes requested):
Still looking things over, but here are some initial comments.

p2p/dial.go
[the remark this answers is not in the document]
lorenzo-dev1: Actually I saw that Seeding is not needed anymore https://pkg.go.dev/math/rand#Seed 

Solving

consensus/tendermint/backend/gossiper.go
```diff
@@ -94,6 +92,15 @@ func (g *Gossiper) gossip(msg message.Msg, recipients []common.Address) {
 	if msg.Originator() == g.address {
 		g.knownMessages.Add(hash, true)
 	}
+	// if it's an aggregate the originator is the representative of the signers, so check in cache first and then add
+	switch m := msg.(type) {
+	case *message.Prevote:
```
lorenzo-dev1, line 97: Fallthrough missing here I believe

consensus/tendermint/router/constants/constants.go
```diff
@@ -0,0 +1,14 @@
+package constants
```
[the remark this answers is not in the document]
ScottRMalley: I think in general, I'm sceptical of the exports from all the subpackages in `router/`, when these are used in other places such as `constants.CacheCleanupInterval` or `interfaces.BlockchainProvider` it's really not clear that these are `router` related. I'd much prefer these be defined on a `router` package level. I'm assuming there was some sort of import cycle problem?

core/types/epoch.go
```diff
@@ -277,60 +279,22 @@ func (c *Committee) MemberByAddress(address common.Address) *CommitteeMember {
...
-	c.lock.Lock()    // Acquire write lock
-	defer c.lock.Unlock()
-
-	// Double-check if the value was initialized while waiting for the lock
-	if c.totalVotingPower == nil {
+	c.votingPowerOnce.Do(func() {
 		total := new(big.Int)
 		for _, m := range c.Members {
-			total.Add(total, m.VotingPower)
+			if m.VotingPower != nil {
```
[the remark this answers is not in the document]
lorenzo-dev1: Solution proposed in https://github.com/autonity/autonity/pull/1182

core/types/epoch.go
```diff
@@ -277,60 +279,22 @@ func (c *Committee) MemberByAddress(address common.Address) *CommitteeMember {
...
-			total.Add(total, m.VotingPower)
-		}
-		c.totalVotingPower = total
-	}
-
-	// return a copy of the cached value to prevent un-expected modification of the cached value.
-	return new(big.Int).Set(c.totalVotingPower)
-}*/
+func (c *Committee) Quorum() *big.Int {
+	return new(big.Int).Set(bft.Quorum(c.TotalVotingPower()))
```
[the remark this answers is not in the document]
lorenzo-dev1: Solution proposed in https://github.com/autonity/autonity/pull/1182

consensus/tendermint/core/committee/committee.go
```diff
@@ -128,20 +128,26 @@ func getMemberIndex(committee *types.Committee, memberAddr common.Address) int64
...
+	quorum           *big.Int
 	previousHeader   *types.Header
 	autonityContract *autonity.ProtocolContracts
 }
 
 func NewWeightedRandomSamplingCommittee(previousHeader *types.Header, committee *types.Committee, autonityContract *autonity.ProtocolContracts) *WeightedRandomSamplingCommittee {
 	return &WeightedRandomSamplingCommittee{
 		committee:        committee,
+		totalVotingPower: committee.TotalVotingPower(),
+		quorum:           new(big.Int).Set(bft.Quorum(committee.TotalVotingPower())),
```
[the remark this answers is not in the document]
lorenzo-dev1: solution proposed in https://github.com/autonity/autonity/pull/1182

ScottRMalley (changes requested):

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,226 @@
...
+// getBucketIndex calculates the bucket index for a given latency
+func (c *Clusters) getBucketIndex(latency uint) int {
+	if c.bucketSize < 1 {
+		return 0
+	}
+	bucketIdx := int(float64(latency-c.minLatency) / c.bucketSize)
```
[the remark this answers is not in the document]
vpiyush: this can not happen, only possible flow is if the self latency is used when calling getbuckeIndex, which we skip currently.

consensus/tendermint/router/network/network.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"math"
+	"strings"
+	"sync"
+
+	"github.com/autonity/autonity/common"
+	"github.com/autonity/autonity/log"
+)
+
+type Network struct {
```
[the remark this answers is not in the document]
vpiyush: It's not only for print function, It's there to maintain a single complete cluster object, this design saves from protecting individual fields of the clusters object. we can create a new object and swap this with old one which has only one contention point.

consensus/tendermint/router/network/network.go
```diff
@@ -0,0 +1,88 @@
...
+	latencyMap map[common.Address]uint,
+	self common.Address,
+) (Clusters, error) {
+	numClusters := int(math.Floor(math.Sqrt(float64(len(committee)))))
+	c, err := createClusters(committee, latencyMap, self, numClusters)
+	if err != nil {
+		return Clusters{}, err
+	}
+
+	remoteBuckets := c.ComputeLatencyBuckets()
```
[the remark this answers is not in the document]
vpiyush: this is updated now.

consensus/tendermint/router/selector/selector.go
```diff
@@ -0,0 +1,433 @@
...
+}
+
+func (s *selector) selectRemoteNodesByLatencySpread() []network.Node {
+	recipients := make([]network.Node, 0)
+	usedClusters := make(map[int]bool)
+
+	clusters := s.networkProvider.Clusters()
+
+	// Select one connected node from each remote cluster
+	for _, nodes := range clusters.BucketNodes() {
```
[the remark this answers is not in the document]
vpiyush: we do have fallbacks in each bucket node.

consensus/tendermint/router/selector/selector.go
```diff
@@ -0,0 +1,433 @@
...
+	for _, recipient := range recipients {
+		if _, ok := s.peerFinder.FindPeer(recipient); !ok {
+			return false
+		}
+	}
+	return true
+}
+
+func (s *selector) SelectPeers(committee *types.Committee, msg message.Msg, from common.Address) ([]common.Address, error) {
+	switch msg.Code() {
```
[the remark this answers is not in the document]
vpiyush: done

consensus/tendermint/router/constants/constants.go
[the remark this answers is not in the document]
vpiyush: > I'm not immediately seeing why this would need to be in a separate package. Wouldn't it make more sense if these constants were in the `router` package?

all the constants being used in router and subpackages of router are kept in one place intentionalky, generaly it's easier to understand the meaning and relate to others when done this way, some of them here are used in multiple sub-packages as well.

The second comment is not clear to me, But when it's defined under router it gives that it's related, how do you propose to refactor this?

consensus/tendermint/backend/aggregator.go
```diff
@@ -21,8 +22,9 @@ import (
 )
 
 const (
-	aggregationPeriod            = 150 * time.Millisecond
+	aggregationPeriod            = 30 * time.Millisecond
```
lorenzo-dev1, line 25: I think that this might significantly lower the performance of the aggregator... Lower number of votes will get verified together. Why reducing it?

consensus/tendermint/backend/aggregator_test.go
```diff
@@ -926,33 +925,7 @@ func TestAggregatorProcess(t *testing.T) {
 		// signature is invalid but proposal is created with `verified`=true, so it is considered valid
 		propose := makeBogusPropose(0, 1, 0)
 		proposeEvent := makeBogusEvent(propose)
-		a.processProposal(proposeEvent, func(_ message.Msg, _ chan<- error) interface{} { return struct{}{} })
-	})
-	t.Run("processProposal, invalid proposal is rejected", func(t *testing.T) {
```
lorenzo-dev1, line 931: Can be left for later, but this test should be moved in the backed_test.go file if we decide to go with the decision of verifying the proposal in the backend handler.

consensus/tendermint/router/constants/constants.go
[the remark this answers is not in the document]
ScottRMalley: So in general, when you are using these constants outside of the `router` package they are used like this:

```go
if numTargets > constants.ScaleThresholdForClustering {
    ...
}
```

This means from a readability perspective, they don't seem to be `router` related constants. You would have to scroll up to the import statement to see where the `constants` package is being imported from in order to realize that this is a router constant.

In addition, because you've created a separate subpackage for `constants` within the `router` package, it means we end up having to expose a bunch of constants that are not used outside of the `router` package.

For instance:
```go
constants.MaxLatencyCapFactor
constants.LatencyDataExpiry
constants.RetryLatencyTimeout
constants.CacheCleanupInterval
```

Are all constants that are exclusively used in the `router` package, but because they are now moved to a `constants` package we need to have them exported in order to use them. This makes it a lot easier for someone else to make a mistake when using constants in other packages, as these are then accessible. In addition, for external users trying to use autonity itself as a go dependency, we are now cluttering the exported namespaces.

I don't think this is generally good practice. Normally, I don't create a separate subpackage for every file in my feature. If we have a `router` package, I'd expect all files to be there, and only the necessary classes/structs/constants to be exported.

event/event.go
```diff
@@ -92,8 +93,13 @@ func (mux *TypeMux) Post(ev interface{}) error {
...
 	mux.mutex.RUnlock()
-	for _, sub := range subs {
-		sub.deliver(event)
+	for _, s := range subs {
+		sub := s
+		if len(subs) > 1 {
+			go sub.deliver(event) // temporary - don't block subscribers of same event
+		} else {
+			sub.deliver(event)
+		}
```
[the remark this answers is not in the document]
vpiyush: yes i have made that change

consensus/tendermint/backend/aggregator.go
```diff
@@ -563,13 +561,15 @@ func (a *aggregator) processBatches(batches [][]events.UnverifiedMessageEvent, e
 				aggregateVotes := message.AggregatePrevotesSimple(validVotes)
 				for _, aggregateVote := range aggregateVotes {
 					a.knownMessages.Add(aggregateVote.Hash(), true) // prevents processing of the same aggregate computed by another peer
-					go a.backend.Post(eventer(aggregateVote, nil))
+					a.backend.MessageToCore(eventer(aggregateVote, events.UnverifiedMessageEvent{Sender: a.backend.Address()}))
```
tbssajal, line 564: Don't we need to include the time as well? Because the `eventer` function will use that time.

tbssajal (changes requested):

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+		if selectedNodeIndex < len(clusterNodes)-1 {
+			// add next node as fallback if it exists
+			n := clusterNodes[selectedNodeIndex+1]
+			if n.Addr != c.self {
+				fallbacks = append(fallbacks, n)
+			}
+		}
+		return fallbacks
+	}
```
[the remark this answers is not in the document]
vpiyush: yes, no fallback is possible

consensus/tendermint/router/routing.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+	m.retryMu.Unlock()
+
+	m.refreshClustersLatencies(m.Latencies())
+	log.Debug("Router: latency measurement completed", "failed_nodes", len(failedNodes))
+	return nil
+}
+
+func (m *Router) retryLatency() error {
+	m.retryMu.Lock()
```
tbssajal, line 229: Why not `RLock()`?

consensus/tendermint/router/routing.go
```diff
@@ -0,0 +1,377 @@
...
+	m.retryMu.Unlock()
+
+	if updated {
+		m.latencyMu.Lock()
+		latMap := make(map[common.Address]uint, len(m.latestLatencies))
+		for addr, lat := range m.latestLatencies {
+			latMap[addr] = lat
+		}
+		m.latencyMu.Unlock()
+		m.refreshClustersLatencies(latMap)
```
tbssajal, line 272: why not use `m.Latencies()`?

consensus/tendermint/router/network/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+				c.maxLatency = latency
+			}
+		}
+		c.base[clusterID] = append(c.base[clusterID], Node{Addr: addr, Lat: latency, ClusterID: clusterID})
+		c.addressToCluster[addr] = clusterID
+	}
+	if c.ownClusterID == -1 {
+		return Clusters{}, errors.New("self address not in committee")
+	}
+	c.maxLatency = uint(float64(c.maxLatency) * constants.MaxLatencyCapFactor)
```
tbssajal, line 76: Just curious, what is the idea behind this? If we are trying to consider only 75% of them then maybe we could do the following instead:

```
c.maxLatency = c.minLatency + uint(float64(c.maxLatency - c.minLatency) * constants.MaxLatencyCapFactor)
```

consensus/tendermint/router/routing.go
```diff
@@ -0,0 +1,377 @@
...
+	m.committee = result
+
+	m.retryMu.Lock()
+	m.nodesToRetry = make(map[common.Address]struct{})
+	for _, addr := range result {
+		if addr != m.self {
+			m.nodesToRetry[addr] = struct{}{}
+		}
+	}
+	m.retryMu.Unlock()
```
tbssajal, line 361: This part seems redundant as there is no separate loop for retries. Basically after exiting `updateCommittee` function, the cluster gets updated immediately with old latency map. And then `m.measureLatency()` is called immediately which will update the latency map, `nodesToRetry` map and also the cluster.

consensus/tendermint/router/routing.go
```diff
@@ -0,0 +1,377 @@
...
+				log.Info("Router: clustering not needed, skipping measurement")
+				continue
+			}
+			m.updateCommittee(epoch)
+			nw, err := network.New(m.committee, m.Latencies(), m.self)
+			if err != nil {
+				log.Error("Router: failed to create network", "err", err)
+				continue
+			}
+			m.updateNetwork(nw)
```
tbssajal, line 339: I think we want the following instead. Because even if we get error, we might still want to call `m.measureLatency()` in the following line.

```suggestion
			m.refreshClustersLatencies(m.Latencies())
```

consensus/tendermint/router/latency/latency.go
```diff
@@ -0,0 +1,120 @@
...
+			continue
+		}
+		if _, ok := f.peerFinder.FindPeer(member); !ok {
+			log.Debug("Node not connected, skipping ping", "peer", member.Hex())
+			pingTargets[i] = ping.Target{}
+			failedNodes = append(failedNodes, member)
+			latency[member] = constants.DefaultLatency
+			continue
+		}
+		if memberNode, ok := enodeByAddress(committeeEnodes, member); ok {
```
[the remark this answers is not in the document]
tbssajal: Addressed in #1191

consensus/tendermint/router/routing.go
vpiyush: because nodesToRetry is never read concurrently

consensus/tendermint/router/routing.go
vpiyush: what do you mean? renaming?

consensus/tendermint/router/selector/selector.go
```diff
@@ -0,0 +1,423 @@
...
+			if usedClusters[node.ClusterID] || node.ClusterID == clusters.ID() {
+				continue
+			}
+			if _, ok := s.peerFinder.FindPeer(node.Addr); ok {
+				recipients = append(recipients, node)
+				usedClusters[node.ClusterID] = true
+				break // Take the first connected node (primary or fallback)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
tbssajal: I see, it's supposed to spread the proposal further.

consensus/tendermint/router/routing.go
vpiyush: agree removed

consensus/tendermint/backend/aggregator.go
```diff
@@ -714,22 +755,43 @@ func (a *aggregator) loop(ctx context.Context) {
...
+
 	// channel where the aggregator will receive msgs from the backend handlers
 	messageCh := a.backend.MessageCh()
+	type evMeta struct {
+		dur   time.Duration
+		count int
+	}
+	eventTracker := make(map[string]evMeta)
 
+	updateEventMeta := func(eventType string, duration time.Duration, _ int) {
```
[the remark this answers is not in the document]
vpiyush: this was temporary for debugging, would be removed

consensus/tendermint/core/message/message.go
```diff
@@ -611,8 +641,8 @@ func AggregateVotes[E Prevote | Precommit](votes []Vote) *E {
 	}
 
 	// use votes[0] as a set representative
-	representative := votes[0]
 
+	representative := deterministicRepresentative(votes)
```
tbssajal, line 645: I think we should make self address the originator and remove originator from the message hash. The way originator is picked here, it creates confusion in the network, specifically at `firstRelayerOriginCluster` layer. Although with current approach or [the one suggested](https://github.com/autonity/autonity/pull/1166#issuecomment-2996297510) tackles this confusion pretty well, it seems error prone. Because at first glance, originator seems like the one who created the msg.

consensus/tendermint/core/handler.go
```diff
@@ -242,9 +259,11 @@ eventLoop:
 					}
 
 					// gossip message. We should arrive here only if we did not already gossip a complex aggregate
-					go c.backend.Gossip(c.CommitteeSet().Committee(), msg)
-					recordMessageProcessingTime(msg.Code(), start)
+					//todo(fix): no gossip for future round messages, the current gossip relies on clustering, which relies on gossiper being the
```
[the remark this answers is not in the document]
vpiyush: updated this, gossiping these messages now

consensus/tendermint/router/network/cluster.go
vpiyush: added comments

consensus/tendermint/backend/handler.go
```diff
@@ -216,6 +218,15 @@ func (sb *Backend) handleDecodedMsg(msg message.Msg, errCh chan<- error, sender
...
 			return true, ErrJailed
 		}
+
+		// early Validation for proposals, so we can forward them to other peers
+		if err := msg.Validate(); err != nil {
+			return true, err
+		}
+		// structured relaying happens after the pre-validation, only unknown msg is relayed.
+		if sb.core.Height().Uint64() == msg.H() { // same height messages early forward
+			go sb.router.Forward(committee, msg, sender, nil)
```
tbssajal, line 228: Shouldn't we add this msg in `sb.knownMessages` before forwarding?

consensus/tendermint/backend/gossiper.go
```diff
@@ -49,31 +67,29 @@ func (g *Gossiper) UpdateStopChannel(stopCh chan struct{}) {
...
-				// This peer had this event, skip it
-				continue
-			}
-			p.Cache().Add(hash, true)
-			go p.SendRaw(code, msg.Payload()) //nolint
-		}
+func (g *Gossiper) Gossip(committee *types.Committee, msg message.Msg) {
+	// self message caching
+	if !g.knownMessages.Contains(msg.Hash()) {
+		g.knownMessages.Add(msg.Hash(), true)
```
tbssajal, line 90: Maybe we should move the caching part in router instead? Beause there are other places where we call `router.Forward` and it seems those msges are not cached.

vpiyush: Messages which are created by node itself are only needed to be cached, forwarding is done for the received messages only so there no need for caching for them. Currently that is the only difference between gossip and forwarding.

consensus/tendermint/backend/handler.go
vpiyush: that's  done earlier in the flow, handleConsensusMessage.

consensus/tendermint/core/handler.go
```diff
@@ -161,66 +161,100 @@ eventLoop:
...
+						break
+					}
 
-					// gossip message. We should arrive here only if we did not already gossip a complex aggregate
-					go c.backend.Gossip(c.CommitteeSet().Committee(), msg)
-					recordMessageProcessingTime(msg.Code(), start)
+					if err != nil && errors.Is(err, constants.ErrOldRoundMessage) {
+						go c.backend.SlowGossip(c.CommitteeSet().Committee(), msg)
+					} else {
+						go c.backend.Router().Forward(c.CommitteeSet().Committee(), msg, e.Sender(), nil)
```
tbssajal, line 241: It seems like a possibility that self created msg end up here. If it's possible, then it won't be cached.

tbssajal: The scenario I am thinking is that we received enough votes early but received the proposal late, so we already go quorum before our own vote is generated.

consensus/tendermint/router/interfaces/interfaces.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/autonity/autonity/core"
+	"github.com/autonity/autonity/core/types"
+	"github.com/autonity/autonity/event"
+	"github.com/autonity/autonity/p2p/enode"
+)
+
+type NetworkProvider interface {
+	Clusters() network.Clusters
+	UpdateClusters(network.Clusters)
+}
```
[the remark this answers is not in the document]
vpiyush: renamed package to cluster and network to manager.

consensus/tendermint/router/network/network.go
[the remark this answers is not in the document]
vpiyush: renamed this

consensus/tendermint/router/constants/constants.go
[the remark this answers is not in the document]
ScottRMalley: Yeah I mean I still find this a strange pattern. You can have a `router/cache.go` without having a whole new package `router/cache/cache.go`, which makes it clear that this is a cache for routing. We don't use this pattern elsewhere in the project, for instance it's not as if we have `backend/aggregator/aggregator.go`, we just put it in the backend package as `backend/aggregator.go`.

I noticed you moved the constants into the `router` package, but removed them from a separate file. I never had an issue with them being a separate file, just a separate package. I'm not going to die on this hill, but I really don't see the usefulness of the structure you've set up here. I think it only adds confusion.

tbssajal (changes requested):

consensus/tendermint/core/message/message.go
vpiyush: originator related discussion is concluded in the separate PR.

ScottRMalley (approved):
CI is green. I think we are good to go.

consensus/tendermint/backend/aggregator.go
```diff
@@ -728,6 +771,7 @@ loop:
 				BackendAggregatorTransitBg.Add(time.Since(event.Posted).Nanoseconds())
 			}
 			a.handleEvent(event)
+			//Note: core events are not sent to the aggregator anymore, code remains here for later evaluation
```
Jason-Zhangxin-Chen, line 774: QQ: Is the aggregation without taking core state changes (round change, power change, etc..) better for performance?

consensus/tendermint/backend/handler.go
```diff
@@ -235,6 +246,7 @@ func (sb *Backend) handleDecodedMsg(msg message.Msg, errCh chan<- error, sender
 		sb.logger.Crit("Tendermint backend processing unknown message")
 	}
 
+	//todo: proposal is already verified, post directly to the core
```
Jason-Zhangxin-Chen, line 249: Is this todo still valid? As the aggregator will forward the proposals, votes both to core and FD.

consensus/tendermint/core/core.go
```diff
@@ -353,7 +371,8 @@ func (c *Core) processFuture(previousRound int64, currentRound int64) {
 
 	for r := previousRound + 1; r <= currentRound; r++ {
 		for _, msg := range c.futureRound[r] {
-			go c.SendEvent(backlogMessageEvent{
+			// only to core, should we send to FD ??
```
Jason-Zhangxin-Chen, line 374: We don't send it to FD anymore since the aggregator did it before right after the MsgToCore forwarding.

consensus/tendermint/core/handler.go
```diff
@@ -352,20 +369,15 @@ eventLoop:
 func (c *Core) SendEvent(ev any) {
 	switch ev := ev.(type) {
 	case events.CoreEvent:
-		c.eventCh <- ev
+		// todo: temporary
```
Jason-Zhangxin-Chen, line 372: If the aggregation has better performance without this state changes from core, then we can remove this.

consensus/tendermint/router/selector/selector.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// determines the routing base of a message
+func routingBase(committee *types.Committee, msg message.Msg) common.Address {
+	var rb common.Address
+	switch m := msg.(type) {
+	case *message.Propose:
+		rb = m.Signer()
+	case *message.Prevote:
+		routingBaseIndex := m.Signers().LeftmostSigner()
```
Jason-Zhangxin-Chen, line 94: Is the aggregator always do this intended assembling to make it deterministic?

consensus/tendermint/router/selector/selector.go
```diff
@@ -0,0 +1,431 @@
...
+func (s *selector) selectBucketBasedNodes(clusters cluster.Clusters, committee *types.Committee, senderType SenderType, ownClusterID int, isProposal bool) []cluster.Node {
+	var recipients []cluster.Node
+	var minNodes, lowLatencyNodes int
+
+	switch senderType {
+	case originator:
+		// additional nodes
+		localNodes := len(clusters.Base()[ownClusterID])
+		minNodes = int(float64(localNodes) * (2.0 / 3.0)) // assuming all cluster of same size, send to 2/3 of cluser size
+		lowLatencyNodes = 4
```
Jason-Zhangxin-Chen, line 217: Name lowLatencyNodes with a constant.

consensus/tendermint/router/routing.go
```diff
@@ -0,0 +1,412 @@
...
+		log.Debug("Router: peers not found", "len", len(lostPeers), "peers", lostPeers)
+	}
+}
+
+func (m *Router) Start(ctx context.Context, chain interfaces.BlockChainProvider) {
+	log.Info("Router: starting")
+
+	curEpoch, err := chain.LatestEpoch()
+	if err != nil {
+		log.Error("Error fetching latest epoch", "err", err)
```
lorenzo-dev1, line 201: @vpiyush do you think it could be better to panic here? I guess this should never happen and if it happens it kind of messes up the whole clustering logic right?

lorenzo-dev1 (approved):
LGTM

consensus/tendermint/core/handler.go
vpiyush: yes these gets clean up in another PR, but we are not sending anything to core for now.

consensus/tendermint/router/routing.go
vpiyush: Discussed offline. We addrese it in remove-simple-aggregate PR

eth/backend.go
```diff
@@ -630,12 +632,57 @@ func (s *Ethereum) validatorController() {
...
+				}
+			}
+		}()
+	}
+
 	committee := epoch.Committee
 	if committee.MemberByAddress(s.address) != nil {
 		updateConsensusEnodes(currentHead)
-		s.miner.Start()
+		//todo: the minor control should move to acn server
```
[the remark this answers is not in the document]
vpiyush: sicne acn is also running the committeeWatcher. this  makes more sense in the ACN flow

consensus/tendermint/backend/backend.go
```diff
@@ -57,24 +59,24 @@ func New(
...
-		isHeightExpired:    isHeightExpired,
+		database:            database,
+		eventMux:            event.NewTypeMuxSilent(evMux, log),
+		nodeKey:             nodeKey,
+		consensusKey:        consensusKey,
+		address:             crypto.PubkeyToAddress(nodeKey.PublicKey),
+		logger:              log,
+		knownMessages:       knownMessages,
+		vmConfig:            vmConfig,
+		MsgStore:            ms, //TODO: we use this only in tests, to easily reach the msg store when having a reference to the backend. It would be better to just have the `accountability` module as a part of the backend object.
```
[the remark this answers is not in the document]
vpiyush: this is getting refactore in aggregation_optimizations.

PR #1168, opened 2025-05-21
move endpoint verification during handshake

vpiyush:
Verifying consensus endpoint is moved to handshake phase. since handshake is done in a separate goroutine this would unblock the server loop. Allowing us to process large number of connection requests in shorter time frame.


tbssajal (approved):
LGTM

lorenzo-dev1 (approved):
LGTM

PR #1170, opened 2025-05-22
ASM - Fixed Genesis Prices

ScottRMalley:
As discussed, this is an ASM update to take into account that there won't be an NTN-USD or NTN-ATN market at genesis.

This fix allows the specification of default NTN-ATN and NTN-USD prices in the genesis config. These will be used by the Stabilization contract as the prices used for deposit/withdraw/borrow until the operator calls the `removeFixedGenesisPrices` function to set these prices to the market rate once oracle reports are ready.


tbssajal (changes requested):

yazzaoui:
I think it wouldn't be a bad idea to have a generalized "fixed price" mode which by default would be toggled on at genesis but could be later on disabled and re-enabled on operator demand. That might come perhaps handy in the future. All we need is some additional operator setters for the fixed prices.

autonity/solidity/contracts/asm/Stabilization.sol
```diff
@@ -679,10 +697,16 @@ contract Stabilization is IStabilization {
     /// and converts it to ACU
     /// @return price Price of Collateral Token in ACU
     function collateralPriceACU() public view returns (uint256) {
-        IOracle.RoundData memory data = _oracle.latestRoundData(StabilizationMath.NTN_USD_SYMBOL);
-        if (data.price <= 0) revert InvalidPrice();
+        uint256 ntnUsdPrice;
+        if (_fixedGenesisPrices) {
+            ntnUsdPrice = _config.defaultNTNATNPrice;
```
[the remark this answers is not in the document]
ScottRMalley: Wow, good catch. Rookie mistake here.

tbssajal (approved):
LGTM

autonity/solidity/contracts/asm/Stabilization.sol
```diff
@@ -531,6 +530,36 @@ contract Stabilization is IStabilization {
         emit IConfigEvents.ConfigUpdateUint("minCollateralizationRatio", _minCollateralizationRatio.value(), newMinCollateralizationRatio);
     }
 
+    /**
+    * @notice Set the default NTN-USD price for use when fixed prices are enabled.
```
cmjc, line 534: NTN-ATN price

autonity/solidity/contracts/asm/Stabilization.sol
```diff
@@ -531,6 +530,36 @@ contract Stabilization is IStabilization {
...
+    * @param defaultNTNATNPrice The new default NTN-USD price
```
cmjc, line 535: The new default NTN-ATN price

autonity/solidity/contracts/asm/Stabilization.sol
```diff
@@ -531,6 +530,36 @@ contract Stabilization is IStabilization {
...
+    * @dev Restricted to the operator.
+    */
+    function setDefaultNTNUSDPrice(uint256 defaultNTNUSDPrice) external onlyOperator {
+        emit IConfigEvents.ConfigUpdateUint("defaultNTNUSDPrice", _config.defaultNTNUSDPrice, defaultNTNUSDPrice);
+        _config.defaultNTNUSDPrice = defaultNTNUSDPrice;
+    }
+
+    /**
+    * @notice Set the default ACU price for use when fixed prices are enabled.
+    * @param defaultACUUSDPrice The new default ACU price
```
cmjc, line 555: The new default ACU-USD price

tbssajal:
As the latest changes are about the operator control, we can also consider toggling the restriction to borrow and deposit.

tbssajal:
> As the latest changes are about the operator control, we can also consider toggling the restriction to borrow and deposit.

I think it will not be fair to someone who already deposited or borrowed. To make it fair it would mean to restrict only the new deposits.

tbssajal:
> > As the latest changes are about the operator control, we can also consider toggling the restriction to borrow and deposit.
> 
> I think it will not be fair to someone who already deposited or borrowed. To make it fair it would mean to restrict only to open the new accounts.

And this will require a lot of changes.

ScottRMalley:
> Everything looks good to me, however I have a doubt. In the Auctioneer we are fetching the "real" price of NTNATN from the Oracle even if the price is fixed in Stabilization. Can that be a problem?

So, the issue is that the Auctioneer requires a previous NTN/ATN price when placing a bid, as the auction price function is dependent upon the time since the position became liquidatable.

In general, the bidder has to specify a specific oracle round where the position became liquidatable. The position itself must be liquidatable both in the specified round and the current round in order to be liquidated. Under a fixed price regime, there could be a scenario where the bidder is able to specify an earlier round, even though the cdp just became liquidatable.

This scenario then falls back to the default case where the account's full collateral is offered as reward to the liquidator attempting to pay off the debt. It does not allow a liquidator to liquidate an account that is not currently liquidatable under the fixed prices.

The auctioneer is a contract that attempts to get a better price for a liquidatable account's collateral in the case of a liquidation, however it should be clear to all CDP holders that when their account becomes liquidatable, there is always the chance that their full collateral is liquidated.

I agree this is a sub-optimal situation, however, considering we don't plan on using fixed prices outside of genesis, and we don't plan on opening up CDP creation to the public until after the ASM V3 is implemented, I think it's a reasonable trade off to keep here.

autonity/solidity/contracts/asm/Stabilization.sol
ScottRMalley: Updated.

lorenzo-dev1:
> > Everything looks good to me, however I have a doubt. In the Auctioneer we are fetching the "real" price of NTNATN from the Oracle even if the price is fixed in Stabilization. Can that be a problem?
> 
> So, the issue is that the Auctioneer requires a previous NTN/ATN price when placing a bid, as the auction price function is dependent upon the time since the position became liquidatable.
> 
> In general, the bidder has to specify a specific oracle round where the position became liquidatable. The position itself must be liquidatable both in the specified round and the current round in order to be liquidated. Under a fixed price regime, there could be a scenario where the bidder is able to specify an earlier round, even though the cdp just became liquidatable.
> 
> This scenario then falls back to the default case where the account's full collateral is offered as reward to the liquidator attempting to pay off the debt. It does not allow a liquidator to liquidate an account that is not currently liquidatable under the fixed prices.
> 
> The auctioneer is a contract that attempts to get a better price for a liquidatable account's collateral in the case of a liquidation, however it should be clear to all CDP holders that when their account becomes liquidatable, there is always the chance that their full collateral is liquidated.
> 
> I agree this is a sub-optimal situation, however, considering we don't plan on using fixed prices outside of genesis, and we don't plan on opening up CDP creation to the public until after the ASM V3 is implemented, I think it's a reasonable trade off to keep here.

Makes sense. Moreover when having a fixed price, I guess that CDP positions can become liquidatable only if the user takes action (e.g. borrow more or withdraw some collateral)

autonity/solidity/contracts/asm/Auctioneer.sol
```diff
@@ -178,12 +178,19 @@ contract Auctioneer is IAuctioneer, IConfigEvents, ReentrancyGuard {
...
+        if (_pendingAllocatedInterest < config.interestAuctionThreshold) {
+            return; // not enough interest to start an auction
         }
+        // check if the NTN_SYMBOL is included in the current oracle round
+        uint256 startRound = _oracle.getRound() - 1;
+        IOracle.RoundData memory latestRound = _oracle.getRoundData(startRound, StabilizationMath.NTN_SYMBOL);
+        if (latestRound.price == 0 || latestRound.timestamp == 0) {
+            // no NTN_SYMBOL in the latest round, so we won't start an interest auction until the next repayment
+            return;
+        }
```
lorenzo-dev1, line 190: Maybe it is worth to emit an event if we end up here

lorenzo-dev1 (approved):
LGTM

tbssajal:
> @ScottRMalley
> 
> I was thinking about another possible edge case when using fixed prices after genesis (and therefore without CDP restrictions).
> 
> Supposing that the NTNATN fixed price set by the operator is > than the actual NTNATN market price (the one reported by Oracles and used by the Auctioneer), couldn't this trigger a sort of arbitrage cycle?
> 
> Like a user could buy ATN interest at a low price on the Auctioneer, repay his loans and the higher fixed rate, and then re-buy his own interest at the lower price on the Auctioneer, and so on?

And if the NTNATN fixed price set by the operator is < than the actual NTNATN market price, then the opposite cycle could happen, right?

lorenzo-dev1:
> > @ScottRMalley
> > I was thinking about another possible edge case when using fixed prices after genesis (and therefore without CDP restrictions).
> > Supposing that the NTNATN fixed price set by the operator is > than the actual NTNATN market price (the one reported by Oracles and used by the Auctioneer), couldn't this trigger a sort of arbitrage cycle?
> > Like a user could buy ATN interest at a low price on the Auctioneer, repay his loans and the higher fixed rate, and then re-buy his own interest at the lower price on the Auctioneer, and so on?
> 
> And if the NTNATN fixed price set by the operator is < than the actual NTNATN market price, then the opposite cycle could happen, right?

Maybe, but I am a bit struggling to see how it could be exploited in that way

ScottRMalley:
> @ScottRMalley
> 
> I was thinking about another possible edge case when using fixed prices after genesis (and therefore without CDP restrictions).
> 
> Supposing that the NTNATN fixed price set by the operator is > than the actual NTNATN market price (the one reported by Oracles and used by the Auctioneer), couldn't this trigger a sort of arbitrage cycle?
> 
> Like a user could buy ATN interest at a low price on the Auctioneer, repay his loans and the higher fixed rate, and then re-buy his own interest at the lower price on the Auctioneer, and so on?

The functioning of the ASM in general is dependent on arbitrage. If the market rate is different than the rate used by the Stabilization contract, there will always be an arbitrage opportunity. You've pointed out an arbitrage opportunity between the auctioneer and the Stabilization contract, but in general, the biggest arbitrage opportunity will be between the open market and the Stabilization contract.

If we set fixed prices that are different than the market rate, it doesn't matter whether the user buys from the auctioneer at the market rate and repays their loan at the higher rate. They can always buy from the open market and do the same thing. In fact, when we turn on fixed rates, the arbitrage from the Stabilization contract will actually drive market rates towards the fixed rate.

In general, this is an unavoidable risk we have by turning on fixed prices. We should only turn on fixed prices if the risk of haulting Stabilization mechanisms due to oracle failure is greater than the price impact of the fixed prices. It's really a last resort emergency valve rather than something that should occur in normal operation.

ScottRMalley:
@tbssajal @lorenzo-dev1 

I rebased the PR and I did end up changing the Auctioneer to fetch the collateral price from the Stabilization contract instead of the oracle. This works in the case of the interest auction because we can record the start price for interest auctions at the moment they are initiated.

It's not possible for debt auctions as there is no transaction that starts a debt auction, so we rely on the timestamp of the oracle round.

I think this should give some more consistent results for interest auctions, and at least plug one of the arbitrage holes created by fixed prices. Let me know what you guys think.

PR #1171, opened 2025-05-22
ci, install docker test dep with --fix-missing.

Jason-Zhangxin-Chen:
This PR addresses the following issues in CI:

1. Resolved dependency issues caused by a recent update in a third-party repository.
2. Adjusted the concurrency level of Docker e2e tests to mitigate the Docker Hub pull rate limit error. As per the Docker Hub usage guidelines (https://docs.docker.com/docker-hub/usage/), unauthenticated users are limited to 100 pulls per IP every 6 hours. Although the limit resets every 6 hours, we occasionally encounter this restriction, which necessitated this adjustment. So, now we narrow down to 1/3 docker pulls to experiment the CI workflow.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Generate test IDs
         id: set-ids
         run: |
-          echo "ids=[$(seq -s ',' 0 26)]" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


lorenzo-dev1 (approved):
LGTM

ScottRMalley (approved):
LGTM as well.

PR #1177, opened 2025-06-02
Autonity Interface

### LINKED ISSUE #1176: Update IAutonity and Move necessary structs there
It's better for IAutonity to have all or necessary public interfaces that are useful for smart contract development. For example, `bond`, `unbond` etc methods. Some structures might also be needed to move from `Autonity.sol` to `IAutonity.sol`. Lets decide it here. I think staking related structs and `Validator` struct should be moved.

tbssajal:
Closes #1176

- Moved all external functions to `IAutonity` except functions with `onlyOperator` or `onlyProtocol` modifier
- Moved all struct to `IAutonity`
- All protocol contracts use `IAutonity` interface instead of `Autonity` contract.


autonity/solidity/contracts/Autonity.sol
```diff
@@ -111,75 +54,6 @@ contract Autonity is IAutonity, IERC20, ReentrancyGuard, ScheduleController, Upg
...
-    // part of the config which the golang client is keeping track of
-    struct ClientAwareConfig {
-        uint256 epochPeriod;
-        uint256 blockPeriod;
-        uint256 gasLimit;
-        Accountability accountability;
-        Eip1559 eip1559;
-    }
 
     struct FinalizeResult {
```
lorenzo-dev1, line 58: Is the thinking that this can stay here since the Finalize is only called by the client itself?

tbssajal: Yes. Also `finalize()` is not available in `IAutonity` for the same reason.

cmjc (commented):
lgtm

ScottRMalley (approved):
LGTM

PR #1181, opened 2025-06-11
Trigger ask sync

Jason-Zhangxin-Chen:
This PR refines the triggering of ask-sync process by respecting tendermint round and step timeouts.

### DIFF
--- a/consensus/tendermint/core/core.go
+++ b/consensus/tendermint/core/core.go
@@ -71,18 +71,14 @@ func (c *Core) SetDefaultHandlers() {
 }
 
 type SyncState struct {
-	// todo: remove this outOfSync as it is not used.
 	outOfSync        atomic.Bool
 	lastValidMsgTime atomic.Int64
 	timeOut          atomic.Int64
 }
 
+// todo: remove this outOfSync state, as it was not used at all.
 func (s *SyncState) SetOutOfSync(val bool) {
 	s.outOfSync.Store(val)
-}
-
-func (s *SyncState) IsOutOfSync() bool {
-	return s.outOfSync.Load()
 }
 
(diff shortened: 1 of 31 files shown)


consensus/tendermint/core/core.go
```diff
@@ -47,7 +48,10 @@ func New(backend interfaces.Backend, services *interfaces.Services, address comm
 		stepChange:             time.Now(),
 		noGossip:               noGossip,
 		eventCh:                make(chan events.CoreEvent, EventQueueSize),
+		syncState:              &SyncState{},
 	}
+	c.syncState.SetLastValidMsgTime(time.Now())
+	c.syncState.SetSyncTimeOut(constants.DefaultSyncTimeout)
```
lorenzo-dev1, line 54: I am not sure whether this is the right place to initialize the lastValidMsgTime. I think there could be edge cases where a node starts not as a validator, so Core gets created but it is not running. Then it becomes a validator and it has a very old time in the `syncState`. Wouldn't it make more sense to set it on `Core.Start`? 

Or maybe even better is to set it at every round start... it doesn't mean we received a message but a round change still signals the network is live and it makes sense to wait at least for the duration of the proposer timeout before asking sync.

lorenzo-dev1:
@Jason-Zhangxin-Chen I opened https://github.com/autonity/autonity/pull/1183 as I had some comments, but I thought it would be faster if I just propose a solution

lorenzo-dev1:
More changes in https://github.com/autonity/autonity/pull/1185 @Jason-Zhangxin-Chen

lorenzo-dev1 (approved):
LGTM

vpiyush (approved):
LGTM

PR #1184, opened 2025-06-12
Update IConfigEvents.sol - add missing NatSpec

cmjc:
Edit to add missing natspec @param line:

>    * @param appliesAtHeight block at which the change will apply

Is:

```
    /**
    * @notice Emitted after updating config parameter of type boolean
    * @param name configuration name
    * @param oldValue old value of configuration
    * @param newValue new value of configuration
    */
    event ConfigUpdateBool(string name, bool oldValue, bool newValue, uint256 appliesAtHeight);
`


lorenzo-dev1 (approved):
LGTM

PR #1185, opened 2025-06-12
Trigger ask sync improvement 2

lorenzo-dev1:
Additional changes I wanted to insert in https://github.com/autonity/autonity/pull/1183 but it got merged prematurely.

Changes:
- re-introduce retries on first askSync to ensure feature parity with 2c819001db4f95d2807f987a737a1a60a08ea9f3, however by implementing the retries in the `livenessTrackerLoop` on first sync only.
- do not update the liveness timer if we receive a redundant message (i.e. a message that doesn't contribute any new voting power to Core). This is because it is not a signal of network liveness. For example we could continue receiving votes aggregated in different ways from members of  our own cluster, but the cluster itself is isolated from the rest of the network. In that case the right action to take is to askSync to everyone to try to advance the chain.
- gossip also backlogged old round messages (can happen if the local node is skipping multiple rounds)- 
- fix tests

### DIFF
--- a/accounts/abi/bind/backend_mock.go
+++ b/accounts/abi/bind/backend_mock.go
@@ -24,6 +24,7 @@ import (
 type MockContractCaller struct {
 	ctrl     *gomock.Controller
 	recorder *MockContractCallerMockRecorder
+	isgomock struct{}
 }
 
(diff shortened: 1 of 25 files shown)


Jason-Zhangxin-Chen (approved):
LGTM.

PR #1187, opened 2025-06-16
Fix contract UX issues

lorenzo-dev1:
This PR fixes some "UX issues" of the protocol contracts. By "UX issues" I mean that the protocol was working correctly, however it was difficult to interpret what was going on for an external user (e.g. a delegator wanting to know how is stake is performing).

More specifically it solves the following:
- round down the delegated rewards instead of the self rewards. This prevents the creation of artificially high conversion ratios due to accumulation of 1 wei per epoch of autobonded delegated stake.
- store the NTN:LNTN conversion ratio for each validator in the validator struct. This allows for clear computation of delegated stake APY from a delegator's perspective.
- Add some more information to the `Rewarded` event, to allow for easy computation of PAS stake APY.


autonity/solidity/contracts/interfaces/IAutonity.sol
```diff
@@ -48,6 +48,7 @@ interface IAutonity is IERC20, IScheduleController {
         uint256 selfUnbondingStakeLocked;
         ILiquid liquidStateContract;
         uint256 liquidSupply;
+        uint256 conversionRatio;
```
tbssajal, line 51: Need to change the offset constants in core/vm/contracts.go

autonity/solidity/contracts/Autonity.sol
```diff
@@ -279,6 +279,7 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
             0,                       // self unbonding stake locked
             ILiquid(address(0)), // liquid token contract
             0,                       // liquid token supply
+            1,                       // conversion ratio
```
tbssajal, line 282: We should have some precision for this I think.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1740,6 +1754,7 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
             uint256 _delegatedStake = _validator.bondedStake - _validator.selfBondedStake;
             if (_delegatedStake == 0) {
                 _liquidAmount = _bonding.amount;
+                _validator.conversionRatio=1;
```
tbssajal, line 1757: Should we do this in `_applyUnbonding` instead? Otherwise some validator with 0 liquid supply can have greater than or less than 1 conversion ration.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -918,6 +920,10 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
         // update the validator struct and send the slashed funds to the autonity treasury
         accounts[config.policy.treasuryAccount] += slashingAmount;
         validators[_nodeAddress] = _slashedVal;
+
+        if(validators[_nodeAddress].liquidSupply != 0) {
```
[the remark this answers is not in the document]
lorenzo-dev1: good point, added

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: Yeah that's what @fjribi and me decided. In any case the value of `conversionRatio` should be interpreted in conjuction with the `liquidSupply`. In general if `liquidSupply == 0`, whatever the conversion ratio value is stored, it should be interpreted as invalid. Then when a bonding comes, the ratio will be restored to 1:1

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1528,20 +1537,25 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
...
-                uint256 _atnDelegationReward = _atnReward - _atnSelfReward;
                 if (_atnDelegationReward > 0 || _ntnDelegationReward > 0) {
                     _transfer(address(this), address(_val.liquidStateContract), _ntnDelegationReward);
                     _val.liquidStateContract.redistribute{value: _atnDelegationReward}(accounts[address(_val.liquidStateContract)]);
                 }
+
+                // update historical conversion ratio
+                if(_val.liquidSupply != 0) {
+                    _val.conversionRatio = _delegatedStake / _val.liquidSupply;
+                }
```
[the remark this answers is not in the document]
tbssajal: So for the current implementation it's fine. But because autobond function provide option to update stakes for both self bonding and delegation, it might be error prone to not update the auto bond function.

autonity/solidity/contracts/Autonity.sol
lorenzo-dev1: Yeah good point I need to add that

lorenzo-dev1: added scaling factor with 4 decimals in https://github.com/autonity/autonity/pull/1187/commits/d61eec314a3a7daba09282d2ee5bb3b0e0be667a

fjribi: For future DeFi use cases, scale (i.e., the number of decimal places) can be important, but when calculating APY, it is the precision (i.e., the total number of significant digits) that matters most.

For example, if the conversion ratio becomes as small as 0.0001, you need to detect relative changes as small as 0.01% (1 basis point). However, setting the scale to only 4 decimal places would not even allow detection of a 10% performance change in such cases.

Therefore, to ensure at least 1 bips resolution in all cases, the key constraint is to maintain a total precision that includes at least 4 significant digits beyond the magnitude of the value being represented.

yazzaoui (approved):
LGTM

PR #1191, opened 2025-06-21
Improve Caching

tbssajal:
This PR possibly improves the caching by reducing the total amount of entry with the following observations:

- Returned set of selected recipients from function `selectBucketBasedNodes` is same for `prevote` and `precommit`, so we can use the same cache key for `prevote` and `precommit`.
- Two recipient set returned from function `selectBucketBasedNodes` for same `senderType` and same msg code, but for different sender, will differ by at most one node. If we ignore the sender from the function `selectBucketBasedNodes` and the cache key, the cache entry will be reduced. Then we just exclude the sender from the cache entry to build the list of recipients. This might exclude the cluster of the sender from the recipient list but it should not be a problem as the sender will broadcast to his own cluster.


tbssajal:
> Caching improvements are good. The latency stuff is bit debatable. generally map creation is expensive operation. also currently we endup generating this map even if those peers are not yet connected. The better way to address this is to create a map on committee change. you can keep the copy of this committeSet in Router and set it on core committee set.

How about we handle this a bit differently and completely avoid creating map then? The idea is to sort both `validators` input of the `Fetch` function and committee enodes from `peerFinder.CommitteeEnodes()` by address. Before entering the loop in the fetch function. Then we can find the endoe with iteration, but the total complexity of the loop will be `O(committeeSize)`. @vpiyush

PR #1193, opened 2025-06-25
Move originator outside of the hash 

lorenzo-dev1:


ScottRMalley (commented):
You have some linting errors, and I had one small comment, but otherwise it's looking good.

consensus/tendermint/router/selector/selector.go
```diff
@@ -84,18 +84,40 @@ func (s *selector) SelectPeers(committee *types.Committee, msg message.Msg, from
...
+	return rb
+}
+
 func (s *selector) selectPeersWithBuckets(committee *types.Committee, msg message.Msg, from common.Address, isProposal bool) ([]common.Address, error) {
 	clusters := s.clustersProvider.Clusters()
 	if len(clusters.Base()) == 0 {
 		return []common.Address{}, errors.New("no clusters")
 	}
 
+	// TODO: make sure committee is always the committee related to msg, otherwise indexes might be wrong
```
tbssajal, line 113: Is it possible to always enforce this? When we change epoch, old height msges will have different committee.

consensus/tendermint/router/selector/selector.go
```diff
@@ -84,18 +84,40 @@ func (s *selector) SelectPeers(committee *types.Committee, msg message.Msg, from
...
+	case *message.Propose:
+		rb = m.Signer()
+	case *message.LightProposal:
+		rb = m.Signer() // TODO: maybe need to panic here, lp is not gossiped
+	case *message.Prevote:
+		routingBaseIndex := m.Signers().LeftmostSigner()
+		rb = committee.Members[routingBaseIndex].Address
+	case *message.Precommit:
+		routingBaseIndex := m.Signers().LeftmostSigner()
+		rb = committee.Members[routingBaseIndex].Address
```
tbssajal, line 100: Just curious, what is the purpose of selecting the originator this way? So it's deterministic?

tbssajal (commented):
What I understand is that we only need deterministic approach to select originator if someone misbehaves. But it comes with a cost of misleading the relayers. If the originator is not the one who created the message, then no one will become `firstRelayerOriginCluster` which is the most important relayer after originator. Although, in current implementation it is handled by broadcasting the votes to everyone by the actual originator, so it's fine. But I am worried that it might cause trouble and we might not notice this issue.

On the other, if we put the originator as the message creator and put it out of hash and nobody misbehaves, it is perfect. Otherwise, there can be some confusion in the network layers and some lack of efficiency. However, I think if someone wants to misbehaves, he can just go mute. Not responding to the msg would be efficient for the malicious node compared to tempering with the msg and broadcasting it I think (unless I am missing something).

core/types/signers.go
```diff
@@ -71,15 +72,15 @@ func NewSigners(committeeSize int) *Signers {
...
-		length:        0,
-		powers:        make(map[int]*big.Int),
-		power:         new(big.Int),
-		validated:     true,
-		powerAssigned: true, // when we are locally creating a sender info, we are ok with power being 0 initially
+		Bits:           NewValidatorBitmap(committeeSize),
+		Coefficients:   make([]uint16, 0),
+		committeeSize:  committeeSize,
+		length:         0,
+		leftmostSigner: committeeSize, // means no signers
```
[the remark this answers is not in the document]
lorenzo-dev1: Yeah, so that when we do `if i < leftmostSigner` we do not have to deal with the edge case of -1

consensus/tendermint/router/selector/selector.go
lorenzo-dev1: yes basically it remains the same as @vpiyush version, but without saving the routing base in the message.

tbssajal:
> From what I understand, since the `firstRelayerRemoteCluster` still sends to nodes in all clusters, we should still hit every senderType even if we don't start with `firstRelayerOriginCluster`.

The `firstRelayerRemoteCluster` sends to all nodes in his own cluster only. We will hit `firstRelayerOriginCluster` but at a higher level, then it's little use if we use clustering instead of broadcasting to everyone.

Jason-Zhangxin-Chen (approved):
LGTM.

consensus/tendermint/router/selector/selector.go
lorenzo-dev1: Discussed offline, determined it is not an issue provided we pass always the correct committee to `Forward` (the one related to the message).

ScottRMalley (approved):

PR #1195, opened 2025-06-30
Delegate Staking

### LINKED ISSUE #1194: Autonity Upgradability
If autonity is upgraded in the future after mainnet launch, it might be an issue for non-upgradable 3rd party smart contracts to use `IAutonity` specially for staking purposes. Although I cannot find any general solution to this, I have a proposal for staking purposes. Lets allow bonding and unbonding through allowance (like `transferFrom`). This way 3rd party smart contract can have an upgradable proxy smart contract for staking purposes without having to upgrade itself.

tbssajal:
This PR includes features to delegate staking to any 3rd party account or smart contract through the concept of "allowance". It introduces the following changes in Liquid and Autonity contract:

- New state variable `bondAllowances` to track allowance for bonding in Autonity contract.
- New state variable `unbondAllowances` to track allowance for unbonding in Liquid contract.
- Necessary methods to approve and query the allowance.
- Two new staking method `bondFrom` and `unbondFrom` to stake from allowance.

The following is restricted

- Calling `bondFrom` or `unbondFrom` to do staking operation on behalf of the treasury of a validator to his own validator will revert. In other words, if treasury wants to bond or unbond to his own validator, then the treasury cannot give someone allowance to get it done by someone else on his behalf. The treasury needs to do it himself by calling the original `bond` or `unbond` function.
- Bonding (unbinding) allowances cannot be used for anything else other than bonding (unbondig).

The following is not changed

- There are no conceptual changes in `bond` or `unbond` functions.

Closes #1194


autonity/solidity/contracts/Autonity.sol
```diff
@@ -92,6 +92,9 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
     string[] internal committeeNodes;
     mapping(address => mapping(address => uint256)) internal allowances;
 
+    mapping(address => mapping (address => uint256)) internal stakeAllowances;
+    mapping(address => mapping (address => uint256)) internal selfStakeAllowances;
```
ScottRMalley, line 96: It's not clear to me why we need a separate `selfStakeAllowance`. If the validator treasury approves some other contract/address to stake on their behalf, I'd expect that to be assumed as `selfStakeAllowance`. What is the reason to have this separate?

autonity/solidity/contracts/Autonity.sol
```diff
@@ -312,22 +315,35 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
         _val.enode = _enode;
     }
 
+    /**
+    * @inheritdoc IAutonity
+    */
+    function bondFrom(address _account, address _validator, uint256 _amount) external virtual nonReentrant returns (uint256) {
```
ScottRMalley, line 321: Is there a reason to even have an allowance on this one? If you look at contracts like the Uniswap V3 liquidity pools, it's common for the `deposit` function on behalf of another account to be unpermissioned. That is essentially the same thing as someone sending you LNTN. Why would we stop that?

autonity/solidity/contracts/Autonity.sol
```diff
@@ -312,22 +315,35 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
...
+        require(_allowed >= _amount, "amount exceeded allowance");
+        _approveStake(_account, msg.sender, _allowed - _amount);
+        return _bond(_validator, _amount, payable(_account));
+    }
+
+    /**
+    * @inheritdoc IAutonity
+    */
+    function unbondFrom(address _account, address _validator, uint256 _amount) external virtual nonReentrant returns (uint256) {
+        return _unbond(_validator, _amount, payable(_account), msg.sender);
```
ScottRMalley, line 332: I think it's better practice to have the permissioning checks in the external definition rather than in the internal one. Is there a reason we don't check allowance as the first thing?

ScottRMalley (changes requested):

autonity/solidity/contracts/Autonity.sol
tbssajal: Checking allowance is a bit complex in case of unbonding. When unbonding LNTN, the check is done on the liquid contract side, so I think it's fine, but let me know if it seems wrong. For self-unbonding, the check is done before we apply the operation.

autonity/solidity/contracts/Autonity.sol
ScottRMalley: Ah, yeah so we are bonding directing from the `_account`'s balance. What normally happens is that if there is an external staking contract, that contract would have the funds, which would then be staked for a given `_account`.

So I would transfer `_amount` from `msg.sender` and bond it to `_account`. This would mean we could remove the bonding allowance.

tbssajal: I don't think we want that. Because we need to think about rewards as well.

tbssajal: I think I understand what you mean. But it seems there are several challanges with transferring the funds rather than just allowing them to use the funds.

- Safety: Once the funds are transferred, they cannot be claimed back. If something goes wrong and the staking does not go through, the user will just need to wait for the staking to go through at some point in the future. On the other hand, if the user just allows some contract to use the funds and the staking does not go through, the user can always set allowance to zero. And the stake-allowance cannot be used for any purpose other than staking.
- Reward distribution (non-self-bonding): If the user transfers his LNTN to some contract to unbond on his behalf, the contract will get the atn rewards for the transferred funds instead of the user, as the unbonding will be applied at the end of the epoch. Then the contract will be responsible to send those atn to the user as well, however we cannot enforce that via our protocol.
- Inconsistency: It seems a bit challenging to handle self-bonding via transferring NTN. We might want to ensure that we are doing self-bond only when the treasury wants. Because the self-bonding might play some crucial role in validator performace (rewards ratio, slashing risk). So we might need to use the concept of "allowance" for self-bonding. That's why I think it's consistent to use the same concept in all bonding and unbonding operations.

autonity/solidity/contracts/Autonity.sol
```diff
@@ -1781,15 +1824,24 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
         _updateConversionRatio(_val);
     }
 
-    function _unbond(address _validatorAddress, uint256 _amount, address payable _recipient) internal virtual returns (uint256) {
+    function _unbond(address _validatorAddress, uint256 _amount, address payable _recipient, address _staker) internal virtual returns (uint256) {
```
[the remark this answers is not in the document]
tbssajal: How about `_caller`? We have the same parameter `_staker` in `_bond`.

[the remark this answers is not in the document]
ScottRMalley: `_caller` is fine yes

autonity/solidity/contracts/interfaces/IAutonity.sol
```diff
@@ -189,25 +189,57 @@ interface IAutonity is IERC20, IScheduleController {
...
+    * @param _account address of the delegator.
+    * @param _validator address of the validator to delegate stake to.
+    * @param _amount total amount of NTN to bond.
+    * @return uint256 id of the bonding request in the bonding queue
+    */
+    function bondFrom(address _account, address _validator, uint256 _amount) external returns (uint256);
+
+    /**
+    * @notice Create an unbonding request with the `_account` as delegator. The caller needs to have required
+    * unbonding-allowance (self-unbonding-allowance) to unbond LNTN (NTN) from the `_account`.
```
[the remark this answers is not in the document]
tbssajal: done

autonity/solidity/contracts/Autonity.sol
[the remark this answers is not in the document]

ScottRMalley (approved):

PR #1196, opened 2025-06-30
Remove vesting contracts

tbssajal:


Jason-Zhangxin-Chen (approved):
LGTM, now we changed the address of below two contract, remind the change with doc and some other teams, such they can be aligned ASAP. @cmjc

lorenzo-dev1 (approved):
LGTM

PR #1202, opened 2025-07-07
Aggregation optimizations

vpiyush:
Aggregator cache, new msg store implementation, fault detector improvements.


consensus/tendermint/accountability/fault_detector.go
```diff
@@ -108,7 +109,8 @@ type FaultDetector struct {
 
 	logger log.Logger
 
-	scanned map[uint64]struct{}
+	scanned     map[uint64]struct{}
+	saveQueueCh chan message.Msg
```
lorenzo-dev1, line 113: Question: since now the processing and saving of the messages is decoupled, I think it is possible that we miss some accountability infractions?

For example if we receive:
1. vote for v from A
2. vote for v' from B

But both votes are saved in the msgStore only after point 2.

consensus/tendermint/backend/aggregator.go
```diff
@@ -394,94 +401,146 @@ func (a *aggregator) processVotesFor(h uint64, r int64, c uint8, v common.Hash)
...
-		for _, precommitsEvent := range roundInfo.precommits {
-			for _, e := range precommitsEvent {
-				vote := e.Message.(message.Vote)
-				for index, power := range vote.Signers().Powers() {
-					roundInfo.power.Set(index, power)
-				}
-			}
-		}
+		roundInfo.power.Subtract(removedPower)
+		roundInfo.prevotesPower.Subtract(removedPower)
```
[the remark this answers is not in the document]
vpiyush: reverting these for now.

consensus/tendermint/accountability/fault_detector.go
Jason-Zhangxin-Chen: I think those SelfIncriminating faults can be still addressed as the logics wasn't changed at all. For those faults scanned by rule engine, it depends on the timing. So I think that is fine.

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -517,9 +546,14 @@ func (fd *FaultDetector) innocenceProofC1(c *Proof, committee *types.Committee)
 		return nil, errNoEvidenceForC1
 	}
 
-	prevotesForV := fd.msgStore.GetPrevotes(height, func(m *message.Prevote) bool {
-		return m.Value() == precommit.Value() && m.R() == precommit.R()
-	})
+	//prevotesForV := fd.msgStore.GetPrevotes(height, func(m *message.Prevote) bool {
```
Jason-Zhangxin-Chen, line 549: So we remove the query flexibility for better performance of msg store I/O?

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -918,93 +1001,118 @@ oldProposalLoop:
...
+	maxRoundForHeight := fd.msgStore.GetMaxRoundSeen(height)
+	if maxRoundForHeight == -1 {
+		return nil
+	}
 
-			// We need to check whether we have proposals from the prevote's round
-			correspondingProposals := fd.msgStore.GetProposals(height, func(m *message.Propose) bool {
-				return m.R() == prevote.R() && m.Value() == prevote.Value()
-			})
+	for r := int64(0); r <= maxRoundForHeight; r++ { // iterate over all prevotes in each round for this height
```
Jason-Zhangxin-Chen, line 1013: We won't have any attack vector with a large round number if the aggregator always buffers future round message, is it assumption true?

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -918,93 +1001,118 @@ oldProposalLoop:
...
-							Rule:          autonity.PVN,
-							Message:       prevote,
-							OffenderIndex: signerIndex,
-						}
-						proofs = append(proofs, accusation)
-						fd.logger.Info("🕵️ Suspicious behavior detected", "rule", "PVN", "suspect", signer)
+	SignersLoop:
+		// for each signer look at their action
+		for signerIndex := 0; signerIndex < committee.Len(); signerIndex++ {
+			voteForSigner := make(map[common.Hash]*message.Prevote)
```
Jason-Zhangxin-Chen, line 1022: Could `votesOfSigner` a better name?

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -918,93 +1001,118 @@ oldProposalLoop:
...
-							Type:          autonity.Accusation,
-							Rule:          autonity.PVN,
-							Message:       prevote,
-							OffenderIndex: signerIndex,
-						}
-						proofs = append(proofs, accusation)
-						fd.logger.Info("🕵️ Suspicious behavior detected", "rule", "PVN", "suspect", signer)
+	SignersLoop:
+		// for each signer look at their action
+		for signerIndex := 0; signerIndex < committee.Len(); signerIndex++ {
```
Jason-Zhangxin-Chen, line 1021: Comparing with the legacy iteraion of prevote.Signers().FlattenUniq(), now we loop all the index of the committee members, even with some omistion faulty node. I am wondering if this can improve the performance or not?

consensus/tendermint/accountability/fault_detector.go
vpiyush: This has been refactored to sequentially write messages now.

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -139,6 +142,7 @@ func NewFaultDetector(
 		txOpts:                txOpts,
 		tendermintMsgSub:      sub,
 		ruleEngineBlockCh:     make(chan core.ChainEvent, 300),
+		messageEventCh:        messageEventCh, //todo size
```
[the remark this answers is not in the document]
vpiyush: perhaps, but as long as we can't dynamically increase system RAM this may not help if we are not processing messages fast enought

consensus/tendermint/accountability/fault_detector.go
vpiyush: yes aggregator doesn't push future round messages to fault-detector.

consensus/tendermint/events/events.go
```diff
@@ -84,99 +84,50 @@ func (m *MessageEvent) SetDisseminated(disseminated bool) {
...
 
-func (f FuturePowerChangeEvent) Height() uint64 {
-	return f.height
+func (o OldMessageEvent) ErrCh() chan<- error {
+	return o.errCh
 }
 
-func (f FuturePowerChangeEvent) Round() int64 {
-	return f.round
+type MessageEventer interface {
```
lorenzo-dev1, line 118: We can leave this for later as it is not a big deal, but I think if we use this interface then we should name:
- `MessageEventer` --> `MessageEvent`
- `MessageEvent` --> `CurrentMessageEvent`

consensus/tendermint/backend/handler.go
```diff
@@ -234,13 +235,20 @@ func (sb *Backend) handleDecodedMsg(msg message.Msg, errCh chan<- error, sender
...
 			}
-		}
+		}, vote.Signers().CommitteeSize())
+		//for _, signerIndex := range vote.Signers().FlattenUniq() {
+		//	signer := committee.Members[signerIndex].Address
+		//	if !sb.IsJailed(signer) {
+		//		allJailed = false
+		//		break
+		//	}
+		//}
```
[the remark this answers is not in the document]
vpiyush: yeah, ideally we should return an indication from the callback to break outer loop

consensus/tendermint/events/events.go
vpiyush: yeah, I was thinking about it. Eventer is pretty wierd but i didn't think of currentMessageEvent. Because we need a method to create a NewMessageEvent

core/types/signers.go
```diff
@@ -383,6 +383,15 @@ func (s *Signers) FlattenUniq() []int {
...
 
+// ForEachDistinctSigner iterates over each signer and calls the callback function.
+func (s *Signers) ForEachDistinctSigner(callback func(signerIndex int), committeeLen int) {
+	for i := 0; i < committeeLen; i++ {
+		if s.Bitmap.IsSet(i) {
+			callback(i)
+		}
+	}
+}
+
```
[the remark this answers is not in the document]
vpiyush: yeah, it's a leftover of previous integration

consensus/tendermint/accountability/fault_detector.go
vpiyush: yeah i think voteForSigner is not semantically correct. i would change it to votesFromSigner

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -515,9 +521,10 @@ func (fd *FaultDetector) innocenceProofC1(c *Proof, committee *types.Committee)
 		return nil, errNoEvidenceForC1
 	}
 
-	prevotesForV := fd.msgStore.GetPrevotes(height, func(m *message.Prevote) bool {
-		return m.Value() == precommit.Value() && m.R() == precommit.R()
-	})
+	prevotesForV, err := fd.msgStore.GetPrevotesByRoundAndValue(height, precommit.R(), precommit.Value())
+	if err != nil {
+		return nil, errNoEvidenceForC1
+	}
```
lorenzo-dev1, line 527: I think that this should never return  an error, worst case it should return an empty set. messages with rounds <0 or > maxRound should never arrive in the fault_detector, but rather be filtered at decoding.

ScottRMalley (approved):
Good to go when the tests are green.

consensus/tendermint/accountability/fault_detector.go
```diff
@@ -904,93 +938,114 @@ oldProposalLoop:
...
-				var proof *Proof
-				if proposal.ValidRound() == -1 {
-					proof = fd.newPrevotesAccountabilityCheck(height, prevote, proposal, signer, signerIndex)
-				} else {
-					proof = fd.oldPrevotesAccountabilityCheck(height, quorum, proposal, prevote, signer, signerIndex)
+			if len(votesFromSigner) > 1 { // skip equivocations
+				continue SignersLoop
+			}
+
+			for value, prevoteForValue := range votesFromSigner {
```
lorenzo-dev1, line 970: So if I understood correctly this loop actually will last always only a single iteration right?

lorenzo-dev1 (approved):
LGTM

PR #1205, opened 2025-07-08
Update aggregator cache

ScottRMalley:
Adds to the aggregator cache.
- Filtered events are cached (cache dropped on core height change)
- Cached events are reinjected upon invalid signature


consensus/tendermint/backend/aggregator.go
```diff
@@ -601,10 +604,29 @@ func (a *aggregator) processBatches(batches [][]events.UnverifiedMessageEvent, e
...
+					filtered = a.signerSetCache.EmptyFiltered(
+						batch[0].Message.H(),
+						batch[0].Message.R(),
+						batch[0].Message.Code(),
+						batch[0].Message.Value(),
+					)
+				default:
+					return
+				}
+				for _, e := range filtered {
```
vpiyush, line 626: this is bit unsafe, if not for a goroutine this would be a potential deadlock. Perhaps we should write a note on the goroutine mentioning that so that it doesn't get removed during future refactors. Alternatively we could also have a larger backlog buffer in combination of a non-bocking send here.

PR #1209, opened 2025-07-17
Fix formatting ASM Auctioneer

ScottRMalley:
This is a very small PR. Idea is just to:
- Fix naming of `startRound` -> `startTimestamp`
- Fix incorrect Natspec format


lorenzo-dev1 (approved):
lgtm

PR #1210, opened 2025-07-17
Add clustering threshold to protocol params

ScottRMalley:
The only goal of this PR is to allow an operator-controlled setting of the clustering threshold. That is, the committee size above which clustering is active. This lets us specify that both during runtime (via the operator), but also at genesis, to allow testing of clustering on different networks (devnet, testnet).


autonity/solidity/contracts/Autonity.sol
```diff
@@ -502,6 +505,16 @@ contract Autonity is IAutonity, ReentrancyGuard, ScheduleController, Upgradeable
         emit EpochPeriodUpdated(_period, _appliedAtBlock);
     }
 
+    /*
+    * @notice Set the clustering threshold for consensus messaging. Restricted to the Operator account.
+    * @param _threshold Positive integer - the committee size at which clustering becomes active.
+    */
+    function setClusteringThreshold(uint256 _threshold) external virtual onlyOperator {
+        require(_threshold > 0, "clustering threshold cannot be 0");
```
lorenzo-dev1, line 513: would be good to enforce this constraint also on the initial config at genesis, just to make sure it is not deployed with invalid params.

consensus/tendermint/router/routing.go
```diff
@@ -221,6 +232,8 @@ func (m *Router) Start(ctx context.Context, chain interfaces.BlockChainProvider)
 	m.committee = addresses
 	m.inCommittee = curEpoch.Committee.MemberByAddress(m.self) != nil
 	nw, err := cluster.New(addresses, m.latestLatencies, m.self)
+	m.chain = chain
+	m.clusteringThreshold = m.clusteringThresholdByHeight(curEpoch.EpochBlock.Uint64())
```
lorenzo-dev1, line 236: I think that to get the new updated clustering threshold, you need to pass epochBlock + 1, since the config change happens at epoch block

lorenzo-dev1 (approved):
LGTM

yazzaoui (approved):
This is looking good to me

PR #1211, opened 2025-07-18
fix an edge case in AggregateDistinctPrecommits as the assumption is no longer valid.

Jason-Zhangxin-Chen:
This PR fixs an edge case in the distinct message aggregator which was used by AFD, as the legacy assumption is not valid anymore.


vpiyush (commented):
you could use the range loop which is cleaner, but functionally It looks good to me.

Jason-Zhangxin-Chen:
> you could use the range loop which is cleaner, but functionally It looks good to me.

resolved in af28db0062263b01470552949054482f7bef1674

vpiyush (approved):

PR #1212, opened 2025-07-22
fix accountability bug in accusation flow

lorenzo-dev1:
Accusation deadline should be computed based on the `accusation.reportingBlock` rather than the `accusation.block`


Jason-Zhangxin-Chen (approved):
LGTM.

Jason-Zhangxin-Chen (approved):

PR #1214, opened 2025-07-24
separate signer maps for precommits and prevotes in msg store - bug fix

vpiyush:
There was a common signer map being used for precommits and prevotes. which is actually a bug because this values can get overwrittern for precommit and prevote msg. 
There is another improvement - introduction of get only methods for height and round store. which is exclusively used for getter methods.


Jason-Zhangxin-Chen (approved):
LGTM.

yazzaoui (approved):

PR #1218, opened 2025-07-29
fix future round power check

lorenzo-dev1:
Fixes future round power check (`<=` instead of `<`)


consensus/tendermint/backend/aggregator_cache.go
```diff
@@ -209,7 +209,7 @@ func (c *aggregatorCache) contains(height uint64, round int64, event events.Unve
 func (c *aggregatorCache) filter(event events.UnverifiedMessageEvent) bool {
 	msg := event.Message
 	switch msg.(type) {
-	case *message.Precommit, *message.Prevote:
+	case *message.Propose, *message.Precommit, *message.Prevote:
```
ScottRMalley, line 212: We discussed this offline, but we don't want to be filtering proposals here.

ScottRMalley (approved):
LGTM once the tests pass.

yazzaoui (approved):
LGTM

PR #1219, opened 2025-07-30
Improve build ux

lorenzo-dev1:
- Avoid pulling the entire influxdb package (which was in turn requiring `pkg-config` and `rust` installed to compile) and pull instead only the separated v1 client (https://github.com/influxdata/influxdb1-client) as that's the only part used in the project.
- fix the Dockerfile to build autonity


yazzaoui (approved):
LGTM

PR #1220, opened 2025-07-30
Limit upgrade

yazzaoui:
This PR limits the power of the Upgrader precompiler contract to only accept as valid target protocol contracts.


Jason-Zhangxin-Chen (approved):
LGTM unless the last contract address is Auctioneer

params/protocol_contracts.go
```diff
@@ -139,6 +139,19 @@ var (
 	InflationControllerContractAddress    = crypto.CreateAddress(DeployerAddress, 7)
 	OmissionAccountabilityContractAddress = crypto.CreateAddress(DeployerAddress, 8)
 	AuctioneerContractAddress             = crypto.CreateAddress(DeployerAddress, 9)
+
+	ProtocolContracts = []common.Address{
```
lorenzo-dev1, line 143: There are also `LiquidLogic` and `Slasher` which get deployed by the autonity contract in the constructor. I think we should precompute and add also those ones

yazzaoui: Those ones don't save state in the current form so we can just use the AC setters.

tbssajal (approved):

PR #1222, opened 2025-07-31
[BUGFIX] fix clustering setup at epoch switch

lorenzo-dev1:
Fixes:
- Do not update clusters at node start if below clustering threshold
- At epoch switch, update committee and clustering threshold before using them


Jason-Zhangxin-Chen (approved):
LGTM.

PR #1229, opened 2025-08-08
test, refine docker test to run it faster in CI.

Jason-Zhangxin-Chen:
This PR improve the running of docker test in CI to make it faster.


lorenzo-dev1 (commented):
Overall LGTM and the CI looks quite sped up!

I noticed a couple things regarding the e2e-tests though:
1. they still take quite a long time > 25m. Do you think it could be possible to separate those and parallelize them
2. Under `e2e_test` we have several folders containing tests which do not get executed in CI. I think that only the `byzantine` folder gets executed, while the `contract`, `economic`, `mining` etc. etc. folders do not get executed. I am not sure whether this is intended or they have just been forgotten.

